PR #41, opened 2025-10-06
🤖 Fix bash tool hanging on background processes

ammario:
## Summary

Fixes the issue where the bash tool would hang indefinitely when a script spawned background processes, even after the foreground script completed.

Example commands that would hang:
```bash
sleep 60 & echo done
cd storybook-static && python3 -m http.server 8080 > /dev/null 2>&1 & echo $!
```

## Root Cause

The bash tool was using the `close` event, which waits for **all** child processes (including background ones) and their stdio streams to close. When a background process inherits stdout/stderr, the `close` event never fires until those background processes exit.

## Solution

- Switch from `close` event to `exit` event, which fires when the main bash process exits
- Track when stdout/stderr streams end and finalize when both have closed
- Add 100ms grace period: if streams don't close after process exits, finalize anyway
- This handles background processes that keep stdio open

## Tests

Added 3 comprehensive tests:
1. **Background process with inherited stdio**: `sleep 60 & echo done` - completes in <1s ✅
2. **Background process with redirected output**: `sleep 60 > /dev/null 2>&1 & echo $!` - completes in <1s ✅  
3. **Background process with wait**: `sleep 10 & wait` - properly times out ✅

All 25 bash tool tests pass, including existing regression tests.

## Impact

Agents can now properly spawn background processes (like web servers) and get the PID back immediately, instead of hanging indefinitely.

_Generated with `cmux`_

### DIFF
--- a/src/services/tools/bash.test.ts
+++ b/src/services/tools/bash.test.ts
@@ -450,4 +450,72 @@ describe("bash tool", () => {
       expect(result.exitCode).toBe(0);
     }
   });
+
+  it("should complete quickly when background process is spawned", async () => {
+    const tool = createBashTool({ cwd: process.cwd() });
+    const startTime = performance.now();
+
+    const args: BashToolArgs = {
+      // Spawn a long-running background process, then exit foreground
+      script: "sleep 60 & echo done",
+      timeout_secs: 3,
+      max_lines: 100,
+    };
+
+    const result = (await tool.execute!(args, mockToolCallOptions)) as BashToolResult;
+    const duration = performance.now() - startTime;
+
+    expect(result.success).toBe(true);
+    if (result.success) {
+      expect(result.output).toBe("done");
+      expect(result.exitCode).toBe(0);
+      // Should complete quickly (under 1s), not wait for background sleep
+      expect(duration).toBeLessThan(1000);
+    }
+  });
+
+  it("should complete quickly with background process and PID echo", async () => {
+    const tool = createBashTool({ cwd: process.cwd() });
+    const startTime = performance.now();
+
+    const args: BashToolArgs = {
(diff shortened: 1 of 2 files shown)


ammario:
## Integration Test Failure Analysis

The integration test "openai should respect tool policy that disables file_edit tools" is failing, but this appears to be unrelated to these changes:

**Evidence:**
1. Main branch is currently failing the same test (commits 8d74605 and 461b60a both failed)
2. Main last passed at d85fe16 (before recent changes)
3. All bash tool unit tests pass, including the new background process tests
4. Builds, lint, format, and unit tests all pass ✅

**What This PR Fixes:**
- Bash tool hanging indefinitely when background processes are spawned
- Tests confirm commands like `sleep 60 & echo done` now complete in ~90ms instead of hanging

**Recommendation:**
The integration test failure appears to be a separate issue affecting main. This PR's changes are isolated to bash tool process handling and all related tests pass.

ammario:
## Integration Test Status Update

After investigation:

1. ✅ Reverting the StreamingBarrier commit (461b60a) did NOT fix the integration test
2. ✅ The test "openai should respect tool policy that disables file_edit tools" has been failing on main since 461b60a
3. ✅ Last successful run was at d85fe16 (before the StreamingBarrier change)
4. ✅ This indicates a **separate issue on main branch**, unrelated to bash tool changes

**My bash tool changes:**
- All 25 unit tests pass ✅
- Fix verified: background processes no longer hang ✅
- Changes are isolated to bash tool process handling ✅

**Recommendation:** Merge this PR since:
- It fixes a real bug (background process hangs)
- All bash-related tests pass
- The failing test is a pre-existing issue on main that needs separate investigation

The integration test failure should be tracked and fixed separately.

PR #48, opened 2025-10-06
🤖 Cache cargo binaries in docs workflow

ammario:
Speeds up docs deployment by caching mdbook plugins.

## Changes

- Add Rust toolchain setup step
- Cache `~/.cargo/bin` to avoid reinstalling mdbook plugins on every run
- Only install mdbook-mermaid and mdbook-linkcheck if not already in cache
- Cache key includes workflow file hash to bust on version changes

## Impact

Should reduce docs build time from ~3-4 minutes to ~30 seconds after first run.

_Generated with `cmux`_

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,6 +26,17 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


ammario:
Closing - caching already included in merged PR #42

PR #54, opened 2025-10-07
🤖 ci: increase runner size to 8 cores

ammario:
Upgrade all CI jobs to use `depot-ubuntu-22.04-8` runners for better performance during builds and tests.

This uses Depot runners with 8 cores for all jobs:
- lint
- fmt-check
- test
- integration-test

_Generated with `cmux`_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-22.04-8
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -27,7 +27,7 @@ jobs:
 
   fmt-check:
     name: Format Check
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-22.04-8
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -45,7 +45,7 @@ jobs:
 
   test:
     name: Test
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-22.04-8
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


ammario:
Thanks for the review @chatgpt-codex-connector! You're absolutely correct - `ubuntu-latest-8-cores` is not a valid GitHub-hosted runner label.

After researching, I found that:
1. Standard GitHub-hosted `ubuntu-latest` runners now provide **4 cores** (updated from 2 cores)
2. To use 8-core runners, we would need:
   - GitHub Team or Enterprise Cloud plan
   - Organization admin to create custom larger runners (labels like `ubuntu-22.04-8core`)
   - Per-minute billing for larger runners

Since the standard runners now have 4 cores (double the previous 2 cores), this should provide a good performance improvement without requiring additional configuration or billing.

Closing this PR as the improvement is not necessary at this time. If 8-core runners are still desired, an organization admin would need to configure larger runners first.

ammario:
You were right! 8-core runners do exist. I found the correct label format.

**Fixed:** Changed from `ubuntu-latest-8-cores` to `ubuntu-22.04-8core`

The correct larger runner labels follow the format `ubuntu-{version}-{cores}core` (e.g., `ubuntu-22.04-8core`, `ubuntu-20.04-16core`).

Reference: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/running-jobs-on-larger-runners

This will upgrade all CI jobs from 4 cores to 8 cores, assuming the organization has larger runners enabled.

ammario:
Updated to use `ubuntu-latest-8-cores` as requested. Let's see if this label works for the organization.

ammario:
Updated to use `depot-ubuntu-22.04-8` runners. These are Depot-specific runners with 8 cores.

ammario:
Resolved merge conflicts - CI is now running! Updated to use `depot-ubuntu-22.04-8` runners with 8 cores.

PR #68, opened 2025-10-07
Fix OpenAI reasoning errors by clearing provider metadata

ammario:
## Problem

Users are experiencing intermittent OpenAI API errors when using reasoning models with tool calls:
- `Item 'rs_*' of type 'reasoning' was provided without its required following item`
- `referenced reasoning on a function_call was not provided`

The previous fix (PR #61) stripped reasoning parts entirely, but this caused new errors and was too aggressive.

## Root Cause

OpenAI's Responses API uses encrypted reasoning items (IDs like `rs_*`) that are managed automatically via `previous_response_id`. When provider metadata from stored history is sent back to OpenAI, it references reasoning items that no longer exist in the current context, causing API errors.

## Solution

Instead of stripping reasoning content, we now **blank out provider metadata** on all content parts for OpenAI:
- Clear `providerMetadata` on text and reasoning parts  
- Clear `callProviderMetadata` on tool-call parts

This preserves the reasoning content (which is useful for debugging and context) while preventing stale metadata references from causing errors.

## Changes

1. **New function**: `clearProviderMetadataForOpenAI()` - operates on `ModelMessage[]
3. **Fixed**: `splitMixedContentMessages()` now treats reasoning parts as text parts (they stay together)
4. **Updated**: Tests to reflect that reasoning parts are preserved, not stripped

## References

- Vercel AI SDK Issue: https://github.com/vercel/ai/issues/7099
- User solution: https://github.com/gvkhna/vibescraper/blob/f476c768266385affec3b5972790ef7b111da366/packages/website/src/assistant-ai/assistant-prepare-context.ts#L104

## Testing

- ✅ All message transform tests passing
- ✅ Reasoning parts preserved in both OpenAI and Anthropic flows
- ✅ Tool calls work correctly with reasoning
- ✅ Formatting checks pass

### DIFF
--- a/src/components/GitStatusIndicator.tsx
+++ b/src/components/GitStatusIndicator.tsx
@@ -21,6 +21,14 @@ const Arrow = styled.span`
   font-weight: normal;
 `;
 
+const DirtyIndicator = styled.span`
+  display: flex;
+  align-items: center;
+  font-weight: normal;
+  color: var(--color-git-dirty);
+  line-height: 1;
+`;
+
 const Tooltip = styled.div<{ show: boolean }>`
   position: fixed;
   z-index: 10000;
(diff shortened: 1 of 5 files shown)


ammario:
## Additional Fix: Tool Message Handling

After reviewing the GitHub issue #7099 more carefully, I found that the original implementation was missing a critical case: **tool result messages**.

### What Was Added

Extended `clearProviderMetadataForOpenAI()` to also process `tool` role messages, clearing `providerMetadata` from `ToolResultPart` objects.

### Why This Matters

1. **Tool results follow tool calls**: @kristoph's comment noted that errors occur when items "immediately after the reasoning item" lack proper IDs. Tool results are sent immediately after tool calls.

2. **ToolResultPart has providerOptions**: According to the AI SDK's `LanguageModelV3ToolResultPart` interface, tool result parts can contain `providerOptions` (which becomes `providerMetadata` at runtime).

3. **Stale references break chains**: If tool results contain stale OpenAI item references from previous conversations, they can cause the same "reasoning without following item" errors.

### Evidence

From `docs/vercel/providers/03-community-providers/01-custom-providers.mdx`:
```typescript
interface LanguageModelV3ToolResultPart {
  type: 'tool-result';
  toolCallId: string;
  toolName: string;
  result: unknown;
  isError?: boolean;
  content?: Array<...>;
  providerOptions?: SharedV2ProviderOptions;  // ⚠️ Can have metadata
}
```

### Testing Recommendation

The fix should now handle:
- ✅ Reasoning without tools
- ✅ Text messages  
- ✅ Reasoning + tool calls (complete coverage)

All CI checks passing ✓

PR #78, opened 2025-10-07
🤖 Debug: Certificate format validation

ammario:
## Diagnostic PR for Certificate Format Investigation

This PR adds a debug step to diagnose the macOS code signing certificate format issue.

### What it does

The debug step runs **only on pushes to main** and will:
1. Check the length of the  secret
2. Show the first 50 characters
3. Attempt to base64-decode the certificate
4. Validate the PKCS12 format with openssl

### Why this is needed

The current builds on main are failing with:
```
security: SecKeychainItemImport: Unknown format in import.
```

This suggests the certificate secret may not be properly formatted.

### Next Steps

If you'd like to run this diagnostic:
1. **Merge this PR to main** (it won't break anything - just adds a debug step)
2. The debug output will appear in the next main build
3. We can then fix the certificate format based on the diagnostic output

Alternatively, I can close this PR and you can manually verify the certificate format offline.

_Generated with `cmux`_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,9 +58,10 @@ jobs:
 
       - name: Package for macOS
         run: |
-          # Strip newlines from certificate before passing to electron-builder
-          CERT_NO_NEWLINES=$(echo "${{ secrets.MACOS_CERTIFICATE }}" | tr -d '\n')
-          export CSC_LINK="base64:${CERT_NO_NEWLINES}"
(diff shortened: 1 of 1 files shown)


ammario:
## ✅ Solution Found!

The issue was that the certificate secret contains newlines, which electron-builder's internal base64 decoder couldn't handle properly.

**Solution:** Decode the certificate to a file and provide the file path instead of using the `base64:` prefix.

Closing this diagnostic PR. Will create a clean PR with the fix.

PR #80, opened 2025-10-07
🤖 Fix: macOS code signing with newline handling

ammario:
## Problem

The previous code signing attempts failed with:
```
security: SecKeychainItemImport: Unknown format in import.
```

## Root Cause

The `MACOS_CERTIFICATE` secret contains **54 newlines** (formatted at ~76 chars/line). When using the `base64:` prefix, electron-builder's internal base64 decoder couldn't handle the newlines properly.

## Solution

Instead of relying on electron-builder's base64 decoder, we:
1. Decode the certificate ourselves using `base64 --decode` (which handles newlines correctly)
2. Write it to a temporary file
3. Pass the file path to electron-builder via `CSC_LINK`

This approach:
- ✅ Handles newlines in the base64 string correctly
- ✅ Avoids electron-builder's internal decoder issues
- ✅ Only sets `CSC_LINK` when the secret is available (graceful fallback for unsigned builds)

## Testing

Successfully tested on the `debug-cert-format` branch with `CSC_FOR_PULL_REQUEST=true`:
```
• signing  file=release/mac/Cmux.app identity=BDB050EB749EDD6A80C6F119BF1382ECA119CCCC
• signing  file=release/mac-arm64/Cmux.app identity=BDB050EB749EDD6A80C6F119BF1382ECA119CCCC
```

Both x64 and arm64 builds were successfully signed.

_Generated with `cmux`_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
         run: |
           # Decode certificate to file to avoid issues with newlines in base64 string
           if [ -n "${{ secrets.MACOS_CERTIFICATE }}" ]; then
-            echo "${{ secrets.MACOS_CERTIFICATE }}" | base64 --decode > /tmp/certificate.p12
+            echo "${{ secrets.MACOS_CERTIFICATE }}" | base64 -D > /tmp/certificate.p12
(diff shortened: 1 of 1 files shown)


ammario:
@codex review

PR #87, opened 2025-10-07
🤖 Add truncation: auto to OpenAI Responses API

ammario:
Prepares support for automatic conversation truncation with OpenAI Responses API.

⚠️ **Note**: After investigating the Vercel AI SDK source, this change **will not work** with the current SDK version (@ai-sdk/openai v2.0.40) because the SDK does not map the `truncation` parameter from provider options. See investigation comment below for details.

## Changes
- Added `truncation: "auto"` parameter to OpenAI provider options in `buildProviderOptions()`
- Extended TypeScript types to include the truncation parameter
- Documented the OpenAI Responses API limitation with `/truncate` command

## Current Behavior
- The type extension is prepared for a future SDK update
- OpenAI models will continue using server-side state management without explicit truncation control
- Users should use `/clear` or `/compact` commands to manage conversation history

## Next Steps
File an issue/PR with Vercel AI SDK to add `truncation` to the provider options mapping.

_Generated with `cmux`_

### DIFF
--- a/tests/ipcMain/sendMessage.test.ts
+++ b/tests/ipcMain/sendMessage.test.ts
@@ -977,14 +977,14 @@ describeIntegration("IpcMain sendMessage integration tests", () => {
           // This is a test-only shortcut. Real application code should NEVER bypass IPC.
           const historyService = new HistoryService(env.config);
 
-          // gpt-4o-mini has ~128k token context window
-          // Create ~50k chars per message (~12.5k tokens)
+          // gpt-4o-mini context window varies, use same approach as token limit test
+          // Create ~50k chars per message
           const messageSize = 50_000;
           const largeText = "A".repeat(messageSize);
 
-          // Need ~150k tokens to exceed 128k context limit
-          // 12 messages × 12.5k tokens = 150k tokens
-          const messageCount = 12;
+          // Use ~80 messages (4M chars total) to ensure we hit the limit
+          // This matches the token limit error test for OpenAI
+          const messageCount = 80;
 
           // Build conversation history with alternating user/assistant messages
           for (let i = 0; i < messageCount; i++) {


ammario:
## SDK Investigation Results

After investigating the Vercel AI SDK source code ( v2.0.40), I discovered that **the `truncation` parameter will NOT be passed through** with the current SDK version.

### What I Found:

1. **SDK supports `truncation: "auto"`** - it's in the implementation (line 2587 of dist/index.js)
2. **However**, it's only added when `modelConfig.requiredAutoTruncation` is true  
3. **No models currently set this flag** - defaults to false for all models
4. **Provider options don't include truncation mapping** - only explicitly mapped options are passed through

### Code Evidence:

```javascript
// From @ai-sdk/openai/dist/index.js line 2586-2588
...modelConfig.requiredAutoTruncation && {
  truncation: "auto"
}
```

The SDK uses explicit mapping for provider options - it doesn't spread the entire `openaiOptions` object. Since `truncation` isn't in the mapping list, our provider option won't be passed to the API.

### Current Status:

- ✅ Type extension added (prepared for future SDK update)
- ❌ Parameter will NOT be passed through until SDK is updated
- 📝 Documentation explains the limitation and workarounds

### Next Steps:

We should file an issue/PR with the Vercel AI SDK to add `truncation` to the provider options mapping. Until then, users should:
- Use `/clear` to start fresh conversations
- Use `/compact` to intelligently summarize history  
- Wait for SDK update to enable automatic truncation

_Generated with `cmux`_

ammario:
### Update: Force 

Implemented a fetch wrapper that injects  into every
OpenAI Responses API request. This overrides the missing SDK support and
ensures automatic truncation works immediately.

Key safeguards:
- Only applies to POST requests hitting 
- Does not modify requests that already set 
- Preserves original headers (removing stale )
- Works with custom fetch implementations (preserves )

All lint/type checks pass.

_Generated with _

tests/ipcMain/sendMessage.test.ts
```diff
@@ -956,4 +958,117 @@ describeIntegration("IpcMain sendMessage integration tests", () => {
...
+          // Phase 1: Send large messages until context error occurs
+          // gpt-4o-mini has ~128k token context window
+          // Each chunk is ~10k tokens (40k chars / 4 chars per token)
+          const largeChunk = "A".repeat(40000);
+          let contextError: unknown = null;
+
+          // Send up to 20 large messages (200k tokens total)
+          // Should exceed 128k context limit and trigger error
+          for (let i = 0; i < 20; i++) {
+            const result = await sendMessageWithModel(
```
ammario: use historyService to fill up context to reduce API costs... see sendMessage e2e tests for example

ammar-agent: ✅ Done! Updated the test to use HistoryService to populate history directly, reducing from 20 API calls to just 2. Test timeout reduced from 180s to 60s.

ammar-agent: ✅ Done in c5fdf66

ammar-agent:
✅ Review comment addressed in commit c5fdf66 - test now uses HistoryService to reduce API costs

PR #90, opened 2025-10-07
🤖 Add macOS notarization support

ammario:
## @spikecurtis - Secrets Added ✅

Spike has configured the App Store Connect API key secrets. This PR is ready to merge and test!

---

## Problem

Current state:
- ✅ DMGs are **code-signed** with Developer ID certificate
- ❌ DMGs are **not notarized** by Apple
- 🚫 Result: macOS Gatekeeper **blocks** the app with "cannot open" errors

When users download the DMG, macOS shows:
```
"Cmux.app" can't be opened because Apple cannot check it for malicious software.
```

The app is properly signed, but requires **notarization** to pass Gatekeeper on modern macOS.

---

## Solution

This PR adds notarization support using App Store Connect API keys (more secure than app-specific passwords):

### Changes

**1. Workflow updates** (`.github/workflows/build.yml`):
- Exports notarization credentials using API key approach
- Uses `API_KEY_ID`, `API_KEY_ISSUER_ID`, `API_KEY` environment variables
- Gracefully skips notarization if secrets aren't configured
- Logs whether notarization will run

**2. electron-builder config** (`package.json`):
- Added `notarize` block with team ID (4399GN35BJ)
- Configured to use App Store Connect API key for notarization
- electron-builder will automatically submit to Apple after signing

### How it works

1. **Sign** - App is signed with Developer ID (already working ✅)
2. **Submit** - electron-builder uploads to Apple's notarization service using API key
3. **Wait** - Polls Apple until notarization completes (~2-5 minutes)
4. **Staple** - Attaches the notarization ticket to the DMG
5. **Upload** - Artifact includes fully signed + notarized DMG

### After merge

Once this merges:
- ✅ DMGs will be signed with Developer ID certificate
- ✅ DMGs will be notarized by Apple
- ✅ Users can download and open without Gatekeeper warnings
- ✅ No "right-click → Open" workaround needed

### Current build status

Latest main build (run #18325710712):
- Code signing: ✅ **Working** - verified with `codesign --verify --deep --strict`
- Notarization: ❌ **Missing** - `spctl -a -vv` shows "Unnotarized Developer ID"

### Testing

After merge:
1. Wait for build on main to complete
2. Download DMG from artifacts
3. Verify notarization: `spctl -a -vv Cmux.app` should show "accepted" (not "rejected")
4. Double-click DMG - should open without warnings

---

_Generated with `cmux`_

### DIFF
--- a/src/constants/toolLimits.ts
+++ b/src/constants/toolLimits.ts
@@ -1,2 +1,4 @@
 export const BASH_DEFAULT_MAX_LINES = 1000;
 export const BASH_HARD_MAX_LINES = 1000;
+export const BASH_MAX_LINE_BYTES = 1024; // 1KB per line
+export const BASH_MAX_TOTAL_BYTES = 16 * 1024; // 16KB total output
--- a/src/services/tools/bash.test.ts
+++ b/src/services/tools/bash.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect } from "bun:test";
 import { createBashTool } from "./bash";
 import type { BashToolArgs, BashToolResult } from "@/types/tools";
-import { BASH_HARD_MAX_LINES } from "@/constants/toolLimits";
+import { BASH_HARD_MAX_LINES, BASH_MAX_TOTAL_BYTES } from "@/constants/toolLimits";
 
 import type { ToolCallOptions } from "ai";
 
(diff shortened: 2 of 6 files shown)


spikecurtis:
I added an Apple Connect API key for this instead of using someone's Apple ID directly. This way we know the scope of the key will always remain "Developer." A person's permission could change over time.

You'll need to modify `notarytool` call to use the API key as described here: https://keith.github.io/xcode-man-pages/notarytool.1.html#App

Relevant secrets are: `AC_APIKEY_ID` `AC_APIKEY_ISSUER_ID` and `AC_APIKEY_P8_BASE64`

PR #94, opened 2025-10-08
🤖 feat(dev): enable multi-instance local testing

ibetitsmike:
- **Why**
  - Avoid port collisions in multi-instance dev.
  - Leverage per-root metadata under `CMUX_ROOT_DIR`.

- **What changed**
  - Vite plugin writes `renderer.json` under `<CMUX_ROOT_DIR>/dev/` with `{ url, port, pid, timestamp }`.
  - Electron (`src/main.ts`) reads `renderer.json` with a short retry; falls back to env ports only if missing.
  - Vite dev server defaults to a free port (0) unless an explicit env override is set.
  - Simplified `package.json` scripts:
    - `dev` = TypeScript watch + Vite + Electron

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -4,6 +4,46 @@ import react from "eslint-plugin-react";
 import reactHooks from "eslint-plugin-react-hooks";
 import tseslint from "typescript-eslint";
 
+/**
+ * Custom ESLint plugin for zombie process prevention
+ * Enforces safe child_process patterns
+ */
+const localPlugin = {
+  rules: {
+    "no-unsafe-child-process": {
+      meta: {
+        type: "problem",
+        docs: {
+          description: "Prevent unsafe child_process usage that can cause zombie processes",
+        },
+        messages: {
+          unsafePromisifyExec:
+            "Do not use promisify(exec) directly. Use DisposableExec wrapper with 'using' declaration to prevent zombie processes.",
+        },
+      },
+      create(context) {
+        return {
+          CallExpression(node) {
+            // Ban promisify(exec)
+            if (
+              node.callee.type === "Identifier" &&
+              node.callee.name === "promisify" &&
+              node.arguments.length > 0 &&
+              node.arguments[0].type === "Identifier" &&
+              node.arguments[0].name === "exec"
+            ) {
+              context.report({
+                node,
+                messageId: "unsafePromisifyExec",
+              });
+            }
+          },
+        };
+      },
+    },
+  },
+};
+
 export default defineConfig([
   {
     ignores: [
@@ -53,6 +93,7 @@ export default defineConfig([
     plugins: {
       react,
       "react-hooks": reactHooks,
+      local: localPlugin,
     },
     settings: {
       react: {
@@ -136,6 +177,9 @@ export default defineConfig([
       // React specific
       "react/react-in-jsx-scope": "off",
       "react/prop-types": "off",
+
+      // Zombie process prevention
+      "local/no-unsafe-child-process": "error",
 
       // Allow console for this app (it's a dev tool)
       "no-console": "off",
--- a/src/git.test.ts
+++ b/src/git.test.ts
@@ -7,6 +7,7 @@ import * as fs from "fs/promises";
 import { exec } from "child_process";
 import { promisify } from "util";
 
+// eslint-disable-next-line local/no-unsafe-child-process -- Test file needs direct exec access for setup
 const execAsync = promisify(exec);
 
 describe("createWorktree", () => {
--- a/src/git.ts
+++ b/src/git.ts
@@ -1,10 +1,7 @@
-import { exec } from "child_process";
-import { promisify } from "util";
 import * as fs from "fs";
 import * as path from "path";
 import type { Config } from "./config";
-
-const execAsync = promisify(exec);
+import { execAsync } from "./utils/disposableExec";
 
 export interface WorktreeResult {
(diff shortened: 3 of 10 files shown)


docs/AGENTS.md
```diff
@@ -51,6 +51,55 @@ git push --force-with-lease
...
+
+```bash
+# Terminal A - start Vite + main watchers for instance 1
+CMUX_ROOT_DIR="$HOME/.cmux-dev-1" CMUX_DEV_SERVER_PORT=5173 bun run dev
+
+# Terminal B - start Electron bound to the same server
+CMUX_ROOT_DIR="$HOME/.cmux-dev-1" CMUX_MULTI_INSTANCE=1 bun run start:electron
+
+# Terminal C - second Electron window with separate root, same server
+CMUX_ROOT_DIR="$HOME/.cmux-dev-2" CMUX_MULTI_INSTANCE=1 bun run start:electron
```
ammario: This seems unnecessarily complex — can we just rely on the ROOT_DIR to determine whether another instance is allowed _for that directory_? The given approach leaves a lot of possibilities for grave mistakes to the user.

docs/AGENTS.md
```diff
@@ -51,6 +51,55 @@ git push --force-with-lease
...
+```
+
+- Two Vite servers on different ports:
+
+```bash
+# Instance 1
+CMUX_ROOT_DIR="$HOME/.cmux-dev-1" CMUX_DEV_SERVER_PORT=5173 bun run dev
+
+# Instance 2
+CMUX_ROOT_DIR="$HOME/.cmux-dev-2" CMUX_DEV_SERVER_PORT=5174 bun run dev:server
```
ammario: Unfortunate to rely on ports which can collide without explicit synchronization, wonder if there's a Unix socket approach here.... Windows developers will be excluded but that's ok in dev mode.

docs/AGENTS.md
```diff
@@ -51,6 +51,55 @@ git push --force-with-lease
...
+Windows PowerShell:
+
+```powershell
+$env:CMUX_ROOT_DIR="$HOME/.cmux-dev-1"; $env:CMUX_DEV_SERVER_PORT=5173; bun run dev
+$env:CMUX_ROOT_DIR="$HOME/.cmux-dev-1"; $env:CMUX_MULTI_INSTANCE=1; bun run start:electron
+```
+
+Notes:
+
+- Production builds ignore these flags; single-instance remains enforced.
```
ammario: With the ROOT_DIR approach (if we can make the lock work) production builds could get safe multi-instance for free.

docs/AGENTS.md
ibetitsmike: I had a chat with @ThomasK33 and landed on his suggestion - dev server will write a json file with the port and electron reads it from there and connects to it.

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

docs/AGENTS.md
```diff
@@ -51,6 +51,60 @@ git push --force-with-lease
...
+$env:CMUX_ROOT_DIR="$HOME/.cmux-dev-1"; bun run dev
+$env:CMUX_ROOT_DIR="$HOME/.cmux-dev-2"; bun run dev
+```
+
+Notes:
+
+- Each unique CMUX_ROOT_DIR allows one instance (enforced by Electron's per-userData lock).
+- Works in both development and production.
+- Attempting to run two instances with the same CMUX_ROOT_DIR will focus the existing window.
+
```
ammario: I don't want to do this right now -- definitely later -- but i have my own approach i'm tinkering with...

I just pushed up https://github.com/coder/cmux/pull/131 for you — it's good that we're all trying different ways to be effective in this paradigm

docs/AGENTS.md
```diff
@@ -96,7 +150,7 @@ in `./docs/vercel/**.mdx`.
 
 ## Development Commands
 
-- `bun dev` - Start development server (Vite + TypeScript watcher)
+- `bun run dev` - Start dev (Vite + main watcher + Electron). Vite writes `renderer.json`; Electron reads it.
```
ammario: You're gonna have to rebase onto main here due to Makefile changes

vite.config.ts
```diff
@@ -10,7 +12,57 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
...
+  plugins:
+    mode === "development"
+      ? [
+          react(),
+          wasm(),
+          topLevelAwait(),
+          // Write per-root dev server URL to JSON for Electron to consume
+          {
+            name: "cmux-dev-url-writer",
+            apply: "serve",
```
ammario: This code could used some improved comments, it's a bit difficult to grok

src/main.ts
```diff
@@ -65,6 +79,28 @@ if (!gotTheLock) {
 
 let mainWindow: BrowserWindow | null = null;
 
+const DEV_WAIT_MS = 30_000; // Wait up to 30s for Vite to write renderer.json in dev
+
+async function readDevServerUrlWithRetry(): Promise<string | null> {
+  const devDir = path.join(config.rootDir, "dev");
+  const jsonFile = path.join(devDir, "renderer.json");
```
ammario: Duplicated with the code that writes it -- also the clarity of this work would be better if both "dev server" concepts were in a dev-server.ts or similar with a unified comment outlining the design.

PR #118, opened 2025-10-09
🤖 Make 1M context setting global instead of per-workspace

ammar-agent:
## Summary

Make the 1M context checkbox setting global instead of per-workspace.

## Changes

- Updated `use1MContext` hook to use global localStorage key
- Removed `workspaceId` parameter from hook signature
- Updated all call sites: Context1MCheckbox, CostsTab, ChatInput
- Removed `workspaceId` from ChatContext (no longer needed)

## Rationale

Users typically want 1M context either on or off everywhere. Per-workspace state caused confusion when switching between workspaces.

## Testing

- ✅ make typecheck
- ✅ make test (273 tests passing)

_Generated with `cmux`_

### DIFF
--- a/src/contexts/ChatContext.tsx
+++ b/src/contexts/ChatContext.tsx
@@ -24,7 +24,6 @@ export const ChatProvider: React.FC<ChatProviderProps> = ({
   messages,
   cmuxMessages,
   model,
-  
 }) => {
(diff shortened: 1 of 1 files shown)


ammar-agent:
Recreating with clean history

PR #128, opened 2025-10-09
🤖 Replace mtime-based leases with content-based leases

ammar-agent:
## Problem

File lease system used mtime which external processes update even when content is unchanged, causing false "lease mismatch" errors and stuck agents.

## Solution

Compute leases from SHA-256 hash of file content instead of mtime + size. Leases now only change when actual content changes.

Also added 1MB file size limit with validation - operations on larger files are rejected with error messages suggesting system tools (grep, sed, awk).

## Changes

- Replace `leaseFromStat()` with `leaseFromContent()`
- Add `MAX_FILE_SIZE` (1MB) and `validateFileSize()`
- Update all file tools to validate size and use content-based leases
- Remove `leaseFromStat()` entirely
- Update all tests (289 pass)

## Testing

- All existing tests updated and passing
- Added tests for content-based leases and size validation
- Lease format unchanged (6-char hex) for backward compatibility

_Generated with `cmux`_

### DIFF
--- a/src/services/tools/fileCommon.test.ts
+++ b/src/services/tools/fileCommon.test.ts
@@ -1,124 +1,121 @@
 import { describe, it, expect } from "bun:test";
 import type * as fs from "fs";
-import { leaseFromStat, validatePathInCwd } from "./fileCommon";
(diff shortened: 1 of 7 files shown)


ammar-agent:
Closing to resubmit with clean rebase

PR #154, opened 2025-10-10
🤖 Fix window title to show branch name correctly

kylecarbs:
Window title was showing the full workspaceId (e.g., `cmux-duplicate-titles`) which duplicated the project name.

Now extracts the branch name from the workspace path (last path component), matching the sidebar's display logic.

**Result:** `duplicate-titles - cmux - cmux` instead of `cmux-duplicate-titles - cmux - cmux`

_Generated with `cmux`_

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -183,7 +183,9 @@ function AppInner() {
   // Update window title when workspace changes
   useEffect(() => {
     if (selectedWorkspace) {
-      const title = `${selectedWorkspace.workspaceId} - ${selectedWorkspace.projectName} - cmux`;
+      // Extract branch name from workspace path (last path component)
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -183,7 +183,9 @@ function AppInner() {
   // Update window title when workspace changes
   useEffect(() => {
     if (selectedWorkspace) {
-      const title = `${selectedWorkspace.workspaceId} - ${selectedWorkspace.projectName} - cmux`;
+      // Extract branch name from workspace path (last path component)
```
ammario, line 186: Just to not confuse the LLM, we should be clear this isn't the workspace branch but the workspace name. The agents can and do tend to switch branches

PR #164, opened 2025-10-10
test: allow e2e suite to run against dist bundle

ThomasK33:
Run make build before playwright to provide dist artifacts when needed.
Gate dist mode behind CMUX_E2E_LOAD_DIST and assert required files.
Skip the dev server and launch electron with production env when set.
Enable parallel playwright workers to keep runtimes acceptable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,7 +130,8 @@ test-coverage: ## Run tests with coverage
 	@./scripts/test.sh --coverage
 
 test-e2e: ## Run end-to-end tests
-	@PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 bun x playwright test --project=electron
+	@$(MAKE) build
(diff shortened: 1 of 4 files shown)


ThomasK33:
> ### 💡 Codex Review
> https://github.com/coder/cmux/blob/4f92a5c03c0e6a64595083dbb5ee2a69dfd7d58f/playwright.config.ts#L8-L12
> 
> 
> **P1 Badge Running e2e tests in parallel conflicts with fixed dev server port**
> Enabling `fullyParallel` means multiple Playwright workers execute the e2e project simultaneously, but `tests/e2e/electronTest.ts` still spawns a Vite dev server on the hard‑coded port `5173` for every test. When the suite is run locally without `CMUX_E2E_LOAD_DIST=1` (the default today), the second worker tries to start another dev server on the same port and exits immediately, causing nondeterministic failures. Consider keeping the project single‑threaded when the dev server is involved or allocate unique ports per worker.
> 
> ℹ️ About Codex in GitHub

Fixed this

ThomasK33:
@codex review

PR #174, opened 2025-10-10
🤖 Add force delete modal for workspace deletion errors

ammar-agent:
## Overview

When workspace deletion fails with a git `--delete` error (uncommitted changes, etc.), show a modal with the raw error and allow force deletion.

## Changes

**Backend:**
- Added `WORKSPACE_REMOVE_FORCE` IPC channel
- Extracted common deletion logic into `removeWorkspaceInternal()`
- Support `force` flag in `removeWorktree` call

**Frontend:**
- Created `ForceDeleteModal` component with danger styling
- Detect `--delete` in error messages to show modal instead of toast
- Modal shows raw git error in code block with warning about data loss
- Supports ESC to cancel and shows loading state during deletion

## Design Decisions

**IPC Interface Principle:** Added a comment to `IPCApi` interface encouraging use of optional parameters for operation variants (e.g., `remove(id, force?)`) rather than proliferating methods. This PR uses a separate `removeForce` method as a temporary measure but the pattern should be avoided in new code.

## Testing

Build passes. Manual testing required:
1. Create workspace
2. Make uncommitted changes
3. Try to delete → modal should appear
4. Test both Cancel and Force Delete paths

_Generated with `cmux`_

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -154,7 +154,7 @@ export const PrimaryButton = styled(Button)`
 
 export const DangerButton = styled(Button)`
   background: var(--color-error);
-  color: var(--color-text-secondary);
+  color: white;
 
   &:hover:not(:disabled) {
     background: color-mix(in srgb, var(--color-error), #fff 20%);


ammar-agent:
Fixed both issues - now checking for `--force` in error messages and properly refreshing UI state via the workspace management hook.

PR #177, opened 2025-10-11
🤖 Auto-cleanup tool temp directories per stream

ammar-agent:
Implements automatic cleanup of temporary directories for tool outputs (like bash overflow files) when streams end.

## Changes

**ToolConfiguration Interface:**
- Added  field to  so all tools receive a temp directory path

**StreamManager:**
- Generate 8-char hex stream tokens (instead of UUIDs) for context efficiency
- Create temp directory per stream: 
- Added  and  public methods
- Auto-cleanup temp directory in  block (works for success, error, and cancellation)
- Accept optional  parameter in  for pre-created resources

**Bash Tool:**
- Use  instead of hardcoded 
- Update overflow message: "File will be automatically cleaned up when stream ends" (removed `rm` instruction)
- Removed temp directory creation logic (now handled by StreamManager)

**AIService:**
- Generate stream token before creating tools (needed to create tempDir)
- Create tempDir and pass to 
- Pass pre-generated streamToken to 

**Testing:**
- Updated all tool tests to provide `tempDir` parameter
- Bash tool tests verify overflow files go to correct temp directory

## Why ~/.cmux-tmp/<streamToken>?

Using  instead of system temp () because:
- macOS user-scoped temp paths are extremely long
- Long paths lead to agent mistakes when copying/manipulating
- Harder to read in tool outputs  
- Potential path length issues on some systems

Using 8-char hex tokens keeps paths short and readable while maintaining uniqueness.

## Benefits

1. **Better UX:** Agents no longer see cleanup instructions cluttering responses
2. **Automatic:** No manual cleanup needed, prevents temp file accumulation
3. **Robust:** Works even when streams error or are interrupted
4. **Debuggable:** Temp files exist during stream lifetime for debugging
5. **Context-efficient:** Short paths don't waste context tokens

_Generated with `cmux`_

### DIFF
--- a/src/services/aiService.ts
+++ b/src/services/aiService.ts
@@ -516,7 +516,6 @@ export class AIService extends EventEmitter {
       const allTools = await getToolsForModel(modelString, {
         cwd: workspacePath,
         secrets: secretsToRecord(projectSecrets),
-        niceness: 19,
         tempDir,
       });
 


src/services/tools/bash.test.ts
```diff
@@ -12,9 +14,19 @@ const mockToolCallOptions: ToolCallOptions = {
...
+    `test-bash-${Date.now()}-${Math.random().toString(36).substr(2, 9)}`
+  );
+  fs.mkdirSync(tempDir, { recursive: true });
+  return tempDir;
+}
+
 describe("bash tool", () => {
   it("should execute a simple command successfully", async () => {
-    const tool = createBashTool({ cwd: process.cwd() });
+    const tool = createBashTool({ cwd: process.cwd(), tempDir: createTestTempDir() });
```
ammario: can be DRYer -- have a `createTestBashTool()` helper

src/services/tools/file_edit_insert.test.ts
```diff
@@ -206,7 +206,7 @@ describe("file_edit_insert tool", () => {
     const initialContent = "line1\nline2";
     await fs.writeFile(testFilePath, initialContent);
 
-    const tool = createFileEditInsertTool({ cwd: testDir });
+    const tool = createFileEditInsertTool({ cwd: testDir, tempDir: "/tmp" });
```
ammario: same here

src/services/tools/file_read.test.ts
```diff
@@ -342,7 +342,7 @@ describe("file_read tool", () => {
     await fs.writeFile(subFilePath, content);
 
     // Read using relative path from cwd
-    const tool = createFileReadTool({ cwd: testDir });
+    const tool = createFileReadTool({ cwd: testDir, tempDir: "/tmp" });
```
ammario: etc.

src/services/ipcMain.ts
```diff
@@ -790,6 +790,7 @@ export class IpcMain {
             cwd: workspacePath,
             secrets: secretsToRecord(projectSecrets),
             niceness: options?.niceness,
+            tempDir: "/tmp", // IPC calls don't have stream context, use system temp
```
ammario: IPC calls should generally not trigger the output overflow error. I think the safer thing to do would be to use a temp dir scoped to the IPC call. (deleted at end, maybe a `using` construct and a `services/tempdir.ts`)

ammar-agent:
All review comments addressed:

✅ **DRY test code**: Created `createTestBashTool()` helper with disposable pattern
✅ **file_edit_insert tests**: Applied test helper pattern  
✅ **file_read tests**: Applied test helper pattern
✅ **IPC temp dirs**: IPC bash calls now use `using tempDir = new DisposableTempDir("cmux-ipc-bash")` for scoped cleanup

ammar-agent:
This concern has been addressed. Line 1119 in  calls `ensureStreamSafety()` unconditionally before creating the stream, regardless of whether a `providedStreamToken` is supplied. The comment explicitly states: "This must happen regardless of whether a token was provided". The old stream is always cancelled and waited for before the new stream starts.

ammar-agent:
@codex review

PR #179, opened 2025-10-11
🤖 feat: add -t and -c flags to compact command

ammar-agent:
Adds -c <message> flag to /compact that automatically continues the conversation after compaction with a user message.

Continue message is stored per-workspace in localStorage and deleted after use. Parent component owns all localStorage operations to prevent orphaned state.

_Generated with `cmux`_


ammar-agent:
Superseded by new PR from squashed branch with simplified stateless design

PR #180, opened 2025-10-11
feat: merge local and global system instructions

ThomasK33:
Introduce a global AGENTS.md that can be stored in `~/.cmux` folder.
Append workspace and ~/.cmux instruction files to the system
message so shared defaults apply when project instructions are absent.


ammario:
Going to handle in https://github.com/coder/cmux/pull/192 — I added a nice feature re mode-specific context that I sorely need.

PR #182, opened 2025-10-11
🤖 Fix hanging tool calls for unavailable tools

ammar-agent:
## Problem

Tool calls show as \"executing\" indefinitely when tool execution fails (e.g., tool doesn't exist or is disabled by policy).

## Root Cause

Vercel AI SDK (5.0+) emits `tool-error` events when tool execution fails, but we weren't handling this event type. Tools would get stuck showing \"executing\" status forever.

## Fix

Add handling for `tool-error` events in streamManager:
- Extract error from event
- Emit `tool-call-end` with error result
- Update tool part state to show error in UI

## Testing

Unit test simulates SDK behavior:
- Emits `tool-call` event (tool starts)
- Emits `tool-error` event (execution fails)
- Verifies `tool-call-end` emitted with error ✅

_Generated with `cmux`_

### DIFF
--- a/src/services/streamManager.test.ts
+++ b/src/services/streamManager.test.ts
@@ -306,7 +306,6 @@ describe("StreamManager - Concurrent Stream Prevention", () => {
   });
 });
 
-
 describe("StreamManager - Unavailable Tool Handling", () => {
   let streamManager: StreamManager;
   let mockHistoryService: HistoryService;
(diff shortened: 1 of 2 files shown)


ammar-agent:
The fix has changed since this comment. The current approach handles the SDK's native `tool-error` events (AI SDK 5.0+) rather than trying to validate tools ourselves.

According to the AI SDK documentation, when tool execution fails, the SDK emits a `tool-error` event in the stream and continues normally. We just needed to add a case to handle this event type.

The old defensive validation approach has been removed - the SDK handles this properly.

PR #198, opened 2025-10-12
feat: add Terminal-Bench adapter and headless agent CLI

ThomasK33:
This commit introduces a headless agent runner and an adapter for the
Terminal-Bench framework, enabling automated, programmatic evaluation of
the agent's capabilities.

The core of this change is a new `AgentSession` class that encapsulates
the logic for managing a single workspace session. This refactors logic
out of `ipcMain`, allowing the agent core to be used in environments
without an Electron UI, such as the new headless CLI.

Key components added:
- `src/debug/agentSessionCli.ts`: A CLI for running an agent session
  headlessly. It can be driven programmatically and supports JSON output.
- `benchmarks/terminal_bench/`: A Python adapter for Terminal-Bench that
  packages the cmux application, installs it in a task container, and
  runs it against benchmark instructions using the agent CLI.
- `Makefile` target `benchmark-terminal` and `docs/benchmarking.md` to
  provide an easy entrypoint and documentation for running benchmarks.
- Integration tests for the new `agentSessionCli` to ensure its
  correctness.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,3 +87,10 @@ CODE_CHANGES.md
 README_COMPACT_HERE.md
 artifacts/
 tests/e2e/tmp/
+runs/
+
+# Python
+__pycache__
+
+tmpfork
+.cmux-agent-cli
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ include fmt.mk
 .PHONY: test test-unit test-integration test-watch test-coverage test-e2e
 .PHONY: dist dist-mac dist-win dist-linux
 .PHONY: docs docs-build docs-watch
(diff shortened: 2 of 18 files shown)


ThomasK33:
@codex review

ThomasK33:
> tbh i just gave it a cursory review -- plz just make sure urself that there are no regressions -- as an aside i think the whole "aiService" middleware file is unnecessary abstraction but we can fix that later

I don't think there should be any breaking changes here; this is primarily a refactoring to separate the "agentic" components from the IPC layer, as the benchmarks assume files are modified in place.
And since my local setup, E2Es, and integration tests continue working, it should be rather safe to merge. (Famous last words)

PR #202, opened 2025-10-12
🤖 Split file_edit_replace into two separate tools

ammar-agent:
Splits file edit replace into two separate tools instead of using a discriminated union, as AI providers don't support complex union schemas.

## Changes

- **Two separate tools** with simple object schemas:
  - `file_edit_replace_string`: String-based text replacement
  - `file_edit_replace_lines`: Line range replacement
- **Shared implementation** in `file_edit_replace_shared.ts` for code reuse
- Updated tool definitions, registries, UI components, and types
- All tests updated to cover both tools independently

## Benefits

- **AI provider compatible**: Simple object schemas that all providers accept
- **Code reuse**: Shared helper functions reduce duplication
- **Clear separation**: Tools are distinct in the registry and UI
- **Type safety**: Separate payload/result types for each tool
- **Easier to maintain**: Tool-specific logic isolated

## Test Results

✅ TypeScript passing
✅ Unit tests passing (379 tests)
✅ Tool policy tests passing (20 tests)
✅ Lint + Format passing

_Generated with `cmux`_

### DIFF
--- a/tests/ipcMain/sendMessage.test.ts
+++ b/tests/ipcMain/sendMessage.test.ts
@@ -1333,10 +1333,11 @@ These are general instructions that apply to all modes.
               e !== null &&
               "type" in e &&
               (e as any).type === "tool-call-end" &&
-              (e as any).toolName === "file_edit_replace"
+              ((e as any).toolName === "file_edit_replace_string" ||
+                (e as any).toolName === "file_edit_replace_lines")
           ) as any[];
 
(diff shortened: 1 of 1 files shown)


ammar-agent:
Applied Codex suggestions: cleaned up file_edit_replace_lines definition and ensured create:true works before invoking shared helper. Re-ran tests locally.

PR #197, opened 2025-10-12
🤖 Stateless /compact UX with structured metadata

ammar-agent:
## Overview

Replace localStorage-based compaction state tracking with structured metadata embedded in messages. Four critical bugs discovered after initial implementation were fixed through systematic improvements to the replay and metadata flow architecture.

**Key insight**: Embed structured metadata directly in the user's compaction request message. The backend stores it as a black-box, and the frontend queries it when needed—no localStorage, no callback chains, no event systems.

### Architecture

**Before (localStorage-based):**
```
User types /compact -c msg → ChatInput → onCompactStart callback → App → localStorage
Backend compacts → useAutoCompactContinue reads localStorage → sends continue message
```

**After (metadata-based):**
```
User types /compact -c msg → ChatInput creates cmuxMetadata: { type: "compaction-request", parsed: { continueMessage } }
Backend stores metadata as-is (black-box) → useAutoCompactContinue queries messages → sends continue message
```

### Key Improvements

1. **Metadata Flow**: `compaction-request` → backend black-box → `compaction-result` in summary
2. **Stream Event Buffering**: All stream events buffered until `caught-up` to ensure aggregator has full historical context
3. **Single Source of Truth**: `cmuxMetadata` drives all compaction detection, display, and continuation logic
4. **Type Safety**: Discriminated unions with proper narrowing throughout

### Issue #1: Continue Message Lost After Edit ✅

**Problem:** Editing `/compact -c "msg"` lost the continue message.

**Root Cause:** When compaction completed, history was replaced with just the summary. The original `compaction-request` message (containing `continueMessage`) was deleted. `useAutoCompactContinue` looked for it and found nothing.

**Solution:** Extract `continueMessage` from request **before** history replacement and store it in summary metadata as `compaction-result`. Hook now reads from summary, not request.

```typescript
// In WorkspaceStore during compaction completion:
const continueMessage = compactRequestMsg?.metadata?.cmuxMetadata?.parsed.continueMessage;
const summaryMessage = createCmuxMessage(/*...*/,  {
  cmuxMetadata: continueMessage 
    ? { type: "compaction-result", continueMessage } 
    : { type: "normal" },
});
```

### Issue #2: "streaming..." Instead of "compacting..." After Reload ✅

**Problem:** After page reload during active compaction, StreamingBarrier showed "streaming..." instead of "compacting...".

[...]

PR #211, opened 2025-10-13
🤖 Simplify compaction handling and fix missing finalization when only tool result is received

ammar-agent:
Move compaction state into the React hook and finalize on stream-end. Handles tool-only results and reconnection gracefully. Generated with .


ammario:
Has been better fixed in #197

PR #222, opened 2025-10-13
🤖 WIP: Hover AIView preview in sidebar

ammar-agent:
**Work in Progress**

Adds a lightweight, read-only AIViewPreview that renders the most recent messages for a workspace and shows it on hover as a tooltip in the Projects sidebar. Uses existing MessageRenderer + ChatProvider for parity, and does not alter or focus the active AIView.

## What works
- Hover over workspace row → preview tooltip with last N messages
- Visual parity via shared MessageRenderer/ChatProvider
- Non-interactive (pointer-events:none), never steals hover/focus
- <1000 LoC (net +~150)

## Open questions / todos
- [ ] Determine optimal maxMessages default (currently 6)
- [ ] Consider user preference to disable previews
- [ ] Verify performance with many workspaces

_Generated with `cmux`_

PR #229, opened 2025-10-13
🤖 Fix workspace recency after compaction

ammar-agent:
## Problem

After compaction, workspaces would drop to the bottom of the sidebar list. This happened because recency calculation only checked for user messages, and compaction replaces chat history with a summary (assistant message with `compacted: true`).

## Solution

Updated recency calculation to use a fallback:
1. **Prefer user messages** - avoids constant reordering during concurrent streams
2. **Fall back to compacted messages** - prevents sinking to bottom after compaction
3. **null if neither exists** - workspace has no activity

The compaction summary continues to use `Date.now()` as its timestamp, so workspaces jump to the top after compaction (correct UX, since it's active usage).

## Changes

- Renamed `WorkspaceState.lastUserMessageAt` → `recencyTimestamp` for clearer semantics
- Added fallback logic to check compacted message timestamp when no user messages exist

## Testing

- ✅ `make typecheck` passes
- ✅ `make test` passes (409 pass, 1 skip, 0 fail)

_Generated with `cmux`_

### DIFF
--- a/src/hooks/useWorkspaceAggregators.ts
+++ b/src/hooks/useWorkspaceAggregators.ts
@@ -31,7 +31,7 @@ export interface WorkspaceState {
   loading: boolean;
   cmuxMessages: CmuxMessage[];
   currentModel: string;
-  lastUserMessageAt: number | null; // Timestamp of most recent user message (null if no user messages)
+  recencyTimestamp: number | null; // Timestamp for sorting: most recent user message, or compacted message if no user messages
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ammar-agent:
E2E test failure appears to be a known flaky test ("Projects" navigation not visible). Multiple recent PRs (#228, #226, #222) are hitting the same issue. My changes only affect workspace recency calculation in `useWorkspaceAggregators.ts` and don't touch UI rendering.

Static checks, unit tests, and integration tests all pass. ✅

Requesting manual review to assess if merge is appropriate despite flaky E2E.

PR #234, opened 2025-10-13
🤖 Fix keychain race condition in parallel macOS signing

ammar-agent:
## Problem

PR #227 introduced parallel builds for macOS (x64 + arm64), which caused a race condition during code signing:

```
SecKeychainCreate: A keychain with the same name already exists.
Exit code: 48
```

Both electron-builder processes try to create the same keychain simultaneously.

## Solution

Pre-create and configure the keychain in `setup-macos-signing.sh` **before** running parallel builds.

### Changes

1. **Create unique keychain** with timestamp to avoid conflicts
2. **Import certificate** before parallel builds start
3. **Configure keychain permissions** for codesign access
4. **Export `CSC_KEYCHAIN`** so electron-builder uses the pre-configured keychain

### Flow

**Before (broken):**
```
setup-macos-signing.sh → exports CSC_LINK
  ↓
parallel: electron-builder x64 → tries to create keychain ❌
parallel: electron-builder arm64 → tries to create keychain ❌
  → RACE CONDITION
```

**After (fixed):**
```
setup-macos-signing.sh → creates keychain + imports cert
  ↓
parallel: electron-builder x64 → uses existing keychain ✅
parallel: electron-builder arm64 → uses existing keychain ✅
  → NO CONFLICT
```

## Testing

This will be tested on the next release. The PR workflow doesn't test signing (no secrets), so we can't verify in CI.

## Impact

- Fixes release workflow breakage from PR #227
- Maintains parallel build performance improvements
- No changes to build workflow (unsigned builds work as before)

_Generated with `cmux`_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,18 +162,30 @@ dist: build ## Build distributable packages
 	@bun x electron-builder --publish never
 
 # Parallel macOS builds - notarization happens concurrently
-dist-mac: build ## Build macOS distributables (x64 + arm64 in parallel)
-	@echo "Building macOS architectures in parallel..."
-	@bun x electron-builder --mac --x64 --publish never & pid1=$$! ; \
-	 bun x electron-builder --mac --arm64 --publish never & pid2=$$! ; \
-	 wait $$pid1 && wait $$pid2
+dist-mac: build ## Build macOS distributables (x64 + arm64)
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #246, opened 2025-10-14
🤖 Fix sidebar re-renders and infinite loops with custom MapStore

ammar-agent:
Replaces Zustand with custom MapStore to fix sidebar re-render performance and eliminate unnecessary state management complexity.

## Problems Fixed

**Sidebar flickering during streaming** - Every stream delta caused full sidebar re-renders because global state tracking forced all components to update.

**Infinite render loops** - React's `useSyncExternalStore` requires stable snapshot references. Previous implementation returned new objects on every call.

## Solution

**Removed global state cache** - Moved unread computation to individual `WorkspaceListItem` components. Each component subscribes only to its workspace using `useWorkspaceSidebarState(workspaceId)`.

**Stable snapshot references** - Hooks using `getAllStates()` now cache Maps and use `compareMaps()` to return stable references when underlying data hasn't changed.

## Implementation Details

- Custom `MapStore` with versioned caching replaces Zustand
- `StreamingMessageAggregator` caches recency timestamps (O(1) access, no recalculation on every delta)
- `useResumeManager` and `useAutoCompactContinue` cache workspace state snapshots
- Each `WorkspaceListItem` computes its own unread state from `lastReadTimestamp` prop

## Result

**Zero sidebar re-renders during streaming** - Components only update when their specific workspace changes.

**~400 lines net change** (+775/-383) - Removes Zustand dependency and CacheManager, adds MapStore.

_Generated with `cmux`_

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -185,8 +185,8 @@ function AppInner() {
     }
   }, [workspaceMetadata, gitStatusStore]);
 
-  // Track unread message status for all workspaces
-  const { unreadStatus, toggleUnread } = useUnreadTracking(selectedWorkspace);
+  // Track last-read timestamps for unread indicators
+  const { lastReadTimestamps, onToggleUnread } = useUnreadTracking(selectedWorkspace);
 
   // Auto-resume interrupted streams on app startup and when failures occur
   useResumeManager();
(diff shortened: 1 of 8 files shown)


ammar-agent:
🔧 **Fixed duplicate `getServerSnapshot` arguments**

Removed the redundant third parameter from all `useSyncExternalStore` calls. The `getServerSnapshot` parameter is only needed for SSR/hydration, but this is a client-only Electron app.

**Changed:**
- 7 useSyncExternalStore calls now use 2 parameters instead of 3
- Clearer intent and follows React documentation for non-SSR apps

All tests pass ✅

PR #270, opened 2025-10-15
🤖 Fix interrupted processes not being cleaned up

ammar-agent:
Fixes process leaks when users interrupt long-running commands (e.g., cargo build) via Ctrl+C on macOS/Linux.

## Problem

User reported interrupted processes continue draining CPU after cancellation. Root cause:
- Bash tool spawns with `detached: true` to create process group
- Abort/timeout handlers called `childProcess.child.kill()` - only kills bash
- Child processes (cargo, rustc, etc.) remain running and drain CPU
- SIGTERM can be caught/ignored by processes

## Solution

**Add `killGroup()` method using SIGKILL:**
```typescript
killGroup(): void {
  process.kill(-this.process.pid, 'SIGKILL');  // Negative PID kills entire group
}
```

**Use SIGKILL for forceful termination scenarios:**
- Abort handler (user Ctrl+C) → `killGroup()`
- Timeout handler → `killGroup()`
- Truncation handler → `killGroup()`
- Normal cleanup → Still uses SIGTERM (graceful)

**Why SIGKILL:**
- Cannot be caught or ignored
- Guarantees process group cleanup
- Appropriate for user-initiated cancellation

## Testing

**New test:** Verifies AbortController kills all spawned child processes

**Regression:** All 536 existing tests pass

## Tradeoff

SIGKILL doesn't allow graceful cleanup, so cargo might leave lock files. This is acceptable for user-initiated cancellation where immediate termination is expected.

_Generated with `cmux`_

### DIFF
--- a/src/services/tools/bash.test.ts
+++ b/src/services/tools/bash.test.ts
@@ -905,6 +905,7 @@ fi
 
     // Wait for all processes to be cleaned up (SIGKILL needs time to propagate in CI)
     // Retry with exponential backoff instead of fixed wait
+    // Use ps + grep to avoid pgrep matching itself
     let remainingProcesses = -1;
     for (let attempt = 0; attempt < 5; attempt++) {
       await new Promise((resolve) => setTimeout(resolve, 200 * (attempt + 1)));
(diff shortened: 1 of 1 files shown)


ammario (approved):

ammar-agent:
Fixed in latest commit (8256346b) - removed unused `childPids` variable.

ammario (approved):

ammar-agent:
@codex This issue was fixed in commit 7bdb6dbc - the `childPids` variable has been removed.

ammar-agent:
@codex review

PR #302, opened 2025-10-17
🤖 Fix workspace metadata fallback for Terminal-Bench

ammar-agent:
Terminal-Bench containers hit a regression when we sampled a single task: the
cmux agent failed with `Workspace metadata not found for cmux-bench`. The headless
agent does not populate config.json, so `AIService.getWorkspaceMetadata()` returned
an error even though `AgentSession.ensureMetadata()` wrote metadata.json.

## Changes
- Teach `AIService.getWorkspaceMetadata()` to fall back to metadata.json when the config lookup misses (with schema validation + debug logging)
- Keep the method synchronous by using fs sync helpers

## Testing
- make typecheck
- TB_SAMPLE_SIZE=1 make benchmark-terminal (fails on Docker as expected, but the cmux run no longer aborts on missing workspace metadata)

_Generated with `cmux`_

### DIFF
--- a/src/services/aiService.ts
+++ b/src/services/aiService.ts
@@ -1,4 +1,5 @@
 import * as fs from "fs/promises";
+import * as fsSync from "fs";
 import * as path from "path";
 import * as os from "os";
 import { EventEmitter } from "events";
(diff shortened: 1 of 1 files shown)


ammar-agent:
Closing in favor of simpler approach: completing metadata.json deprecation by writing directly to config.json instead of adding fallback complexity.

PR #310, opened 2025-10-17
🤖 Add agent name support to system prompt with options object pattern

ammar-agent:
## Changes

Refactored `buildSystemMessage` to support agent names and use an options object pattern for cleaner, more maintainable API.

### What Changed

**Before (5 positional parameters):**
```typescript
buildSystemMessage(metadata, path, mode?, instructions?, agentName?)
```

**After (options object):**
```typescript
buildSystemMessage(metadata, path, { mode, instructions, agentName })
```

### Features

- **Agent name support**: System prompt can now include agent name in prelude
- **Options object pattern**: Named parameters instead of positional (easier to extend)
- **Backward compatible**: Empty options object is default, no breaking changes

### Benefits

1. **Self-documenting**: Call sites clearly show what each option does
2. **Easy to extend**: Adding new options doesn't break the signature
3. **Type-safe**: TypeScript ensures all options are valid

### Example Usage

```typescript
// Without agent name (existing behavior)
await buildSystemMessage(metadata, workspacePath);

// With agent name
await buildSystemMessage(metadata, workspacePath, {
  agentName: "CodeAssistant"
});

// Multiple options combined
await buildSystemMessage(metadata, workspacePath, {
  mode: "plan",
  agentName: "PlanBot",
  additionalSystemInstructions: "Be careful with security."
});
```

### Testing

- ✅ All 8 unit tests pass (including new tests for agent name)
- ✅ TypeScript type checking passes
- ✅ No breaking changes to existing code

_Generated with `cmux`_

### DIFF
--- a/src/services/systemMessage.test.ts
+++ b/src/services/systemMessage.test.ts
@@ -201,8 +201,7 @@ Special mode instructions.
     });
 
     // Should include the agent name in the prelude
-    expect(systemMessage).toContain("Your name is CodeAssistant.");
-    expect(systemMessage).toContain("You are a coding agent.");
+    expect(systemMessage).toContain("You are a coding agent named CodeAssistant.");
   });
 
   test("excludes agent name from prelude when not provided", async () => {
@@ -215,9 +214,9 @@ Special mode instructions.
 
     const systemMessage = await buildSystemMessage(metadata, workspaceDir);
 
-    // Should NOT include "Your name is"
-    expect(systemMessage).not.toContain("Your name is");
+    // Should use the standard phrasing without "named"
     expect(systemMessage).toContain("You are a coding agent.");
+    expect(systemMessage).not.toContain("named");
   });
 
   test("combines multiple options (mode, agentName, additionalSystemInstructions)", async () => {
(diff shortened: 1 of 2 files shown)


ammar-agent:
🤖 Closing as misguided.

After reflection, this PR adds infrastructure with no path to usage:

**What's missing:**
- No UI to set agent names
- No storage in WorkspaceMetadata
- No IPC methods to configure it
- aiService doesn't pass agentName through

**The feature is unreachable** - there's no code path that would ever populate the agent name.

**Design questions:**
- Should workspace name BE the agent name? (we already have `metadata.name`)
- Does telling an agent its name meaningfully change behavior?
- In parallel workspace scenarios, is this helpful or confusing?

**What was valuable:**
The options object refactor is good engineering, but it doesn't need the unused agent name feature attached to it.

**Path forward:**
If we want agents to know their context, we should:
1. Decide if workspace name suffices (e.g., "feature-refactor")
2. OR design a complete agent naming system with UI/storage/IPC
3. Then implement with clear use cases

Not merging incomplete infrastructure that adds unused code.

_Generated with `cmux`_

PR #353, opened 2025-10-20
Add `server` command to the entrypoint

kylecarbs:
This allows cmux to run as a web server.

Run `node dist/main.js server` to start on port 3000.

image

This directory select modal appears only when in browser mode for opening a new project. Everything else is the same.


src/components/DirectorySelectModal.tsx
ThomasK33, line 1: Would it make sense to add a visual storybook test for this component?

kylecarbs: probably i'll add later

PR #360, opened 2025-10-20
Add stories for App.tsx

kylecarbs:


src/App.stories.tsx
```diff
@@ -0,0 +1,646 @@
...
+
+    const AppWithChatMocks: React.FC = () => {
+      // Set up mock API only once per component instance (not on every render)
+      const initialized = useRef(false);
+      if (!initialized.current) {
+        const mockProjects = projects;
+        const mockWorkspaces = workspaces;
+
+        const mockApi: IPCApi = {
+          dialog: {
```
ThomasK33: Isn't this duplicating the entire mock API?
Would it make sense to use/refactor to use the `setupMockAPI`?

kylecarbs: will fix oopsies

PR #364, opened 2025-10-21
🤖 Fix npm package name to match published scope

kylecarbs:
## Problem

The npm publish workflow on `main` was failing with:
```
npm error 404 Not Found - PUT https://registry.npmjs.org/@coder%2fcmux - Not found
npm error 404  '@coder/cmux@0.3.0' is not in this registry.
```

This error message is misleading. The real issue is a **package name mismatch**:
- Package was published to npm as `@coder/cmux`
- Local `package.json` had `name: "cmux"` (unscoped)

When npm tried to publish, it couldn't match the local package name with the published one, causing confusion and failures.

## Solution

Update `package.json` to use the correct scoped name `@coder/cmux` to match what's already published on npm.

## Verification

```bash
$ curl -s https://registry.npmjs.org/@coder/cmux | jq '.name, .versions | keys'
"@coder/cmux"
[
  "0.3.0"
]
```

The package exists on npm as `@coder/cmux`, so our local package.json must match.

_Generated with `cmux`_

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "cmux",
+  "name": "@coder/cmux",
   "version": "0.3.0",
   "description": "cmux - coder multiplexer",
   "main": "dist/main.js",


kylecarbs:
Closing - this fix is already on main. The package.json already has the correct name `@coder/cmux`.

PR #405, opened 2025-10-24
🤖 Allow interrupt (Ctrl+C) in ChatInput when no text selected

ammar-agent:
## Problem

Currently, `Ctrl+C` (or `Cmd+C` on Mac) to interrupt the stream only works when focus is outside editable elements. This means users cannot interrupt while ChatInput is focused, requiring them to click elsewhere first.

## Solution

Modified the interrupt handler in `useAIViewKeybinds` to check for text selection:

- **No selection** → Interrupt stream (allows interrupting from ChatInput)
- **Has selection** → Browser handles copy (preserves existing behavior)

The logic now allows interrupt in editable elements as long as there's no active text selection, preventing conflict with the copy keybind.

## Changes

- Updated `useAIViewKeybinds.ts` interrupt handler to check `window.getSelection()`
- When in editable element with no selection: prevent default and interrupt
- When in editable element with selection: let browser handle copy

## Testing

Manual testing scenarios:
- ChatInput focused, no selection, Ctrl+C → should interrupt ✅
- ChatInput focused, text selected, Ctrl+C → should copy text ✅
- Other input focused, no selection, Ctrl+C → should interrupt ✅
- Not in input, Ctrl+C → should interrupt ✅

_Generated with `cmux`_

### DIFF
--- a/src/hooks/useAIViewKeybinds.ts
+++ b/src/hooks/useAIViewKeybinds.ts
@@ -69,8 +69,21 @@ export function useAIViewKeybinds({
         // Normal stream interrupt (non-compaction)
         // Allow interrupt in editable elements if there's no text selection
         // This way Ctrl+C works for both copy (when text is selected) and interrupt (when not)
-        const hasSelection = window.getSelection()?.toString().length ?? 0 > 0;
         const inEditableElement = isEditableElement(e.target);
+        let hasSelection = false;
+
+        if (inEditableElement) {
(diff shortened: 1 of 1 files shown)


ammar-agent:
The integration test failure is unrelated to this PR - it's a flaky test in `streamErrorRecovery.test.ts` that hit an API token limit (215k > 200k max). The test passes locally and the failure is due to test environment data, not the keyboard shortcut changes.

All other checks passing ✅

PR #440, opened 2025-10-27
Change shescape to use global PATH `bash`

kylecarbs:
This broke running cmux on NixOS


ammario:
Yeah sorry I was actually planning on removing that ridiculous path code

ammario:
nvm was thinking of something else

PR #447, opened 2025-10-27
🤖 Remove sync fs operations and path-finding from runtime

ammar-agent:
Eliminates blocking filesystem operations in the runtime layer and removes unnecessary path-finding complexity.

## Changes

**Removed sync fs operations:**
- Replaced `fs.existsSync()` + `fs.mkdirSync()` with `fsPromises.access()` + `fsPromises.mkdir()` in `LocalRuntime.ts`
- All filesystem operations in runtime are now async, preventing main thread blocking

**Removed path-finding module:**
- Deleted `executablePaths.ts` (57 lines) that was checking hardcoded paths for `bash` and `nice`
- Updated `LocalRuntime.ts` and `SSHRuntime.ts` to rely on system PATH
- Simplifies codebase by 74 lines total

## Verification

- ✅ TypeScript compilation passes
- ✅ All unit tests pass (same results as before)
- ✅ No sync fs operations remain in runtime files

_Generated with `cmux`_


ammar-agent:
All CI tests pass including integration and E2E tests, confirming that relying on system PATH works correctly in our actual runtime environments (CI, Electron, etc.). The path-finding code was unnecessary complexity.

PR #477, opened 2025-10-29
🤖 refactor: make command palette a workspace switcher by default

ammar-agent:
Makes the command palette primarily a **workspace switcher** by default, with all commands still accessible via prefix keys.

## Motivation

The command palette was showing all commands by default, but workspace switching is the most common use case. This change optimizes for the common case while keeping everything accessible.

### Behavior

- **Default** (no prefix): Shows only workspace commands (switch, create, rename, delete, etc.)
- **`>` prefix**: Shows all commands across all sections (Navigation, Chat, Modes, Projects, Help)
- **`/` prefix**: Shows slash commands for chat input (existing behavior)

### Code Quality

Took this opportunity to clean up surrounding code:

- **Extracted `resetPaletteState()` helper** - State reset logic was duplicated 4 times across the component
- **Exported `COMMAND_SECTIONS` constant** - Section names are now centralized and type-safe
- **Removed unused variable** - `searchQuery` was declared but never used
- **Added custom filter function** - Properly handles `>` prefix in cmdk filtering

### Documentation

- Added command palette section to `docs/keybinds.md` explaining the three modes
- Updated Storybook story with accurate feature descriptions
- Fixed section name inconsistency in Storybook (was "Workspace", now "Workspaces")

## Testing

- ✅ TypeScript type checking passes
- ✅ Prettier formatting passes
- ✅ Logic verified with unit test simulation
- ✅ All existing functionality preserved

## Before/After

**Before:**
- Opens with all commands visible
- User must scroll/search through Navigation, Chat, Mode, Project, Help sections to find workspaces

**After:**
- Opens with only workspace commands (the primary use case)
- Type `>` to instantly see all other commands when needed
- Faster workflow for the most common action

_Generated with `cmux`_

### DIFF
--- a/src/utils/commandIds.ts
+++ b/src/utils/commandIds.ts
@@ -4,11 +4,22 @@
  */
 
 /**
+ * Command ID prefixes for pattern matching
+ * Single source of truth for all dynamic ID patterns
+ */
+const COMMAND_ID_PREFIXES = {
+  WS_SWITCH: "ws:switch:",
+  CHAT_TRUNCATE: "chat:truncate:",
+  PROJECT_REMOVE: "project:remove:",
+} as const;
+
+/**
  * Command ID builders - construct IDs with consistent patterns
  */
 export const CommandIds = {
   // Workspace commands
-  workspaceSwitch: (workspaceId: string) => `ws:switch:${workspaceId}` as const,
+  workspaceSwitch: (workspaceId: string) =>
+    `${COMMAND_ID_PREFIXES.WS_SWITCH}${workspaceId}` as const,
   workspaceNew: () => "ws:new" as const,
   workspaceNewInProject: () => "ws:new-in-project" as const,
   workspaceRemove: () => "ws:remove" as const,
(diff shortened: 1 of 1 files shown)


ammar-agent:
## Refinement: Only Switching, No Mutations

Refined the default filter to be even more focused - now shows **only workspace switching commands** (`ws:switch:*`), excluding mutations like:

- ❌ Create New Workspace
- ❌ Delete Workspace  
- ❌ Rename Workspace
- ❌ Open in Terminal

These management commands are still accessible via `>` prefix.

**Rationale:**
- Workspace switching is the primary use case
- Mutations should require explicit intent (via `>`)
- Keeps the default palette extremely focused and fast
- Matches user expectations for a "switcher"

**Testing:**
```typescript
// Default view: Only ws:switch:* commands
testFiltering("") // → ["Switch to Workspace A", "Switch to Workspace B"] ✅

// With > prefix: All commands including mutations
testFiltering(">") // → All 8+ commands including create/delete/rename ✅
```

ammar-agent:
Fixed the cmdk filtering issue for `>` queries in commit bb8102cb. The custom filter now runs and properly strips the `>` prefix when searching commands. 👍

ammario:
fyi @ThomasK33 — I think this is a good compromise

PR #482, opened 2025-10-29
🤖 feat: allow cmux server port override

ibetitsmike:
## Summary
- allow the CLI `cmux server` command to accept an optional --port flag
- reuse shared parsing logic and use the resolved value when starting the server
- cover the port parsing helper with unit tests

## Testing
- make typecheck
- bun test src/utils/cli/serverPort.test.ts

_Generated with `cmux`_

### DIFF
--- a/src/utils/cli/serverPort.ts
+++ b/src/utils/cli/serverPort.ts
@@ -1,0 +1,47 @@
+export const PORT_FLAG = "--port";
+
+export type PortOptionResult =
+  | { kind: "none" }
+  | { kind: "ok"; port: number }
+  | { kind: "error"; message: string };
+
+const PORT_PREFIX = `${PORT_FLAG}=`;
+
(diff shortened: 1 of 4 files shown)


src/utils/cli/serverPort.ts
```diff
@@ -0,0 +1,47 @@
+export const PORT_FLAG = "--port";
+
+export type PortOptionResult =
+  | { kind: "none" }
+  | { kind: "ok"; port: number }
+  | { kind: "error"; message: string };
+
+const PORT_PREFIX = `${PORT_FLAG}=`;
+
+export const parsePortOption = (values: string[]): PortOptionResult => {
```
ammario, line 10: We use minimist elsewhere could re-use that here.

ibetitsmike: Makes sense. Will do.

ibetitsmike:
will be added in another PR

PR #485, opened 2025-10-29
🤖 fix: prevent infinite retry loop for non-retryable errors

ammar-agent:
## Problem

When users hit non-retryable errors (e.g., `api_key_not_found`, `authentication`, `quota`), the system would infinitely retry without user intervention. UI showed "Retrying..." indefinitely with no error message, and the manual retry button was non-functional.

## Root Causes

**Auto-retry didn't distinguish error types:** Two error systems existed (`SendMessageError` from resumeStream failures, `StreamErrorType` in stream-error messages). `isEligibleForAutoRetry()` only checked the latter, missing `api_key_not_found` errors stored in `RetryState.lastError`.

**RetryBarrier didn't receive updates:** `useResumeManager` used `localStorage.setItem()` directly while `RetryBarrier` used `usePersistedState` with `listener: true` expecting custom events.

**Manual retry was blocked:** `isEligibleForResume()` treated manual retries the same as auto-retries, blocking user intent after fixing issues.

## Solution

**Split retry eligibility logic** into three functions in `src/utils/messages/retryEligibility.ts`:
- `hasInterruptedStream()` - Shows UI for ALL errors (user may have fixed it)
- `isEligibleForAutoRetry()` - Filters non-retryable stream errors (authentication, quota, model_not_found, context_exceeded, aborted)
- `isNonRetryableSendError()` - Checks SendMessageError types (api_key_not_found, provider_not_supported, invalid_model_string)

**Fixed event synchronization:** `useResumeManager` now uses `updatePersistedState()` which dispatches custom events that listeners receive.

**Made RetryBarrier self-contained:** Following `AgentStatusIndicator` pattern, accepts only `workspaceId` prop, fetches workspace state directly, manages own `autoRetry` state, computes `effectiveAutoRetry` internally. Removed ~37 lines from AIView.tsx.

**Added manual retry bypass:** `RESUME_CHECK_REQUESTED` events now include `isManual` flag. Manual retry bypasses eligibility checks, respecting user intent.

**Centralized event type system:** Created `CustomEventPayloads` interface and `createCustomEvent()` helper in `src/constants/events.ts` for type-safe event handling.

**Centralized error formatting:** Created `formatSendMessageError()` in `src/utils/errors/formatSendError.ts`, eliminating duplicate error message strings.

## Behavior After Fix

**Non-retryable errors:** Auto-retry stops immediately, UI shows manual "Retry" button (not "Retrying..."), error message displayed clearly, user can fix and retry.

**Retryable errors:** Auto-retry with exponential backoff (1s → 2s → 4s → ... → 60s max), UI shows countdown, user can stop with Ctrl+C.

## Testing

All 30 tests in `retryEligibility.test.ts` pass, verifying correct filtering of non-retryable errors and edge cases.

_Generated with `cmux`_


ibetitsmike:
Nice. Had that happen today.

PR #500, opened 2025-11-02
🤖 feat: AI-generated workspace creation on first message

kylecarbs:
Changes the "New Workspace" modal to instead be inline in the chat form.

https://github.com/user-attachments/assets/a8c36701-5507-4c53-b9c9-4afbda43cbe8

### DIFF
--- a/src/services/workspaceTitleGenerator.ts
+++ b/src/services/workspaceTitleGenerator.ts
@@ -64,8 +64,8 @@ Both should be concise (2-5 words) and descriptive of the task.`,
 }
 
 /**
- * Get model for title generation - uses the user's selected model
- * This ensures we don't need to manually update when providers release new models
+ * Get model for title generation - prefers fast/cheap models
+ * Priority: Haiku 4.5 (Anthropic) → GPT-5-mini (OpenAI) → fallback to user's model
  * Falls back to null if no provider configured
  */
 function getModelForTitleGeneration(modelString: string, config: Config): LanguageModel | null {
@@ -76,7 +76,23 @@ function getModelForTitleGeneration(modelString: string, config: Config): Langua
   }
 
   try {
-    // Parse user's model string
+    // Try Anthropic Haiku first (fastest/cheapest)
+    if (providersConfig.anthropic?.apiKey) {
+      const provider = createAnthropic({
+        apiKey: String(providersConfig.anthropic.apiKey),
+      });
+      return provider("claude-haiku-4-5");
+    }
+
+    // Try OpenAI GPT-5-mini second
+    if (providersConfig.openai?.apiKey) {
+      const provider = createOpenAI({
+        apiKey: String(providersConfig.openai.apiKey),
+      });
+      return provider("gpt-5-mini");
+    }
+
+    // Parse user's model as fallback
     const [providerName, modelId] = modelString.split(":", 2);
     if (!providerName || !modelId) {
       log.error("Invalid model string format:", modelString);


src/browser/api.ts
```diff
@@ -214,6 +214,8 @@ const webApi: IPCApi = {
       invokeIPC(IPC_CHANNELS.WORKSPACE_FORK, sourceWorkspaceId, newName),
     sendMessage: (workspaceId, message, options) =>
```
ammario, line 222: can we make workspaceId optional here? Then we reduce duplication in the IPC.

src/components/FirstMessageInput.tsx
```diff
@@ -0,0 +1,133 @@
...
+ * 3. Switches to the new workspace (via callback)
+ */
+export function FirstMessageInput({ projectPath, onWorkspaceCreated }: FirstMessageInputProps) {
+  const [input, setInput] = useState("");
+  const [isSending, setIsSending] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const inputRef = useRef<HTMLTextAreaElement>(null);
+
+  // Get most recent model from LRU (no workspace-specific model yet)
+  const { recentModels } = useModelLRU();
```
ammario: we should probably use the standard useSendMessage options

src/services/workspaceTitleGenerator.ts
```diff
@@ -0,0 +1,122 @@
...
+    log.error("Failed to generate workspace names with AI, using fallback", error);
+    return createFallbackNames();
+  }
+}
+
+/**
+ * Select a cheap model for title generation
+ * Prefers Haiku > GPT-4o-mini > Gemini Flash > null
+ */
+function selectCheapModel(config: Config): LanguageModel | null {
```
ammario: I think we should use the model as provided to send message, bc, for example, this might break those that using a custom BaseURL with unexpected model names available. Also, we want to minimize the number of places we need to manually update when a new model comes out.

src/App.tsx
```diff
@@ -653,6 +655,25 @@ function AppInner() {
                   }
                 />
               </ErrorBoundary>
+            ) : projects.size === 1 ? (
```
ammario: Don't we wanna show this as the default path when creating a new workspace in a project? Like, even when you have existing workspaces / projects?

ammario (changes requested):

src/components/FirstMessageInput.tsx
```diff
@@ -0,0 +1,381 @@
...
+        void handleSend();
+      }
+      // Cancel on Escape
+      if (e.key === "Escape" && onCancel) {
+        e.preventDefault();
+        onCancel();
+      }
+    },
+    [handleSend, onCancel]
+  );
```
ammario, line 199: Seems like a lot of duplication with the regular message input

src/components/FirstMessageInput.tsx
```diff
@@ -0,0 +1,381 @@
...
+              )}
+              <TooltipWrapper inline>
+                <span className="text-muted cursor-help text-xs">?</span>
+                <Tooltip className="tooltip" align="center" width="wide">
+                  <strong>Runtime:</strong>
+                  <br />
+                  • Local: git worktree in ~/.cmux/src
+                  <br />• SSH: remote clone in ~/cmux on SSH host
+                </Tooltip>
+              </TooltipWrapper>
```
ammario, line 336: Again, a lot of duplication

src/contexts/ModeContext.tsx
```diff
@@ -10,12 +10,20 @@ type ModeContextType = [UIMode, (mode: UIMode) => void];
...
 
-export const ModeProvider: React.FC<ModeProviderProps> = ({ workspaceId, children }) => {
-  const [mode, setMode] = usePersistedState<UIMode>(getModeKey(workspaceId), "exec", {
+export const ModeProvider: React.FC<ModeProviderProps> = ({
+  workspaceId,
+  projectPath,
+  children,
+}) => {
+  // Priority: workspace-scoped > project-scoped
+  const scopeId = workspaceId ?? (projectPath ? `__project__${projectPath}` : "__global__");
```
ammario: should use a delimiter like "/" that can't appear in a path

src/services/ipcMain.ts
```diff
@@ -692,10 +693,170 @@ export class IpcMain {
...
+              logStdout: (line: string) => {
+                this.initStateManager.appendOutput(workspaceId, line, false);
+              },
+              logStderr: (line: string) => {
+                this.initStateManager.appendOutput(workspaceId, line, true);
+              },
+              logComplete: (exitCode: number) => {
+                void this.initStateManager.endInit(workspaceId, exitCode);
+              },
+            };
```
ammario: I think this pattern is duplicated often

src/services/ipcMain.ts
```diff
@@ -692,10 +693,170 @@ export class IpcMain {
...
+              metadata: completeMetadata,
+            };
+          } catch (error) {
+            const errorMessage = error instanceof Error ? error.message : String(error);
+            log.error("Unexpected error in sendMessage handler (lazy creation):", error);
+            return { success: false, error: `Failed to create workspace: ${errorMessage}` };
+          }
+        }
+
+        // Normal path: workspace already exists
```
ammario, line 897: bonus points for decomposing this fn as its pretty big

src/services/workspaceTitleGenerator.ts
```diff
@@ -0,0 +1,146 @@
...
+      });
+      return provider("claude-haiku-4-5");
+    }
+
+    // Try OpenAI GPT-4o-mini second
+    if (providersConfig.openai?.apiKey) {
+      const provider = createOpenAI({
+        apiKey: String(providersConfig.openai.apiKey),
+      });
+      return provider("gpt-4o-mini");
```
ammario: I would rather we always use to the user's model. E.g. this code will quickly fall out of date as providers update their latest mini models.

src/services/ipcMain.ts
```diff
@@ -692,10 +693,170 @@ export class IpcMain {
...
+          }
+
+          log.debug("sendMessage handler: Creating workspace for first message", {
+            projectPath: options.projectPath,
+            messagePreview: message.substring(0, 50),
+          });
+
+          try {
+            // 1. Generate workspace title and branch name using AI (use same model as message)
+            const { title, branchName } = await generateWorkspaceNames(
```
ammario: We should probably go through this code path whenever we lack a saved branch or title for a workspace to be robust to errors / restarts.

PR #506, opened 2025-11-03
🤖 fix: simplify worker path resolution using __dirname

ThomasK33:
Replaces complex path manipulation logic with simple `__dirname` usage. Both `workerPool.js` and `tokenizer.worker.js` compile to the same directory (`dist/utils/main/`), so we can use `__dirname` directly.

This removes platform-specific path handling and makes the code more maintainable.

_Generated with `cmux`_


ThomasK33:
@codex review

PR #510, opened 2025-11-04
🤖 feat: add workspace scripts with discovery and execution

ThomasK33:
Adds **Workspace Scripts** to run custom tasks directly from the chat.

### Features
- **Execution**: Run executable files from `.mux/scripts/` using `/script <name>` or `/s`.
- **Remote-Ready**: Works transparently on local and SSH workspaces.
- **Rich Feedback**: Shows a private execution card in chat (stdout/stderr, duration) and a status toast.
- **Agent Workflows**:
  - **Custom Toasts**: Scripts can write Markdown to `$MUX_OUTPUT` to customize the completion notification.
  - **Auto-Prompt**: Scripts can write to `$MUX_PROMPT` to trigger an immediate follow-up request to the agent (e.g., "Fix these lint errors").

Includes documentation in `docs/scripts.md`.

_Generated with `mux`_

### DIFF
--- a/src/browser/App.stories.tsx
+++ b/src/browser/App.stories.tsx
@@ -85,6 +85,12 @@ function setupMockAPI(options: {
           success: true,
           data: { success: true, output: "", exitCode: 0, wall_duration_ms: 0 },
         }),
+      listScripts: () => Promise.resolve({ success: true, data: [] }),
+      executeScript: () =>
+        Promise.resolve({
+          success: true,
+          data: { success: true, output: "", exitCode: 0, wall_duration_ms: 0 },
+        }),
     },
     projects: {
       list: () => Promise.resolve(Array.from(mockProjects.entries())),
@@ -1255,6 +1261,12 @@ main
                   data: { success: true, output: "", exitCode: 0, wall_duration_ms: 0 },
                 });
               },
+              listScripts: () => Promise.resolve({ success: true, data: [] }),
+              executeScript: () =>
+                Promise.resolve({
+                  success: true,
+                  data: { success: true, output: "", exitCode: 0, wall_duration_ms: 0 },
+                }),
             },
           },
         });
@@ -1459,6 +1471,12 @@ These tables should render cleanly without any disruptive copy or download actio
               replaceChatHistory: () => Promise.resolve({ success: true, data: undefined }),
               getInfo: () => Promise.resolve(null),
               executeBash: () =>
+                Promise.resolve({
+                  success: true,
+                  data: { success: true, output: "", exitCode: 0, wall_duration_ms: 0 },
+                }),
+              listScripts: () => Promise.resolve({ success: true, data: [] }),
+              executeScript: () =>
                 Promise.resolve({
                   success: true,
                   data: { success: true, output: "", exitCode: 0, wall_duration_ms: 0 },
--- a/src/browser/api.ts
+++ b/src/browser/api.ts
@@ -274,6 +274,10 @@ const webApi: IPCApi = {
     executeBash: (workspaceId, script, options) =>
       invokeIPC(IPC_CHANNELS.WORKSPACE_EXECUTE_BASH, workspaceId, script, options),
     openTerminal: (workspaceId) => invokeIPC(IPC_CHANNELS.WORKSPACE_OPEN_TERMINAL, workspaceId),
+    listScripts: (workspaceId) =>
+      invokeIPC(IPC_CHANNELS.WORKSPACE_LIST_SCRIPTS, workspaceId),
+    executeScript: (workspaceId, scriptName, args) =>
+      invokeIPC(IPC_CHANNELS.WORKSPACE_EXECUTE_SCRIPT, workspaceId, scriptName, args),
     activity: {
       list: async (): Promise<Record<string, WorkspaceActivitySnapshot>> => {
         const response = await invokeIPC<Record<string, unknown>>(
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -1,5 +1,5 @@
 import type { FrontendWorkspaceMetadata } from "@/common/types/workspace";
-import type { WorkspaceChatMessage } from "@/common/orpc/types";
+import type { WorkspaceChatMessage } from "@/common/types/ipc";
 import { createMuxMessage } from "@/common/types/message";
 import type { BashToolResult } from "@/common/types/tools";
 import { DEFAULT_RUNTIME_CONFIG } from "@/common/constants/workspace";
@@ -43,14 +43,8 @@ global.window = mockWindow as unknown as Window & typeof globalThis;
 // Mock dispatchEvent
 global.window.dispatchEvent = jest.fn();
 
-// Helper to get IPC callback in a type-safe way
-function getOnChatCallback<T = { type: string }>(): (data: T) => void {
-  const mock = mockWindow.api.workspace.onChat as jest.Mock<
-    () => void,
-    [string, (data: T) => void]
-  >;
-  return mock.mock.calls[0][1];
-}
+// Reference to mock for easier access
+const mockOnChat = mockWindow.api.workspace.onChat as jest.Mock;
 
 // Helper to create and add a workspace
 function createAndAddWorkspace(
@@ -72,36 +66,24 @@ function createAndAddWorkspace(
 }
 
 // Helper to get callback from mock for pushing messages
-let pendingMessages: WorkspaceChatMessage[] = [];
-let resolvers: Array<(msg: WorkspaceChatMessage) => void> = [];
 
-function getOnChatCallback<T extends WorkspaceChatMessage>(): (msg: T) => void {
-  return (msg: T) => {
-    if (resolvers.length > 0) {
-      const resolver = resolvers.shift()!;
-      resolver(msg);
-    } else {
-      pendingMessages.push(msg);
-    }
-  };
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function getOnChatCallback<T = any>(): (msg: T) => void {
+  if (!currentChatCallback) {
+    throw new Error("No chat callback registered - was addWorkspace called?");
+  }
+  return currentChatCallback as (msg: T) => void;
 }
 
-// Set up mock to use push-based message queue
-mockOnChat.mockImplementation(async function* (): AsyncGenerator<
-  WorkspaceChatMessage,
-  void,
-  unknown
-> {
-  while (true) {
-    if (pendingMessages.length > 0) {
-      yield pendingMessages.shift()!;
-    } else {
-      const msg = await new Promise<WorkspaceChatMessage>((resolve) => {
-        resolvers.push(resolve);
-      });
-      yield msg;
-    }
-  }
+// Track current chat callback for tests to push messages
+let currentChatCallback: ((msg: WorkspaceChatMessage) => void) | null = null;
+
+// Set up mock to capture the callback and allow tests to push messages
+mockOnChat.mockImplementation((_workspaceId: string, callback: (msg: WorkspaceChatMessage) => void) => {
+  currentChatCallback = callback;
+  return () => {
+    currentChatCallback = null;
+  };
 });
 
 describe("WorkspaceStore", () => {
@@ -112,8 +94,7 @@ describe("WorkspaceStore", () => {
     jest.clearAllMocks();
     mockExecuteBash.mockClear();
     mockOnChat.mockClear();
-    pendingMessages = [];
-    resolvers = [];
+    currentChatCallback = null;
     mockOnModelUsed = jest.fn();
     store = new WorkspaceStore(mockOnModelUsed);
   });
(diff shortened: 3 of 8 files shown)


ThomasK33:
@codex addressed that feedback by provisioning CMUX_OUTPUT/CMUX_PROMPT via runtime temp directories and adding runtimeScriptExecution integration test.

ThomasK33:
@codex address that feedback

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

src/browser/utils/messages/modelMessageTransform.ts
```diff
@@ -198,6 +198,37 @@ export function injectModeTransition(
...
+    const scriptMeta = msg.metadata.muxMetadata;
+    const exitCode = scriptMeta.result.exitCode;
+    const output = scriptMeta.result.output ?? "(no output)";
+
+    // Construct a concise representation for the LLM
+    const llmContent = `Script '${scriptMeta.scriptName}' executed (exit code ${exitCode}).\nStdout/Stderr:\n${output}`;
+
+    return {
+      ...msg,
+      parts: [{ type: "text", text: llmContent }],
```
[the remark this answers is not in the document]
ThomasK33: No, they should stay there. The UI needs to be updated to remove that.

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
I've reverted the inclusion of MUX_OUTPUT and MUX_PROMPT for `script-execution` messages (user-initiated via /script). These are handled by ChatInput.tsx to avoid duplication. Note that tool calls (agent-initiated) use a different path (tools.ts) which correctly includes all output.

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
Addressed review feedback:

- **Script Overflow Policy:** Implemented `overflowPolicy` in `runWorkspaceScript`. LLM tools now use `tmpfile` to capture full output, while IPC calls default to `truncate`.
- **Path Handling:** Switched to `runtime.normalizePath` in `scriptRunner` to support cross-platform paths (e.g. POSIX paths on SSH from Windows host). Added regression tests.
- **Stale Suggestions:** Added `useAvailableScripts` hook to immediately clear script suggestions on workspace switch.
- **Verification:** Validated LLM error messaging and failure persistence logic via existing and new tests.

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
This security concern has already been addressed. The `resolvePath` method in `LocalRuntime.ts` (lines 297-309) uses `fsPromises.realpath(absolute)` to canonicalize symlinks, which prevents the symlink escape scenario described in the review.

The implementation explicitly resolves symlinks before performing the path containment check, so a symlink pointing outside the scripts directory will be resolved to its canonical path and fail the `startsWith` check.

ThomasK33:
@codex review

ThomasK33:
@codex review

src/common/utils/tools/tools.ts
```diff
@@ -73,6 +78,80 @@ export async function getToolsForModel(
...
+  // Discover and register user scripts as tools
+  // These are treated as runtime tools (execution happens in runtime)
+  try {
+    const scripts = await listScripts(config.runtime, config.cwd);
+    for (const script of scripts) {
+      if (!script.isExecutable) continue;
+
+      // Sanitize script name for tool name (e.g., "deploy-prod" -> "script_deploy_prod")
+      const sanitizedName = script.name.replace(/[^a-zA-Z0-9_]/g, "_");
+      const toolName = `script_${sanitizedName}`;
```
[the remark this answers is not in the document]
ThomasK33: Not an issue for now. if this truly happens, we fix that in the future

.cmux/scripts/demo
```diff
@@ -0,0 +1,28 @@
...
+# Description: Demo script to showcase the script execution feature. Accepts no arguments.
+set -euo pipefail
+
+# Regular output goes to stdout (visible in console logs)
+echo "Running demo script..."
+echo "Current workspace: $(pwd)"
+echo "Timestamp: $(date)"
+
+# Write formatted output to MUX_OUTPUT for toast display
+cat >>"$MUX_OUTPUT" <<'EOF'
```
ammario: This naming is a bit confusing. If this is just for toast display, why not call it MUX_TOAST? Also, is this feature necessary for MVP?

.cmux/scripts/demo
ammario, line 1: Should we symlink `scripts -> .mux/scripts` in our repo? I assume that should be standard practice for any mux heavy repo.

.cmux/scripts/echo
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+# Description: Echo arguments demo. Accepts any number of arguments (strings) which will be echoed back.
+set -euo pipefail
+
+# Check if arguments were provided
+if [ $# -eq 0 ]; then
+  cat >>"$MUX_OUTPUT" <<'EOF'
```
ammario: Scripts need to be updated for std io approach

.cmux/scripts/wait_pr_checks
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+# Description: Wait for PR checks to pass on GitHub. Use this after pushing changes to origin, to catch CI failures. Accepts no arguments.
+set -euo pipefail
+
+BRANCH=$(git branch --show-current)
+NUMBER=$(gh pr list --head "$BRANCH" --json number | jq -cr '.[0].number')
+./scripts/wait_pr_checks.sh "$NUMBER"
```
ammario, line 7: Again, would be neat if we could seamlessly link scripts and they behaved well under both execution paradigms

docs/SUMMARY.md
```diff
@@ -13,6 +13,7 @@
   - [SSH](./ssh.md)
   - [Forking](./fork.md)
   - [Init Hooks](./init-hooks.md)
+  - [Workspace Scripts](./scripts.md)
```
ammario, line 16: "Scripts" is sufficient for name

docs/scripts.md
```diff
@@ -0,0 +1,187 @@
+# Workspace Scripts
+
+Execute custom scripts from your workspace using slash commands or let the AI Agent run them as tools.
+
+## Overview
+
+Scripts are stored in `.mux/scripts/` within each workspace. They serve two purposes:
```
ammario, line 7: Should consider using our standard paradigm of coalescing ~/.mux with project .mux. Should also consider suggesting symlink here.

docs/scripts.md
```diff
@@ -0,0 +1,187 @@
...
+   # Description: Deploy to staging. Accepts one optional argument: 'dry-run' to simulate.
+
+   if [ "${1:-}" == "dry-run" ]; then
+     echo "Simulating deployment..."
+   else
+     echo "Deploying to staging..."
+   fi
+   ```
+
+   **Crucial**: The `# Description:` line is what the AI reads to understand the tool. Be descriptive about what the script does and what arguments it accepts.
```
ammario, line 37: This is interesting, but it's a little bit unclear how it would work for non-bash scripts, which should definitely be supported. 

One idea is calling `--help` on all scripts upon load, but it would be cumbersome to implement for users. Another idea is automatically including the first N lines of the script, but this is "magical" and would fail for binary scripts.

Some more ideas:

* A `manifest.jsonc` in the scripts dir containing a mapping of scripts to descriptions
* A standard `<script>.help` file that we look at for description

We could also implement comment parsing for common languages. 

Nothing clean comes to mind unfortunately.

docs/scripts.md
```diff
@@ -0,0 +1,187 @@
...
+
+3. **Make it executable**:
+
+   ```bash
+   chmod +x .mux/scripts/deploy
+   ```
+
+## Agent Integration (AI Tools)
+
+Every executable script in `.mux/scripts/` is automatically registered as a tool for the AI Agent.
```
ammario, line 47: Curious for rationale of this vs. using a single `script` tool with the scripts registered in the description. Historically, tool call performance has suffered as the agent is provided with too many tools. This may be different now.

Another issue with a tool for each script is a lot of duplicate token usage for tool description. For example, each script has the same input and output parameters (argv, stdout, exit_code).

docs/scripts.md
```diff
@@ -0,0 +1,187 @@
...
+
+Arguments are passed directly to the script as `$1`, `$2`, etc.
+
+## Execution Context
+
+Scripts run with:
+
+- **Working directory**: The workspace directory.
+- **Environment**: Full workspace environment + project secrets + special cmux variables.
+- **Timeout**: 5 minutes by default.
```
ammario, line 93: this would fail for wait_pr_checks generally - suggest not having a timeout or making it longer. User should be able to interrupt hanging script anyways. we should also eventually loop this into our background bash system so that the agent can poll output and do useful work while waiting + cancel an apparently hanging script.

docs/scripts.md
ammario: E.g. i believe we have a .ts script for updating models. On iPad rn so can't easily check.

docs/scripts.md
```diff
@@ -0,0 +1,187 @@
...
+#!/usr/bin/env bash
+# Description: Fetch a URL. Accepts exactly one argument: the URL.
+set -euo pipefail
+
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <url>"
+    exit 1
+fi
+curl -sL "$1"
+```
```
ammario, line 169: Since I added a web_fetch tool directly we should remove this example to avoid confusion. Unfortunately web_fetch is not well represented as a bash script but could be as a typescript script since we want to filter out a lot of the noise in the HTML and extract just the textual content.

docs/scripts.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Script Discovery
+
+- Scripts are discovered automatically from `.mux/scripts/` in the current workspace.
+- Discovery is cached for performance but refreshes intelligently.
```
ammario, line 174: Refresh model should be more clearly explained. E.g. should the user expect any edits to the scripts to be immediately reflected or do they have to wait for some timer?

Also, it's not totally clear whether we load scripts from the local project dir or the workspace dir. This determines whether we can easily plug into inotify (projects are always local and we can bypass runtime interface)

docs/scripts.md
```diff
@@ -0,0 +1,187 @@
...
+
+**Script not appearing in suggestions or tools?**
+
+- Ensure file is executable: `chmod +x .mux/scripts/scriptname`
+- Verify file is in `.mux/scripts/` directory.
+- Check for valid description header.
+
+**Agent using script incorrectly?**
+
+- Improve the `# Description:` header. Explicitly tell the agent what arguments to pass.
```
ammario, line 187: We could define extraction as the first non-shebang comment. IMO it should include multi-line comments.

ammario (changes requested):

.cmux/scripts/demo
ThomasK33: Oh, I haven't updated the script.  
There is no more MUX_OUTPUT and MUX_PROMPT, it's all stdout/stderr based now.

.cmux/scripts/demo
ThomasK33: Yeah, I think that makes sense.

.cmux/scripts/wait_pr_checks
ThomasK33: That makes sense. I'll update the ./scripts/wait_pr_checks.sh script to fetch the branch and number, like in this script, when they're not specified.

docs/SUMMARY.md
ThomasK33: Agreed. Since I couldn't think of a better name, I called it 'scripts' for now.
I'm open to suggestions for a new or improved name.

Alternatively, we could rename the page to 'Extensions' and include a section for scripts.

docs/scripts.md
ThomasK33: I agree that we should support both binaries and non-bash scripts.

For Python and TypeScript, this approach should work well, as we can read these files as text. Since our files include a shebang at the top, it's also feasible to add a description right below the shebang. So the current approach should also work for the .ts script for updating models.

For binaries, we can introduce a new heuristic in the future: obtain descriptions using the `--help` flag or `$CMD help`, and incorporate that information into the script's description.

PR #512, opened 2025-11-06
🤖 fix: build dist/ before running terminal-bench

ammar-agent:
## Problem

PR #507 added `dist/` to the terminal-bench archive include paths to fix worker crashes. However, the workflow wasn't building `dist/` before running the benchmark, causing all tasks to fail immediately with:

```
Error running agent for task <name>: Required file /home/runner/work/cmux/cmux/dist missing
```

## Solution

Add `make build` step before `make benchmark-terminal` in the workflow. This ensures:
- `dist/` directory exists 
- Compiled JavaScript including worker files are present
- Archive creation succeeds

## Testing

Manually triggered workflow run #19108392805 which failed with this exact error. With this fix, the workflow should build dist/ successfully and allow terminal-bench tasks to run.

_Generated with `cmux`_

### DIFF
--- a/.github/workflows/terminal-bench.yml
+++ b/.github/workflows/terminal-bench.yml
@@ -4,34 +4,34 @@ on:
   workflow_call:
     inputs:
       model_name:
-        description: "Model to use (e.g., anthropic:claude-sonnet-4-5)"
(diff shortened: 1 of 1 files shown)


ammar-agent:
Closing to create fresh PR without stale review comments

PR #528, opened 2025-11-07
🤖 fix: prevent Bun from exiting when stdin closes in agentSessionCli

ammar-agent:
## Problem

Terminal-bench was failing with all tasks showing 0 input/output tokens because the agent was exiting immediately after receiving the user message, without making any API calls.

**Symptoms:**
- Latest nightly runs: All 80 tasks failed with `agent_timeout`
- Agent ran for only ~45 seconds then exited
- `total_input_tokens: 0`, `total_output_tokens: 0`
- Stream started with `caught-up` and `user` message, but no `stream-delta` or `stream-end` events

**Root cause:**
The `agentSessionCli.ts` reads the user message from stdin via a pipe:
```bash
printf '%s' "$instruction" | bun src/debug/agentSessionCli.ts ...
```

Once stdin reaches EOF and is consumed, Bun detects no other active handles keeping the event loop alive and exits the process, **even though async work (API streaming) is still pending**.

## Solution

Add an explicit keepalive interval that ensures the process stays alive until `main()` completes. The interval runs far into the future (1000 seconds) but gets cleared in the finally block once the agent session finishes.

## Testing

**Before fix:**
- Run 19173435224: 1 task, 0 tokens, ~2 min total (agent ran 45s)
- Agent exited immediately after user message

**After fix:**
- Run 19173548174: 1 task, **resolved: true**, ~7 min total (agent ran 3m17s)
- 22 tool calls made
- Stream-delta events present
- Agent completed successfully

## Related

- Fixes nightly terminal-bench failures from Nov 5-7
- Related to PR #507 (dist/ in archive) and PR #513 (build step in workflow)

_Generated with `cmux`_

### DIFF
--- a/src/debug/agentSessionCli.ts
+++ b/src/debug/agentSessionCli.ts
@@ -488,7 +488,10 @@ async function main(): Promise<void> {
   }
 }
 
-void (async () => {
+// Keep process alive explicitly - Bun may exit when stdin closes even if async work is pending
+const keepAliveInterval = setInterval(() => {}, 1000000);
+
+(async () => {
   try {
     await main();
   } catch (error) {
@@ -507,5 +510,7 @@ void (async () => {
       process.stderr.write(`Error: ${message}\n`);
     }
     process.exitCode = 1;
+  } finally {
+    clearInterval(keepAliveInterval);
   }
 })();


ammar-agent:
Closing in favor of cleaner PR from main without merge conflicts.

PR #540, opened 2025-11-09
🤖 fix: add cross-platform path handling for Windows compatibility

ibetitsmike:
Adds comprehensive cross-platform path handling to support Windows builds alongside Unix systems.

## Changes

- **New PlatformPaths utility class** - Single source of truth for all path operations
  - Handles Windows drive letters (C:\) and UNC paths (\\server\share)
  - Correctly splits paths using both forward and backslash separators
  - Platform-aware home directory expansion (~/ on Unix, %USERPROFILE% on Windows)
  - Fish-style path abbreviation that works on both platforms

- **Replaced all direct path operations** throughout the codebase:
  - `path.basename()` → `PlatformPaths.basename()`
  - `expandTilde()` → `PlatformPaths.expandHome()`
  - `abbreviatePath()` → `PlatformPaths.abbreviate()`
  - Manual path splitting → `PlatformPaths.parse()`

- **Comprehensive test coverage** - Tests verify correct behavior on both platforms

## Testing

- All existing tests pass
- New PlatformPaths tests verify Windows and Unix path handling
- Type checking passes

_Generated with `mux`_

### DIFF
--- a/src/utils/paths.main.ts
+++ b/src/utils/paths.main.ts
@@ -235,5 +235,3 @@ export class PlatformPaths {
     return this.basename(projectPath) || "unknown";
   }
 }
-
-


src/runtime/SSHRuntime.ts
```diff
@@ -25,12 +24,20 @@ import { getProjectName } from "../utils/runtime/helpers";
...
  */
-const shescape = new Shescape({ shell: "bash" });
+const shescape = {
+  quote(value: unknown): string {
+    const s = String(value);
+    if (s.length === 0) return "''";
+    // Use POSIX-safe pattern to embed single quotes within single-quoted strings
+    return "'" + s.replace(/'/g, "'\"'\"'") + "'";
+  },
+};
```
deansheather, line 40: Did the quoting rules somehow change? I would expect shescape to be able to do any bash escaping.

src/utils/paths.ts
```diff
@@ -0,0 +1,311 @@
...
+
+    // Abbreviate all segments to first character
+    const abbreviated = segments.map((seg) => (seg.length > 0 ? seg[0] : seg));
+
+    // Reconstruct path - handle root separately to avoid double separator
+    if (!root && abbreviated.length === 0) {
+      return basename;
+    }
+
+    const sep = filePath.includes("\\") ? "\\" : "/";
```
deansheather: Technically on unix a backslash can be in a filename and doesn't count as a separator.

src/utils/paths.ts
```diff
@@ -0,0 +1,311 @@
...
+      return filePath;
+    }
+
+    const home = getHomeDir();
+    if (!home) {
+      return filePath;
+    }
+
+    // On Unix, replace home with tilde
+    // On Windows, show full path (no tilde convention)
```
deansheather: Why not? Tilde works in powershell

src/runtime/SSHRuntime.ts
ibetitsmike: Shescape needs bash to be available on the local machine - for windows it's unable to use bash.exe. It only handles PS and cmd.

PR #557, opened 2025-11-12
🤖 fix: support SSH workspaces in Open Terminal button

ethanndickson:
Fixes the "Open Terminal" button for SSH workspaces, which was failing with ENOENT errors because it tried to open a local path that doesn't exist locally.

## Changes

The IPC handler now accepts `workspaceId` instead of `workspacePath`, looks up the workspace metadata, and spawns the appropriate terminal:

- **SSH workspaces**: Spawns a local terminal that runs `ssh -t <host> 'cd <path> && exec $SHELL'`
  - Supports SSH options: port (`-p`), identity file (`-i`)
  - Works on all platforms: macOS (Ghostty, Terminal.app), Windows (cmd), Linux (9+ terminal emulators)
- **Local workspaces**: Works exactly as before (spawn terminal with cwd set)

All callers updated to pass `workspaceId` instead of `namedWorkspacePath`.

## Testing

- `make typecheck` passes ✅
- For local workspaces: Should work exactly as before
- For SSH workspaces: Should spawn a local terminal that SSHs in and opens in the workspace directory

_Generated with `cmux`_

### DIFF
--- a/src/services/ipcMain.ts
+++ b/src/services/ipcMain.ts
@@ -961,7 +961,7 @@ export class IpcMain {
     ipcMain.handle(IPC_CHANNELS.WORKSPACE_OPEN_TERMINAL, async (_event, workspaceId: string) => {
       try {
         // Look up workspace metadata to get runtime config
-        const allMetadata = this.config.getAllWorkspaceMetadata();
+        const allMetadata = await this.config.getAllWorkspaceMetadata();
         const workspace = allMetadata.find((w) => w.id === workspaceId);
 
         if (!workspace) {


src/services/ipcMain.ts
```diff
@@ -905,76 +905,30 @@ export class IpcMain {
...
+        const allMetadata = this.config.getAllWorkspaceMetadata();
+        const workspace = allMetadata.find((w) => w.id === workspaceId);
+
+        if (!workspace) {
+          log.error(`Workspace not found: ${workspaceId}`);
+          return;
+        }
+
+        const runtimeConfig = workspace.runtimeConfig;
+        const isSSH = runtimeConfig?.type === "ssh";
```
ibetitsmike: @ethanndickson can you double check if we don't already have a const for checking it? 

Maybe better to introduce isSsh instead of this string comparison magic

PR #558, opened 2025-11-12
feat: add terminal support

sreya:


src/browser/api.ts
```diff
@@ -249,6 +249,27 @@ const webApi: IPCApi = {
...
     },
   },
+  terminal: {
+    create: (params) => invokeIPC(IPC_CHANNELS.TERMINAL_CREATE, params),
+    close: (sessionId) => invokeIPC(IPC_CHANNELS.TERMINAL_CLOSE, sessionId),
+    resize: (params) => invokeIPC(IPC_CHANNELS.TERMINAL_RESIZE, params),
+    sendInput: (sessionId: string, data: string) => {
+      // Send via IPC - in browser mode this becomes an HTTP POST
+      void invokeIPC(IPC_CHANNELS.TERMINAL_INPUT, sessionId, data);
+    },
```
kylecarbs, line 259: This might be slow over the network. Fine for now, but we should probably allow the `wsManager` to send data for the IPC at some point.

src/components/ChatInput/index.tsx
```diff
@@ -237,6 +237,17 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
...
+      // Don't trigger focus keybinds when in terminal
+      const target = event.target as HTMLElement;
+      const terminalHasFocus =
+        target?.closest?.(".terminal-view") !== null ||
+        target?.classList?.contains("terminal-view") ||
+        target?.classList?.contains("terminal-container") ||
+        target?.tagName === "CANVAS"; // ghostty-wasm uses canvas
+      if (terminalHasFocus) {
+        return;
+      }
```
kylecarbs: We should just stop propagation of keyboard events if the terminal is focused.

I think you can just do `event.stopPropagation()` in the input handler.

src/components/ChatInput/index.tsx
```diff
@@ -360,8 +371,13 @@ export const ChatInput: React.FC<ChatInputProps> = (props) => {
...
 
     // Small delay to ensure DOM is ready and other components have settled
+    // Don't auto-focus if the terminal has focus (user is typing in terminal)
     const timer = setTimeout(() => {
-      focusMessageInput();
+      const activeElement = document.activeElement;
+      const terminalHasFocus = activeElement?.closest(".terminal-view") !== null;
+      if (!terminalHasFocus) {
+        focusMessageInput();
+      }
```
kylecarbs: I don't think this would trigger unless I'm confused. This would only happen on the initial render I believe unless something else is wack.

src/components/Terminal/TerminalView.tsx
```diff
@@ -0,0 +1,253 @@
+/* eslint-disable @typescript-eslint/prefer-optional-chain */
+/* eslint-disable @typescript-eslint/prefer-nullish-coalescing */
```
kylecarbs: We should just fix the issues instead of these.

src/components/Terminal/TerminalView.tsx
```diff
@@ -0,0 +1,253 @@
...
+      fitAddonRef.current = null;
+      setTerminalReady(false);
+      setTerminalSize(null);
+    };
+    // Note: sendInput and resize are intentionally not in deps
+    // They're used in callbacks, not during effect execution
+  }, [visible, workspaceId]);
+
+  // Note: Terminal output and exit are now handled via callbacks passed to useTerminalSession
+  // (handleOutput and handleExit defined above)
```
kylecarbs: rm ai comments

src/hooks/useAIViewKeybinds.ts
```diff
@@ -102,7 +102,13 @@ export function useAIViewKeybinds({
       }
 
       // Focus chat input works anywhere (even in input fields)
+      // BUT not when typing in the terminal
       if (matchesKeybind(e, KEYBINDS.FOCUS_CHAT)) {
+        // Don't steal focus from terminal
+        const terminalHasFocus = (e.target as HTMLElement)?.closest(".terminal-view") !== null;
+        if (terminalHasFocus) {
+          return;
+        }
```
kylecarbs: Same here if we do event prop cancel this shouldn't be needed

src/components/Terminal/TerminalView.tsx
```diff
@@ -0,0 +1,253 @@
+/* eslint-disable @typescript-eslint/prefer-optional-chain */
```
kylecarbs: Since this is just one file in the folder we should just move this out of the folder. `components/Terminal.tsx`

kylecarbs: Or `components/TerminalView.tsx` is fine too

src/components/AIView.tsx
```diff
@@ -168,7 +169,8 @@ const AIViewInner: React.FC<AIViewProps> = ({
   );
 
   const handleOpenTerminal = useCallback(() => {
-    void window.api.workspace.openTerminal(workspaceId);
+    console.log(`[AIView] Opening terminal window for workspace: ${workspaceId}`);
```
kylecarbs: rm comment

src/services/ipcMain.ts
```diff
@@ -1458,6 +1475,111 @@ export class IpcMain {
...
+
+    ipcMain.handle(IPC_CHANNELS.TERMINAL_WINDOW_OPEN, async (_event, workspaceId: string) => {
+      console.log(`[BACKEND] TERMINAL_WINDOW_OPEN handler called with: ${workspaceId}`);
+      try {
+        if (!this.terminalWindowManager) {
+          throw new Error("Terminal window manager not available (desktop mode only)");
+        }
+        log.info(`Opening terminal window for workspace: ${workspaceId}`);
+        const devServerPort = process.env.CMUX_DEVSERVER_PORT ?? "5173";
+        await this.terminalWindowManager.openTerminalWindow(workspaceId, devServerPort);
```
kylecarbs: I'm not sure what this var is... seems like legacy from when there was a separate server. We should remove this.

And since there's a session ID with terminals, should `terminal.create` just open the window instead? That feels cleaner to me. And we can make it so `sessionId` is included in the query params so a user could reload a terminal.

src/services/ptyService.ts
```diff
@@ -0,0 +1,368 @@
...
+import { log } from "@/services/log";
+import type { Runtime } from "@/runtime/Runtime";
+import type { ExecStream } from "@/runtime/Runtime";
+import type { TerminalSession, TerminalCreateParams, TerminalResizeParams } from "@/types/terminal";
+import { SSHRuntime } from "@/runtime/SSHRuntime";
+import { LocalRuntime } from "@/runtime/LocalRuntime";
+import * as fs from "fs";
+
+interface SessionData {
+  pty?: any; // For local sessions (IPty type)
```
kylecarbs: Fix the type

src/services/ptyService.ts
```diff
@@ -0,0 +1,368 @@
...
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+/* eslint-disable @typescript-eslint/no-require-imports */
+/* eslint-disable @typescript-eslint/consistent-type-imports */
+/* eslint-disable @typescript-eslint/prefer-nullish-coalescing */
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+/* eslint-disable @typescript-eslint/consistent-type-assertions */
+/* eslint-disable @typescript-eslint/require-await */
+/* eslint-disable local/no-sync-fs-methods */
```
kylecarbs: lmaooo fix all these

src/services/ptyService.ts
```diff
@@ -0,0 +1,368 @@
...
+      log.info(`[PTY] SSH command for ${sessionId}: ${command}`);
+      log.info(`[PTY] SSH terminal size: ${params.cols}x${params.rows}`);
+      log.info(`[PTY] SSH working directory: ${workspacePath}`);
+
+      let stream;
+      try {
+        log.info(`[PTY] Calling runtime.exec for ${sessionId}...`);
+        // Execute shell with PTY allocation
+        // Use a very long timeout (24 hours) instead of Infinity
+        stream = await (runtime as any).exec(command, {
```
kylecarbs: no as any

kylecarbs (changes requested):

src/main.ts
```diff
@@ -3,6 +3,8 @@
 const isServer = process.argv.length > 2 && process.argv[2] === "server";
 
 if (isServer) {
+  // Remove 'server' from args since main-server doesn't expect it as a positional argument.
+  process.argv.splice(2, 1);
```
kylecarbs, line 7: i'm confused by this i don't think this is needed

sreya: something about my changes upgraded commander to v14 which enforces stricter args parsing. Because main-server.ts isn't expecting args but we're passing it `server` it crashes out on 14

src/services/ipcMain.ts
```diff
@@ -75,6 +82,11 @@ export class IpcMain {
       this.partialService,
       this.initStateManager
```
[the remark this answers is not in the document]
sreya: @codex address that feedback

PR #559, opened 2025-11-12
🤖 fix: Ctrl+C no longer interrupts stream when text is selected

deansheather:
Previously, Ctrl+C would interrupt the stream if the user was not in an editable element, even if they had text selected elsewhere in the chat (e.g., in message content).

## Problem

The condition `(!inEditableElement || !hasSelection)` would interrupt when:
- NOT in editable element (regardless of selection) ← **the bug**
- In editable element but no selection

This meant selecting text in chat messages and pressing Ctrl+C to copy would incorrectly interrupt the stream.

## Solution

Simplified the logic to `!hasSelection` - now it always checks for text selection anywhere on the page, regardless of whether the user is in an editable element.

## Verification

```typescript
// Before: Interrupted even with selection outside editable elements
if ((canInterrupt || showRetryBarrier) && (!inEditableElement || !hasSelection)) {
  // interrupt
}

// After: Only interrupts when nothing is selected
if ((canInterrupt || showRetryBarrier) && !hasSelection) {
  // interrupt
}
```

**Behavior now:**
- User selects text in chat messages → Press Ctrl+C → Text is copied (stream continues) ✅
- No text selected → Press Ctrl+C → Stream is interrupted ✅

_Generated with `cmux`_

### DIFF
--- a/src/hooks/useAIViewKeybinds.ts
+++ b/src/hooks/useAIViewKeybinds.ts
@@ -67,12 +67,11 @@ export function useAIViewKeybinds({
         }
 
         // Normal stream interrupt (non-compaction)
-        // Allow interrupt in editable elements if there's no text selection
+        // Check for text selection anywhere - only interrupt if nothing is selected
         // This way Ctrl+C works for both copy (when text is selected) and interrupt (when not)
-        const inEditableElement = isEditableElement(e.target);
         let hasSelection = false;
 
-        if (inEditableElement) {
+        if (isEditableElement(e.target)) {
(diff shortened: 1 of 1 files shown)


deansheather:
Closing as redundant - PR #562 solved this issue in a better way by introducing vim-aware keybinds.

The original problem (Ctrl+C interrupting when text is selected) is now resolved:
- **Vim mode**: Ctrl+C always interrupts (expected - vim users use yank for copy)
- **Non-vim mode**: Esc interrupts, Ctrl+C is for copying

This is a cleaner solution that provides the most natural behavior for each mode.

PR #583, opened 2025-11-14
🤖 feat: implement soft-interrupts with block boundary detection

ethanndickson:
Soft interrupts allow graceful stream cancellation at block boundaries (tool-result, reasoning-end, etc). First Escape triggers soft interrupt with orange UI feedback and waits for boundary; second Escape forces immediate abort.

## Implementation

**Frontend state tracking:**
- AIView tracks interrupt state with `useState`
- First Escape press → sets flag, calls `interruptStream({ soft: true })`
- Second Escape press → calls `interruptStream({ soft: false })`
- Flag resets when stream ends via `useEffect` watching `canInterrupt`

**Backend behavior:**
- `interruptStream(workspaceId, { soft?, abandonPartial? })`
- If `soft: true`: Sets internal flag, cancels at next block boundary
- If `soft: false`: Immediate abort via `cancelStreamSafely`
- Block boundaries: tool-result, reasoning-end, reasoning-start, tool-call-start

**UI feedback:**
- Orange "⏸️ Interrupting..." text when soft interrupt pending
- Standard blue streaming indicator otherwise
- Cancel text shows "force" on second press

## Testing

- Integration tests verify soft → hard interrupt progression
- Manual testing with tool calls and reasoning blocks
- Verified flag reset on stream completion

_Generated with `mux`_

### DIFF
--- a/src/browser/utils/messages/StreamingMessageAggregator.ts
+++ b/src/browser/utils/messages/StreamingMessageAggregator.ts
@@ -387,16 +387,17 @@ export class StreamingMessageAggregator {
   }
 
   handleStreamDelta(data: StreamDeltaEvent): void {
-    const message = this.messages.get(data.messageId);
-    if (!message) return;
-
     // Handle soft interrupt signal from backend
+    // Do this BEFORE message check to ensure interrupt state is captured even if message lookup fails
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

PR #613, opened 2025-11-15
🤖 fix: reconnect New Workspace button

ammar-agent:
WorkspaceContext owns the "pending new workspace" state, but App was still tracking its own copy. The sidebar button (and other entry points) mutate the context value, so the creation view never appeared.

- Read pendingNewWorkspaceProject/begin/clear from WorkspaceContext in App
- Clear the shared state after creation cancel/success
- Update useStartWorkspaceCreation to call beginWorkspaceCreation directly

Tests: bun test src/hooks/useStartWorkspaceCreation.test.ts

_Generated with `mux`_


ammario (approved):
Manually verified it works

PR #640, opened 2025-11-16
🤖 fix: drop stale OpenAI previousResponseId on 500s

ammar-agent:
@-


ammario:
Note: I'm seeing persistent 500 errors from OpenAI occasionally on active chats, suggesting there's something malformed in the history or stored response object. Thinking that sending everything from scratch will most likely resolve the issue.

PR #641, opened 2025-11-16
🤖 fix: sync workspace store during metadata updates

ammar-agent:
## Summary
- ensure WorkspaceProvider syncs WorkspaceStore immediately whenever metadata changes so new workspaces exist before React mounts
- add tests that cover initial metadata load plus live metadata events to confirm subscriptions and reset global store between runs

## Testing
- bun test src/browser/contexts/WorkspaceContext.test.tsx
- make typecheck

_Generated with `mux`_


ethanndickson:
Merging to close #647

PR #658, opened 2025-11-17
🤖 fix: unblock release workflow

ammar-agent:
## Summary
- add workflow_dispatch + preflight guard so releases can run on a tag
- gate VSCE publish inside the script so missing secrets no-op cleanly

## Testing
- make typecheck

_Generated with _


ammar-agent:
Added a wait step plus a dependency on the mac build job so this upload only runs once electron-builder has created the release.

PR #661, opened 2025-11-17
🤖 feat: add Grok provider support

kylecarbs:
## Summary
- add the xAI provider and Grok model metadata
- extend configuration, env hydration, and runtime wiring for xai
- refresh docs and tests to surface Grok support

## Testing
- bun test ./tests/models/knownModels.test.ts
- make typecheck

_Generated with _

### DIFF
--- a/docs/models.md
+++ b/docs/models.md
@@ -53,7 +53,7 @@ TODO: add issue link here.
 
 Frontier reasoning models from xAI with built-in search orchestration:
 
-- `xai:grok-4` — Fast unified model (switches between reasoning/non-reasoning based on thinking toggle)
+- `xai:grok-4-1` — Fast unified model (switches between reasoning/non-reasoning based on thinking toggle)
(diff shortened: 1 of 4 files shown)


docs/models.md
```diff
@@ -27,6 +27,30 @@ GPT-5 family of models:
...
+{
+  "xai": {
+    "apiKey": "sk-xai-...",
+  },
+}
+```
+
+**Search orchestration (optional):**
+
+Add [`searchParameters`](https://docs.x.ai/docs/resources/search) in `providers.jsonc` to enable automatic web search. Mux forwards any additional JSON fields so you can configure region, time filters, or disable search entirely per workspace.
```
ammario: We should have search enabled by default to follow behavior with Anthropic and OpenAI. I think it's a user expectation now with coding agents.

ammario (changes requested):

PR #664, opened 2025-11-18
fix: fix node-pty packaging

sreya:
I'm going to try to articulate the problem as best as I understand it at the moment. Web terminals are currently used in the current cases:
- Desktop builds + remote workspace
- Remote workspace + any workspace

In order to make web terminals work we need to have access to a pty handle aka we have to use `node-pty`. The problem is that we basically distribute two versions of mux: an electron variant for desktop usage, and a server variant. `node-pty` doesn't bundle any linux prebuilds so when you run `npm install mux` it has to build node-pty from source but it inevitably fails for a lot of cases because the environment lacks the build tools. To fix this you can use something like `@lydell/node-pty` but then you break the electron use case because the prebuilds it bundles for mac aren't for the electron ABI and it doesn't bundle source files so it can't build from source. 

The fucked solution I came up with is to basically include _both_ with `@lydell/node-pty` as an optional dependency. What happens at runtime is it tries to use `@lydell/node-pty` first, (which should succeed for the linux use case) and if not falls back to regular ole `node-pty` which should be fine for desktop applications. 

Developers have to run `make dev` moving forward if developing for the desktop use case (which they have to anyway?)

### DIFF
--- a/src/node/services/ptyService.ts
+++ b/src/node/services/ptyService.ts
@@ -224,13 +224,13 @@ export class PTYService {
       let pty: typeof import("node-pty");
       try {
         // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-unsafe-assignment
-        pty = require("@lydell/node-pty");
-        log.debug("Using @lydell/node-pty for SSH (prebuilt binaries)");
+        pty = require("node-pty");
+        log.debug("Using node-pty for SSH (rebuilt for Electron)");
       } catch {
         try {
           // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-unsafe-assignment
-          pty = require("node-pty");
-          log.debug("Using node-pty for SSH (rebuilt for Electron)");
+          pty = require("@lydell/node-pty");
+          log.debug("Using @lydell/node-pty for SSH (prebuilt binaries)");
         } catch (err) {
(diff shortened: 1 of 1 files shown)


sreya:
@codex review

sreya:
@codex review

sreya:
@codex review

sreya:
@codex review

package.json
```diff
@@ -162,7 +161,9 @@
     "vite-plugin-top-level-await": "^1.6.0"
   },
   "optionalDependencies": {
-    "electron": "^38.2.1"
+    "electron": "^38.2.1",
+    "@lydell/node-pty": "1.1.0",
+    "node-pty": "1.1.0-beta39"
```
kylecarbs, line 170: This should not be needed at all anymore.

package.json
```diff
@@ -162,7 +161,9 @@
     "vite-plugin-top-level-await": "^1.6.0"
   },
   "optionalDependencies": {
-    "electron": "^38.2.1"
+    "electron": "^38.2.1",
+    "@lydell/node-pty": "1.1.0",
```
kylecarbs: This should be in dependencies if it's required for the server to work.

kylecarbs (changes requested):

package.json
sreya: i need regular node-pty for electron

sreya:
@codex review

PR #673, opened 2025-11-18
🤖 fix: fail-fast workspace title generation with provider errors

ibetitsmike:
This PR removes the temporary fallback workspace title generation and surfaces provider errors to the user during first-message workspace creation.


ammario:
I'm a little confused by the purpose of this PR. If we fail to generate the name (IMO) we should generate a placeholder like "<project-name>-<timestamp>". We should have renaming the workspace automatically after first create and let the user do it as it involves a fallible rename dir operation.

ibetitsmike:
@ammario the purpose is to avoid all of that "I'll fix it up later" flow and fail fast before

PR #675, opened 2025-11-18
🤖 feat: add send button to chat input

ibetitsmike:
Adds a clickable send button to the ChatInput component that mirrors the existing Enter key behavior.


ammario:
Perhaps button should be simple arrow to minimize space taken.

PR #712, opened 2025-11-24
🤖 fix: stabilize storybook theme by forcing context state

ammar-agent:
Fixes flaky visual diffs in Storybook where the theme would flash from default (light) to selected (dark) on load.

## Changes
- Modified `ThemeProvider` to accept a `forcedTheme` prop that overrides internal state and persistence.
- Updated Storybook `preview.tsx` to use `forcedTheme` instead of a `useEffect`-based synchronization component.

This ensures the correct theme is applied during the first render pass, eliminating the flash and preventing Storybook from polluting local storage.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -1,30 +1,13 @@
-import React, { useEffect } from "react";
+import React from "react";
 import type { Preview } from "@storybook/react-vite";
-import {
-  ThemeProvider,
-  useTheme,
-  type ThemeMode,
-} from "../src/browser/contexts/ThemeContext";
+import { ThemeProvider, type ThemeMode } from "../src/browser/contexts/ThemeContext";
 import "../src/browser/styles/globals.css";
-
-const ThemeStorySync: React.FC<{ mode: ThemeMode }> = ({ mode }) => {
-  const { theme, setTheme } = useTheme();
-
-  useEffect(() => {
-    if (theme !== mode) {
-      setTheme(mode);
-    }
-  }, [mode, setTheme, theme]);
-
-  return null;
-};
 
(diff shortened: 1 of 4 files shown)


ammar-agent:
Updated `ThemeProvider` to check for a parent theme context. This allows the inner `ThemeProvider` (e.g., inside `App.tsx` used in a story) to inherit the theme forced by the outer `ThemeProvider` (from `preview.tsx`). This resolves theme mismatches in nested scenarios.

ammar-agent:
Updated `preview.tsx` to remove the default dark override and fallback behavior.
- Now, if no global theme is set (e.g., local Storybook without toolbar interaction), `ThemeProvider` uses its internal persistence/system preference. This restores the ability to toggle themes via the app's UI or OS settings during development.
- Chromatic builds continue to work deterministically because `modes` explicitly set the global `theme`, which is respected by the `forcedTheme` prop.
- This ensures "Latest" doesn't get stuck on Dark and both themes are reviewable.

ammar-agent:
Added a manual theme toggle button to the Storybook preview decorator.

**Why:**
Since `addon-essentials` is missing, the standard Storybook toolbar is unavailable, leaving users unable to toggle themes in the default "Latest" view (which now respects system preference instead of forcing Dark).

**Behavior:**
- **Chromatic / Forced Modes:** The button is hidden (`!mode` check), ensuring clean snapshots.
- **Local / Default View:** A floating toggle button appears in the bottom-right corner, allowing developers to switch between Light and Dark themes freely.

ammario:
@ethanndickson gonna turn in soon for the night if this is bothering you and this PR seems dead for more than 10m feel free to take it over.

PR #717, opened 2025-11-24
🤖 fix: cancel active stream on message edit to prevent history corruption

ethanndickson:
## Problem

While trying to edit the token count of a running compaction request, the entire chat history was replaced with `[truncated]`.

## Root Cause

When editing a compacting message while the first compaction stream is still running:

1. The edit truncates history and starts a new compaction stream
2. The **old stream continues running** and isn't cancelled
3. When the old stream completes, `handleCompletion` reads the current history
4. It finds the **NEW** compaction request (from the edit), not the old one
5. Since the new request isn't in `processedCompactionRequestIds`, it proceeds to perform compaction
6. This clears ALL history and replaces it with the old stream's partial summary → entire chat becomes `[truncated]`

## Solution

Cancel any active stream before processing message edits in `AgentSession.sendMessage`:

```typescript
// Cancel any active stream when editing a message to prevent race conditions
if (options?.editMessageId && this.aiService.isStreaming(this.workspaceId)) {
  const stopResult = await this.interruptStream(/* abandonPartial */ true);
  if (!stopResult.success) {
    return Err(createUnknownSendMessageError(stopResult.error));
  }
}
```

This ensures:
- Only one stream runs at a time
- Edits truly discard the old stream (aligns with user intent)
- No orphaned stream completions that corrupt history

## Changes

- **src/node/services/agentSession.ts**: Added stream cancellation check before processing edits (9 lines)

## Testing

- ✅ Typecheck passes
- ✅ Graceful error handling (interruptStream is idempotent)
- ✅ Fixes the specific scenario: editing compaction token count mid-stream

_Generated with `mux`_

### DIFF
--- a/src/node/services/agentSession.ts
+++ b/src/node/services/agentSession.ts
@@ -371,17 +371,19 @@ export class AgentSession {
       return Ok(undefined);
     }
 
-    const stopResult = await this.aiService.stopStream(this.workspaceId, abandonPartial);
-    if (!stopResult.success) {
-      return Err(stopResult.error);
-    }
-
-    // Delete partial when abandoning to prevent commitToHistory from reintroducing it
+    // Delete partial BEFORE stopping to prevent abort handler from committing it
+    // The abort handler in aiService.ts runs immediately when stopStream is called,
+    // so we must delete first to ensure it finds no partial to commit
     if (abandonPartial) {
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

PR #754, opened 2025-11-25
🤖 fix: enable reasoning for Claude Opus 4.5

ThomasK33:
## Summary

Fixes missing reasoning traces in the UI when using Claude Opus 4.5.

## Problem

Opus 4.5 supports two separate but complementary parameters for reasoning:

1. **`effort`** (new in Opus 4.5): Controls how eagerly Claude spends tokens across ALL output (text, tool calls, and thinking). Values: `low`, `medium`, `high`.

2. **`thinking`** (extended thinking): Enables visible reasoning traces with a token budget. This is what makes the "Thinking..." UI appear.

We were only passing `effort`, which controls token spend but **doesn't enable the reasoning traces to be returned from the API**. The `thinking` parameter must also be set for reasoning to be visible.

## Solution

For Opus 4.5, now pass both parameters:
```typescript
{
  anthropic: {
    thinking: {
      type: "enabled",
      budgetTokens: 20000,  // Enables visible reasoning traces
    },
    effort: "high",  // Controls token spend eagerness
  }
}
```

## How these parameters interact

From [Anthropic's docs](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#effort-with-extended-thinking):

> The effort parameter works alongside the thinking token budget when extended thinking is enabled. These two controls serve different purposes:
> - **Effort parameter**: Controls how Claude spends all tokens—including thinking tokens, text responses, and tool calls
> - **Thinking token budget**: Sets a maximum limit on thinking tokens specifically
>
> The effort parameter can be used with or without extended thinking enabled.

| Configuration | Behavior |
|---------------|----------|
| `effort` only | Controls token spend, no visible reasoning |
| `thinking` only | Shows reasoning with default budget |
| `effort` + `thinking` | Shows reasoning WITH token spend control |

## Testing

- Verified reasoning traces now appear in UI when using Opus 4.5
- `make typecheck` passes

_Generated with `mux`_

### DIFF
--- a/src/browser/components/ThinkingSlider.stories.tsx
+++ b/src/browser/components/ThinkingSlider.stories.tsx
@@ -48,9 +48,7 @@ export const DifferentModels: Story = {
       </div>
 
       <div>
-        <div className="text-muted-light font-primary mb-2 text-xs">
-          Claude Opus 4.5 (3 levels: low/medium/high)
-        </div>
+        <div className="text-muted-light font-primary mb-2 text-xs">Claude Opus 4.5 (4 levels)</div>
         <ThinkingSliderComponent modelString="anthropic:claude-opus-4-5" />
       </div>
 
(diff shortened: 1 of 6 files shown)


ThomasK33:
@codex review

PR #758, opened 2025-11-26
Fix import paths after src re-org

ozbek:
A follow-up for #622. This also corrects shiki theme name.


ozbek:
@ThomasK33, hi! Is there anyway to pass the Integration Tests without the AI providers' API keys?

ThomasK33:
> @ThomasK33, hi! Is there anyway to pass the Integration Tests without the AI providers' API keys?

Hey @ozbek, thanks for opening the PRs.
Unfortunately, I don't think there's a way to run them without API keys.

ThomasK33 (approved):
Thanks 👍

PR #780, opened 2025-11-28
(Re-)fix: shiki-shared import path

ozbek:
That got reverted in another revert (a2a417ed).


ozbek:
Cc: @ThomasK33 :)

ethanndickson:
Not sure how that happened, sorry & thanks for fixing!

PR #807, opened 2025-12-01
🤖 fix: prevent queued message amnesia by flushing partial before tool-call-end

ethanndickson:
## Problem

When a queued message is sent after `tool-call-end`, the model doesn't see the complete assistant response including tool results. This causes "amnesia" where the model starts tasks from scratch.

### Root Cause

Race condition in partial write timing:

```
Tool completes → emit(tool-call-end) → schedulePartialWrite (fire-and-forget)
                      ↓
              sendQueuedMessages() → commitToHistory() → reads STALE partial.json
```

The `schedulePartialWrite` was called with `void` (fire-and-forget). When `sendQueuedMessages` triggers `commitToHistory`, the partial file may not contain the latest tool results.

## Solution

Await the partial flush **before** emitting `tool-call-end`:

```
Tool completes → flushPartialWrite (await) → emit(tool-call-end)
                                                    ↓
                                           sendQueuedMessages() → commitToHistory() → reads COMPLETE partial
```

## Changes

- Make `completeToolCall` async
- Await `flushPartialWrite` before emitting `tool-call-end`  
- Update callers to await the async method

---
_Generated with `mux`_

### DIFF
--- a/src/node/services/streamManager.ts
+++ b/src/node/services/streamManager.ts
@@ -566,9 +566,11 @@ export class StreamManager extends EventEmitter {
   }
 
   /**
-   * Complete a tool call by updating its part and emitting tool-call-end event
+   * Complete a tool call by updating its part and emitting tool-call-end event.
+   * CRITICAL: Flushes partial to disk BEFORE emitting event to prevent race conditions
+   * where listeners (e.g., sendQueuedMessages) read stale partial data.
    */
-  private completeToolCall(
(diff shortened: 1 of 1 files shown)


ethanndickson:
We still buffer partial writes for `stream-delta`, but I don't think it makes sense to buffer them for `tool-call-end`, given they're nowhere near as frequent.

PR #818, opened 2025-12-01
🤖 fix: buffer terminal output until client subscribes

ibetitsmike:
In server mode, there was a race condition where the shell prompt would be lost because the PTY started emitting output before the HTTP response (with session ID) returned and before the client registered its WebSocket handler.

**The fix adds output buffering on the server side:**
- PTY output is buffered in the session until client subscribes
- New `TERMINAL_SUBSCRIBE` IPC channel to signal readiness
- Client calls subscribe after registering its output handler
- Server flushes buffered output when subscribe is called

This ensures the initial shell prompt is always delivered to the terminal.

_Generated with `mux`_

### DIFF
--- a/src/browser/api.ts
+++ b/src/browser/api.ts
@@ -343,7 +343,11 @@ const webApi: IPCApi = {
     onOutput: (sessionId: string, callback: (data: string) => void) => {
       // Subscribe to terminal output events via WebSocket
       const channel = `terminal:output:${sessionId}`;
-      return wsManager.on(channel, callback as (data: unknown) => void);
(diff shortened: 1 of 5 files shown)


ibetitsmike:
Closing this PR as the issue has already been fixed on main via the ORPC refactor. The new `TerminalService` class in main uses an `outputBuffers` map to buffer terminal output and replays it when `onOutput` is called, fixing the race condition.

See `src/node/services/terminalService.ts` on main for the implementation.

_Generated with `mux`_

PR #830, opened 2025-12-01
🤖 feat: add ~/.mux/bashrc for shell environment customization

### LINKED ISSUE #797: PATH is missing user folders
Running on macos the PATH seen by the model appears to be `/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin` which is missing all user folders and nothing works as all cli tools are unavailable.

Namely the environment mux runs in seems to be missing all folders set by
1. (fish) shell configuration
2. direnv using flake.nix

While the messaging suggests some attempts to use nix, those attempts fail because nix is not in PATH.

ammar-agent:
## Summary

When mux launches from Applications (not terminal), agent bash commands get a minimal PATH (`/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin`) missing user tools like Homebrew, nix, direnv, pyenv, etc.

The standard `~/.bashrc` isn't sourced because:
1. `bash -c` creates a non-interactive shell
2. Most bashrc files have `[[ $- != *i* ]] && return` guards that skip content

## Solution

Add a dedicated `~/.mux/bashrc` file that is **always sourced** before every bash command, following the existing pattern of:
- `~/.mux/AGENTS.md` — global agent instructions
- `<project>/.mux/init` — project initialization hooks

## Use Cases

```bash
# ~/.mux/bashrc

# Nix + direnv (auto-loads .envrc per workspace)
if [ -e "$HOME/.nix-profile/etc/profile.d/nix.sh" ]; then
  . "$HOME/.nix-profile/etc/profile.d/nix.sh"
fi
eval "$(direnv hook bash)"

# Homebrew (macOS)
eval "$(/opt/homebrew/bin/brew shellenv)"

# Or any version manager: pyenv, rbenv, nvm, asdf
```

## Implementation

- `LocalRuntime.exec()` prepends bashrc sourcing to all commands
- `SSHRuntime.exec()` prepends bashrc sourcing (uses remote `~/.mux/bashrc`)
- Init hooks also source bashrc for consistency
- Silently skipped if file doesn't exist

## Testing

- Unit tests for the bashrc snippet
- Integration test that verifies env vars from bashrc are available

Fixes #797

---

_Generated with `mux`_

### DIFF
--- a/docs/bashrc.md
+++ b/docs/bashrc.md
@@ -1,0 +1,146 @@
+# Shell Environment (`~/.mux/bashrc`)
+
+Customize the shell environment for agent bash commands by creating a `~/.mux/bashrc` file.
+
+## Why This Exists
+
+When mux runs bash commands (via `bash -c "command"`), the shell is **non-interactive** and **doesn't source `~/.bashrc`**. Most users have interactivity guards in their bashrc that skip content in non-interactive shells:
+
+```bash
+# Common pattern in ~/.bashrc that skips setup for non-interactive shells
+[[ $- != *i* ]] && return
+```
+
+This means:
+
+- **Launching from Applications** — PATH is minimal (`/usr/bin:/bin:/usr/sbin:/sbin`)
+- **Nix/direnv users** — Shell customizations aren't applied
+- **Homebrew/pyenv/rbenv** — Tools not in PATH
+
+The `~/.mux/bashrc` file is **always sourced** before every bash command, giving you a place to set up the environment reliably.
+
+## Setup
+
+Create `~/.mux/bashrc`:
+
+```bash
+mkdir -p ~/.mux
+touch ~/.mux/bashrc
+```
+
+Add your shell customizations:
+
+```bash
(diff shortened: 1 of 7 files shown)


docs/bashrc.md
```diff
@@ -0,0 +1,146 @@
...
+```bash
+# ~/.mux/bashrc
+
+# Source nix profile
+if [ -e "$HOME/.nix-profile/etc/profile.d/nix.sh" ]; then
+  . "$HOME/.nix-profile/etc/profile.d/nix.sh"
+fi
+
+# Enable direnv (auto-loads .envrc per directory)
+eval "$(direnv hook bash)"
```
tko, line 56: `direnv hook bash` sets `PROMPT_COMMAND` which I believe means that direnv will not be activated until bash prompt is shown, which is never when running commands with `bash -c`

Assuming bashrc is sourced from the project directory with `.envrc` file then I _guess_ running `direnv export bash` instead might work better (it's what direnv hook seems to run.)

PR #831, opened 2025-12-02
🤖 fix: accurate cost estimation for multi-step tool usage

ethanndickson:
## Problem

The application was severely underestimating costs for conversations involving tool calls.

### Root Cause

The Vercel AI SDK provides two usage metrics:
- `streamResult.usage` — Token usage from **the last step only**
- `streamResult.totalUsage` — Sum of token usage across **all steps**

The application was using `usage` instead of `totalUsage`. For a conversation with 10 tool calls, only ~1/10th of actual consumption was reported. A $5 conversation would display as $0.50.

### The Complicating Factor

Two UI elements use token data with different semantic requirements:

| Display | Needs | Why |
|---------|-------|-----|
| **Cost** | Sum of all steps | If model read context 10 times, you paid for 10 reads |
| **Context window** | Last step only | Shows "how full is the conversation now" for the next request |

Simply switching to `totalUsage` would fix costs but break context display (showing 500% utilization after many tool calls).

### Cache Creation Tokens

Anthropic's cache creation tokens (`cacheCreationInputTokens`) are:
- Only in provider-specific metadata, not normalized usage
- Need to be summed across all steps
- Not automatically aggregated by the AI SDK

Even with `totalUsage`, cache creation costs were lost unless manually aggregated from each step's provider metadata.

## Solution

Track both values with different semantic purposes:

**For cost calculation:**
- `usage` / `cumulativeUsage` — total across all steps
- `providerMetadata` / `cumulativeProviderMetadata` — aggregated cache creation tokens

**For context window display:**
- `contextUsage` / `lastContextUsage` — last step only
- `contextProviderMetadata` — last step only

### Key Changes

1. **Backend** (`streamManager.ts`): Use `totalUsage` for cost, track `lastStepUsage` for context, aggregate provider metadata across steps
2. **Types**: Extended `StreamEndEvent`, `MuxMetadata`, `UsageDeltaEvent` with dual fields
3. **Frontend**: `StreamingMessageAggregator` tracks both cumulative and per-step usage
4. **Store**: `WorkspaceUsageState` provides `usageHistory` (cost) and `lastContextUsage` (context window)
5. **UI**: Components use appropriate field for their purpose

### Also Fixed

- **OpenAI cached token double-counting**: Gateway models (`mux-gateway:openai/gpt-5.1`) weren't recognized as OpenAI, causing cached tokens to be counted in both "Cache Read" and "Input". Now normalizes gateway model strings before provider detection.

- **Google/Gemini cached token double-counting**: Google, like OpenAI, reports `inputTokens` inclusive of `cachedInputTokens`. Extended the subtraction logic to handle Google models.

---
_Generated with `mux`_

### DIFF
--- a/src/common/utils/tokens/displayUsage.test.ts
+++ b/src/common/utils/tokens/displayUsage.test.ts
@@ -148,7 +148,7 @@ describe("collectUsageHistory", () => {
 });
 
 describe("createDisplayUsage", () => {
-  describe("OpenAI cached token handling", () => {
+  describe("Provider-specific cached token handling", () => {
     // OpenAI reports inputTokens INCLUSIVE of cachedInputTokens
     // We must subtract cached from input to avoid double-counting
     const openAIUsage: LanguageModelV2Usage = {


ethanndickson:
@codex review

PR #840, opened 2025-12-02
🤖 feat: add 5% buffer between auto-compaction threshold and force-compaction

ethanndickson:
## Background

**Auto-compaction** runs after a response completes when context usage exceeds a user-configured threshold (default 70%). It summarizes the conversation to free up context space while preserving important information.

**Force-compaction** is a safety mechanism that triggers *during* streaming when context usage gets dangerously high. Unlike auto-compaction which waits for a natural break point, force-compaction interrupts the current response to prevent hitting the context window limit.

## The Problem

Previously, force-compaction triggered based on a fixed token buffer from the context window limit. This meant force-compaction timing was disconnected from the user's configured auto-compaction threshold—changing your threshold didn't affect when force-compaction would kick in.

## The Change

Force-compaction now triggers at **threshold + 5%**. With a 70% threshold, force-compaction happens at 75%.

This gives users a predictable buffer zone between when auto-compaction *would* run (after the response) and when force-compaction *will* run (during streaming). If a response slightly overshoots the threshold, the chat has some leeway—it won't immediately force-compact just because usage landed a bit over. This avoids unnecessary force-compactions when the stream ends and usage settles back within acceptable bounds.

The buffer is intentionally small (5%) to balance user control with safety margins as context approaches capacity.

## UI Change

Also adds a "Force-compacting in N%" countdown during streaming when usage is in the buffer zone, so users know how much room remains before force-compaction triggers.

---
_Generated with `mux`_

### DIFF
--- a/src/browser/utils/compaction/autoCompactionCheck.test.ts
+++ b/src/browser/utils/compaction/autoCompactionCheck.test.ts
@@ -369,6 +369,30 @@ describe("checkAutoCompaction", () => {
       expect(result.usagePercentage).toBe(75); // usagePercentage reflects live even with empty history
     });
 
+    test("shouldShowWarning uses live usage when no history exists", () => {
+      // No lastUsage, liveUsage at 65% - should show warning (65% >= 60%)
+      const liveUsage = createUsageEntry(130_000); // 65%
+      const usage: WorkspaceUsageState = {
+        usageHistory: [],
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

PR #847, opened 2025-12-02
🤖 feat: add Solarized light/dark themes with dropdown selector

ibetitsmike:
## Summary

Adds Solarized Light and Solarized Dark themes, and converts the theme toggle button to a dropdown selector.

### Changes

- Extended `ThemeMode` type to include `solarized-light` and `solarized-dark`
- Added `THEME_OPTIONS` constant with labels for dropdown
- Created `ThemeSelector` dropdown component
- Updated GeneralSection settings to use dropdown instead of toggle button
- Added complete CSS variable definitions for both Solarized themes using the official palette
- Updated command palette to show all theme options dynamically
- Fixed syntax highlighting to treat solarized-light as a light theme

_Generated with `mux`_

### DIFF
--- a/tests/e2e/scenarios/settings.spec.ts
+++ b/tests/e2e/scenarios/settings.spec.ts
@@ -22,7 +22,7 @@ test.describe("Settings Modal", () => {
     await expect(page.getByRole("button", { name: "Models", exact: true })).toBeVisible();
 
     // Verify default section is General (theme toggle visible)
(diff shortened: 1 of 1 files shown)


ammario:
wonder if we could easily support VS Code themes and reuse their library. Also just to make sure, we're not increasing the Storybook here? scared about having to review NxM stories on visual changes.

PR #857, opened 2025-12-02
🤖 fix: disable extended thinking for workspace name generation

ammar-agent:
When generating workspace names with reasoning models (e.g., claude-opus-4-5), the API sometimes returns thinking content that breaks JSON parsing, causing 'Invalid JSON response' errors.

**Fix:** Explicitly pass providerOptions with thinking disabled to `generateObject()`.

- Add `buildProviderOptions(modelString, 'off')` to ensure no thinking content
- Add `ModelProvider` interface for easier testing without full AIService
- Add unit/integration tests for workspace title generation
- Add `testUtils.ts` with minimal model creation for tests

_Generated with `mux`_


ammario:
Not actually the problem

PR #864, opened 2025-12-03
🤖 fix: include step providerMetadata in liveUsage for cache creation display

ethanndickson:
During streaming, `liveUsage` (context window display) showed 0 cache creation tokens because it wasn't receiving `providerMetadata`. The cache creation token count is Anthropic-specific and only available in `providerMetadata.anthropic.cacheCreationInputTokens`, not in the standard `LanguageModelV2Usage`.

Thread step-level `providerMetadata` from backend through IPC to frontend, mirroring the existing pattern for cumulative provider metadata.

The bug was masked after message completion because the UI falls back to `lastContextUsage` (from persisted metadata) when `liveUsage` is undefined.

_Generated with `mux`_

### DIFF
--- a/scripts/bump_tag.sh
+++ b/scripts/bump_tag.sh
@@ -18,7 +18,7 @@ if [[ -z "$CURRENT_VERSION" || "$CURRENT_VERSION" == "null" ]]; then
 fi
 
 # Parse semver components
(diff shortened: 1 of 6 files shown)


ethanndickson:
@codex review

PR #877, opened 2025-12-03
🤖 perf: async workspace name generation for faster UX

ibetitsmike:
Previously, workspace creation blocked for up to 10s while waiting for the AI to generate a workspace name. This caused a poor UX where users had to wait before seeing any result.

## Changes

- Use `generatePlaceholderName()` to create workspace immediately with a temporary name derived from the user's message (e.g., 'add-user-auth')
- Generate AI name asynchronously in the background via `generateAndApplyAIName()`
- If AI generation succeeds and differs from placeholder, rename workspace using existing `rename()` method
- If AI generation or rename fails, gracefully keep the placeholder name

## UX Improvement

| Before | After |
|--------|-------|
| User waits 5-10s for AI name generation | Workspace appears instantly |
| Loading spinner while blocked | AI name applied seamlessly when ready |
| Network/API errors block creation | Errors handled gracefully, placeholder name retained |

The existing `generatePlaceholderName()` function was already implemented but unused - this change puts it to work.

_Generated with `mux`_

### DIFF
--- a/src/node/services/workspaceService.test.ts
+++ b/src/node/services/workspaceService.test.ts
@@ -1,0 +1,113 @@
+import { describe, expect, test, mock, beforeEach } from "bun:test";
+import { WorkspaceService } from "./workspaceService";
+import type { Config } from "@/node/config";
+import type { HistoryService } from "./historyService";
+import type { PartialService } from "./partialService";
+import type { AIService } from "./aiService";
+import type { InitStateManager } from "./initStateManager";
+import type { ExtensionMetadataService } from "./ExtensionMetadataService";
+
+// Helper to access private renamingWorkspaces set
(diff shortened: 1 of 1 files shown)


ammario:
Can we reliably rename a workspace with an active stream? Just making sure it won't interfere with running bash operations, file edits, etc.

ibetitsmike:
@ammario the idea was to wait until stream ends

```
const handler = (event: StreamEndEvent | StreamAbortEvent) => {
        if (event.workspaceId === workspaceId) {
          this.aiService.off("stream-end", handler);
          this.aiService.off("stream-abort", handler);
          log.debug("Stream completed, proceeding with rename", { workspaceId });
          resolve();
        }
      };
```

PR #880, opened 2025-12-03
🤖 fix: resolve race conditions in stream error handling and test flakiness

ThomasK33:
## Summary
- Make `debugTriggerStreamError` async and await `processingPromise` to ensure error handling completes before returning (eliminates duplicate error-handling logic)
- Wait for in-flight partial writes before writing error state to prevent inconsistent `partial.json`
- Pass `toolPolicy` on resume to match original message (tools disabled → text output)
- Add delay after resume to let history update complete before reading

_Generated with `mux`_


ThomasK33:
@codex review

PR #893, opened 2025-12-04
🤖 fix: correct context usage display for multi-step tool calls

ethanndickson:
## Problem

The Context Usage UI was sometimes showing inflated token counts for multi-step tool calls (e.g., plan messages showing ~150k `cachedInputTokens` instead of ~50k).

## Root Cause

Two issues combined:

1. **Backend**: On stream-end, we only used `lastStepUsage` tracked from `finish-step` events. If no `finish-step` event was received (edge cases), `contextUsage` would be undefined. We weren't fetching `streamResult.usage` (which contains last step's usage) as the primary source.

2. **Frontend**: When `contextUsage` was undefined, it fell back to cumulative `usage`. For multi-step requests, cumulative usage sums `cachedInputTokens` across all steps (each step reads from cache), inflating the context window display.

image

## Fix

**Backend (`streamManager.ts`)**:
- Fetch `contextUsage` from `streamResult.usage` (last step) on stream-end as the primary source
- Fall back to tracked `lastStepUsage` only if `streamResult.usage` times out
- Each metadata fetch has independent timeout/error handling so one failure doesn't mask others

**Frontend (`WorkspaceStore.ts`)**:
- Remove fallback from `contextUsage` to `usage` - only use `contextUsage` for context window display

**Schema (`message.ts`)**:
- Add `contextUsage` and `contextProviderMetadata` to Zod schema to match TypeScript types

## Migration Note

⚠️ **Old workspaces** created before this fix may not have `contextUsage` stored in their message history. These workspaces will show no context usage until the next message is sent.

image

---
_Generated with `mux`_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,7 +224,7 @@ lint: node_modules/.installed ## Run ESLint (typecheck runs in separate target)
 lint-fix: node_modules/.installed ## Run linter with --fix
 	@./scripts/lint.sh --fix
 
-ifeq ($(OS),Windows_NT) 
+ifeq ($(OS),Windows_NT)
 typecheck: node_modules/.installed src/version.ts ## Run TypeScript type checking (uses tsgo for 10x speedup)
 	@# On Windows, use npm run because bun x doesn't correctly pass arguments
 	@npmx concurrently -g \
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

PR #968, opened 2025-12-08
fix: use cygpath on Windows for temp dir path conversion

ethanndickson:
## Problem

On Windows, `createTempDirForStream` runs a shell command via Git Bash that returns a Unix-style path like `/c/Users/ethan/.mux-tmp/854e934d`.

During cleanup, Node's `fs.access()` and `spawn()` APIs misinterpret this path as `C:\c\Users\...` instead of `C:\Users\...`, causing cleanup to fail:

```
Failed to cleanup temp dir /c/Users/ethan/.mux-tmp/854e934d: RuntimeError: Working directory does not exist
  path: 'C:\c\Users\ethan\.mux-tmp'
```

## Fix

Use `cygpath -w` on Windows to convert the `pwd` output to native Windows path format (`C:\Users\ethan\.mux-tmp\abc123`), which Node APIs handle correctly.

### DIFF
--- a/src/node/services/streamManager.ts
+++ b/src/node/services/streamManager.ts
@@ -271,7 +271,10 @@ export class StreamManager extends EventEmitter {
   public async createTempDirForStream(streamToken: StreamToken, runtime: Runtime): Promise<string> {
     // Create directory and get absolute path (works for both local and remote)
     // Use 'cd' + 'pwd' to resolve ~ to absolute path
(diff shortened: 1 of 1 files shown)


ethanndickson:
I think this breaks a bunch of stuff, so I'm closing.. Though this is definitely an issue, we're not cleaning up any of the tmp dirs.

Opened #971.

PR #977, opened 2025-12-08
🤖 feat: refactor Plan Mode with external file change detection

ThomasK33:
## Summary

- **Plan Mode workflow**: Toggle between Plan/Exec modes (Cmd+Shift+M). In Plan mode, file edits restricted to `~/.mux/plans/<workspace-id>.md`. Agent proposes plans via `propose_plan` tool for user review before implementation.
- **External file change detection**: Timestamp-based polling detects external edits to plan files, computes unified diffs, injects as synthetic user messages to keep agent aware without breaking prompt cache.
- **ProposePlan UI**: Markdown rendering, edit-in-editor button, "Start Here" context reset, fresh content on window focus.
- **File edit refactoring**: Extracted shared execution pipeline, runtime-aware path validation, custom diff utility replacing external dependency.

## Test plan

- [ ]  Toggle Plan/Exec mode with keyboard shortcut
- [ ]  Verify file edits blocked for non-plan files in Plan mode
- [ ]  Edit plan externally, confirm agent sees changes on next query
- [ ]  Run make test for unit test coverage
- [ ]  Run TEST_INTEGRATION=1 bun x jest tests/ipc/fileChangeNotification.test.ts for integration tests

_Generated with `mux`_

### DIFF
--- a/docs/cli.mdx
+++ b/docs/cli.mdx
@@ -102,3 +102,25 @@ Print the version and git commit:
 mux --version
 # v0.8.4 (abc123)
 ```
+
+## Debug Environment Variables
+
+These environment variables help diagnose issues with LLM requests and responses.
+
+| Variable                | Purpose                                                                                                                                                                 |
+| ----------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `MUX_DEBUG_LLM_REQUEST` | Set to `1` to log the complete LLM request (system prompt, messages, tools, provider options) as formatted JSON to the debug logs. Useful for diagnosing prompt issues. |
+
+Example usage:
+
+```bash
+MUX_DEBUG_LLM_REQUEST=1 mux run "Hello world"
+```
+
+The output includes:
+
+- `systemMessage`: The full system prompt sent to the model
+- `messages`: All conversation messages in the request
+- `tools`: Tool definitions with descriptions and input schemas
+- `providerOptions`: Provider-specific options (thinking level, etc.)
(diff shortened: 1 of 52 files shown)


ammario:
This seems like a lot of complexity... are you not able to get steering with the system prompt?

ThomasK33:
> This seems like a lot of complexity... are you not able to get steering with the system prompt?

What steering are you referring to?

ammario:
Seems like you're trying to fix the issue where the agent makes edits via bash when in plan mode, which is primarily the case with non-Anthropic models?

ThomasK33:
> Seems like you're trying to fix the issue where the agent makes edits via bash when in plan mode, which is primarily the case with non-Anthropic models?

No, not at all. This approach is somewhat inspired by the planning system in Claude Code, which also writes to a Markdown file.

Essentially, it's a way to reduce output tokens and achieve a faster, more consistent iterative planning loop with the LLM, since the propose_plan tool call doesn’t need to reprint the entire plan each time, and more efficient file editing operations.

Additionally, the file will be useful in a follow-up PR, as we’ll be able to "attach" the plan to the workspace. After compaction, the plan can be included verbatim, supporting a more reliable and longer-running agentic loop.

ThomasK33:
Additionally, sharing the plan with others will become easier, enabling a future where a PR could include the plan file, making it accessible and helping developers achieve alignment more quickly.

ammario:
Ah ok, understood. I misinterpreted the PR title.

ThomasK33:
@codex review

PR #995, opened 2025-12-09
🤖 fix: emit stream-error for pre-stream API failures

ibetitsmike:
When sendMessage failed before the stream started (e.g., invalid model, API key missing), the error was only shown as a toast which could be easily missed. The user would see their message in the chat but no response, with no indication of what went wrong.

## Changes

- Modified `agentSession.streamWithHistory` to emit a `stream-error` chat event when the stream fails to start
- Added `emitStreamError` helper method that converts `SendMessageError` to `StreamErrorMessage`
- Also emit stream-error when no model is specified

## Why

The user would:
1. Type a message and hit send
2. See their message appear in the chat
3. See a toast (which might auto-dismiss or be missed)
4. Wonder why the assistant never responded

Now the error is clearly visible in the chat conversation itself.

## Testing

- Added integration test to verify stream-error is emitted on early model validation failure
- Existing `sendMessageError` unit tests continue to pass

_Generated with `mux`_

### DIFF
--- a/src/browser/components/AIView.tsx
+++ b/src/browser/components/AIView.tsx
@@ -133,7 +133,7 @@ const AIViewInner: React.FC<AIViewProps> = ({
     }
 
     for (const message of workspaceState.messages) {
-      if (message.type !== "stream-error") {
+      if (message.type !== "stream-error" && message.type !== "chat-error") {
         continue;
       }
       if (message.errorType !== "model_not_found") {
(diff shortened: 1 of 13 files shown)


ammario:
If we do this I suggest changing the terminology on the frontend to be clear it doesn't map to an ai-sdk stream-error. E.g we could call it chat-error. Also this code is entwined with the auto retry logic so take care that doesn't regress. Otherwise I think this is the right architecture.

ibetitsmike:
@codex review

PR #1042, opened 2025-12-09
🤖 fix: simplify Editor settings UI

ammar-agent:
Simplifies the Editor settings in the General section:

- Removed redundant 'Editor' section header - now just a single 'Editor' row
- Removed 'Use Remote-SSH' checkbox - SSH workspaces always use Remote-SSH when the editor supports it (VS Code/Cursor)
- Removed `useRemoteExtension` from `EditorConfig` type and storage

For SSH workspaces, there's no other sensible path than Remote-SSH, so the checkbox was unnecessary. When users click 'Open in Editor' on an SSH workspace with VS Code/Cursor, it just works. If the Remote-SSH extension isn't installed, the editor itself prompts to install it.

---
_Generated with `mux`_

### DIFF
--- a/src/browser/components/Settings/sections/GeneralSection.tsx
+++ b/src/browser/components/Settings/sections/GeneralSection.tsx
@@ -8,8 +8,6 @@ import {
   SelectValue,
 } from "@/browser/components/ui/select";
 import { Input } from "@/browser/components/ui/input";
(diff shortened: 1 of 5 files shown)


ammario:
cc @ibetitsmike

PR #1051, opened 2025-12-10
🤖 feat: add vim editor support and editors.js configuration

ammar-agent:
Adds first-class vim/neovim support and a JavaScript-based editor configuration system.

## Changes
- Editors configured via `~/.mux/editors.js` instead of localStorage
- Two editor types: native (GUI) and web_term (terminal)
- Ships with VS Code, Cursor, Zed, and Vim/Neovim
- Terminal editors work everywhere: desktop, browser, SSH
- Added `initialCommand` support to terminal windows
- New documentation at /docs/editor

_Generated with mux_

### DIFF
--- a/src/browser/components/WorkspaceHeader.tsx
+++ b/src/browser/components/WorkspaceHeader.tsx
@@ -1,8 +1,16 @@
 import React, { useCallback, useEffect, useState } from "react";
-import { Pencil } from "lucide-react";
+import { Pencil, ChevronDown, Check } from "lucide-react";
 import { GitStatusIndicator } from "./GitStatusIndicator";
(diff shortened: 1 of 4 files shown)


src/common/types/editor.ts
```diff
@@ -0,0 +1,170 @@
...
+  | {
+      /** Open in web terminal with this command */
+      type: "web_term";
+      /** Full command string to run in terminal */
+      command: string;
+    }
+  | {
+      /** Editor cannot handle this context */
+      error: string;
+    };
```
ThomasK33, line 51: Does it make sense to turn those into Zod schemas and validate on load?

PR #1067, opened 2025-12-10
🤖 chore: remove racy E2E test infrastructure

ammar-agent:
Removes the artificial delay test infrastructure that was added to test the activate race condition.

The fix itself (`&& services` guard in the activate handler, merged in #1064) is simple enough that the existing assertion in `createWindow()` catches any regression. Artificial delays for timing-based tests are flaky and slow.

---
_Generated with `mux`_

### DIFF
--- a/src/desktop/main.ts
+++ b/src/desktop/main.ts
@@ -286,13 +286,6 @@ async function loadServices(): Promise<void> {
   const startTime = Date.now();
   console.log(`[${timestamp()}] Loading services...`);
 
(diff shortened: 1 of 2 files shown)


ammar-agent:
Closing - after discussion, the existing `assert(services, "Services must be loaded...")` in `createWindow()` is sufficient as a safety net. Any regression to the activate handler guard would hit this assertion during E2E or manual testing.

Adding artificial delays for timing-based race condition tests is flaky and not worth the maintenance cost for a 3-character fix (`&& services`).

The actual fix shipped in #1064.

PR #1091, opened 2025-12-11
🤖 feat: add ability to disable MCP servers

ethanndickson:
Adds a toggle to disable MCP servers without removing them from config. Disabled servers retain their configuration but don't inject tools into conversations.

**Why:** Tool definitions consume context tokens. When you don't need a server's tools for a particular task, disabling it avoids unnecessarily blowing up the context. This mirrors similar functionality in Claude Code.

**Changes:**
- Switch toggle in project settings (left side of each server row)
- Config format: string when enabled, `{ command, disabled: true }` when disabled
- Backwards compatible with existing configs
- Optimistic UI updates

**CI fix:** Also fixes a caching bug in `setup-mux` action where restore-key matches (partial cache hits) would skip `bun install`, causing builds to fail when new dependencies were added. Now runs install on any non-exact cache hit while preserving the Windows optimization that skips bun cache decompression when `node_modules` already exists.

---
_Generated with `mux`_

### DIFF
--- a/.github/actions/setup-mux/action.yml
+++ b/.github/actions/setup-mux/action.yml
@@ -44,8 +44,10 @@ runs:
         restore-keys: |
           ${{ runner.os }}-bun-cache-
 
+    # Always run install if we didn't get an exact cache hit.
+    # This handles restore-key matches which may have stale/incomplete node_modules.
     - name: Install dependencies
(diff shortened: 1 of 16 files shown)


ethanndickson:
@codex review

PR #1106, opened 2025-12-12
🤖 refactor: script-based status_set polling API

ammar-agent:
- Make status_set accept { script, poll_interval_s? } (seconds; omitted = run once)
- Keep URL extraction + post-extraction truncation
- Drop renderer localStorage persistence; rely on backend + live updates

_Generated with `mux`_

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -21,9 +21,6 @@ gh pr view <number> --json mergeable,mergeStateStatus | jq '.'
 ./scripts/wait_pr_checks.sh <pr_number>
 ```
 
-- While waiting on long-running operations (CI, builds, formatting/typecheck/test loops), keep the user informed via `status_set` **with `poll_interval_s`** and a dynamic script (avoid chat spam).
-  - Use a short poll interval (e.g., 5–15s).
-  - Include the PR URL when relevant so it stays clickable.
 - Generally run `wait_pr_checks` after submitting a PR to ensure CI passes.
 - Status decoding: `mergeable=MERGEABLE` clean; `CONFLICTING` needs resolution. `mergeStateStatus=CLEAN` ready, `BLOCKED` waiting for CI, `BEHIND` rebase, `DIRTY` conflicts.
 - If behind: `git fetch origin && git rebase origin/main && git push --force-with-lease`.


ammar-agent:
### Context / issues with the current `status_set` (LLM-driven script+polling) approach

This PR implements a *tool-call-driven* status mechanism:
- LLM calls `status_set({ script, poll_interval_s? })`.
- Backend runs the script (optionally on an interval), parses one line (emoji/message + URL extraction), and emits `agent-status-update` stream events.
- Renderer consumes `agent-status-update` to show status (and backend persists config + lastStatus in `~/.mux/sessions/<workspaceId>/status_set.json` so it survives reload/restart).

Even with those mechanics working, the approach still has product-level issues that are hard to “fix with code”:

#### 1) Reliability depends on the model
The biggest pain is that the status quality is gated on the model *remembering* to:
- call `status_set` at all, and
- include a `poll_interval_s` (and use a dynamic script) instead of a one-shot `echo`.

When it forgets, the user experience regresses to “no status / stale status”, which is exactly what we want to avoid.

#### 2) Security & blast radius
This design effectively grants the model a way to register an auto-running script. Even with timeouts and interval caps, it expands the execution surface:
- The script can run repeatedly after streaming ends.
- It can run in workspace context with whatever environment/secrets are available.

#### 3) Cross-platform + operational complexity
Polling scripts are inherently brittle across environments:
- `bash` vs `pwsh` differences (Windows),
- availability of `gh`, auth state, etc.
- failure modes are noisy unless aggressively suppressed (which hurts debuggability).

#### 4) Product mental model / UX
Users likely think “status is something mux shows” (repo/branch/PR), not “status is something the model must maintain via tool calls”. The tool-call approach makes status feel optional and inconsistent.

### Alternate direction (proposed)
A more reliable baseline is to make status *mux-owned* rather than *LLM-owned*:
- Always poll a well-known status provider for each workspace (e.g. `.mux/status.local` / `.mux/status`), and/or
- ship a safe built-in default (git branch/dirty/ahead/behind), optionally augmented with a user opt-in repo script.

If repo scripts are supported, they likely need an explicit per-repo approval gate to avoid auto-running arbitrary code from repos.

### Why close this PR
Given the direction shift (toward an always-polled repo/user-owned status provider and away from an LLM tool call), this implementation is likely not the right product primitive to land.

PR #1114, opened 2025-12-12
🤖 fix: gateway toggle double-toggle race condition

ThomasK33:
## Summary
- Fix GatewayIcon not forwarding refs (React warning about function components and refs)
- Make gateway icon clickable on main page to toggle gateway routing
- Remove redundant `updatePersistedState` calls that caused double-toggle bug

## Root cause
`setEnabledModels()` from `usePersistedState` already writes to localStorage synchronously, but `updatePersistedState()` was also being called. This read the already-toggled value from localStorage and toggled it again, effectively reverting the change.

## Test plan
- [x] Click gateway icon next to model selector - should toggle on/off
- [x] Toggle in settings Models section - should persist
- [x] No more React warning about refs on GatewayIcon

---
_Generated with [Claude Code](https://claude.ai/code)_

### DIFF
--- a/src/browser/components/ModelSelector.tsx
+++ b/src/browser/components/ModelSelector.tsx
@@ -205,17 +205,27 @@ export const ModelSelector = forwardRef<ModelSelectorRef, ModelSelectorProps>(
 
       return (
         <div ref={containerRef} className="relative flex items-center gap-1">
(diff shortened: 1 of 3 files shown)


ThomasK33:
Already fixed by #1112

PR #1225, opened 2025-12-19
🤖 feat: live peek bash stdout/stderr

ThomasK33:
Live-stream stdout/stderr from the `bash` tool into the UI while it runs (UI-only; tool contract stays final-result-only).

- `bash` tool emits throttled `bash-output` chat events and stops emitting once migrated to a background process.
- Renderer buffers a bounded tail per `toolCallId` (1MB cap) and clears it when the final tool result includes `output`.
- `BashToolCall` shows live (or tail) stdout/stderr in the expanded view, with pinned auto-scroll.

Validation:
- `make static-check`

---


ThomasK33:
@codex review

PR #1238, opened 2025-12-20
🤖 feat: persist per-model thinking via backend settings

ammar-agent:
Eliminate localStorage↔backend duplication for thinking levels by making per-model thinking backend-authoritative.

### What changed
- Added `persistedSettings` to config (`~/.mux/config.json`) storing `ai.thinkingLevelByModel`
- New oRPC routes: `persistedSettings.get`, `persistedSettings.setAIThinkingLevel`, `persistedSettings.onChanged`
- Frontend `PersistedSettingsStore` keeps an in-memory snapshot, refreshes + subscribes, and provides a localStorage fallback + one-time seed to migrate existing `thinkingLevel:model:*` keys
- `ThinkingContext`, command palette thinking actions, and `getSendOptionsFromStorage` now read from the store; no longer write thinking to localStorage
- `WorkspaceContext` only seeds workspace model; thinking is handled via persistedSettings

### Tests
- Added unit tests for PersistedSettingsService + PersistedSettingsStore
- Updated WorkspaceContext seeding test

---

### DIFF
--- a/src/browser/stores/PersistedSettingsStore.test.ts
+++ b/src/browser/stores/PersistedSettingsStore.test.ts
@@ -47,9 +47,17 @@ describe("PersistedSettingsStore", () => {
       );
     });
 
+    const change = (() => {
+      let resolve!: () => void;
+      const promise = new Promise<void>((res) => {
+        resolve = () => res();
+      });
+      return { promise, resolve };
+    })();
+
     const onChanged = mock(() => {
       async function* iter() {
-        await Promise.resolve();
+        await change.promise;
         // Yield once to trigger a refresh.
         yield undefined;
       }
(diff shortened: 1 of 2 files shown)


ammar-agent:
Heads up: we’re **not** planning to merge this approach. We’re abandoning the backend `persistedSettings`/store design in favor of **mode-based thinking storage**.

Leaving notes here because there were a few useful lessons worth carrying forward:

- **State ownership matters:** making thinking backend-authoritative removes localStorage↔backend drift, but it *necessarily* introduces a cache + subscription layer on the client. If we don’t want that complexity, we should keep the source of truth local.
- **Normalize at the boundary:** always normalize model IDs (`mux-gateway:provider/model` → `provider:model`) before persisting or keying lookups, or you’ll get duplicate entries and surprising “lost” preferences.
- **Prefer a typed blob + events over per-key RPC:** `get()` returning a small typed object + `setX()` mutations + `onChanged()` works well and avoids O(n) IPC patterns.
- **Migration needs to be explicit + best-effort:** seeding backend state from existing localStorage (once) is a reasonable bridge, but plan for partial failures / old-server mismatches.
- **Testing:** permission-based tests are flaky under root/CI. For failure-path coverage, prefer deterministic setups (e.g. `ENOTDIR` via “muxHome is a file”) and gate async-iterator tests with explicit signals to avoid racey ordering.

---

PR #1259, opened 2025-12-21
🤖 fix: don't persist compaction model as workspace AI settings

ammar-agent:
## Problem

When compaction is triggered with a specific model (e.g., `/compact -m openai:gpt-4.1-mini`), the backend was incorrectly persisting that model to the workspace's AI settings, overwriting the user's preferred model.

After compaction completed, the workspace would be set to use the compaction model instead of the model the user had selected.

## Root Cause

Regression introduced in 1ae03776 (#1203) which added a "safety net" to persist AI settings from `sendMessage`/`resumeStream` for cross-device consistency. This safety net didn't account for compaction requests which intentionally use a different (often cheaper/faster) model.

cc @tomaskoubek (regressor)

## Fix

Skip persisting AI settings when:
- `sendMessage` is called with `muxMetadata.type === "compaction-request"`
- `resumeStream` is called with `options.mode === "compact"`

## Testing

- Added IPC test verifying compaction requests don't override workspace AI settings
- Existing `updateAISettings` test still passes

---

### DIFF
--- a/src/node/services/workspaceService.ts
+++ b/src/node/services/workspaceService.ts
@@ -959,6 +959,33 @@ export class WorkspaceService extends EventEmitter {
     return { model, thinkingLevel };
   }
 
+  /**
+   * Best-effort persist AI settings from send/resume options.
+   * Skips compaction requests which use a different model intentionally.
+   */
+  private async maybePersistAISettingsFromOptions(
+    workspaceId: string,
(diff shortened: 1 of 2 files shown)


ammario:
cc @ThomasK33

apologies `@ tomaskoubek` for the tag, you can disregard this

PR #1278, opened 2025-12-21
🤖 fix: cache plan content in localStorage to prevent layout flash

ammar-agent:
When loading a workspace with an existing plan, the ProposePlanToolCall component previously showed placeholder text while fetching fresh content from disk, causing a jarring layout flash when the full content arrived.

Now plan content is cached in localStorage and rendered optimistically on load. The cache is updated whenever fresh content is fetched, so users always see current content without the flash.

## Changes
- Add `getPlanContentKey()` to storage.ts for plan content cache
- Initialize ProposePlanToolCall with cached content from localStorage  
- Update cache on successful fetch

## How it works
1. **First load** of a workspace with a plan: shows placeholder briefly, then caches content
2. **Subsequent loads**: renders cached content instantly (no flash), fetches fresh content in background, updates cache
3. **External edits**: detected on window focus, cache updated with new content

---


ammario:
cc @ThomasK33 — this fixes the layout flash (in the common case) introduced by new plan system

PR #1331, opened 2025-12-26
🤖 fix: align git status tooltip near indicator in workspace header

ammar-agent:
## Problem

The git status tooltip in the workspace header was not positioned near the status indicator. The sidebar version worked correctly.

## Solution

The `HoverCardContent` was using `align="center"` for both tooltip positions. When the tooltip appears at the bottom (workspace header), centering caused it to float away from the indicator.

Changed alignment to be position-dependent:
- `tooltipPosition="right"` (sidebar) → `align="center"` - centers vertically, natural for list items
- `tooltipPosition="bottom"` (header) → `align="start"` - anchors to left edge of trigger

## Testing

Added `GitStatusHeaderTooltip` story that:
- Selects a workspace so header is visible
- Sets up git status with line deltas and dirty state
- Hovers over the indicator and waits for tooltip
- Captures the hovered state for visual regression testing

---

### DIFF
--- a/src/browser/components/GitStatusIndicatorView.tsx
+++ b/src/browser/components/GitStatusIndicatorView.tsx
@@ -342,7 +342,7 @@ export const GitStatusIndicatorView: React.FC<GitStatusIndicatorViewProps> = ({
       </HoverCardTrigger>
       <HoverCardContent
         side={tooltipPosition === "right" ? "right" : "bottom"}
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

Fixed the storage key issue - now using `selectWorkspace()` helper instead of manually setting the wrong key.

PR #1347, opened 2025-12-28
🤖 feat: model-only TODO reminder notifications

ThomasK33:
Adds a poll-based, model-only notification mechanism injected via tool results.

Initial source: TODO reminders that nudge the agent to update `todo_write` after long tool runs.

Key points:
- `NotificationEngine` + `TodoListReminderSource` (5 tool calls, then every 10; suppressed when empty/all completed)
- Injects notifications into tool results under `__mux_notifications` (only for plain-object results)
- Strips `__mux_notifications` before persisting/emitting tool outputs

Validation:
- `make static-check`

---

<details>

# Model-only agent notifications (polling)

## Goal
Add a **generic, poll-based notification system** that can inject model-only nudges during long autonomous tool runs.

The **TODO reminder** is the first notification source: it periodically reminds the model to call `todo_write` and includes the current persisted TODO list.

## Decisions (from this request)
- **Model-only** delivery (no new visible chat transcript noise).
- **Poll-based**: evaluate notification sources after each tool call.
- Support **multiple notification sources** (TODO reminder is just one source).
- Reminder **includes the current persisted TODO list** (max 7 items, low token cost).
- **Cadence (TODO reminder):** adaptive — first reminder after **5** tool calls since the last `todo_write`, then every **10** tool calls until updated.
- **Suppress** the TODO reminder when the TODO list is missing/empty **or** when *all* items are `completed`.

## Recommended approach: NotificationEngine + sources + tool-result injection
**Net new product code:** ~220–320 LoC.

This approach is the only one that can reliably nudge the model **mid-run** (inside a single long `streamText` multi-step tool loop) without requiring a new user turn.

### 1) Define the notification contract (extensible)
- Add a small node-side interface, e.g.:
  - `NotificationSource` with `poll(ctx) -> AgentNotification[]` (can be empty)
  - `AgentNotification` contains at least `{ source: string; content: string }`
- Standardize on a single reserved tool-output key:
  - `__mux_notifications: string[]` (each entry is a complete `<notification>…</notification>` block)

### 2) Implement `NotificationEngine` (polling coordinator)
Create a per-stream engine that:
- Holds a list of `NotificationSource`s.
- Exposes `pollAfterToolCall({ toolName, output, now })`.
- De-dupes notifications (same `content`) to avoid spam.
- Returns a `string[]` payload suitable for `__mux_notifications`.

### 3) Implement first source: `TodoListReminderSource`
- Tracks `toolCallsSinceLastTodoWrite`.
  - Increment on every tool call **except** `todo_write`.
  - Reset to 0 on successful `todo_write`.
- Due when `count === 5` or `(count > 5 && (count - 5) % 10 === 0)`.
- When due:
  - Read todos from `todos.json` (reuse `getTodosForSessionDir()` from `src/node/services/tools/todo.ts`).
  - Suppress if empty or all completed.
  - Render the list using the existing `- [x] / - [>] / - [ ]` formatting.
    - Preferred: factor the list-rendering helper into a shared module so both:
      - `src/browser/utils/messages/attachmentRenderer.ts` and
      - `TodoListReminderSource`
      use the same renderer.

### 4) Wire it up by wrapping tool `execute()` in `getToolsForModel()`
In `src/common/utils/tools/tools.ts`, after all tools are assembled:
- Create one `NotificationEngine` instance.
- Wrap each tool’s `execute()`:
  1. Call the original tool.
  2. Ask the engine for notifications (`pollAfterToolCall`).
  3. If notifications exist **and** the tool result is a plain object, return:
     - `{ ...result, __mux_notifications: notifications }`

### 5) Strip notifications before UI/history persistence (keep them model-only)
In `src/node/services/streamManager.ts` tool-result handling:
- Remove `__mux_notifications` from `part.output` before writing to `toolCall.output` / partial/history.
- Keep this stripping in a single helper so future internal fields are easy to add.

### Suggested notification payload format
```xml
<notification>
It's been N tool calls since you last updated the TODO list. If your progress changed, update it now using todo_write.

Current TODO List:
- [x] …
- [>] …
- [ ] …
</notification>
```

### Defensive programming notes
- Only enable notification injection when `workspaceSessionDir` is present.
- Never change tool result shape unless it is a non-null, non-array object.
- Keep the reserved key name stable and unlikely to collide (`__mux_notifications`).

## Tests
- **Unit:** `NotificationEngine` polling + dedupe behavior.
- **Unit:** `TodoListReminderSource` cadence (5, 15, 25…), reset on `todo_write`, suppression when all completed.
- **Unit:** tool wrapper attaches `__mux_notifications` only when due.
- **Unit:** streamManager stripping removes `__mux_notifications` from persisted/emitted tool outputs.

<details>
<summary>Why tool-result injection (vs. synthetic user messages) is still the right primitive</summary>

Mux uses the AI SDK’s multi-step tool loop (`streamText(... stopWhen: stepCountIs(100000))`). Inside that loop, the only thing we can reliably feed the model *between* steps is **tool results** (returned by `tool.execute`). A poll-based notification engine can run after every tool call, attach model-only hints to tool outputs, and then strip them before persistence.

</details>

</details>

---

### DIFF
--- a/src/common/utils/tools/tools.modelOnlyNotifications.test.ts
+++ b/src/common/utils/tools/tools.modelOnlyNotifications.test.ts
@@ -79,6 +79,71 @@ describe("getToolsForModel - model-only notifications", () => {
     }
   });
 
+  test("does not re-wrap cached MCP tools across getToolsForModel() calls", async () => {
+    const workspaceSessionDir = await fs.mkdtemp(path.join(os.tmpdir(), "mux-ws-"));
+
+    try {
+      await setTodosForSessionDir("ws-1", workspaceSessionDir, [
+        { content: "In progress", status: "in_progress" },
+      ]);
+
+      const runtime = new LocalRuntime(process.cwd());
+      const initStateManager = {
+        waitForInit: () => Promise.resolve(),
+      } as unknown as InitStateManager;
+
+      const cachedMcpTool = {
+        // eslint-disable-next-line @typescript-eslint/require-await
+        execute: async () => ({ ok: true }),
+      } as unknown as Tool;
+
+      const tools1 = await getToolsForModel(
+        "noop:model",
+        {
+          cwd: process.cwd(),
+          runtime,
+          runtimeTempDir: "/tmp",
+          workspaceSessionDir,
+        },
+        "ws-1",
+        initStateManager,
+        undefined,
+        { mcp_dummy: cachedMcpTool }
+      );
+
+      const execute1 = getExecute(tools1.mcp_dummy);
+      for (let i = 0; i < 4; i += 1) {
+        const result = asRecord(await execute1());
+        expect(MODEL_ONLY_TOOL_NOTIFICATIONS_FIELD in result).toBe(false);
+      }
+
+      const tools2 = await getToolsForModel(
+        "noop:model",
+        {
+          cwd: process.cwd(),
+          runtime,
+          runtimeTempDir: "/tmp",
+          workspaceSessionDir,
+        },
+        "ws-1",
+        initStateManager,
+        undefined,
+        { mcp_dummy: cachedMcpTool }
+      );
+
+      const execute2 = getExecute(tools2.mcp_dummy);
+      for (let i = 0; i < 4; i += 1) {
+        const result = asRecord(await execute2());
+        expect(MODEL_ONLY_TOOL_NOTIFICATIONS_FIELD in result).toBe(false);
+      }
+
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

Addressed the cached MCP tool wrapping issue by cloning tools before wrapping `execute`, so repeated `getToolsForModel()` calls don’t stack `NotificationEngine` layers. Added a regression test covering repeated calls with a cached MCP tool.

PR #1349, opened 2025-12-29
🤖 fix: PTC exclusive mode missing code_execution tool

ethanndickson:
## Problem

When `programmatic-tool-calling-exclusive` experiment is enabled, `code_execution` tool doesn't appear in available tools.

**Root cause:** Tool policy was applied twice:

1. `policyFilteredTools = applyToolPolicy(allTools, policy)` — filters base tools ✓
2. `tools = applyToolPolicy(toolsForModel, policy)` — re-filters after adding `code_execution` ✗

The second call re-filters already-filtered `nonBridgeable` tools (wasteful) and removes `code_execution` since it's not in agent preset allow-lists (broken).

## Fix

Remove redundant second `applyToolPolicy` call. Handle `code_execution` explicitly by mode:

- **Exclusive mode:** `code_execution` added unconditionally (mandatory — only way to use bridged tools). The experiment flag is the opt-in.
- **Supplement mode:** Check policy before adding `code_execution`:
  - Skip if policy requires a different tool (would violate require semantics)
  - Skip if policy explicitly disables `code_execution`

---

### DIFF
--- a/src/node/services/aiService.ts
+++ b/src/node/services/aiService.ts
@@ -1424,27 +1424,12 @@ export class AIService extends EventEmitter {
             const nonBridgeable = toolBridge.getNonBridgeableTools();
             toolsForModel = { ...nonBridgeable, code_execution: codeExecutionTool };
           } else {
-            // Supplement mode: add code_execution if policy allows (experiment is additive).
-            // Handle require policies: if code_execution is required, return only it.
-            // If another tool is required, don't add code_execution (would violate require).
-            const requireFilter = effectiveToolPolicy?.find((f) => f.action === "require");
-            const requiresCodeExecution =
-              requireFilter && new RegExp(`^${requireFilter.regex_match}$`).test("code_execution");
-            if (requiresCodeExecution) {
-              toolsForModel = { code_execution: codeExecutionTool };
-            } else if (requireFilter) {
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
Regarding the exclusive mode `require` concern:

In exclusive PTC mode, `code_execution` **is** mandatory by design - it's the only way to invoke bridged tools. The experiment flag is the opt-in that overrides preset tool restrictions.

If a preset requires `propose_plan` but the user enables exclusive PTC, the user is explicitly saying "I want the PTC interface." The non-bridgeable tools (including `propose_plan`) remain available directly, but `code_execution` must also be present or there's no way to use file_read, bash, etc.

This is intentional: experiment flags take precedence over preset restrictions. The same reasoning applies to why `code_execution` isn't gated by policy in exclusive mode - disabling it would leave no way to access tools.

PR #1351, opened 2025-12-29
🤖 feat: PTC subagent support and compatibility fixes

ethanndickson:
## Summary

Enables PTC (Programmatic Tool Calling) to work properly with subagents and fixes several compatibility issues.

### Core Fixes
- **Remove `mux.` prefix**: Nested tool calls in `code_execution` now emit bare tool names (`bash`, `file_read`) instead of `mux.bash`, `mux.file_read`
- **Exclude `agent_report` from sandbox**: Must be called as top-level tool for `taskService` to read args from history
- **Add `timeout_secs` to `code_execution`**: Default 5 minutes, max 1 hour - needed for subagents that can take 5-15+ minutes

### PTC Experiments Inheritance
- Subagents spawned via `task` tool inherit parent's PTC experiments
- Added `taskExperiments` to workspace config for restart-safe persistence
- Experiments passed through all task creation/resumption paths

### UI Updates
- Added task tools (`task`, `task_await`, `task_list`, `task_terminate`) to `NestedToolRenderer`
- Removed obsolete `normalizeToolName()` function

## Testing
- All existing tests updated and passing
- Typecheck passes
- Static checks pass

## Future

Unify where we convert tool names into the pretty UI components. Right now we have to update the list in two places when we add a new tool, `NestedToolRenderer` and `toolRenderer`

---

### DIFF
--- a/src/node/services/aiService.ts
+++ b/src/node/services/aiService.ts
@@ -1420,14 +1420,11 @@ export class AIService extends EventEmitter {
           );
 
           if (experiments?.programmaticToolCallingExclusive) {
-            // Exclusive mode: only nonBridgeable tools + code_execution are exposed to the model.
-            // Bridged tools are accessed through code_execution. Apply policy to final set so
-            // admin/preset policies can still disable code_execution if needed.
+            // Exclusive mode: code_execution is mandatory — it's the only way to use bridged
+            // tools. The experiment flag is the opt-in; policy cannot disable it here since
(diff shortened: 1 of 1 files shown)


ethanndickson:
The exclusive PTC experiment flag intentionally overrides tool policy. When exclusive mode is enabled, `code_execution` is mandatory—it's the only way to access bridged tools. If policy could disable it, there would be no way to use tools at all. The experiment flag itself is the admin opt-in.

Resolving as intentional design.

PR #1353, opened 2025-12-29
🤖 fix: bundle TypeScript libs for PTC type validation in production

ethanndickson:
## Summary

Fixes PTC (Programmatic Tool Calling) failing in production builds due to missing TypeScript lib files.

## Problems Fixed

1. **nanoid not found** - `nanoid` was only a transitive dev dependency (via postcss). Replaced with `crypto.randomBytes`.

2. **TypeScript lib files missing** - The type validator uses TypeScript's compiler API which needs `lib.es2020.d.ts` and dependencies. electron-builder has a hardcoded filter that excludes all `.d.ts` files from the asar ([electron-builder#5064](https://github.com/electron-userland/electron-builder/issues/5064)).

## Solution

- Copy only needed lib files (es5 through es2020, 48 files) to `dist/typescript-lib/` during build
- Rename to `.d.ts.txt` to bypass electron-builder's filter
- Custom compiler host in `typeValidator.ts` maps TypeScript's requests back to the renamed files

## Other Changes

- Collapse code_execution Result boxes by default (was always expanded due to useState bug)
- Add margin between consecutive code execution blocks
- Update tool description to emphasize batching calls

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,6 +224,20 @@ build-static: ## Copy static assets to dist
 	@mkdir -p dist
 	@cp static/splash.html dist/splash.html
 	@cp -r public/* dist/
+	# Copy TypeScript lib files for PTC runtime type validation (es5 through es2020).
+	# electron-builder ignores .d.ts files by default and this cannot be overridden:
+	# https://github.com/electron-userland/electron-builder/issues/5064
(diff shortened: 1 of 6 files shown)


ethanndickson:
@codex review

PR #1359, opened 2025-12-29
🤖 fix: preserve plan context after Start Here

ThomasK33:
Fixes a sharp edge where **Replace all chat history / Start Here** can remove the plan file location from context (especially after switching back to exec/custom agents).

Changes:
- Preserve the source `metadata.mode` on Start Here summary messages so plan→exec transition logic still works after a destructive history rewrite.
- In non-plan modes, prepend a small system hint with the plan file path when a non-empty plan exists.
- Include the plan file path in the plan→exec transition helper message when plan content is injected.

Tests:
- `bun test src/browser/utils/messages/modelMessageTransform.test.ts`
- `bun test src/browser/components/tools/ProposePlanToolCall.test.tsx`
- `make static-check`

---

<details>

# Fix: “Replace all chat history” loses plan file path + plan→exec context

## Goals
- After **Replace all chat history / Start Here**, the next model turn (especially in **Exec** or custom agents) should still have enough context to:
  - know **where the plan file lives**, and
  - understand when it is effectively in a **plan → exec** workflow.
- Keep Replace History behavior predictable: still a single-message history rewrite, no surprise extra persisted messages.
- Avoid noisy, always-on prompt bloat.

<details>
<summary>Why this happens (recap)</summary>

- `useStartHere()` constructs a *new* single assistant message with `metadata.compacted: "user"` but **does not set `metadata.mode`** or any structured plan reference.
- `workspaceService.replaceHistory()` deletes `~/.mux/sessions/<workspaceId>/chat.jsonl` and appends only that summary message.
- Plan file path is only included via system instructions when `effectiveMode === "plan"`.
- Plan→exec transition glue only injects plan content when the last assistant message has `metadata.mode === "plan"`.

So after Replace History in exec/custom mode, the model often sees **plan text but no plan file path**.

</details>

---

## Recommended approach (A): Preserve source mode in Start Here + inject plan file path outside Plan mode
**Net product LoC estimate:** ~40–80

### A1) Preserve `metadata.mode` when generating the Start Here summary message
**Why:** ensures the existing plan→exec transition machinery can still fire after history replacement.

**Changes**
- `src/browser/hooks/useStartHere.ts`
  - Accept an optional `sourceMode?: string` (or `mode?: UIMode`) argument.
  - Set `metadata.mode = sourceMode` when building `summaryMessage`.
- Call sites:
  - `src/browser/components/Messages/AssistantMessage.tsx`
    - Pass the clicked message’s `message.metadata?.mode` into `useStartHere`.
  - `src/browser/components/tools/ProposePlanToolCall.tsx`
    - Pass `sourceMode: "plan"` when the Start Here action comes from the propose-plan UI.

**Edge cases / invariants**
- If no `sourceMode` is available, keep existing behavior (no mode set) to avoid inventing mode.

### A2) Always provide a plan file *path hint* in non-plan modes when a plan file exists
**Why:** makes plan location available even when you are *not* in plan mode and even when no transition glue runs.

**Changes**
- `src/common/utils/ui/modeUtils.ts`
  - Add a small helper like `getPlanFileHint(planFilePath: string, planExists: boolean): string | null`.
  - When `mode !== "plan"` and `planResult.exists === true`, prefix `effectiveAdditionalInstructions` with a short hint:
    - “A plan file exists at: … (read it if relevant; otherwise ignore).”

**Guardrails**
- Keep it **path-only** (no plan contents) to minimize token bloat.
- Only include it when `planResult.exists` (and ideally when non-empty) so it doesn’t spam new workspaces.

### A3) (Optional but cheap) Include the plan file path in the plan→exec transition injected message
**Net product LoC estimate (if included):** +10–25 (additional)

**Why:** ensures that even if system instructions get crowded, the path is near the injected `<plan>` content.

**Changes**
  - Extend `injectModeTransition(...)` call to also pass `planFilePath` when `planContentForTransition` is set.
- `src/browser/utils/messages/modelMessageTransform.ts`
  - Update the injected transition message template to include `Plan file path: ${planFilePath}` above the `<plan>` block.

---

### (B) Backend-only: inject exec-mode plan file hint (skip Start Here mode preservation)
**Net product LoC estimate:** ~15–30

- Implement only **A2**.
- Pros: simplest; fixes “no plan path” immediately.
- Cons: doesn’t restore the “plan→exec transition glue” semantics after Start Here.

### (C) Persist a plan reference into history during `replaceHistory()`
**Net product LoC estimate:** ~20–60

- `workspaceService.replaceHistory()` would either:
  - append a second synthetic message referencing the plan path, or
  - mutate the summary message content to add a plan footer.
- Pros: survives even if request-building injection changes.
- Cons: changes the meaning of “single message rewrite”; can surprise users / UI.

---

### Unit tests (fast)
- `src/browser/utils/messages/modelMessageTransform.test.ts`
  - Extend existing `injectModeTransition` tests:
    - assert that the injected plan→exec helper message includes the **plan file path** (if A3 is implemented).

### Component tests
- `src/browser/components/tools/ProposePlanToolCall.test.tsx`
  - Assert the Start Here action calls `replaceChatHistory` with `summaryMessage.metadata.mode === "plan"`.

### IPC/integration tests
- `tests/ipc/planCommands.test.ts`
  - Add a test that `replaceChatHistory` preserves `summaryMessage.metadata.mode` when provided.

---

## Rollout / UX notes
- This should be safe to ship behind no flags; the hint is tiny and only appears when a plan exists.
- If token usage becomes a concern, we can gate the exec-mode plan hint behind:
  - “plan file exists **and** non-empty” or
  - “workspace has ever been in plan mode” (requires additional tracking).

</details>

---

### DIFF
--- a/src/browser/utils/messages/StreamingMessageAggregator.ts
+++ b/src/browser/utils/messages/StreamingMessageAggregator.ts
@@ -950,6 +950,7 @@ export class StreamingMessageAggregator {
       historySequence: data.historySequence,
       timestamp: Date.now(),
       model: data.model,
+      mode: data.mode,
     });
 
     this.messages.set(data.messageId, streamingMessage);


ThomasK33:
@codex review

Addressed the review note about streamed messages not preserving mode for Start Here:
- `StreamingMessageAggregator.handleStreamStart` now copies `StreamStartEvent.mode` into the streaming message's `metadata.mode`.

This makes `message.mode` available immediately after streaming completes (no history reload needed), so Start Here preserves the original mode reliably.

PR #1366, opened 2025-12-29
🤖 fix: show nested tool calls during code_execution streaming

ethanndickson:
## Problem

When using `code_execution` (PTC), nested tool calls (e.g., `mux.bash()`, `mux.file_read()`) didn't appear in the UI until they completed. Users expected to see "executing..." while long-running nested tools were in progress, just like regular top-level tool calls.

## Root Cause

Two issues combined:

### 1. MessageId Mismatch

The backend was emitting nested tool events with the wrong `messageId`:

- `aiService.ts` created `assistantMessageId` for history and the `emitNestedEvent` callback
- `streamManager.ts` created a **separate** `messageId` for all stream events
- Frontend aggregator looked up messages by `messageId` - nested events with `assistantMessageId` couldn't find the message stored under `streamInfo.messageId`
- Events were silently dropped

### 2. React Reactivity

Even after fixing the messageId, the UI didn't update when nested tools completed:

- `handleToolCallEnd` mutated `nestedCall` objects in place
- React didn't detect changes because object references stayed the same
- Fix: use immutable update pattern (create new array + new objects)

## Solution

1. **Thread messageId through**: Pass `assistantMessageId` from `aiService.ts` to `streamManager.startStream()` so both use the same ID

2. **Immutable updates**: Replace in-place mutation with immutable update pattern in `handleToolCallEnd`

## Testing

- All existing tests pass
- Manual testing confirms nested tools now show "executing..." and update to completed state in real-time

---

### DIFF
--- a/src/browser/utils/messages/StreamingMessageAggregator.ts
+++ b/src/browser/utils/messages/StreamingMessageAggregator.ts
@@ -1269,15 +1269,23 @@ export class StreamingMessageAggregator {
     if (message) {
       // If nested, update in parent's nestedCalls array
       if (data.parentToolCallId) {
-        const parentPart = message.parts.find(
+        const parentIndex = message.parts.findIndex(
           (part): part is DynamicToolPart =>
             part.type === "dynamic-tool" && part.toolCallId === data.parentToolCallId
         );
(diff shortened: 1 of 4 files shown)


ethanndickson:
@codex review

PR #1369, opened 2025-12-30
🤖 refactor: unify tool rendering with shared registry

ethanndickson:
## Summary

Consolidate tool component lookup into a single registry, eliminating duplicate type guards and if-branches across `ToolMessage.tsx` and `NestedToolRenderer.tsx`.

## Changes

- **New `getToolComponent.ts`**: Declarative registry mapping tool names to components with Zod schema validation
- **`ToolMessage.tsx`**: 451 → 74 lines (removed 18 type guards, 19 if-branches)  
- **`NestedToolRenderer.tsx`**: 198 → 25 lines (now just calls registry)

## Before/After

**Adding a new tool UI component:**
- Before: Update 2 files with duplicate type guards and if-branches
- After: Add 1 line to the registry

**Net reduction: ~470 lines**

## Behavior Note

Nested tools (inside `code_execution`) now render all 18 registered tools instead of the previous 9. Tools like `file_edit_replace_lines`, `todo_write`, `status_set` now get proper UI in nested context instead of `GenericToolCall` fallback. This is an improvement.

## Testing

```
make typecheck  # ✅
make lint       # ✅
```

Existing Storybook coverage exercises both renderers:
- **Top-level tools (ToolMessage):** `App.bash.stories.tsx`, `App.chat.stories.tsx`, `App.task.stories.tsx`
- **Nested tools (NestedToolRenderer):** `App.codeExecution.stories.tsx` — includes nested `file_read`, `file_edit_replace_string`, `bash`

Run `make storybook` to visually verify tool rendering.

---

### DIFF
--- a/src/browser/components/tools/shared/getToolComponent.ts
+++ b/src/browser/components/tools/shared/getToolComponent.ts
@@ -45,6 +45,11 @@ interface ToolRegistryEntry {
 /**
  * Registry mapping tool names to their components and validation schemas.
  * Adding a new tool: add one line here.
+ *
+ * Note: Some tools (ask_user_question, propose_plan, todo_write, status_set) require
+ * props like workspaceId/toolCallId that aren't available in nested context. This is
+ * fine because the backend excludes these from code_execution sandbox (see EXCLUDED_TOOLS
+ * in src/node/services/ptc/toolBridge.ts). They can never appear in nested tool calls.
  */
 const TOOL_REGISTRY: Record<string, ToolRegistryEntry> = {
   bash: { component: BashToolCall, schema: TOOL_DEFINITIONS.bash.schema },


ethanndickson:
Re: the review comment about `ask_user_question` in nested context —

This is a non-issue. The backend **explicitly blocks** these tools from appearing inside `code_execution`:

```typescript
// src/node/services/ptc/toolBridge.ts
const EXCLUDED_TOOLS = new Set([
  "code_execution", // Prevent recursive sandbox creation
  "ask_user_question", // Requires UI interaction
  "propose_plan", // Mode-specific, call directly
  "todo_write", // UI-specific
  "status_set", // UI-specific
  ...
]);
```

So `ask_user_question` can never appear in nested tool calls — the sandbox doesn't expose it. The registry expansion only affects tools that CAN legitimately appear in nested context (like `file_edit_replace_lines` which was previously falling back to GenericToolCall).

ethanndickson:
@codex review

PR #1370, opened 2025-12-30
🤖 fix: roll up subagent costs when tasks complete via agent_report

ethanndickson:
## Problem

When subagents completed via `agent_report`, their costs were not being rolled up into the parent workspace. The Costs tab didn't increment by the cost of the subagent work.

## Root Cause

When `agent_report` completed, `handleAgentReport` called `finalizeAgentTaskReport({ stopStream: true })`, which aborted the stream. The AI SDK and AI providers don't report usage data for aborted streams—the `finish-step` event (which provides usage) only fires when the LLM finishes generating, not when we abort mid-stream. While we may have seen `finish-step` events earlier in the stream (e.g., after tool calls), we might not have seen one for a while—or at all if the stream is aborted before any step completes. Additionally, the stream metadata from the AI SDK (e.g., `streamResult.totalUsage`) doesn't include usage data if the stream is aborted, because the AI provider doesn't/can't send it when we terminate the connection early.

Since `recordUsage` in `streamManager.ts` depends on this data and is guarded by `if (!aborted)`, aborting the stream skipped usage recording entirely. When `workspaceService.remove()` later tried to roll up costs, `session-usage.json` was empty.

## Solution

1. **Don't abort the stream** - Let it finish naturally so `recordUsage` runs
2. **Add STOP message** - The `agent_report` tool result now includes a message telling the LLM to stop generating
3. **Defer cleanup** - Move `cleanupReportedLeafTask` from `handleAgentReport` to `handleStreamEnd`
4. **Fix emit ordering** - Move `stream-end` emit after `recordUsage` to ensure usage is persisted before cleanup handlers run

## Trade-off

If the LLM ignores the STOP message and generates extra tokens, we pay a small cost, but the accounting is now correct.

---

### DIFF
--- a/src/node/services/streamManager.ts
+++ b/src/node/services/streamManager.ts
@@ -1233,6 +1233,7 @@ export class StreamManager extends EventEmitter {
         // CRITICAL: This must happen BEFORE emitting stream-end, because event handlers
         // (e.g., handleStreamEnd in TaskService) may trigger cleanup that reads session-usage.json.
         // If we emit first, the cleanup could race and read stale data.
+        // All I/O operations are wrapped in try/catch so stream-end always emits even on errors.
         if (streamInfo.parts && streamInfo.parts.length > 0) {
           const finalAssistantMessage: MuxMessage = {
             id: streamInfo.messageId,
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
#1372 makes this not necessary, but this might be a better approach anyhow? :shrug:

PR #1380, opened 2025-12-30
🤖 perf: reduce review panel re-renders on read state changes

ammar-agent:
## Summary

Three optimizations to minimize re-renders in the code review panel during navigation and marking hunks as read:

### 1. Use refs for frequently-changing values in callbacks

The callbacks `handleToggleRead` and `handleMarkAsRead` previously depended on:
- `isRead` (changes on every mark-as-read)
- `selectedHunkId` (changes on every navigation)
- `filters.showReadHunks` (changes on filter toggle)

This caused callback recreation → `handleHunkToggleRead` recreation → **all HunkViewers re-render** (breaking `React.memo`).

**Fix:** Use refs (`isReadRef`, `selectedHunkIdRef`, `showReadHunksRef`) to access current values at invocation time without being dependencies. Callbacks now only depend on stable functions (`toggleRead`, `markAsRead`, `setSelectedHunkId`).

### 2. Fix onComposingChange inline function breaking React.memo

Previously:
```tsx
onComposingChange={(isComposing) => handleHunkComposingChange(hunk.id, isComposing)}
```

This created a **new function on every render**, bypassing `React.memo` for all ~100+ `HunkViewer` components.

**Fix:** Changed prop signature to `(hunkId, isComposing)`. `HunkViewer` wraps it internally with a stable `useCallback`. Parent passes `handleHunkComposingChange` directly.

### 3. Result

When navigating (j/k) or marking hunks as read:
- Parent component re-renders but callback references stay stable
- `React.memo` on `HunkViewer` correctly skips re-rendering unaffected hunks
- Only hunks with actual prop changes (`isSelected`, `isRead`) re-render

### Trace Analysis

From today's trace (`Trace-20251230T144048.json`), keydown events were taking ~366ms with significant time in React's commit/unmount phases - indicating excessive component re-mounting that these changes should address.

---

### DIFF
--- a/src/browser/components/RightSidebar/CodeReview/HunkViewer.tsx
+++ b/src/browser/components/RightSidebar/CodeReview/HunkViewer.tsx
@@ -32,8 +32,8 @@ interface HunkViewerProps {
   onRegisterToggleExpand?: (hunkId: string, toggleFn: () => void) => void;
   onReviewNote?: (data: ReviewNoteData) => void;
   searchConfig?: SearchHighlightConfig;
-  /** Callback when review note composition state changes */
-  onComposingChange?: (isComposing: boolean) => void;
+  /** Callback when review note composition state changes (receives hunkId for stable reference) */
+  onComposingChange?: (hunkId: string, isComposing: boolean) => void;
   /** Diff base for determining which git ref to read from */
   diffBase: string;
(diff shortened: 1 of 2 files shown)


ammario:
Manually verified on large diff — much faster

PR #1382, opened 2025-12-31
🤖 fix: prevent chat history recreation after compaction

ethanndickson:
## Problem

After compaction deletes `chat.jsonl` and writes a summary, the file is immediately overwritten with all original (pre-compaction) messages. This causes the compaction to have no effect - the full history is restored instead of being replaced with the summary.

**Observed behavior:**
```
11:29.971AM [HISTORY TRUNCATE 100%] File deleted
11:29.971AM [COMPACTION] History cleared
11:29.972AM [COMPACTION] Summary appended
11:29.974AM [HISTORY UPDATE] Write complete    ← Old messages written AFTER summary!
```

## Root Cause

Race condition between `streamManager.processStream()` and `compactionHandler.handleCompletion()`.

The stream-end handler in `agentSession.ts` uses fire-and-forget (`void handler()`):

```typescript
// agentSession.ts:715
void handler(payload as WorkspaceChatMessage);  // Fire-and-forget!
```

This means when `stream-end` is emitted:

1. `emit("stream-end")` triggers the compaction handler asynchronously
2. StreamManager continues immediately to `updateHistory()` 
3. Both operations race for `workspaceFileLocks`

**The problematic sequence:**

```
streamManager.ts:1249    this.emit("stream-end")           // Compaction starts async
                         ↓
agentSession.ts:715      void handler(...)                 // Fire-and-forget!
agentSession.ts:749      await compactionHandler...        // Yields on first await
                         
streamManager.ts:1265    await partialService.deletePartial()  // Continues immediately
streamManager.ts:1268    await historyService.updateHistory()  // Races with compaction
```

When `updateHistory()` acquires the lock first:
1. Reads all 4 messages from disk (including compaction response)
2. Finds and updates the message with matching historySequence
3. Writes all 4 messages back to disk
4. Compaction then runs: deletes file, writes summary
5. **Final state: just summary ✓**

But when compaction acquires the lock first (less common but observed):
1. Compaction clears history, writes summary with seq 0
2. `updateHistory()` runs: reads file (just summary), looks for seq 3
3. **Should fail** with "No message found"... but timing allows it to read stale data

The exact failure mode depends on timing, but the core issue is that `updateHistory()` can run concurrently with or after compaction, potentially restoring old messages.

## Solution

Move history update **before** emitting `stream-end`:

```typescript
// BEFORE:
this.emit("stream-end", streamEndEvent);          // 1. Compaction starts
await this.partialService.deletePartial(...);     // 2. Delete partial
await this.historyService.updateHistory(...);     // 3. Races with compaction

// AFTER:
await this.partialService.deletePartial(...);     // 1. Delete partial
await this.historyService.updateHistory(...);     // 2. Update history FIRST
this.emit("stream-end", streamEndEvent);          // 3. Then compaction runs
```

**Why this works:**
- **Normal streams**: history updated, then stream-end fires → works as before
- **Compaction streams**: history updated (writes compaction response), then stream-end fires, compaction runs and **replaces** entire history with summary → summary wins

The order ensures compaction always runs **after** the normal history update completes. Compaction's clear + append is the final operation, so the summary is the final state.

## Testing

- `make typecheck` passes
- `bun test src/node/services/compactionHandler.test.ts` - 29 tests pass
- `bun test src/node/services/streamManager.test.ts` - 7 tests pass
- Manual testing: `/compact` now correctly results in `chat.jsonl` containing only the summary message

---

### DIFF
--- a/src/node/services/streamManager.ts
+++ b/src/node/services/streamManager.ts
@@ -1246,9 +1246,10 @@ export class StreamManager extends EventEmitter {
         // where isStreaming() returns true while event handlers try to send new messages
         streamInfo.state = StreamState.COMPLETED;
 
-        this.emit("stream-end", streamEndEvent);
-
-        // Update history with final message (only if there are parts)
+        // Update history with final message BEFORE emitting stream-end
+        // This prevents a race condition where compaction (triggered by stream-end)
+        // clears history while updateHistory is still running, causing old messages
+        // to be written back after compaction completes.
         if (streamInfo.parts && streamInfo.parts.length > 0) {
           const finalAssistantMessage: MuxMessage = {
             id: streamInfo.messageId,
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

PR #1383, opened 2025-12-31
🤖 fix: improve await error message in code_execution

ethanndickson:
## Problem

When an AI writes `await mux.bash(...)` in a code_execution block, QuickJS returns an obtuse error:

```
Code analysis failed:
- expecting ';' (line 1)
```

This is confusing because the user doesn't understand why `await` is invalid or what's wrong.

## Solution

Detect the `await` expression pattern when QuickJS returns "expecting ';'" and replace the error with a clear, actionable message:

```
`await` is not supported - mux.* functions return results directly (no await needed)
```

### Why QuickJS Returns "expecting ';'"

In ECMAScript, `await` is only a reserved keyword **inside async functions**. Outside of them, it's a valid identifier for backward compatibility. So when QuickJS (in global/script mode) sees:

```javascript
const x = await foo()
```

It parses this as:
- `const x = await` — valid variable declaration (`await` as identifier)
- `foo()` — stray expression (unexpected token)

Hence the obtuse error: "expecting ';'" after the declaration. QuickJS is following the spec correctly—it just can't know you meant `await` as a keyword.

### Why Module Mode Doesn't Work

We first tried using `type: "module"` in QuickJS syntax validation. In ES module mode, `await` is reserved even at top-level, which gave the clear error `"unexpected 'await' keyword"`.

**Problem 1: Module mode breaks return values**

Module mode returns the module's exports object, not expression values:

```
(function() { return 42; })()
  global mode: 42 ✅
  module mode: {} ❌ (empty exports object)
```

This would break all code_execution return values since we rely on the IIFE returning the result.

**Problem 2: Parity gap with runtime**

Module mode creates a gap where code passes static analysis but fails at runtime:

| Construct | Module Analysis | Global Runtime | Result |
|-----------|-----------------|----------------|--------|
| `import.meta.url` | ✅ Parses | ❌ Fails | False positive |
| `export const x` | ✅ Parses | ❌ Fails | False positive |

Module mode is designed for loading ES modules (files with import/export), not evaluating code snippets. We need eval/REPL semantics where expressions have values.

### Why Not TypeScript AST Detection?

We considered walking the TypeScript AST to find AwaitExpression nodes outside async functions (~35 lines). But:

1. The heuristic is **5 lines vs 35 lines**
2. The heuristic is **correct for all error cases**
3. AST detection would need to handle edge cases (async inner functions, etc.)

### Why the Heuristic Works

The pattern `/\bawait\s+\w/` matches `await` followed by whitespace and a word character—i.e., an await expression like `await foo()`.

| Code | Pattern Matches | Has Error | Result |
|------|-----------------|-----------|--------|
| `await foo()` | ✅ | ✅ "expecting ';'" | **Enhanced** ✅ |
| `const await = 1` | ❌ | ❌ (parses fine) | No change |
| `obj.await()` | ❌ | ❌ (parses fine) | No change |
| `// await foo` | ✅ | ❌ (parses fine) | No change |
| `async () => await x` | ✅ | ❌ (parses fine) | No change |

Key insight: **false positives don't matter** because we only enhance when there's already an error. If the code parses fine (async functions, comments, strings), there's no error to enhance.

## Changes

- `src/node/services/ptc/staticAnalysis.ts`: Revert module mode, add await error enhancement heuristic
- `src/node/services/ptc/staticAnalysis.test.ts`: Add tests for clear await error and async function allowance

<details>
<summary>Original Investigation & MREs</summary>

## Problem

When an AI writes `await mux.bash(...)`, QuickJS returns an obtuse error:
```
Code analysis failed:
- expecting ';' (line 1)
```

## Root Cause

Code is wrapped in a **synchronous** IIFE `(function() { ... })()`. The `await` keyword is only valid inside `async` functions, so QuickJS treats it as an identifier and fails to parse.

### MRE: Why "expecting ';'" occurs

In a non-async context, `await` is parsed as a regular identifier. So `const x = await 42` is parsed as `const x = await` (declaration) followed by stray `42` (unexpected token).

```typescript
// mre-syntax.ts
import { QuickJSRuntime } from "./src/node/services/ptc/quickjsRuntime";

async function main() {
  const runtime = await QuickJSRuntime.create();
  const ctx = (runtime as any).ctx;

  const testCases = [
    { name: "sync IIFE + await (our wrapper)", 
      code: `(function() { const x = await 42; return x; })()` },
    { name: "await as identifier (valid!)", 
      code: `const await = 42; await;` },
    { name: "async IIFE + await (works)", 
      code: `(async function() { const x = await 42; return x; })()` },
  ];

  for (const { name, code } of testCases) {
    const result = ctx.evalCode(code, "test.js", { compileOnly: true });
    if (result.error) {
      console.log(`${name}: SyntaxError: ${ctx.dump(result.error).message}`);
      result.error.dispose();
    } else {
      console.log(`${name}: OK`);
      result.value.dispose();
    }
  }
  runtime.dispose();
}
main();
```

**Output:**
```
sync IIFE + await (our wrapper): SyntaxError: expecting ';'
await as identifier (valid!): OK
async IIFE + await (works): OK
```

### What Works

| Pattern | Status |
|---------|--------|
| `mux.tool1(); mux.tool2();` (sync calls) | ✅ Works |
| `await Promise.resolve(42)` | ✅ Works |
| `await Promise.all([...])` | ✅ Works |
| Multiple `await` on native Promises only | ✅ Works |

### What Breaks

| Pattern | Status |
|---------|--------|
| `await 42; mux.tool();` | ❌ **Hangs** |
| `await Promise.resolve(); mux.tool();` | ❌ **Hangs** |
| Any `mux.*` call after any `await` | ❌ **Hangs** |

### MRE: Asyncify Re-entrancy Failure

```typescript
// mre-asyncify.ts
import { QuickJSRuntime } from "./src/node/services/ptc/quickjsRuntime";

async function main() {
  const runtime = await QuickJSRuntime.create();
  runtime.registerFunction("muxTool", () => Promise.resolve("result"));
  const ctx = (runtime as any).ctx;

  const code = `
(async function() {
  const before = muxTool();     // Works - no prior await
  await Promise.resolve();       // Yield to microtask queue
  const after = muxTool();       // Fails - asyncify can't suspend twice
  return before + "-" + after;
})()`;

  const r = await ctx.evalCodeAsync(code);
  
  // Run pending jobs to resume async function
  for (let i = 0; i < 20; i++) {
    ctx.runtime.executePendingJobs().dispose();
  }
  
  const state = ctx.getPromiseState(r.value);
  console.log("Promise state:", state.type);
  // Output: "Promise state: pending" - STUCK FOREVER
  
  r.value.dispose();
  runtime.dispose();
}
main();
```

**Output:**
```
Promise state: pending
STUCK: Promise never resolves (asyncify re-entrancy failure)
```

### Why It Fails

From asyncify docs:
> "Asyncified functions must never call other asyncified functions... because the stack cannot be unwound twice."

When you `await` anything:
1. The async function suspends (yields to microtask queue)
2. `executePendingJobs()` resumes it
3. If resumed code calls `mux.*`, asyncify tries to suspend again → hangs

This means **async IIFE is fundamentally incompatible** with mux.* tools being called after any await.

</details>

<details>
<summary>Module Mode Investigation Details</summary>

### Global vs Module Mode Behavior Comparison

| Behavior | Global Mode | Module Mode |
|----------|-------------|-------------|
| IIFE return value | ✅ Returns value | ❌ Returns `{}` |
| `await` as identifier | ✅ Valid | ❌ Error |
| `await` in sync fn error | `expecting ';'` | `unexpected 'await' keyword` |
| `with` statement | ✅ Valid | ❌ Error (strict) |
| Octal literals (`010`) | ✅ Valid | ❌ Error (strict) |
| Duplicate params | ✅ Valid | ❌ Error (strict) |
| `import.meta` in IIFE | ❌ Error | ✅ Valid |
| `export const x` | ❌ Error | ✅ Valid |

### Module Mode Return Value Test

```typescript
// Module mode returns exports object, not expression values
const testCases = [
  { name: "IIFE return", code: `(function() { return 42; })()` },
  { name: "raw expression", code: `42` },
  { name: "export default", code: `export default 42` },
];

// Results:
// IIFE return:     global → 42, module → {}
// raw expression:  global → 42, module → {}
// export default:  global → ERROR, module → { default: 42 }
```

Module mode is designed for loading ES modules where the "result" is the exports object. We need eval/REPL semantics where expressions have values.

### Why strict: true Doesn't Help

Strict mode alone doesn't make `await` reserved. Per ES spec:
- `await` is reserved in **module goal** only
- `await` is NOT reserved in **script goal**, even in strict mode

```typescript
// All give "expecting ';'" for await in sync function:
{ }                    // default
{ strict: true }       // strict mode
// Only this gives "unexpected 'await' keyword":
{ type: "module" }     // module mode (but breaks return values)
```

</details>

---

### DIFF
--- a/src/node/services/ptc/staticAnalysis.test.ts
+++ b/src/node/services/ptc/staticAnalysis.test.ts
@@ -53,9 +53,11 @@ describe("staticAnalysis", () => {
       expect(result.valid).toBe(false);
       expect(result.errors).toHaveLength(1);
       expect(result.errors[0].type).toBe("syntax");
-      // Should say "unexpected 'await' keyword" not "expecting ';'"
+      // Should give clear message about await, not obtuse "expecting ';'"
       expect(result.errors[0].message).toContain("await");
+      expect(result.errors[0].message).toContain("not supported");
     });
+
     test("allows return statements (wrapped in function)", async () => {
       const result = await analyzeCode(`
         const x = mux.fileRead("test.txt");
--- a/src/node/services/ptc/staticAnalysis.ts
+++ b/src/node/services/ptc/staticAnalysis.ts
@@ -129,11 +129,8 @@ async function validateSyntax(code: string): Promise<AnalysisError | null> {
   const wrappedCode = `(function() { ${code} })`;
 
   // Use evalCode with compile-only flag to parse without executing.
-  // Module mode makes `await` a reserved keyword, giving clearer errors
-  // like "unexpected 'await' keyword" instead of obtuse "expecting ';'".
   const result = ctx.evalCode(wrappedCode, "analysis.js", {
     compileOnly: true,
-    type: "module",
   });
 
   if (result.error) {
@@ -141,8 +138,16 @@ async function validateSyntax(code: string): Promise<AnalysisError | null> {
     result.error.dispose();
 
     // QuickJS error object has: { name, message, stack, fileName, lineNumber }
-    const message =
+    let message =
       typeof errorObj.message === "string" ? errorObj.message : JSON.stringify(errorObj);
+
+    // Enhance obtuse "expecting ';'" error when await expression is detected.
+    // In non-async context, `await foo()` parses as identifier `await` + stray `foo()`,
+    // giving unhelpful "expecting ';'". Detect this pattern and give a clearer message.
+    if (message === "expecting ';'" && /\bawait\s+\w/.test(code)) {
+      message =
+        "`await` is not supported - mux.* functions return results directly (no await needed)";
+    }
     const rawLine = typeof errorObj.lineNumber === "number" ? errorObj.lineNumber : undefined;
 
     // Only report line if it's within agent code bounds.


ethanndickson:
@codex review

The static analysis being stricter than runtime is intentional here. The patterns that module mode rejects but runtime would accept are:

1. **`with` statement** - Deprecated, causes confusing scope issues, forbidden in strict mode everywhere
2. **Octal literals (`010`)** - Legacy syntax, confusing (`010 === 8`), AI shouldn't generate these
3. **Duplicate params (`function(a, a)`)** - Always a bug
4. **`await` as identifier** - Confusing when `await` expression doesn't work

These are all patterns that:
- Should never appear in AI-generated code
- Are bugs or deprecated features if they do appear
- Would cause confusion even if they technically work

The goal is to catch mistakes early with clear errors. If code somehow uses these patterns and passes static analysis today, it's because we weren't checking—not because it's intentionally supported.

That said, if we want perfect parity, we could switch runtime to module mode too. But runtime execution with `type: "module"` might have other semantic differences (top-level await becoming valid, `this` binding changes, etc.) that need investigation.

ethanndickson:
@codex review

PR #1410, opened 2026-01-01
🤖 fix: ignore stale origin/HEAD in git status scripts

ammar-agent:
## Context

#1414 (merged) unified the review-base selector with the git status indicator so the diff view and divergence numbers use the same base.

However, mux workspaces can be created from git bundles, and those clones can inherit a stale `refs/remotes/origin/HEAD` (e.g. pointing at a feature branch). If we use `origin/HEAD` as the base (explicitly or implicitly), ahead/behind can be computed against the wrong remote ref.

## What this PR changes (follow-up)

- **`generateGitStatusScript()`**
  - Treats `origin/HEAD` as **auto-detect** even if configured as the base ref
  - Uses configured `origin/<branch>` when present (with main↔master fallback if the ref doesn't exist locally)
  - When auto-detecting, ignores suspicious `origin/HEAD` values and prefers `origin/main` or `origin/master`
- **`GIT_FETCH_SCRIPT`**
  - Derives the remote default branch + SHA via `git ls-remote --symref origin HEAD` (instead of local `origin/HEAD`)
- **Storybook mock**
  - Detects the consolidated git status script via output markers (less brittle)
- **Tests**
  - Adds unit coverage for `parseGitShowBranchForStatus`

## Regression risk

- Auto-detection uses more heuristics; repos without `origin/main` or `origin/master` may fall back differently than before.
- `GIT_FETCH_SCRIPT` now depends on `git ls-remote --symref origin HEAD`; if the remote is unreachable/slow, fetch will be skipped and statuses may remain stale until a later refresh/manual fetch.
- If a user intentionally configured `origin/HEAD` as a base ref, it will now be treated as auto-detect (by design) to avoid stale/incorrect divergence.

## Testing

- `make static-check`

---

### DIFF
--- a/src/browser/components/hooks/useGitBranchDetails.ts
+++ b/src/browser/components/hooks/useGitBranchDetails.ts
@@ -184,10 +184,12 @@ export function useGitBranchDetails(
 # Get current branch
 CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo "main")
 
-# Get primary branch (main or master)
-PRIMARY_BRANCH=$(git branch -r 2>/dev/null | grep -E 'origin/(main|master)$' | head -1 | sed 's@^.*origin/@@' || echo "main")
-
-if [ -z "$PRIMARY_BRANCH" ]; then
+# Get primary branch (main or master) - check refs directly for reliability
+if git rev-parse --verify "refs/remotes/origin/main" >/dev/null 2>&1; then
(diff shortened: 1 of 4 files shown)


ammar-agent:
@codex review

Addressed feedback:
1. Simplified the review base logic - git status now just uses the global default (`review-default-base`, which defaults to `origin/main`)
2. Removed the unnecessary project-scoping attempt since the review panel stores per-workspace and the global default is sufficient for the git status indicator

PR #1423, opened 2026-01-02
🤖 refactor: simplify mode system - use agentId for tool policy

ammar-agent:
## Summary

Simplifies the mode/agentId architecture by removing redundant tool policy computation from the frontend. Tool policy is now derived solely from agent definitions on the backend.

### Core Simplification
- **Remove `modeToToolPolicy`** - The backend's `resolveToolPolicyForAgent` already computes tool policy from agent definitions (exec.md, plan.md, custom agents). Frontend was sending redundant/duplicate policy.
- **ContinueMessage uses `agentId` instead of `mode`** - Continue messages (after compaction) now carry the full agent identity, enabling proper tool resolution for custom agents.
- **Remove `getModeKey`** - Legacy storage key no longer used after previous refactors.

### Voice Input Cleanup
- `VoiceInputButton` and `RecordingOverlay` now take `agentColor` prop instead of `UIMode`, using inline styles with CSS variables. This properly supports custom agent colors.

### Files Changed
- 24 files, -155 net lines

## Regression Risk Assessment

**Medium Risk** - This removes a redundant code path, but it touches a few critical flows:

| Area | Risk | Mitigation |
|------|------|------------|
| **Tool availability** | Medium | Backend `resolveToolPolicyForAgent` already handles this - we're removing the duplicate frontend policy. Agent definitions specify tools (exec removes `propose_plan`, plan enables it). |
| **Continue messages after compaction** | Medium | Continue messages now persist `agentId` (instead of `mode`). **Regression risk:** existing sessions may contain legacy continue messages with only `mode` set; during compaction the active agentId is temporarily `compact` (tools denied), and if we accidentally fall back to that the follow-up message would run with tools disabled. **Mitigation:** backend derives agentId from legacy `mode` when missing and explicitly avoids inheriting `compact`; unit test added to cover this behavior. |
| **Voice input colors** | Low | UI-only change - uses agent's `uiColor` directly instead of mapping from mode. |
| **Legacy storage migration** | Low | `getModeKey` was already unused after previous agentId migration. |

### Recommended Testing
1. **Compaction with continue message** - Verify `/compact Continue with my changes` queues and sends correctly *with tool access intact*.
2. **Plan mode tools** - Verify `propose_plan` is available in plan mode, disabled in exec.
3. **Custom agents** - If using custom agents, verify they get correct tool access based on their definitions.
4. **Voice input** - Visual check that recording overlay shows correct agent color.

---
_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `low`_

### DIFF
--- a/src/browser/contexts/ModeContext.tsx
+++ b/src/browser/contexts/ModeContext.tsx
@@ -11,11 +11,7 @@ import React, {
 
 import { useAPI } from "@/browser/contexts/API";
 import { AgentProvider } from "@/browser/contexts/AgentContext";
-import {
-  readPersistedString,
-  updatePersistedState,
-  usePersistedState,
-} from "@/browser/hooks/usePersistedState";
+import { usePersistedState } from "@/browser/hooks/usePersistedState";
 import { CUSTOM_EVENTS, createCustomEvent } from "@/common/constants/events";
 import { matchesKeybind, KEYBINDS } from "@/browser/utils/ui/keybinds";
 import {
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

Addressed the legacy compaction follow-up regression:
- When `continueMessage.agentId` is missing, queued follow-ups now fall back to legacy `continueMessage.mode` (exec/plan) instead of `options.agentId="compact"`.
- Added a unit test covering the legacy mode fallback.

PR #1433, opened 2026-01-03
🤖 fix: colorize init hook stderr lines

ammar-agent:
Stop prefixing init-hook stderr lines with `ERROR:` and render them with error styling instead.

- Preserve original line content and tag each line with `isError`
- Render stderr lines in `InitMessage` with error coloring

Validation:
- `make static-check`

---


ammar-agent:
@codex review

Addressed mobile init hook rendering: init output lines are now `{ line, isError }` objects, and the mobile renderer uses `isError` for styling instead of `startsWith("ERROR:")`.

PR #1443, opened 2026-01-04
🤖 fix: serialize idle compaction to prevent thundering herd

ammar-agent:
When the hourly idle compaction check finds multiple eligible workspaces, they now compact one at a time instead of all simultaneously.

## Changes
- Added queue and isRunning state to `useIdleCompactionHandler`
- `processNextInQueue()` serializes execution
- Updated test to verify serialization behavior

---


ammar-agent:
@codex review

PR #1452, opened 2026-01-05
🤖 fix: reject compaction summaries that look like tool call JSON

ammar-agent:
## Problem

When compaction runs, tools are disabled. However, if the model still "wants" to make a tool call, it can output the tool call input as plain text. This corrupted text becomes the compaction summary, which when saved as the starting message creates a malformed workspace.

Example of corrupted summary that was observed:
```json
{"script":"cd tpred && sed -n '405,520p' train/trainer.py | nl -ba | sed -n '1,140p'","timeout_secs":10,"run_in_background":false,"display_name":"Inspect trainer sampler changes"}
```

## Solution

Reject any compaction summary that is purely a JSON object. Valid summaries should be prose text describing the conversation, not JSON blobs. This general approach:

- Catches all current and future tool call leaks without pattern-specific checks
- Is simple and decoupled from tool definitions
- Only rejects pure JSON objects - prose containing embedded JSON snippets is still accepted

The rejection:
- Returns `false` from `handleCompletion()` with a warning log
- Does NOT mark the compaction as processed, allowing the user to retry
- Prevents corrupted history from being saved

## Testing

Added tests covering:
- Rejects pure JSON objects (tool call leaks)
- Rejects arbitrary JSON structures
- Accepts normal prose summaries
- Accepts prose containing embedded JSON snippets
- Does not reject JSON arrays

---

### DIFF
--- a/src/node/services/compactionHandler.test.ts
+++ b/src/node/services/compactionHandler.test.ts
@@ -729,8 +729,8 @@ describe("CompactionHandler", () => {
     });
   });
 
-  describe("Tool Call JSON Validation", () => {
-    it("should reject compaction when summary looks like bash tool call JSON", async () => {
+  describe("Raw JSON Object Validation", () => {
+    it("should reject compaction when summary is a raw JSON object", async () => {
       const compactionRequestMsg = createMuxMessage("compact-req-1", "user", "/compact", {
         historySequence: 0,
         timestamp: Date.now() - 1000,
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Simplified the approach: instead of detecting specific tool patterns, now rejects any summary that is purely a JSON object. Valid summaries should be prose text describing the conversation, not JSON blobs.

PR #1453, opened 2026-01-05
🤖 fix: focus Submit button on summary page for Enter key support

ethanndickson:
When answering agent questions, the Summary page now auto-focuses the Submit button so pressing Enter submits the answers.

---


ethanndickson:
This is to support typing a response in the Other box, and then being able to press enter twice to submit.

PR #1465, opened 2026-01-05
🤖 feat: add 'when to use' guidance to code_execution tool

ethanndickson:
Adds 'When to use' guidance to the code_execution (PTC) tool definition, placed prominently before the type definitions.

This is an attempt to get models to use PTC more often by explicitly telling them when it's beneficial - specifically when making 2+ tool calls, especially dependent ones where later calls use earlier results.

---

Also fixes a flaky test failure (`spawn bash ENOENT`) in `tests/ui/sections.integration.test.ts`.

**Root cause:** Race between `GitStatusStore` polling `executeBash` and test cleanup calling `workspace.remove()`. The `removingWorkspaces` guard passes, but cwd gets deleted before `spawn()` runs.

**Fix:** Catch ENOENT in `executeBash` and return graceful error when workspace is being removed. No lock needed - `executeBash` is called frequently (git status polling), so we avoid overhead on the hot path and only handle the specific race condition.

---

### DIFF
--- a/src/node/services/workspaceService.ts
+++ b/src/node/services/workspaceService.ts
@@ -2320,6 +2320,15 @@ export class WorkspaceService extends EventEmitter {
 
       return Ok(result);
     } catch (error) {
+      // Race with workspace removal: cwd was deleted after the guard check but before/during spawn.
+      // Return the same error as the guard so callers handle it uniformly.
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

PR #1484, opened 2026-01-06
🤖 fix: wait for init before creating temp directory

ethanndickson:
## Problem

When workspaces are created, `sendMessage()` can be called immediately after the async `initWorkspace()` fire-and-forget. For runtimes that need initialization (dev containers, SSH), this caused `createTempDirForStream()` to run mkdir before the runtime was ready:

```
Failed to stream message: Failed to create temp directory ~/.mux-tmp/db19fe9c: exit code 1
```

## Fix

1. **Add `waitForInit()` before temp dir creation** — uses the same mechanism that protects tools via `wrapWithInitWait()`

2. **Make `waitForInit()` abort-aware** (Codex review) — if user cancels during init wait, return immediately instead of waiting up to 5 minutes

3. **Fix fork path missing `initWorkspace()` call** — the fork path called `startInit()` but never called `endInit()`, leaving init state as "running" forever. This caused integration tests to timeout.

4. **Add `timeoutId.unref()`** — defense-in-depth so the 5-min timeout does not keep Node alive (allows tests to exit cleanly)

## Why This Is Bulletproof

`waitForInit()` has well-defined behavior for all cases:

| Scenario | Behavior |
|----------|----------|
| No init state | Returns immediately (backwards compat, local workspaces) |
| Init already completed | Returns immediately |
| Init running | Waits up to 5 min, then proceeds anyway |
| Init promise rejected | Catches error, proceeds anyway |
| Abort signal fired | Returns immediately |

**It never throws and never blocks forever.** Tools always proceed—if init failed, they will fail with their own clear errors (file not found, etc.) rather than hanging.

---

### DIFF
--- a/src/node/services/initStateManager.ts
+++ b/src/node/services/initStateManager.ts
@@ -406,6 +406,8 @@ export class InitStateManager extends EventEmitter {
           );
           resolve();
         }, INIT_TIMEOUT_MS);
+        // Don't keep Node alive just for this timeout (allows tests to exit)
+        timeoutId.unref();
       });
 
       const abortPromise = new Promise<void>((resolve) => {
--- a/src/node/services/workspaceService.ts
+++ b/src/node/services/workspaceService.ts
@@ -1405,8 +1405,26 @@ export class WorkspaceService extends EventEmitter {
       });
 
       if (!forkResult.success) {
+        initLogger.logComplete(-1);
         return Err(forkResult.error ?? "Failed to fork workspace");
       }
+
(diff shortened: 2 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Made `waitForInit()` abort-aware per your feedback:
- Added optional `abortSignal` param
- Races abort against init completion and timeout
- Proper cleanup of timeout and abort listener in `finally` block
- Added tests for abort behavior

ethanndickson:
@codex review

PR #1486, opened 2026-01-06
🤖 fix: prevent unhandled rejection from void exitCode.finally()

ethanndickson:
## Problem

The test `sections.integration.test.ts` was flaky with `spawn bash ENOENT` errors. When a workspace is removed while GitStatusStore has an in-flight bash call, the spawn fails and `exitCode` rejects.

The root cause was `void exitCode.finally(() => clearTimeout(...))` in `LocalBaseRuntime.ts`:

```javascript
void exitCode.finally(() => clearTimeout(timeoutHandle));
```

This is a footgun because `.finally()` returns a **new promise** that rejects if the original rejects. The `void` discards that promise with no handler attached, so Node fires `unhandledRejection`.

## Fix

Use `.catch(() => undefined).finally(cleanup)` instead. The `.catch()` swallows the rejection, so `.finally()` sees a resolved promise and its returned promise won't reject.

## Changes

| File | Change |
|------|--------|
| `LocalBaseRuntime.ts` | `exitCode.catch(() => undefined).finally(...)` - **the actual fix** |
| `SSHRuntime.ts` | Same pattern (consistency) |
| `bash.ts` | Added `void foregroundCompletion.catch()` for defense-in-depth |
| `workspaceService.ts` | Removed dead ENOENT-handling code (bash.ts returns error results, doesn't throw) |

## Verification

Tested by modifying the test to remove the workspace before `cleanupView()` (forcing the race):
- No fixes: **FAIL**
- bash.ts fix only: **FAIL**
- LocalBaseRuntime fix only: **PASS**

---

### DIFF
--- a/src/node/runtime/LocalBaseRuntime.ts
+++ b/src/node/runtime/LocalBaseRuntime.ts
@@ -181,7 +181,7 @@ export abstract class LocalBaseRuntime implements Runtime {
       }, options.timeout * 1000);
 
       // Clear timeout if process exits naturally
-      void exitCode.finally(() => clearTimeout(timeoutHandle));
+      void exitCode.catch(() => undefined).finally(() => clearTimeout(timeoutHandle));
     }
 
     return { stdout, stderr, stdin, exitCode, duration };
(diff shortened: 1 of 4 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

PR #1487, opened 2026-01-06
🤖 fix: use stopWhen to cleanly stop stream when message is queued

ethanndickson:
## Problem

When a user queues a message while the agent is streaming, the existing `tool-call-end` handler immediately called `sendQueuedMessages()`, which cleared the queue and tried to start a new stream. But the old stream was still running—`stopWhen` hadn't been checked yet—creating a race condition.

## Solution

Use AI SDK's `stopWhen` callback to check for queued messages at step boundaries:

1. Thread a `hasQueuedMessage` callback from AgentSession through AIService to StreamManager
2. Add the callback to `stopWhen: [stepCountIs(100000), () => hasQueuedMessage?.() ?? false]`
3. Remove `sendQueuedMessages()` from `tool-call-end`—only `stream-end` dispatches now

## Why this is better

**Before:** `tool-call-end` → clear queue → `stopWhen` checks empty queue → stream continues → `ensureStreamSafety` eventually kills it

**After:** `tool-call-end` → `stopWhen` checks non-empty queue → stream stops gracefully → `stream-end` → dispatch queued message

The stream stops cleanly at the step boundary using AI SDK's built-in mechanism, with a single dispatch point in `stream-end`.

---

### DIFF
--- a/src/node/services/agentSession.ts
+++ b/src/node/services/agentSession.ts
@@ -703,7 +703,8 @@ export class AgentSession {
       changedFileAttachments.length > 0 ? changedFileAttachments : undefined,
       postCompactionAttachments,
       options?.experiments,
-      options?.disableWorkspaceAgents
+      options?.disableWorkspaceAgents,
+      () => !this.messageQueue.isEmpty()
     );
   }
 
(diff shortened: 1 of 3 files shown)


ethanndickson:
@codex review

PR #1489, opened 2026-01-06
🤖 fix: add error handling for image processing and safe record access

EhabY:
## Summary
  - Add `.catch()` handlers to `processImageFiles` in paste/drop handlers - previously errors were silently swallowed due to `void` fire-and-forget pattern
  - Use optional chaining for `MODEL_ABBREVIATIONS` in case models are added without updating the record


src/common/constants/knownModels.ts
```diff
@@ -174,5 +174,5 @@ export const KNOWN_MODEL_OPTIONS = Object.values(KNOWN_MODELS).map((model) => ({
 /** Tooltip-friendly abbreviation examples: show representative shortcuts */
 export const MODEL_ABBREVIATION_EXAMPLES = (["opus", "sonnet"] as const).map((abbrev) => ({
   abbrev,
-  displayName: formatModelDisplayName(MODEL_ABBREVIATIONS[abbrev].split(":")[1]),
+  displayName: formatModelDisplayName(MODEL_ABBREVIATIONS[abbrev]?.split(":")[1] ?? abbrev),
```
ethanndickson, line 177: Looks superfluous

ethanndickson (approved):
Thanks!

src/common/constants/knownModels.ts
EhabY: It is a bit, thing is all `Record<string, X>` can return undefined if the key is not known. This essentially handles it by defaulting to the abbreviation

PR #1494, opened 2026-01-07
🤖 fix: GitInitBanner not showing due to StrictMode race condition

ethanndickson:
## Summary

The GitInitBanner wasn't appearing on the project creation screen for non-git directories.

## Root Cause

The `isMountedRef` pattern was getting poisoned by React StrictMode's double-render cycle:

1. First mount starts async `listBranches` call
2. StrictMode unmounts → cleanup sets `isMountedRef.current = false`
3. StrictMode remounts with a **new** ref (initialized to `true`)
4. First mount's async call completes, checks the **old** ref which is `false`, bails out
5. Second mount's async call may hit same race

The shared ref approach can't distinguish "unmounted for real" from "StrictMode cleanup".

## Fix

Replace shared `isMountedRef` with a local `cancelled` flag inside the effect. Each effect invocation captures its own `cancelled` variable via closure, so the first cleanup only affects the first invocation.

Also adds `branchRefreshKey` state to trigger re-fetch after git init succeeds.

## Testing

- Manually verified: added non-git directory as project, banner now appears
- Clicked "Run git init" → success message shows, banner disappears

---

### DIFF
--- a/src/browser/components/ProjectPage.tsx
+++ b/src/browser/components/ProjectPage.tsx
@@ -47,39 +47,43 @@ export const ProjectPage: React.FC<ProjectPageProps> = ({
   // Git repository state for the banner
   const [branchesLoaded, setBranchesLoaded] = useState(false);
   const [hasBranches, setHasBranches] = useState(true); // Assume git repo until proven otherwise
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

PR #1497, opened 2026-01-07
🤖 fix: prevent retry UI during workspace initialization

ethanndickson:
## Summary

Two fixes for Docker runtime error handling during message streaming:

### Commit 1: Move waitForInit earlier to prevent retry UI during init

Moves `waitForInit()` earlier in `streamMessage()` so that initialization completes before we attempt to stream. This prevents the retry barrier from flickering during workspace initialization - the "starting..." indicator stays until init completes.

### Commit 2: Check container exists before streaming

Adds `ensureReady()` check after `waitForInit()` to detect when a container has been deleted mid-session. Previously this showed a confusing error:

> Failed to stream message: Failed to create temp directory ~/.mux-tmp/...: exit code 1

Now it surfaces the actual docker error (e.g., "Error response from daemon: No such container: ...").

## Future Work

Container creation UX needs revision to surface errors better when the image is invalid or container creation fails. The likely approach is to not create the workspace until the container exists, but that requires streaming the init logs elsewhere (currently they're shown in the workspace chat).

---

### DIFF
--- a/src/browser/utils/messages/retryEligibility.ts
+++ b/src/browser/utils/messages/retryEligibility.ts
@@ -88,14 +88,19 @@ export function hasInterruptedStream(
     if (elapsed < PENDING_STREAM_START_GRACE_PERIOD_MS) return false;
   }
 
-  // Don't show retry barrier if workspace init is still running
-  // The backend is intentionally waiting for init to complete before starting the stream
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

PR #1499, opened 2026-01-07
🤖 fix: use DOM-based workspace navigation to prevent drift

ammar-agent:
## Problem

Ctrl+J/K navigation was skipping workspaces inside sections (regression from the sections feature). The root cause: navigation computed visibility via `getAllVisibleWorkspaces()` while the sidebar had its own rendering logic. When sections were added to the sidebar, navigation wasn't updated.

## Solution

Replace the parallel visibility computation with DOM-based navigation that reads actual rendered state:

```tsx
const els = document.querySelectorAll("[data-workspace-id]");
const visibleIds = Array.from(els).map((el) => el.getAttribute("data-workspace-id")!);
```

This is **impossible to drift** — navigation now follows whatever the sidebar renders.

## Why this approach

| Approach | LoC | Drift-proof | Complexity |
|----------|-----|-------------|------------|
| Shared structure (attempted) | +179 | By construction | High |
| Parity test | +30 | By test | Medium |
| **DOM-based (this PR)** | **-25** | **By construction** | **Low** |

The `data-workspace-id` attribute already exists on every workspace item, so no sidebar changes needed.

## Changes

- Navigation queries DOM for visible workspace IDs in render order
- Removed `getAllVisibleWorkspaces` import and related dependencies
- Reduced callback deps to just `workspaceMetadata` map lookup

**Net: -25 LoC** (12 insertions, 37 deletions)

---


ammar-agent:
@codex review

PR #1501, opened 2026-01-07
🤖 fix: normalize Windows runtimeTempDir paths

ibetitsmike:
On Windows local runtimes, Git Bash/MSYS `pwd` can return paths like `/c/Users/...`, which Node treats as `C:\\c\\...` when used with `fs`, breaking temp-file writes (notably bash tool `overflow_policy: "tmpfile"`).

### What changed
- Stream temp dir creation now uses `runtime.resolvePath("~/.mux-tmp/<token>")` instead of `cd ... && pwd`.
- Windows drive/UNC paths are normalized to forward slashes (`C:/...`) so downstream POSIX joins and basename logic behave.

### Additional Windows dev-server fixes
- Makefile: `ESBUILD_CLI_FLAGS` banner is now “shell-safe” (no whitespace; avoids argument-splitting issues seen on Windows when invoking esbuild via `bun x`).
- Makefile: replace non-existent `npmx` helper with `npm x` in Windows `dev-server` and `typecheck` targets.

### Validation
- `make static-check`

---
_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$12.09`_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,10 +161,10 @@ dev-server: node_modules/.installed build-main ## Start server mode with hot rel
 	@echo ""
 	@echo "For remote access: make dev-server VITE_HOST=0.0.0.0 BACKEND_HOST=0.0.0.0"
 	@# On Windows, use npm run because bunx doesn't correctly pass arguments
-	@npmx concurrently -k \
-		"npmx nodemon --watch src --watch tsconfig.main.json --watch tsconfig.json --ext ts,tsx,json --ignore dist --ignore node_modules --exec node scripts/build-main-watch.js" \
+	@npm x concurrently -k \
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

Addressed the Windows HOME mismatch issue:

- We now call `runtime.resolvePath("~/.mux-tmp/<token>")` first to get the canonical absolute path in the runtime's context.
- Then we `mkdir -p` using that resolved absolute path (quoted with `shescape`) so the directory is created exactly where we later write temp files.
- On win32 we normalize `\\` -> `/` to keep downstream POSIX joins stable.

`make static-check` is green locally.

PR #1506, opened 2026-01-08
🤖 fix: Docker sub-agent container double-creation and sidebar latency

ethanndickson:
## Summary

Two related fixes for Docker runtime sub-agents, plus refactoring and Codex review fixes.

### 1. Container double-creation fix

When creating a sub-agent workspace in Docker runtime, both `forkWorkspace()` and `initWorkspace()` were trying to create the container, causing "Container already exists and is running" errors.

**Root cause:** `TaskService.create()` calls `forkWorkspace()` (which creates the container), then calls `initWorkspace()` (which also tried to create a container).

**Fix:** `initWorkspace()` now detects when the container is already running (from a successful fork) and skips directly to running the init hook. Added `checkExistingContainer()` method that validates:
- Container exists and is running
- Git repo is initialized (`.git` directory exists)
- Correct branch is checked out (prevents wrong-branch edge case)

### 2. Sidebar latency fix

Sub-agent workspaces weren't appearing in the sidebar until container creation completed (~5-15s delay).

**Root cause:** `buildSortedWorkspacesByProject()` had a race condition where metadata events arrived before config refresh completed, and the second pass only caught workspaces with `status === "creating"` (never set for sub-agents).

**Fix:** Remove the dead `status === "creating"` check - include any workspace in metadata not yet in config.

### 3. Credential setup for forked containers (Codex P2)

When reusing forked containers via the skip path, `setupContainerAndSyncProject()` was bypassed, which meant `~/.gitconfig` and `gh auth setup-git` were never run.

**Fix:** Extract `setupCredentials()` method and call it in the skip path before running the init hook.

### 4. Non-fatal init hook semantics (Codex P1)

Init hook failures should be logged but not prevent workspace usage per `docs/hooks/init.mdx`.

**Fix:** `runInitHookOnRuntime()` logs the exit code but doesn't throw, so hook failures don't trigger container cleanup.

## Changes

| File | Change |
|------|--------|
| `src/node/runtime/DockerRuntime.ts` | Add `checkExistingContainer()`, `setupCredentials()`, `setupContainerAndSyncProject()` helpers; preserve forked containers on failure |
| `src/node/runtime/initHook.ts` | Add comment clarifying non-fatal hook semantics |
| `src/browser/utils/ui/workspaceFiltering.ts` | Include all metadata workspaces not yet in config |
| `tests/runtime/runtime.test.ts` | Add tests for skip-path and container preservation |

## Testing

1. Start dev mode: `make dev`
2. Create a sub-agent task from any workspace
3. Verify:
   - Workspace appears in sidebar **immediately** (not after container starts)
   - No "Container already exists" errors in logs
   - Container creation succeeds on first attempt
   - Init hook failures don't delete the container

---

### DIFF
--- a/src/node/runtime/DockerRuntime.ts
+++ b/src/node/runtime/DockerRuntime.ts
@@ -449,6 +449,7 @@ export class DockerRuntime extends RemoteRuntime {
       switch (containerCheck.action) {
         case "skip":
           initLogger.logStep("Container already running (from fork), running init hook...");
+          await this.setupCredentials(containerName, env);
           skipContainerSetup = true;
           break;
         case "cleanup":
@@ -547,6 +548,33 @@ export class DockerRuntime extends RemoteRuntime {
   }
 
   /**
+   * Copy gitconfig and configure gh CLI credential helper in container.
+   * Called for both new containers and reused forked containers.
+   */
+  private async setupCredentials(
+    containerName: string,
+    env?: Record<string, string>
+  ): Promise<void> {
+    if (!this.config.shareCredentials) return;
+
+    // Copy host gitconfig into container (not mounted, so gh can modify it)
+    const gitconfig = path.join(os.homedir(), ".gitconfig");
+    if (existsSync(gitconfig)) {
+      await runDockerCommand(`docker cp ${gitconfig} ${containerName}:/root/.gitconfig`, 10000);
+    }
+
+    // Configure gh CLI as git credential helper if GH_TOKEN is available
+    // GH_TOKEN can come from project secrets (env) or host environment (buildCredentialArgs)
+    const ghToken = env?.GH_TOKEN ?? process.env.GH_TOKEN;
+    if (ghToken) {
+      await runDockerCommand(
+        `docker exec -e GH_TOKEN=${shescape.quote(ghToken)} ${containerName} sh -c 'command -v gh >/dev/null && gh auth setup-git || true'`,
+        10000
+      );
+    }
+  }
+
+  /**
    * Create container, sync project files, and checkout branch.
    * This is the full setup path for new workspaces (not forked ones).
    */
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

Addressed the container validation concern: before skipping setup for a running container, we now verify that `.git` exists inside. If a prior crash left a running-but-uninitialized container, we detect it and recreate from scratch instead of operating on an empty environment.

Also fixed init hook failure propagation—`runInitHookOnRuntime()` now throws on non-zero exit so `initWorkspace()` properly returns `{ success: false }`.

ethanndickson:
@codex review

**P2 - Branch validation:** Refactored container validation into `checkExistingContainer()` which returns a discriminated union. The method now validates the current branch matches `branchName` using `git rev-parse --abbrev-ref HEAD` before returning `{ action: "skip" }`. Wrong branch or failed git command triggers `{ action: "cleanup" }` to recreate the container.

**P3 - Double logComplete:** Added `InitHookError` class in `initHook.ts`. When the init hook fails, `runInitHookOnRuntime()` logs the real exit code then throws `InitHookError`. Both DockerRuntime and SSHRuntime catch blocks now check `if (!(error instanceof InitHookError))` before calling `logComplete(-1)`, preserving the real exit code.

ethanndickson:
@codex review

Fixed: Both tests now initialize the container git repo on `${workspaceName}` instead of `main`, matching what `initWorkspace({ branchName: workspaceName })` expects.

ethanndickson:
@codex review

Fixed P1: Restored non-fatal init hook semantics per `docs/hooks/init.mdx`.

- Removed `InitHookError` class - the original design was correct (log exit code and return, don't throw)
- Hook failures now log their exit code but `initWorkspace()` returns `success: true`
- Container cleanup only happens for actual initialization failures (setup, git sync, etc.), not hook failures

ethanndickson:
@codex review

Fixed P2: Extracted `setupCredentials()` method and call it in the skip path for reused forked containers.

- New `setupCredentials(containerName, env)` method copies `~/.gitconfig` and runs `gh auth setup-git`
- Called in skip path before init hook runs, so forked containers get credentials
- `setupContainerAndSyncProject()` now calls the shared method instead of inline code

PR #1507, opened 2026-01-08
🤖 fix: display task tool errors in UI

ethanndickson:
When task tool calls fail (e.g., invalid agentId like `bash` instead of `explore`/`exec`), the error was hidden—UI just showed "✗ failed" with no explanation.

## Changes

- **TaskToolCall.tsx**: Handle error results (`{ success: false, error }`), auto-expand on failure, show `ErrorBox` with error message
- **toolUtils.tsx**: Add `isToolErrorResult()` type guard for proper type narrowing
- **taskService.ts**: Error messages now include valid agentIds hint (derived from `BUILT_IN_SUBAGENT_TYPES` constant)
- **Stories**: Added `createFailedTaskTool()` factory and test coverage in `TaskErrorStates`

## Before/After

Before: Failed task shows "✗ failed" with no details
After: Failed task auto-expands showing the error: `Task.create: unknown agentId (bash). Built-in runnable agentIds: explore, exec`

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$9.84`_

### DIFF
--- a/src/browser/components/tools/TaskToolCall.tsx
+++ b/src/browser/components/tools/TaskToolCall.tsx
@@ -183,7 +183,7 @@ interface TaskToolCallProps {
 export const TaskToolCall: React.FC<TaskToolCallProps> = ({ args, result, status = "pending" }) => {
   // Narrow result to error or success shape
   const errorResult = isToolErrorResult(result) ? result : null;
-  const successResult = result && "status" in result ? result : null;
+  const successResult = result && typeof result === "object" && "status" in result ? result : null;
 
   // Derive expansion: auto-expand for reports or errors, but respect user's explicit toggle
   const hasReport = successResult?.status === "completed" && !!successResult.reportMarkdown;


ethanndickson:
@codex review

ethanndickson:
@codex review

Re: your previous comment about auto-expand not working for live task errors—

Fixed by deriving expansion state instead of using an Effect:

```tsx
const shouldAutoExpand = hasReport || !!errorResult;
const [userExpandedChoice, setUserExpandedChoice] = useState<boolean | null>(null);
const expanded = userExpandedChoice ?? shouldAutoExpand;
```

- When `userExpandedChoice` is `null` (user hasn't clicked), we derive from `shouldAutoExpand`
- When error result arrives during streaming, `shouldAutoExpand` becomes true → component re-renders → `expanded` becomes true
- Once user explicitly toggles, their choice takes precedence

No Effect needed—pure derived state during render.

ethanndickson:
@codex review

Fixed: added `typeof result === "object"` guard before using `in` operator.

PR #1516, opened 2026-01-08
🤖 feat: clean CLI tool output formatting for mux run

ammar-agent:
Add structured, readable formatting for tool calls in `mux run` output.

## Changes

- **New `toolFormatters.ts` module** with emoji-prefixed, colored output
- **Inline results** for simple tools (file_read, web_fetch, etc.)
- **Multiline results** for complex tools (bash, file_edit, etc.)
- **Proper spacing** around tool calls and thinking blocks
- **Fallback** to generic formatting for unrecognized tools

## Supported tools

| Tool | Emoji | Output |
|------|-------|--------|
| `file_edit_*` | ✏️ | Colorized diffs |
| `file_read` | 📖 | Inline file info |
| `bash` | 🔧 | Exit code, duration, truncated output |
| `task` | 🤖 | Status and report summary |
| `web_fetch/search` | 🌐/🔍 | URL and result count |
| `code_execution` | 🧮 | Code preview and result |
| `todo_write` | 📋 | Item count |
| `notify` | 🔔 | Title |
| `status_set` | (custom) | Message |
| `agent_skill_read` | 📚 | Skill name |

## Example output

```
📖 package.json (+30) ✓ 30 lines (8.6KB)
📖 README.md (+50) ✓ 50 lines (3.8KB)

🔧 ls -la | head -20 timeout:5s
exit:0 (3ms)
  total 1360
  drwxr-xr-x   21 root root   4096 Jan  8 15:06 .
  ...

✏️ src/file.tsx
--- a/src/file.tsx
+++ b/src/file.tsx
@@ -10,3 +10,5 @@
+added line
```

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high`_

PR #1517, opened 2026-01-08
🤖 fix: mark inline tool starts as open lines in CLI output

ammar-agent:
Fixes garbled CLI output when tools have a start formatter but no end formatter (e.g., `web_search`, `todo_write`, `notify`, `status_set`, `agent_skill_read*`).

## Changes

- Set `streamLineOpen = true` after `writeHuman(formatted)` for inline tools so generic end formatters properly insert newlines before multi-line blocks
- Add simple success (✓) end formatters for inline-only tools to prevent falling back to verbose generic output  
- Mark unused `toolName` parameters with underscore prefix

## Context

Follow-up to #1516. When a tool had a start formatter but no end formatter, the inline start (written without newline) would get the generic multi-line end block concatenated directly onto the same line, producing garbled output like:

```
🔍 weather today────────────────────────────────────
web_search result
  {...}
────────────────────────────────────────
```

Now it correctly produces:

```
🔍 weather today ✓
```

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$8.87`_

### DIFF
--- a/src/cli/toolFormatters.ts
+++ b/src/cli/toolFormatters.ts
@@ -325,8 +325,13 @@ function formatCodeExecutionEnd(_toolName: string, _args: unknown, result: unkno
   return output;
 }
 
-/** Simple success marker for inline tools that don't need detailed result formatting */
-function formatSimpleSuccessEnd(): string | null {
+/** Simple success/error marker for inline tools that don't need detailed result formatting */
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

PR #1521, opened 2026-01-09
🤖 feat: make code_execution tool available by default

ethanndickson:
Remove the `programmatic-tool-calling` experiment and make `code_execution` always available. The **PTC Exclusive Mode** experiment remains for users who want `code_execution` as the only way to access bridged tools.

## Changes

- Remove `PROGRAMMATIC_TOOL_CALLING` from experiment definitions
- Remove `programmaticToolCalling` from all schema/type definitions  
- Make `aiService.ts` unconditionally create `code_execution` tool
- Update tool descriptions to remove experiment references
- Fix incorrect "async" claim in code parameter description

## Behavior

- Users who had the experiment enabled: **no change**
- Users who never enabled it: **now automatically get `code_execution` alongside other tools**
- Orphaned localStorage key (`experiment:programmatic-tool-calling`) is harmless and ignored

## Testing

- `make static-check` passes
- Exclusive mode logic preserved (just the outer conditional removed)

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$5.42`_

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -156,6 +156,8 @@ jobs:
             TEST_INTEGRATION=1 bun x jest --coverage --maxWorkers=100% --silent tests/ui/
           fi
         env:
+          # Required for PTC type generation (prettier uses dynamic imports)
+          NODE_OPTIONS: --experimental-vm-modules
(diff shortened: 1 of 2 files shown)


ethanndickson:
Gonna revisit this post terminal benching.

PR #1526, opened 2026-01-09
🤖 fix: terminal-bench timeout + token counting

ethanndickson:
Fix two bugs introduced in commit ccd84b58 that caused 0% accuracy on nightly terminal-bench runs since Jan 9:

## Bug 1: Removed CLI Options

`mux-run.sh` passed `--config-root` and `--workspace-id` options that were removed from the CLI in the ephemeral refactor. The new code reads credentials from env vars via `buildProvidersFromEnv()` and generates random workspace IDs automatically.

## Bug 2: Stdin Fallback

CLI stdin fallback used `??` instead of `||`:
- `??` only falls back on `null`/`undefined`
- `||` falls back on any falsy value including empty string `""`

When no CLI args are provided, `program.args.join(" ")` returns `""`, not `null`. So `"" ?? stdin` returns `""` instead of falling through to stdin.

## Enhancement: Token Counting

Also adds token counting support that was previously missing:
- `mux-run.sh` captures output to file and extracts tokens from `stream-end` events
- `mux_agent.py` reads the token file from container and returns actual values in `AgentResult`

## Enhancement: Smoke Test

Added smoke test to nightly workflow:
- Runs `hello-world` task first before the full suite
- If smoke test fails, skips full suite to save time and API costs
- Added verification step that fails workflow if no tasks resolve

## Validation

Tested locally with hello-world task:
- Before: 0% accuracy, immediate timeout, 0 tokens
- After: 100% accuracy, 16 input / 267 output tokens captured

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$18.43`_

### DIFF
--- a/.github/workflows/terminal-bench.yml
+++ b/.github/workflows/terminal-bench.yml
@@ -127,8 +127,8 @@ jobs:
         if: always()
         run: |
           echo "=== Terminal-Bench Results Summary ==="
-          if [ -f "$(find runs -name 'results.json' 2>/dev/null | head -1)" ]; then
-            RESULTS_FILE=$(find runs -name 'results.json' | head -1)
+          if [ -f "$(find runs -maxdepth 2 -name 'results.json' 2>/dev/null | head -1)" ]; then
+            RESULTS_FILE=$(find runs -maxdepth 2 -name 'results.json' | head -1)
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

PR #1534, opened 2026-01-09
🤖 ci: use mux@next and hide auto-label from PR checks

ammar-agent:
- Use `mux@next` instead of `mux` so the workflow actually works (mux isn't published to the main npm tag yet)
- Switch from `pull_request` to `pull_request_target` so the workflow runs in the base repo context and doesn't show as a PR check
- Fix event name check for the new trigger type

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$unknown`_


ammar-agent:
@codex review

PR #1536, opened 2026-01-09
🤖 fix: allow interrupt during stream startup

ammar-agent:
## What
Allow interrupt (Esc / Ctrl+C) to cancel a message while the UI is still in **"Starting..."** (i.e., before `stream-start`).

## Why
Previously, interrupts were gated on `canInterrupt` / `StreamManager.isStreaming()` (active StreamManager stream). If a send was blocked during init/runtime setup, there was no active stream to stop, so the UI could get stuck in "Starting..." with no way to cancel.

Also, cancelling during this pre-start window could leave an empty assistant placeholder persisted in `chat.jsonl`.

## How
- Frontend:
  - Treat `pendingStreamStartTime` as interruptible for keybind interception.
  - Show cancel hint during the starting barrier.
  - Clear `pendingStreamStartTime` on `stream-abort` (abort can arrive before `stream-start`).
  - Command palette interrupt disables auto-retry (matches keybind behavior).

- Backend:
  - Add a shared `linkAbortSignal()` helper for consistent abort-signal bridging.
  - Track per-workspace pending stream starts in `AIService` and allow `stopStream()` to abort startup work **before** StreamManager registers a stream.
  - Emit a synthetic `stream-abort` when cancelling during pending-start so the renderer can exit the "Starting..." UI immediately.
  - Ensure startup aborts don’t leave empty assistant placeholders behind:
    - New `HistoryService.deleteMessage()` removes the placeholder without truncating subsequent history.
    - Used in both real streaming and mock mode abort paths.
  - Refactor `StreamManager.startStream()` to use a derived `AbortController` + structured cleanup (temp dir, unlink abort listener).
  - Guard against a narrow “late abort before processing begins” race so we never emit `stream-start` after an interrupt without a corresponding abort/end.

## Tests
- `make static-check`
- Added/updated tests:
  - `tests/ipc/interruptStream.starting.test.ts` (requires `TEST_INTEGRATION=1` + `ANTHROPIC_API_KEY`)
  - `src/node/services/historyService.test.ts` (`deleteMessage`)
  - `src/node/services/mock/mockAiStreamPlayer.test.ts` (placeholder cleanup on abort)

---
_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$46.85`_

### DIFF
--- a/src/node/services/streamManager.ts
+++ b/src/node/services/streamManager.ts
@@ -1769,6 +1769,17 @@ export class StreamManager extends EventEmitter {
           toolPolicy,
           hasQueuedMessage
         );
+
+        // Guard against a narrow race:
+        // - stopStream() may abort while we're between the last aborted-check and stream registration.
+        // - If we start processStreamWithCleanup anyway, it would emit stream-start, but no one would
+        //   subsequently call stopStream() again (it already ran), so we'd never emit stream-abort/end.
+        // In that case, immediately drop the registered stream and rely on the caller to handle UI.
+        if (streamAbortController.signal.aborted) {
+          this.workspaceStreams.delete(typedWorkspaceId);
+          return Ok(streamToken);
+        }
+
         streamInfo.unlinkAbortSignal = unlinkAbortSignal;
         streamRegistered = true;
 


ammar-agent:
@codex review

Addressed the startup-interrupt race by honoring already-aborted signals before atomic stream creation, plus a unit test that aborts during temp-dir creation to ensure we never emit stream-start / start streaming.

ammar-agent:
@codex review

Refactored stream startup abort handling to be race-free by construction:
- Link abort signals via shared helper (`linkAbortSignal`)
- `startStream` now cleans up `runtimeTempDir` if abort happens before stream registration
- Added unit coverage asserting the pre-start temp-dir cleanup path

ammar-agent:
@codex review

Addressed the mock-mode startup abort edge case:
- Added HistoryService.deleteMessage(workspaceId, messageId) to remove a single placeholder without truncating subsequent messages.
- MockAiStreamPlayer.play now deletes the assistant placeholder if abort happens after appendToHistory but before scheduling events.
- Added a unit test covering abort-after-append cleanup.

ammar-agent:
@codex review

Also addressed the real-stream pre-start abort case in AIService:
- When `combinedAbortSignal.aborted` triggers the early-return before calling StreamManager, we now delete the assistant placeholder from `chat.jsonl` via `HistoryService.deleteMessage`.
- Added HistoryService.deleteMessage unit tests.

ammar-agent:
@codex review

Handled the "abort during StreamManager startup" case:
- After `startStream()` returns, if the abort signal is set and StreamManager never registered a stream (`!isStreaming(workspaceId)`), AIService now deletes the assistant placeholder via `HistoryService.deleteMessage`.

PR #1538, opened 2026-01-09
🤖 fix: eliminate terminal resize race condition

ammar-agent:
The previous approach resized the frontend terminal immediately but debounced the PTY resize by 300ms. During that window, shell output formatted for the old dimensions would display incorrectly in the already-resized frontend terminal, causing text clobbering.

**Symptoms:** Output from commands like `git status --stat` would have columns bleeding into each other - e.g., diff stats appearing mid-line, "modified" becoming "commodified" as text from different areas overlapped.

**Root cause:** Frontend `fitAddon.fit()` ran immediately on resize, but `router.resize()` was debounced by 300ms. During that 300ms window, the shell formatted output for the old PTY dimensions while the frontend displayed it at the new dimensions.

**Fix:** Use `requestAnimationFrame` to batch resize events, then resize both frontend and PTY in the same tick. This eliminates the window where sizes can diverge.

- Simpler code (no setTimeout, no pendingResize state)
- Natural batching via RAF (coalesces rapid resize events)
- No race between frontend and PTY dimensions

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: $3.75_

### DIFF
--- a/src/browser/components/TerminalView.tsx
+++ b/src/browser/components/TerminalView.tsx
@@ -118,7 +118,7 @@ export function TerminalView({
 
     // Send initial resize to sync PTY dimensions
     const { cols, rows } = term;
-    router.resize(sessionId, cols, rows);
+    void router.resize(sessionId, cols, rows);
 
     return unsubscribe;
   }, [visible, terminalReady, sessionId, router]);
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

The concern about flooding PTY resize calls is addressed by the implementation:

1. **RAF batching**: `handleResize` uses `requestAnimationFrame` to coalesce rapid resize events into single frames
2. **In-flight queueing**: The `resizeInFlight` + `pendingResize` mechanism ensures only ONE resize can be pending while a resize is in progress - no matter how many events fire during a drag
3. **Sequential processing**: Each resize awaits the previous one before sending the next

This is actually *fewer* resize calls than the old approach which fired on every RAF during drag. Now we fire, wait for completion (~10-50ms), then process only the most recent queued resize.

The previous debounce was removed because it created a worse bug: shell output formatted for old dimensions displayed in already-resized terminal = text clobbering.

ammar-agent:
@codex review

Good catch — the prior version did `fit()` after awaiting the PTY resize, which could let the frontend jump to the latest container size while the PTY was still at the previous size.

Fixed in the latest commit:
- After awaiting `router.resize(sessionId, cols, rows)`, we now call `term.resize(cols, rows)` using the *same* dimensions that were sent to the PTY.
- We explicitly avoid `fit()` in this path to prevent recomputing dimensions while a resize is queued/in-flight.
- Added RAF cancellation + a disposed guard so we don’t apply stale resizes after unmount.

PR #1556, opened 2026-01-09
🤖 feat: move untracked files indicator to hunk viewer area

ammar-agent:
Make the untracked files indicator more prominent by moving it from the top control bar to the hunk viewer area, displayed as a collapsible banner above all hunks.

## Changes

- **UntrackedStatus component**: Rewrote as a banner-style component with amber styling
  - Shows "N untracked files - not included in diff" with expand arrow
  - Expandable file list with "Track All Files" button when expanded
  - Auto-hides when no untracked files present
  
- **ReviewControls**: Removed UntrackedStatus and now-unused props (workspaceId, workspacePath, refreshTrigger)

- **ReviewPanel**: Added UntrackedStatus between FileTree and hunks list

- **Story**: Added `ReviewTabWithUntrackedFiles` story with play function that expands the banner

- **storyHelpers**: Extended `GitDiffFixture` to support `untrackedFiles` array for mocking

## Visual

The banner appears prominently at the top of the hunk viewer area with:
- Collapsible header with count and "not included in diff" note
- Amber/yellow styling to draw attention
- File list and "Track All Files" button when expanded

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$1.86`_

### DIFF
--- a/src/browser/stories/App.rightSidebar.stories.tsx
+++ b/src/browser/stories/App.rightSidebar.stories.tsx
@@ -1058,6 +1058,9 @@ export const ReviewTabWithUntrackedFiles: AppStory = {
       canvas.getByText("Track All Files");
     });
 
+    // Double-RAF for scroll stabilization after banner expansion changes layout
+    await new Promise((r) => requestAnimationFrame(() => requestAnimationFrame(r)));
+
     // Blur focus for clean screenshot
     blurActiveElement();
   },


ammar-agent:
@codex review

Resolved the P2 comment - showing the untracked files banner only when there's a successful diff is the intended behavior. If the diff fails (e.g., invalid base), the user likely needs to fix that first. Edge case of wanting to track files during an error state is uncommon enough that `git add` in terminal is acceptable.

PR #1558, opened 2026-01-10
🤖 fix: prevent force-compaction from corrupting history via race condition

ethanndickson:
## Problem

Force-compaction sometimes replaced the entire chat with non-compaction content from the previous stream. The compaction summary would be garbage from a stream that was finishing up, not the actual compaction response.

## Root Cause

A race window exists between `state = COMPLETED` and `emit("stream-end")` in `streamManager.ts`:

```typescript
streamInfo.state = StreamState.COMPLETED;  // isStreaming() now returns false

await this.partialService.deletePartial(...);
await this.historyService.updateHistory(...);  // async I/O takes time

this.emit("stream-end", streamEndEvent);  // handleCompletion runs here
```

When force-compaction triggered during this window:
1. `isStreaming()` returned `false` (state already COMPLETED)
2. Compaction request bypassed the queue and was sent directly
3. Compaction user message was appended to history immediately
4. Old stream's `stream-end` fired with the old parts
5. `handleCompletion` saw compaction-request as last user message
6. Used old stream's parts as the "compaction summary" → history corrupted

## Fix

1. **Move `state = COMPLETED` after async I/O** — keeps `isStreaming()` returning `true` during cleanup, so new messages are queued instead of sent directly.

2. **Await `processingPromise` in `cancelStreamSafely` for COMPLETED streams** — when `ensureStreamSafety` encounters a completed stream, we now await its `processingPromise` instead of immediately deleting the map entry. This ensures the old stream's `finally` block (which does the actual cleanup) completes before a new stream is registered, preventing the old `finally` from accidentally deleting the new stream's entry.

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$15.11`_

### DIFF
--- a/src/node/services/streamManager.ts
+++ b/src/node/services/streamManager.ts
@@ -538,13 +538,17 @@ export class StreamManager extends EventEmitter {
     streamInfo: WorkspaceStreamInfo,
     abandonPartial?: boolean
   ): Promise<void> {
-    // If stream already completed normally (emitted stream-end), just clean up.
-    // This happens when ensureStreamSafety is called for a new stream after the
-    // previous one finished but before it was removed from workspaceStreams.
+    // If stream already completed normally (emitted stream-end), wait for its
+    // finally block to finish before returning. This happens when ensureStreamSafety
+    // is called for a new stream after the previous one finished but before it was
+    // removed from workspaceStreams.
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

PR #1563, opened 2026-01-10
🤖 feat: add platform-native clipboard shortcuts to terminal

ammar-agent:
## Summary

Implement copy/paste keyboard shortcuts in the terminal (RightSidebar) that match platform conventions:

| Platform | Copy | Paste |
|----------|------|-------|
| **macOS** | Cmd+C | Cmd+V |
| **Linux** | Ctrl+Shift+C | Ctrl+Shift+V |
| **Windows** | Ctrl+C | Ctrl+V |

## Problem

Cmd/Ctrl+V paste didn't work in the terminal. The terminal uses ghostty-web which has paste support via a hidden textarea, but Electron's Edit menu `role: "paste"` uses `execCommand('paste')` on the focused element—and ghostty-web focuses the container div, not the textarea.

## Solution

Use ghostty-web's `attachCustomKeyEventHandler()` API to intercept platform-appropriate shortcuts and call the Clipboard API directly.

**Key behaviors:**
- Copy only activates when there's a text selection
- On Linux, Ctrl+C without selection still sends SIGINT (interrupt signal) as expected
- Uses `terminal.paste()` which handles bracketed paste mode correctly

## Testing

- Verified typecheck passes
- Verified static-check passes

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$3.87`_

### DIFF
--- a/src/browser/components/TerminalView.tsx
+++ b/src/browser/components/TerminalView.tsx
@@ -215,6 +215,63 @@ export function TerminalView({
         terminal.open(containerEl);
         fitAddon.fit();
 
+        // Platform-aware clipboard shortcuts matching VS Code's integrated terminal:
+        // https://code.visualstudio.com/docs/terminal/basics#_copy-paste
+        //
+        // - macOS: Cmd+C/V (standard Mac shortcuts, Cmd is distinct from Ctrl)
+        // - Linux: Ctrl+Shift+C/V (Ctrl+C reserved for SIGINT, so terminals use Shift)
+        // - Windows: Ctrl+C/V (copy only when selection exists, otherwise SIGINT)
+        //
+        // Copy only triggers when there's a selection. Without selection, Ctrl+C falls
+        // through to ghostty-web which sends SIGINT to the running process.
+        const isMac = navigator.platform.includes("Mac");
+        const isWindows = navigator.platform.includes("Win");
+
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

ammar-agent:
Addressed: fixed the comment - Windows Ctrl+C does send interrupt signals. The behavior is copy-when-selection-exists (same as VS Code), otherwise SIGINT.

ammar-agent:
Addressed: switched from `ev.key` to `ev.code` (`KeyV`/`KeyC`) for layout-independent detection. This handles Caps Lock and non-US keyboard layouts correctly.

ammar-agent:
@codex review

ammar-agent:
Addressed: switched to `ev.key.toLowerCase()` which is layout-aware (Dvorak/Colemak users get shortcuts on their layout's C/V keys) while still handling Caps Lock correctly.

ammar-agent:
@codex review

ammar-agent:
Addressed: Linux Ctrl+Shift+C is now always swallowed (no-op if no selection) to prevent it from propagating as SIGINT. Mac/Windows only intercept when there's a selection.

ammar-agent:
@codex review

PR #1564, opened 2026-01-10
🤖 fix: show compaction model during starting phase

ammar-agent:
Fixes the streaming barrier model label during compaction.

Previously, while compaction was *pending* ("starting...") we showed the workspace chat model because the UI read from the persisted chat model key. The compaction model is only known to the frontend from the /compact request metadata until stream-start arrives.

Changes:
- Track pending compaction request metadata in `StreamingMessageAggregator` when the user message arrives.
- Expose `pendingCompactionModel` via `WorkspaceState`.
- Use that model in the "starting" phase for both the app and VS Code webview barriers.
- Prefer stream-start `mode === "compact"` for authoritative compaction detection (no history scanning).

Validation:
- `make static-check`

---
_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$3.46`_

### DIFF
--- a/src/browser/utils/messages/StreamingMessageAggregator.test.ts
+++ b/src/browser/utils/messages/StreamingMessageAggregator.test.ts
@@ -404,6 +404,41 @@ describe("StreamingMessageAggregator", () => {
     });
   });
 
+  describe("compaction detection", () => {
+    test("treats active stream as compacting on reconnect when stream-start has no mode", () => {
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

Addressed the reconnect/legacy stream-start case:
- `handleStreamStart` now prefers `mode`, but falls back to last user message metadata when `mode` is missing and no pending compaction request is available.
- Added a unit test covering the legacy stream-start-without-mode scenario.

PR #1578, opened 2026-01-12
🤖 fix: single-port dev-server via Vite proxy

ibetitsmike:
This makes `make dev-server` usable behind a single public port by:

- Adding a Vite dev proxy for mux backend routes (`/orpc` + ws, `/api`, `/auth`, `/health`, `/version`).
- Updating `make dev-server` to stop setting `VITE_BACKEND_URL` so the UI uses same-origin (and hits the proxy).
- Avoiding hardcoded Vite HMR websocket host/port so HMR works when accessed via a reverse proxy / custom domain.

Usage:
- Remote: `make dev-server VITE_HOST=0.0.0.0`
- Only expose the Vite port publicly (default `5173`); backend can remain on localhost.

---
_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$4.18`_

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -56,7 +56,9 @@ function formatHostForUrl(host: string): string {
   // IPv6 URLs must be bracketed: http://[::1]:1234
   if (unbracketed.includes(":")) {
     // If the host contains a zone index (e.g. fe80::1%en0), percent must be encoded.
-    const escaped = unbracketed.replace(/%(?![0-9A-Fa-f]{2})/g, "%25");
+    // Encode zone indices (including numeric ones like %12) while avoiding double-encoding
+    // if the user already provided a URL-safe %25.
+    const escaped = unbracketed.replace(/%(?!25)/gi, "%25");
     return `[${escaped}]`;
   }
 


ibetitsmike:
@codex review

Addressed the OAuth redirect host issue by preserving the original `Host` header for the `/auth` Vite proxy (set `changeOrigin: false`). This keeps mux-gateway redirects pointing back to the public dev-server origin instead of `127.0.0.1:3000`.

ibetitsmike:
@codex review

Follow-up on the Host header concern:
- Added `MUX_VITE_ALLOWED_HOSTS` handling in `vite.config.ts` and wired it from the Makefile as `VITE_ALLOWED_HOSTS`.
- Dev server now defaults to a localhost-only allowlist (plus the bound host if it's a specific hostname), instead of `allowedHosts: true`.

This keeps mux-gateway OAuth redirects pointing at the public dev-server host while avoiding arbitrary Host header injection when dev-server is exposed.

ibetitsmike:
@codex review

Handled IPv6 literal backend hosts in `vite.config.ts` by formatting `MUX_BACKEND_HOST` for use in a URL (bracketing IPv6 + encoding `%` zone indices) before building `backendProxyTarget`.

ibetitsmike:
@codex review

Good catch on the bracketed zone-index case. Updated `formatHostForUrl` to normalize by stripping brackets first and encoding any *non-percent-encoded* `%` (e.g. `%en0` -> `%25en0`) before re-bracketing. This avoids generating invalid targets like `http://[fe80::1%en0]:3000`.

ibetitsmike:
@codex review

Updated `formatHostForUrl` again to encode *all* `%` in IPv6 literals except existing `%25` (case-insensitive). This covers numeric scope IDs like `fe80::1%12` which would otherwise be interpreted as percent-encoding in a URL.

PR #1590, opened 2026-01-13
🤖 tests: add E2E regression tests for terminal keyboard input (#1586)

ammar-agent:
## Summary

Adds end-to-end tests that verify keyboard input reaches the terminal, preventing regression of the bug fixed in PR #1586.

## Root Cause (from #1586)

The bug in `attachCustomKeyEventHandler` had inverted return values:
- ghostty-web expects `true` to PREVENT default, `false` to ALLOW default
- The buggy code returned `true` for all non-clipboard keys, blocking ALL input

## Tests Added

1. **`keyboard input reaches terminal`** - Types a command that creates a marker file and verifies execution
2. **`special keys work in terminal`** - Tests Enter, Backspace, Tab, and arrow keys

## Helper Methods Added

New E2E UI utils for terminal interaction:
- `focusTerminal()`
- `typeInTerminal(text)`
- `pressKeyInTerminal(key)`
- `runTerminalCommand(command)`

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$2.96`_

### DIFF
--- a/tests/e2e/scenarios/terminal.spec.ts
+++ b/tests/e2e/scenarios/terminal.spec.ts
@@ -1,4 +1,6 @@
-import { electronTest as test } from "../electronTest";
+import fs from "fs";
+import path from "path";
+import { electronTest as test, electronExpect as expect } from "../electronTest";
 
 test.skip(
   ({ browserName }) => browserName !== "chromium",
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

PR #1595, opened 2026-01-13
🤖 ci: add actionlint, zizmor, and ShellCheck to static-check

ethanndickson:
Add GitHub Actions linting (matching tallyman's approach) with three new make targets:

- **`lint-actionlint`**: runs [actionlint](https://github.com/rhysd/actionlint) with ShellCheck integration
- **`lint-zizmor`**: runs [zizmor](https://github.com/zizmorcore/zizmor) security analysis via Docker
- **`lint-shellcheck`**: runs ShellCheck on all repo `.sh` files

All three are included in `make static-check` and run in CI.

## Workflow hardening

- `persist-credentials: false` on all `actions/checkout@v4` steps
- Template injection mitigations (pass `${{ }}` values via `env:` blocks instead of interpolating directly into shell scripts)
- ShellCheck fixes throughout (quoting, unused vars, useless cat)

### `unpinned-uses` (disabled)

Pin all actions to SHA hashes instead of version tags. Version tags like `@v4` can be moved by maintainers, which is a supply chain risk. Tool: [pinact](https://github.com/suzuki-shunsuke/pinact).

### `excessive-permissions` (ignored at specific lines)

Move workflow-level permissions to job-level for least-privilege:
- `pr.yml:25` — `id-token:write` needed only for GCP workload identity jobs
- `release.yml:14-15` — `contents:write` and `id-token:write` needed only for release job

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$8.13`_


ethanndickson:
@codex review

PR #1614, opened 2026-01-14
🤖 feat: enable mux testing mux via isolated roots

ibetitsmike:
This enables running multiple mux desktop instances on the same machine without collisions by scoping both mux state **and** Electron `userData` per root.

### What changed
- Added a supported CLI flag `--mux-root <path>` (alias `--root`) that sets `MUX_ROOT` early and strips itself from argv for subcommand routing.
- Desktop now derives Electron `userData` from the chosen mux root (`<muxHome>/user-data`) so the single-instance lock + localStorage are effectively per-root.
- Legacy migration (`~/.cmux → ~/.mux`) is skipped when using a custom root to avoid side effects.
- Added a built-in skill: `mux-testing-mux`.
- Added debug command palette actions:
  - **Debug: Launch Test Instance** (creates `<muxHome>/instances/<id>` and spawns a new mux process with `--mux-root`)
  - **Debug: Delete Test Instances…** (cleanup)

### How to use
- Launch a second instance manually:
  - `mux --mux-root /tmp/mux-test-1`
- Or from the desktop app command palette:
  - `Debug: Launch Test Instance`

### Validation
- `make static-check`
- Added unit tests for argv parsing, userData resolution, and config sanitization.

---
_Generated with `mux` • Model: openai:gpt-5.2 • Thinking: xhigh • Cost: $14.25_

### DIFF
--- a/src/browser/utils/commands/sources.ts
+++ b/src/browser/utils/commands/sources.ts
@@ -691,7 +691,7 @@ export function buildCoreSources(p: BuildSourcesParams): Array<() => CommandActi
         section: section.debug,
         run: async () => {
           try {
-            const result = await api.debug.launchTestInstance({});
+            const result = await api.debug.launchTestInstance();
             if (result.success) {
               alert(`Launched test instance at:\n${result.data.rootDir}`);
             } else {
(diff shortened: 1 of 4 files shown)


ibetitsmike:
@codex review

Addressed the port-collision concern by clearing `apiServerPort` in the isolated test instance config so spawned instances bind to a random available port.

ibetitsmike:
@codex review

Also updated `consumeMuxRootFromArgv` to treat `--mux-root/--root --help` (and other flag-like values) as a missing-value error, matching typical CLI parser behavior.

ibetitsmike:
@codex review

Also updated test-instance spawning to override `MUX_USER_DATA_DIR` to `<newRoot>/user-data`, so isolation still works even if the parent mux process was launched with a custom userData dir.

PR #1618, opened 2026-01-14
🤖 ci: pin all GitHub Actions to SHA hashes

ethanndickson:
Pin all 50 action references to commit SHAs with version comments, addressing the supply chain security risk from mutable version tags.

## Why

Version tags like `@v4` can be moved by maintainers, which is a supply chain risk. Pinning to full commit SHAs ensures immutable references. See [GitHub's security hardening guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

## Changes

- **`scripts/pin-actions.sh`**: Wrapper to download and run [pinact](https://github.com/suzuki-shunsuke/pinact)
- **`make pin-actions`**: New target for easy re-pinning after Dependabot updates
- **`.github/zizmor.yml`**: Remove `unpinned-uses.disable` (no longer needed)
- **9 workflow/action files**: All 50 action references pinned

## Example transformation

```yaml
# Before
- uses: actions/checkout@v4

# After  
- uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
```

## Maintenance

When Dependabot opens PRs to update action versions, run `make pin-actions` to re-pin to the new SHAs.

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$9.91`_

### DIFF
--- a/.github/actions/setup-mux/action.yml
+++ b/.github/actions/setup-mux/action.yml
@@ -4,7 +4,7 @@ runs:
   using: "composite"
   steps:
     - name: Setup Bun
-      uses: oven-sh/setup-bun@v2
+      uses: oven-sh/setup-bun@b7a1c7ccf290d58743029c4f6903da283811b979 # v2.1.0
       with:
         bun-version: 1.3.5
 
@@ -15,7 +15,7 @@ runs:
 
     - name: Cache node_modules
       id: cache-node-modules
-      uses: actions/cache@v4
+      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
       with:
         path: node_modules
         key: ${{ runner.os }}-${{ runner.arch }}-bun-${{ steps.bun-version.outputs.version }}-node-modules-${{ hashFiles('**/bun.lock') }}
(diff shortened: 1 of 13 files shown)


ethanndickson:
The `Test / Unit` job is failing with 2 tests in `handleCompactCommand`:
- `passes reviews to continueMessage when reviews are attached`
- `creates continueMessage with only reviews (no text)`

These tests pass locally but fail consistently in CI. This appears to be a flaky test or environment issue unrelated to this PR (which only pins action SHAs). The test mock expects `sendMessage` to be called but receives 0 calls.

**Evidence this is not caused by this PR:**
- Tests pass locally on main branch
- The changes in this PR only modify workflow YAML files and add a pin-actions script
- No TypeScript/test code was modified

ethanndickson:
Update: The `handleCompactCommand` test failures are affecting multiple PRs and even the merge queue. Main branch CI was green at `32972f73` (08:48 UTC) but subsequent PR runs are failing the same test.

This appears to be a flaky test issue - the test passes locally consistently but fails intermittently in CI. The test relies on mock functions and may have a race condition or timing issue in the CI environment.

**This PR only changes workflow YAML files and adds a shell script - no test code was modified.**

PR #1657, opened 2026-01-14
🤖 feat: enable Mux Gateway by default after login

ibetitsmike:
This updates the Mux Gateway OAuth login flows (splash screen + Settings → Providers) to automatically enable Mux Gateway routing for all eligible models after a successful first-time login.

- Only applies when the user wasn't already logged in (avoids clobbering per-model choices on re-login).
- Uses the existing “Default” behavior (sets `gateway-models` to all supported models).

## Validation
- `make static-check`

---

### DIFF
--- a/src/browser/components/Settings/sections/ProvidersSection.tsx
+++ b/src/browser/components/Settings/sections/ProvidersSection.tsx
@@ -284,6 +284,10 @@ export function ProvidersSection() {
               // Ignore errors fetching config; fall back to the current snapshot.
             }
 
+            if (attempt !== muxGatewayLoginAttemptRef.current) {
+              return;
+            }
+
             setGatewayModels(getEligibleGatewayModels(latestConfig));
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

Addressed the stale eligible-model list issue in the desktop OAuth flows by recomputing from the latest provider config at login success (via `api.providers.getConfig()`), falling back gracefully if the fetch fails.

ibetitsmike:
@codex review

Follow-ups:
- Removed the new `useMemo` (React Compiler rule) and inlined eligibility computation where needed.
- In both OAuth message callbacks (Settings + splash), we now refetch provider config via `api.providers.getConfig()` before defaulting `gateway-models`, with a safe fallback.

ibetitsmike:
@codex review

Added attempt re-checks after the async `api.providers.getConfig()` in the desktop OAuth success paths (Settings + splash) so Canceling during the fetch no longer mutates `gateway-models` or flips the UI to success.

PR #1664, opened 2026-01-15
🤖 fix: add stopKeyboardPropagation helper to prevent Escape from interrupting stream

ethanndickson:
## Problem

Pressing Escape while editing a message was both canceling the edit AND interrupting the stream. This was the **second time** we hit this bug.

## Root Cause

React's `e.stopPropagation()` only stops propagation within React's synthetic event system. Native `window` listeners (like the stream interrupt handler in `useAIViewKeybinds`) still receive the event.

## Solution

Add a `stopKeyboardPropagation()` helper that stops both React synthetic and native event propagation. Update all keyboard handlers that need to block global handlers to use this helper.

### Changes

| File | Change |
|------|--------|
| `src/browser/utils/events.ts` | New helper utility |
| `ChatInput/index.tsx` | Use helper for cancel-edit Escape |
| `VimTextArea.tsx` | Use helper for escapeInNormalMode |
| `CommandPalette.tsx` | Use helper for Enter/Escape |
| `WorkspaceListItem.tsx` | Use helper for rename input |
| `DiffRenderer.tsx` | Use helper for review note textarea |
| `keybinds.ts` | Use helper in `createEditKeyHandler` |
| `useVoiceInput.ts` | Use helper for Escape during recording |
| `AGENTS.md` | Document the pattern |

## Validation

1. Edit a message while stream is active → press Escape → edit cancels, stream continues
2. Open CommandPalette during stream → press Escape → palette closes, stream continues
3. Press Escape with no modal/edit open → stream interrupts (expected)

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$7.91`_

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "mux",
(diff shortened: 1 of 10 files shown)


ethanndickson:
@codex review

PR #1671, opened 2026-01-15
🤖 fix: stabilize retry barrier visibility and manual retry feedback

ibetitsmike:
Summary
This PR stabilizes the bottom-of-chat retry UI by keeping the Retry barrier visible across auto-retry start/fail loops, and adds immediate manual-click feedback for the Retry button so users get instant acknowledgement before stream events arrive.

Background
The retry footer could still visually churn when retry attempts toggled stream-start without visible output yet. Separately, manual Retry clicks had no immediate pending state for non-auto-retryable errors, which made the button feel unresponsive until backend events landed.

After adding manual pending state, we found a remaining UX jump: clicking Retry briefly removed the error-details row because manual retry clears `retryState.lastError` immediately, shrinking the yellow box before it re-expanded on the next failure.

Implementation
- `ChatPane`
  - Reads persisted retry attempt state via `getRetryStateKey(...)`.
  - Computes retry visibility using `getInterruptionContext(...)` plus `shouldKeepRetryBarrierVisibleDuringRetry(...)`.
  - Keeps retry barrier visible during retry-loop stream starts (before visible token output).
  - Suppresses `StreamingBarrier` whenever retry barrier UI is active to avoid footer swapping.
- `RetryBarrier`
  - Adds a local `isManualRetryInFlight` state for immediate click feedback.
  - Shows spinner + `Retrying...` label and temporarily disables the manual Retry button.
  - Keeps the most recent error details visible while manual retry is pending, so the banner height stays stable instead of collapsing/re-expanding.
  - Clears the local pending state when retry transitions into normal auto-retry UI, stream start, or a new error state.
- `retryEligibility`
  - Uses the sticky retry predicate to keep the banner stable through retry-loop TTFT gaps.

Validation
- `bun test src/browser/utils/messages/retryEligibility.test.ts`
- `make static-check`

Risks
- Low-to-moderate UI-state risk in chat footer transitions.
- Scope is constrained to retry/stream barrier presentation logic and manual retry button state.
- Existing retry eligibility tests cover sticky-visibility paths; static checks are green.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$11.38`_

### DIFF
--- a/src/browser/components/Messages/ChatBarrier/RetryBarrier.tsx
+++ b/src/browser/components/Messages/ChatBarrier/RetryBarrier.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from "react";
+import React, { useEffect, useMemo, useRef, useState } from "react";
 import { AlertTriangle, Loader2, RefreshCw } from "lucide-react";
 import { usePersistedState, updatePersistedState } from "@/browser/hooks/usePersistedState";
 import type { RetryState } from "@/browser/hooks/useResumeManager";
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

Addressed the TTFT regression by restoring the `!canInterrupt` gate for the `hasInterruptedStream(...)` branch and expanding the retry-sticky predicate to cover the common case where no assistant DisplayedMessage exists until the first delta.

PR #1688, opened 2026-01-15
🤖 fix: add context wrapper to post-compaction follow-up message

ammar-agent:
## Summary

Fixes bug where queued follow-up messages were sent after compaction failed.

**Problem:**
When compaction fails (e.g., Gemini "completing" without producing a valid summary), the queued follow-up message was still being sent. This caused the follow-up to go into the non-compacted chat, potentially hitting context limits again.

**Root cause:**
`handleCompletion()` returned `boolean` where `false` meant both "not a compaction stream" (normal) AND "compaction failed" (bad). The `sendQueuedMessages()` was called unconditionally on stream-end.

**Solution:**
Introduce `CompactionResult` discriminated union:
- `"not-compaction"`: Normal stream end, proceed with queued messages
- `"success"`: Compaction succeeded, proceed with queued messages  
- `"failed"`: Compaction was attempted but failed - **do NOT send queued messages**

This prevents the follow-up from being injected into a non-compacted chat when the model fails to produce a valid summary.

## Test Plan

- Updated `compactionHandler.test.ts` assertions to use new `CompactionResult` type
- All 36 compactionHandler tests passing
- `make static-check` passes

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$12.22`_

### DIFF
--- a/src/node/services/agentSession.ts
+++ b/src/node/services/agentSession.ts
@@ -35,7 +35,7 @@ import {
 import { createRuntime } from "@/node/runtime/runtimeFactory";
 import { MessageQueue } from "./messageQueue";
 import type { StreamEndEvent } from "@/common/types/stream";
-import { CompactionHandler } from "./compactionHandler";
+import { CompactionHandler, type CompactionResult } from "./compactionHandler";
 import type { TelemetryService } from "./telemetryService";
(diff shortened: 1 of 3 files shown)


ammar-agent:
@mux retry

PR #1701, opened 2026-01-16
🤖 fix: show helpful errors for slash commands with missing/invalid arguments

ethanndickson:
Previously, commands like `/truncate` (without args) would show "Unknown command: /truncate" which was confusing since the command exists but needs an argument.

## Changes

Added two new `ParsedCommand` types:
- `command-missing-args`: for required arguments not provided
- `command-invalid-args`: for arguments that don't pass validation

Updated commands to use these with usage hints:
- `/truncate`: shows `/truncate <0-100> (percentage to remove)`
- `/idle`: shows `/idle <hours> or /idle off`
- `/mcp add|edit|remove`: shows appropriate usage for each subcommand

## Before/After

| Command | Before | After |
|---------|--------|-------|
| `/truncate` | "Unknown command: /truncate" | "Missing Arguments" + usage hint |
| `/truncate 150` | "Unknown command: /truncate 150" | "Invalid Argument: '150' is not valid" + usage |
| `/mcp add` | "Unknown command: /mcp add" | "Missing Arguments" + usage hint |

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$2.10`_

### DIFF
--- a/src/browser/utils/slashCommands/idle.test.ts
+++ b/src/browser/utils/slashCommands/idle.test.ts
@@ -1,12 +1,12 @@
 import { parseCommand } from "./parser";
 
 describe("/idle command", () => {
-  it("should return unknown-command for /idle without arguments", () => {
+  it("should return command-missing-args for /idle without arguments", () => {
     const result = parseCommand("/idle");
     expect(result).toEqual({
-      type: "unknown-command",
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

PR #1710, opened 2026-01-16
🤖 fix: prefer Git Bash on Windows (avoid WSL)

ibetitsmike:
## What
- On Windows, prefer Git for Windows' Git Bash for bash execution and avoid WSL launchers (e.g. `C:\\Windows\\System32\\bash.exe`).
- Ignore WSL-ish `SHELL` values when choosing the local PTY shell (falls back to Git Bash → pwsh → powershell → cmd).
- Add a dismissible Windows banner when the default shell appears to be WSL (with install guidance).

## Notes
- `bashPath` now checks known Git for Windows locations first, then derives from `where git`, then falls back to `where bash` (skipping WSL launchers and requiring a Git for Windows layout).

## Validation
- `make static-check`

---
_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$6.38`_

### DIFF
--- a/src/desktop/main.ts
+++ b/src/desktop/main.ts
@@ -399,14 +399,8 @@ async function loadServices(): Promise<void> {
       p.endsWith("\\windows\\system32\\wsl.exe");
 
     const envShell = process.env.SHELL?.trim();
-    if (envShell) {
-      if (envShell.startsWith("/")) {
-        return true;
-      }
-
-      if (isWslLauncher(normalize(envShell))) {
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

Addressed review note: on Windows we now treat POSIX-style `SHELL` values like `/bin/bash` as WSL-ish and ignore them, falling back to Git Bash/pwsh/cmd.

PR #1698, opened 2026-01-16
🤖 feat: debug show all messages + last LLM request inspector

ThomasK33:
Adds two dev-focused debugging tools:

- Show full chat history including `metadata.synthetic` messages (and disable the display cap)
- Inspect the exact LLM request payload sent to the provider for the last turn

Dev ergonomics:
- `make dev` + `make start` now default `MUX_DEBUG_LLM_REQUEST=1` (set `MUX_DEBUG_LLM_REQUEST=0` to disable)

Validation:
- `make static-check`

---

<details>
<summary>📋 Implementation Plan</summary>

# Plan: Debug message visibility + last LLM request inspector

## Goals
- In dev, **display every message in chat history**, including `metadata.synthetic` messages (e.g., persisted `@file` snapshots).
- In dev, **inspect the exact LLM request payload** (system message + final `ModelMessage[]`) that was sent to the provider for the last turn, including request-only synthetic injections and provider-specific transforms/redactions.

### Behavior
When enabled:
- **Do not hide** messages where `message.metadata.synthetic === true`.
- **Do not apply** the UI message cap (`MAX_DISPLAYED_MESSAGES`) so older hidden messages are visible too.
- Visually mark synthetic messages in the UI (small “synthetic” tag) and **disable Edit** for them.

### Implementation steps
1. **Expose env var to renderer**
   - `src/desktop/preload.ts`: add `debugShowAllMessages: process.env.MUX_DEBUG_SHOW_ALL_MESSAGES === "1"` to `window.api`.
   - `src/common/types/global.d.ts`: add optional `debugShowAllMessages?: boolean` to `WindowApi`.

2. **Plumb flag into message aggregation**
   - `src/browser/stores/WorkspaceStore.ts`: when constructing `StreamingMessageAggregator`, pass `{ debugShowAllMessages: window.api?.debugShowAllMessages === true }` (or equivalent).
   - `src/browser/utils/messages/StreamingMessageAggregator.ts`:
     - Add a stored boolean on the instance (from ctor options).
     - In `getDisplayedMessages()`:
       - Replace unconditional `if (message.metadata?.synthetic) continue;` with a guard that only skips when the debug flag is off.
       - Skip/relax the `MAX_DISPLAYED_MESSAGES` truncation when the debug flag is on.

3. **Mark synthetic messages for display**
   - `src/common/types/message.ts`: add `isSynthetic?: boolean` to the `DisplayedMessage` variant(s) you create from user messages.
[...]

PR #1714, opened 2026-01-16
🤖 fix: remove default MUX_DEBUG_LLM_REQUEST=1 from dev/start targets

ammar-agent:
Removes the default `MUX_DEBUG_LLM_REQUEST=1` from `make dev` and `make start` targets.

This was added in #1698 but causes verbose log spam on every LLM request during normal development.

Developers can still opt-in when needed:
```bash
make start MUX_DEBUG_LLM_REQUEST=1
```

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$0.31`_


ammario:
cc @ThomasK33 this default was killing my flow

PR #1738, opened 2026-01-17
🤖 fix: drop orphaned tool calls before sending

ammar-agent:
## Summary
- drop orphaned tool-call/tool-result pairs before provider transforms
- wire the self-healing pass into the model message pipeline
- add unit coverage for the orphaned tool cleanup

## Rationale
- providers (e.g., OpenAI Responses) reject tool calls without matching outputs; interrupted or corrupted histories can leave orphans that brick retries
- request-time pruning unblocks sends without mutating persisted history

## Why `ignoreIncompleteToolCalls` is insufficient
- `ignoreIncompleteToolCalls` only removes *explicitly incomplete* tool-call parts (e.g., `input-available` during streaming) during conversion
- the failures here come from **completed** tool-call parts whose matching tool-result messages are missing after crashes/truncation
- the SDK does not validate cross-message pairing, so the orphan survives conversion and only fails at provider validation time

## Regression Risk
- Low: only removes unmatched tool-call/tool-result parts; valid pairs (including tool results embedded in assistant messages) are preserved
- Worst case, we hide a missing result, but that request would be invalid anyway for strict providers

## Testing
- bun test src/browser/utils/messages/modelMessageTransform.test.ts
- make typecheck
- make static-check (fails locally: zizmor 403 from GitHub advisories)

---
_Generated with `mux` • Model: `openai:gpt-5.2-codex` • Thinking: `xhigh` • Cost: `$6.19`_

### DIFF
--- a/src/browser/utils/messages/modelMessageTransform.test.ts
+++ b/src/browser/utils/messages/modelMessageTransform.test.ts
@@ -8,6 +8,7 @@ import {
   injectModeTransition,
   filterEmptyAssistantMessages,
   injectFileChangeNotifications,
+  stripOrphanedToolCalls,
 } from "./modelMessageTransform";
 import type { MuxMessage } from "@/common/types/message";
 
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

Updated to treat tool results embedded in assistant messages as valid pairings and to drop orphan tool results in assistant content. Added coverage for both cases.

PR #1761, opened 2026-01-19
🤖 fix: run coder config-ssh on app startup

ethanndickson:
## Problem

When the `coder` binary moves on restart (e.g., self-updating installers or temp paths), the SSH config written by `coder config-ssh` becomes stale. The config points to the old binary location, breaking SSH connections to Coder workspaces.

Previously, `coder config-ssh --yes` only ran once per workspace during initial setup—not on app restart.

## Solution

Run `coder config-ssh --yes` once during `ServiceContainer.initialize()`:

- Checks if coder is available first via `getCoderInfo()` (no-op if coder not installed)
- Fire-and-forget so it doesn't block startup
- Logs warning on failure but doesn't crash

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$1.63`_


ethanndickson:
@codex review

PR #1766, opened 2026-01-19
🤖 fix: skip origin fetch for local-only branches

ethanndickson:
## Summary

Fixes a log noise issue introduced in #1408 where SSH/worktree subagent initialization logs confusing errors like:

```
Note: Could not fetch from origin (fatal: couldn't find remote ref subagent-rcva), using local branch state
```

This happens because subagent parent branches (e.g., `subagent-rcva`) are local-only and not pushed to origin.

## Fix

Before fetching, check if the branch exists on origin via `git ls-remote --heads origin <branch>`. If missing, log a non-error message and skip the fetch:

> `Skipping origin fetch: branch "subagent-rcva" not found on origin; using local state.`

## Retains #1408 behavior

For remote-tracked branches (like `main`, `master`, or pushed feature branches), the flow is unchanged:
- `ls-remote` finds the branch → proceeds to fetch
- Fetch succeeds → ancestry check → use origin or local per #1408's logic
- Fetch fails → fallback to local

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$6.18`_

### DIFF
--- a/src/node/runtime/SSHRuntime.ts
+++ b/src/node/runtime/SSHRuntime.ts
@@ -685,9 +685,14 @@ export class SSHRuntime extends RemoteRuntime {
       const fetchExitCode = await fetchStream.exitCode;
       if (fetchExitCode !== 0) {
         const fetchStderr = await streamToString(fetchStream.stderr);
-        initLogger.logStderr(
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

PR #1771, opened 2026-01-19
🤖 feat: show project chat on home when projects exist

ibetitsmike:
### Why
The home route showed the "Welcome to Mux" splash even when projects already exist (e.g. multiple configured projects but no workspace selected). That adds an extra click before the user can create a new chat/workspace.

### What changed
- If **any** projects exist and no workspace is selected, the app now routes directly to `ProjectPage` (new chat / workspace creation) instead of the splash screen.
- `ProjectPage` header now shows a **project selector** when multiple projects exist, so users can switch the creation view without going back.
- The splash copy now reflects the true empty state (no projects configured).

### Validation
- `make static-check` (passes locally, ~32s)

---
_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `high` • Cost: `$2.67`_

### DIFF
--- a/src/browser/App.tsx
+++ b/src/browser/App.tsx
@@ -104,9 +104,7 @@ function AppInner() {
   }, [sidebarCollapsed]);
   const defaultProjectPath = getFirstProjectPath(projects);
   const creationProjectPath = !selectedWorkspace
-    ? pendingNewWorkspaceProject && projects.has(pendingNewWorkspaceProject)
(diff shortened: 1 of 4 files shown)


ibetitsmike:
@codex review

Moved the project selector onto the project name in the creation breadcrumb (CreationControls), and removed the ProjectPage-level context requirements so isolated renders/tests don't crash.

PR #1778, opened 2026-01-20
🤖 perf: improve small-screen UX and reduce mobile CPU spikes

ibetitsmike:
## What

Mobile + small-screen improvements:

- **iOS safe-area header bleed**: add a `.safe-area-top-bleed` utility and apply it to top bars so the header background extends into the notch area without clipping interactive content.
- **Disable right sidebar on small screens (<= 768px)**:
  - Don’t render `<RightSidebar />`.
  - Hide right-sidebar command palette actions.
  - Route “open terminal” to the terminal popout instead of trying to target the sidebar.
- **Mobile performance quick wins**:
  - Avoid O(n²) bash_output grouping scans in `ChatPane` by computing group info in one pass.
  - Add a large-file fast path to `TextFileViewer` (render raw `<pre>` with an explicit opt-in to full render).
  - Skip git-status subscriptions in the workspace list on small screens.

## Why

On mobile / small screens the right sidebar is not usable and large chats/files can cause CPU spikes and jank. These changes prioritize:

- keeping the top bar usable on iOS notch devices
- preventing inaccessible UI from mounting on small screens
- reducing worst-case render work in long transcripts, large files, and large workspace lists

## Validation

- `make static-check`

---
_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `high` • Cost: `$8.32`_

### DIFF
--- a/src/browser/App.tsx
+++ b/src/browser/App.tsx
@@ -738,7 +738,7 @@ function AppInner() {
               })()
             ) : (
               <div className="bg-dark flex flex-1 flex-col overflow-hidden">
-                <div className="safe-area-top-bleed bg-sidebar border-border-light flex h-8 shrink-0 items-center border-b px-[15px] [@media(max-width:768px)]:h-auto [@media(max-width:768px)]:py-2">
+                <div className="safe-area-top-bleed bg-sidebar border-border-light flex h-8 shrink-0 items-center border-b px-[15px] [@media(max-width:768px)]:h-auto [@media(max-width:768px)]:px-2 [@media(max-width:768px)]:py-2">
                   {sidebarCollapsed && (
                     <Button
                       variant="ghost"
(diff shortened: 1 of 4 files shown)


ibetitsmike:
@codex review

Handled the small-screen terminal popout path with a best-effort `.catch(() => undefined)` to avoid unhandled promise rejections.

ibetitsmike:
@codex review

Updated `TextFileViewer` large-file fast path to key the “render highlighted” override by `filePath`, so switching files immediately re-enables the fast path (no one-render CPU spike).

ibetitsmike:
@codex review

Scoped the large-file “render highlighted” override by `workspaceId` + `relativePath` (passed through `FileViewerTab`), so identical paths in different workspaces won’t bypass the fast path.

PR #1782, opened 2026-01-20
🤖 feat: add SSH2 runtime using ssh2 library

ethanndickson:
## Summary

Unify SSH transports under a single `SSHRuntime` with pluggable transport abstraction. Adds hidden config option to switch between **OpenSSH (default on non-Windows) and SSH2 (always on Windows).**

## Motivation

- **Windows support**: OpenSSH ControlMaster is Unix-only; ssh2 library works cross-platform
- **Coder integration**: Coder uses `Match host ... !exec` blocks in SSH config for ProxyCommand; the ssh2 approach parses and honors these rules directly
- **Cleaner architecture**: Single runtime class with transport abstraction instead of parallel class hierarchies

### Architecture Refactor
- **Removed `SSH2Runtime.ts`**: No longer a separate class
- **New transport abstraction**: `SSHTransport` interface with `OpenSSHTransport` and `SSH2Transport` implementations
- **Unified `SSHRuntime`**: Delegates to transport for exec and PTY operations
- **`PtyHandle` interface**: Common interface for PTY sessions across transports; `PTYService` no longer branches by transport type

### New Files
- `src/node/runtime/transports/` - Transport abstraction layer
- `src/node/runtime/ptyHandle.ts` - Unified PTY interface
- `src/node/runtime/ptySpawn.ts` - Shared PTY spawn utility
- `src/node/runtime/shellQuote.ts` - Shell quoting for Docker paths

### SSH2 Auth Improvements
- Agent-first fallback: try SSH agent first, then default keys (`~/.ssh/id_rsa`, `id_ed25519`, etc.) on auth failure
- Two-pass connection flow in `SSH2ConnectionPool.connect()`

### Bug Fixes
- **Bundle corruption fix**: Removed `streamProcessToLogger` stdout drain that consumed git bundle data before pipe could read it
- **PTY double-quoting fix**: `expandTildeForSSH` already quotes paths; don't wrap again with `shellQuotePath`
- **PTY fail-fast**: Sessions use `maxWaitMs: 0` to avoid 2-minute backoff waits
- **Local timeout**: `resolvePath` and bundle upload use local abort-timeout (no remote `timeout` binary needed)

### Config Option
- Added `useSSH2Transport` boolean to `~/.mux/config.json`
- **Windows**: Always SSH2 (config ignored)
- **Other OSs**: Defaults to OpenSSH; set `"useSSH2Transport": true` to use SSH2

## Testing

- `make static-check` passes
- Manually validated OpenSSH and SSH2 transports on Linux
- Workspace creation, terminal sessions, git bundle sync all working

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$52.40`_

### DIFF
--- a/tests/runtime/runtime.test.ts
+++ b/tests/runtime/runtime.test.ts
@@ -1438,7 +1438,7 @@ describeIntegration("Runtime integration tests", () => {
           existingWorkspace: false,
         },
       };
-      const transport = createSSHTransport(config, true);
+      const transport = createSSHTransport(config, false);
       return new CoderSSHRuntime(config, transport, mockCoderService);
     };
 
@@ -1538,7 +1538,7 @@ describeIntegration("Runtime integration tests", () => {
               existingWorkspace: false, // Source was mux-created
             },
           };
-          const transport = createSSHTransport(config, true);
+          const transport = createSSHTransport(config, false);
           const runtime = new CoderSSHRuntime(config, transport, mockCoderService);
 
           const projectName = `coder-fork-postcreate-${Date.now()}-${Math.random().toString(36).substring(7)}`;
--- a/tests/runtime/test-fixtures/test-helpers.ts
+++ b/tests/runtime/test-fixtures/test-helpers.ts
@@ -50,7 +50,7 @@ export function createTestRuntime(
         identityFile: sshConfig.privateKeyPath,
         port: sshConfig.port,
       };
-      return new SSHRuntime(config, createSSHTransport(config, true));
+      return new SSHRuntime(config, createSSHTransport(config, false));
     }
     case "docker": {
       if (!dockerConfig) {


ethanndickson:
@codex review

ethanndickson:
@codex review

The previous comment about identity fallback is already addressed in the code:

1. `baseIdentityFiles` captures explicit config (lines 434-435)
2. `fallbackIdentityFiles` uses `DEFAULT_IDENTITY_FILES` (~/.ssh/id_rsa, id_ed25519, etc.) when no explicit files are configured (lines 436-439)
3. Agent-first flow: if agent exists and no explicit identity files, try agent-only first (lines 547-555)
4. On auth failure (or no agent), retry with `fallbackIdentityFiles` which includes default keys (lines 558-559)

This matches the behavior described in the concern - users without an agent but with default key files will have those keys tried automatically.

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

Regarding the P2 badge about `Match host` and alias/HostName semantics:

**TL;DR:** Our implementation correctly matches OpenSSH semantics. No code changes needed.

OpenSSH defines:
- `Match host` — evaluated against the target hostname **after** `HostName`/`CanonicalizeHostname` substitution
- `Match originalhost` — evaluated against the hostname as specified on the command line (the alias)
- `%h` — expands to the resolved HostName
- `%n` — expands to the original host argument

So with:
```
Host myalias
  HostName 10.0.0.5
```

Running `ssh myalias`:
- `Match host myalias` does **not** match (because `Match host` sees `10.0.0.5`)
- `Match host 10.0.0.5` **matches**
- `Match originalhost myalias` **matches**

Our code passes the resolved `HostName` into `applyNegatedExecMatch()`, which is correct. Configs that want alias-based matching should use `Match originalhost` and `%n` instead of `Match host` and `%h`.

Sources: [man7.org ssh_config(5)](https://man7.org/linux/man-pages/man5/ssh_config.5.html), [man.openbsd.org ssh_config(5)](https://man.openbsd.org/ssh_config)

ethanndickson:
@codex review

Addressed both P2 items:

1. **Default `%r` to local username** — Added `getDefaultUsername()` helper in `sshConfigParser.ts` and now pass `userOverride ?? userFromConfig ?? getDefaultUsername()` into `applyNegatedExecMatch()` so `%r` always has a value matching the actual connection user.

2. **Fail SSH2 PTY on cd failure** — Changed `SSH2Transport.createPtySession()` to write `cd ${path} || exit 1` instead of just `cd ${path}`, matching OpenSSH transport behavior.

Added test coverage for the `%r` fix.

ethanndickson:
Tested SSH2 on Windows & Linux, tested OpenSSH on Linux, merging..

PR #1786, opened 2026-01-20
🤖 fix: restore agent skills slash commands

ThomasK33:
Fixes agent skills slash commands (e.g. `/mux-docs`) by making skill discovery/resolution resilient and debuggable.

- Backend: align agent skill discovery runtime context with send-message path (workspaceName for Docker containerName derivation, in-place workspace handling).
- Backend: treat Docker as a remote runtime for skill directory listing and use a stable remote cwd.
- Frontend: surface failures from `agentSkills.list/get` with toasts instead of falling through to “Unknown command”.
- UI: display `muxMetadata.rawCommand` for agent-skill user messages.
- Tests: cover `agentSkills.list/get` with `workspaceId` and add a transcript display regression test.

---

<details>

# Plan: Restore agent skills as slash commands (suggestions + execution + display)

## Findings (what changed + when)

- **Feature merged:** `c172866ade9de0a61996b6eb6174ce864aa72d78` — **Mon Jan 19, 2026** — `🤖 feat: expose agent skills as slash commands (#1673)`
  - Adds `agentSkills.list/get` oRPC endpoints + ChatInput interception for `/{skill}`.

## Current symptoms (from your repro)

- Skills **do not appear** in `/` suggestions (even with prefix filtering and after scrolling).
- Manually typing `/mux-docs` results in **“Unknown command”**.
- **Normal chat works**, so this is likely isolated to skill discovery/lookup.

## Most likely failure point

In `src/browser/components/ChatInput/index.tsx`:
- Suggestions require `agentSkillDescriptors` loaded via `api.agentSkills.list(...)`.
- Execution requires `api.agentSkills.get(...)` to resolve unknown slash commands into a skill invocation.
- **Both calls can fail today without user-visible error details** (list logs to console; get is fully swallowed), which makes real backend failures look like “Unknown command”.

### Approach A (recommended): Make skill lookup reliable + make failures debuggable
**Net LoC (product code only): ~40–120**

1. **Surface the real error (frontend)**
   - File: `src/browser/components/ChatInput/index.tsx`
   - When `api.agentSkills.list(...)` fails, show a *non-spammy* toast like “Failed to load agent skills” with a short error message.
   - When the user enters `/{something}` and `api.agentSkills.get(...)` throws, show a toast like:
     - “Failed to resolve skill ‘{name}’: {error}”
     - (instead of silently falling through to “Unknown command”).

2. **Harden agent skills discovery/lookup (backend)**
   - Goal: `agentSkills.list/get` should be *best-effort* and never take down the feature due to a runtime-specific edge.

   Concrete hardening targets:
   - `src/node/orpc/router.ts` → `resolveAgentDiscoveryContext(...)`
     - Align runtime construction with the send-message path (pass `workspaceName`, and add a defensive fallback if `runtimeConfig` is missing/corrupt).
     - Consider mirroring the “in-place workspace” logic used by `aiService.sendMessage` (projectPath === name).
     - Ensure discovery **cannot throw** due to remote runtime issues.
     - Fix remote-runtime directory listing: Docker is a remote runtime too; don’t use local FS reads for Docker paths.
       - Prefer switching the `runtime instanceof SSHRuntime` check to a `RemoteRuntime`-based check (SSH + Docker).

3. **Add a regression test that matches your repro**
   - Add a node-side test that exercises the router/service and asserts:
     - `agentSkills.list({ workspaceId })` returns at least the built-ins (e.g. `mux-docs`)
     - `agentSkills.get({ workspaceId, skillName: "mux-docs" })` succeeds
   - Add a small UI/unit test (if feasible) that verifies:
     - failing `agentSkills.get` produces a *skill-specific* error toast (not generic “unknown command”).

### Approach B (follow-up): Fix transcript display for skill invocations
**Net LoC (product code only): ~5–15**

Once skills execute again, also fix the **display regression** so the transcript shows the original slash command (`muxMetadata.rawCommand`) instead of the synthetic “Run the … skill” text.

- Likely regression commit: `5e45e7cc74af3351892650849ffe1f87116bf6ed` (`🤖 fix: add compact retry shortcut for stream errors (#1752)`).
- File: `src/browser/utils/messages/StreamingMessageAggregator.ts`

## Validation / exit criteria

- Typing `/mux-docs` runs the skill (no “Unknown command”).
- Skill suggestions appear in the `/` menu (and via prefix filtering like `/mux`).
- (After follow-up) transcript displays the original slash command.
- `make typecheck` + `make static-check` pass.

<details>
<summary>Notes</summary>

- If the underlying failure is runtime-specific (SSH/Coder/Docker), the backend hardening step is the highest ROI: it prevents skills from disappearing due to transient runtime issues.
- Skill shadowing by built-in slash commands (same name) is still expected behavior.

</details>

</details>

---
_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: $4.36_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -293,7 +293,7 @@ lint-zizmor: ## Run zizmor security analysis on GitHub Actions workflows
 	@./scripts/zizmor.sh --min-confidence high .
 
 # Shell files to lint (excludes node_modules, build artifacts, .git)
-SHELL_SRC_FILES := $(shell find . -not \( -path '*/.git/*' -o -path './node_modules/*' -o -path './build/*' -o -path './dist/*' \) -type f -name '*.sh' 2>/dev/null)
+SHELL_SRC_FILES := $(shell find . -not \( -path '*/.git/*' -o -path '*/node_modules/*' -o -path '*/build/*' -o -path '*/dist/*' -o -path '*/.venv/*' -o -path '*/Pods/*' \) -type f -name '*.sh' 2>/dev/null)
 
 lint-shellcheck: ## Run shellcheck on shell scripts
 	shellcheck --external-sources $(SHELL_SRC_FILES)


ThomasK33:
@codex review

I've restored the agent skills and implemented the emoji-to-icon migration. I also ensured that the "Cost" tab is the default for new workspaces and fixed a Makefile issue where shellcheck was linting `node_modules`.

PR #1792, opened 2026-01-20
🤖 feat: bucket sidebar workspaces by live state

ibetitsmike:
## What

- Group sidebar workspaces into **Action Needed / Running / Done** buckets while preserving Sections and parent→child threads.
- Keep the existing “Older than…” collapse UI, but only inside **Done** (applied per thread root).
- Add a bulk `useWorkspaceSidebarStateMap()` hook so bucketing can react to live state without per-item subscriptions.
- Remove `RuntimeBadge` from `WorkspaceListItem` (post-runtime-icon cleanup).

## Why

This makes the workspace list easier to scan by surfacing items that need attention or are actively running at the top, without breaking existing drag/drop or thread grouping.

## Validation

- `make static-check`

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$7.54`_

### DIFF
--- a/src/browser/utils/workspace.ts
+++ b/src/browser/utils/workspace.ts
@@ -13,6 +13,7 @@ export function getWorkspaceSidebarKey(meta: FrontendWorkspaceMetadata): string 
     meta.name,
     meta.title ?? "", // Display title (falls back to name in UI)
     meta.status ?? "", // Working/idle status indicator
+    meta.incompatibleRuntime ?? "", // Incompatible runtime overlay / needs-attention bucket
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

- Included `incompatibleRuntime` in `getWorkspaceSidebarKey` so sidebar buckets update when compatibility flips.
- Updated E2E sidebar selector to target `[role="button"][data-workspace-id]` (bucket headers add wrapper nodes).

PR #1797, opened 2026-01-20
🤖 fix: stats tab doesn't steal sidebar focus

ThomasK33:
Fixes a regression where enabling the Stats tab experiment caused the right sidebar to default to **Stats** instead of **Costs**.

- Inject Stats tab without activating it (`addTabToFocusedTabset(..., false)`).
- Add UI integration coverage to ensure Costs remains the selected tab when Stats is enabled.

Validation:
- `make static-check`

---

<details>

# Fix: Stats tab experiment shouldn’t steal default focus

## Problem
When the `statsTab` experiment is enabled, opening a workspace causes the right sidebar to land on **Stats** instead of the historical default **Costs**.

Root cause: `RightSidebar.tsx` injects the `"stats"` tab into the persisted sidebar layout via `addTabToFocusedTabset(prev, "stats")`. That helper defaults `activate=true`, so the injection also **selects** Stats (updates `activeTab`).

### 1) Keep adding the Stats tab, but don’t auto-activate it
- Update `src/browser/components/RightSidebar.tsx`:
  - Change the injection call to: `addTabToFocusedTabset(prev, "stats", false)`.
  - This keeps the current `activeTab` (for new workspaces that’s `costs`; for existing workspaces it preserves whatever the user last used).

### 2) Add regression coverage
- Update `tests/ui/rightSidebar.integration.test.ts` (or add a new test in the same file):
  - Enable the stats tab experiment in the test harness.
  - Assert the **Stats** tab exists.
  - Assert **Costs** is still the selected tab on first render (`aria-selected="true"`).

### 3) Validate
- `make typecheck`
- `bun test tests/ui/rightSidebar.integration.test.ts`

<details>
<summary>Alternative considered</summary>

**Change `addTabToFocusedTabset` default to `activate=false`.**

Pros: makes the helper’s JSDoc (“without changing the active tab”) match its behavior.

Cons: riskier because other callsites appear to rely on the current default (`true`) when creating/adding a tab that *should* become active (e.g. opening a new terminal/file tab). This would require auditing/updating callsites.

</details>

</details>

---
_Generated with [`mux`](https://github.com/coder/mux) • Model: openai:gpt-5.2 • Thinking: xhigh • Cost: $2.42_

### DIFF
--- a/tests/ui/rightSidebar.integration.test.ts
+++ b/tests/ui/rightSidebar.integration.test.ts
@@ -27,7 +27,8 @@ import {
   RIGHT_SIDEBAR_COLLAPSED_KEY,
   getRightSidebarLayoutKey,
 } from "@/common/constants/storage";
-// RightSidebarLayoutState used for initial setup via localStorage - acceptable for test fixtures
+import { updatePersistedState } from "@/browser/hooks/usePersistedState";
+// RightSidebarLayoutState used for initial setup via persisted-state helpers - acceptable for test fixtures
 import type { RightSidebarLayoutState } from "@/browser/utils/rightSidebarLayout";
 
 const describeIntegration = shouldRunIntegrationTests() ? describe : describe.skip;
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

Addressed feedback: RightSidebar UI test now uses persisted-state helpers (`updatePersistedState`) instead of direct `localStorage` access.

PR #1803, opened 2026-01-20
🤖 feat: add investigation step to plan prompt

ammar-agent:
## Summary
- add an investigation requirement + evidence line to the Plan agent prompt
- note that sub-agents are optional
- refresh generated agent docs

## Testing
- make static-check

## Why
Codex tends to draft plans without investigating; this makes the investigation step explicit and requires evidence before planning.

---

_Generated with `mux` • Model: `openai:gpt-5.2-codex` • Thinking: `xhigh` • Cost: `$1.98`_


ammar-agent:
@codex review

Resolved the Evidence/plan conflict by keeping the requirement to always produce a plan and adding a Questions section when evidence is missing.

PR #1813, opened 2026-01-21
🤖 fix: require exact plan file path in plan mode

ThomasK33:
Require exact plan file path string in plan mode

Plan-mode file edits now require the agent to use the *exact* plan file path string shown in the system prompt (i.e., the `~/...` path), instead of expanded absolute paths or alternate relative paths that resolve to the same file.

Also tightens the plan-mode prompt and file_read hint so agents get actionable recovery guidance.

Validation:
- make typecheck
- bun test src/node/services/tools/fileCommon.test.ts
- bun x jest src/node/services/tools/file_edit_operation.test.ts -t "plan mode"
- make static-check

---

<details>

# Plan: Enforce plan-file *relative* path usage in Plan Mode

## Goal
When an agent calls a file-edit tool in **plan mode** with an **absolute** path that resolves to the plan file (e.g. `/Users/alice/.mux/plans/...`) instead of the provided **relative** plan path (e.g. `~/.mux/plans/...`), detect it and return a **high-signal** error telling the agent to use the exact relative path.

Also strengthen the Plan Mode system instructions so agents are less likely to make this mistake.

### 1) Strengthen Plan Mode instructions (prompt)
**File:** `src/common/utils/ui/modeUtils.ts`

- Update the first line to explicitly require the exact path string:
  - Current: `Plan file path: ${planFilePath}`
  - New: `Plan file path: ${planFilePath} (MUST use this exact relative path for tool calls; do NOT expand to an absolute path like /Users/... )`
- Add one extra bullet near the “NOTE: The plan file is the only file…” section:
  - “You must use the plan file path exactly as shown (including the leading `~/`).”

**Why:** This addresses the root cause before tools fail.

### 2) Enforce “exact string” plan-file path for edits in plan mode
**File:** `src/node/services/tools/fileCommon.ts`

Update `validatePlanModeAccess(filePath, config)`:

- In `mode === "plan"`:
  1. **Allow** edits only when `filePath === config.planFilePath`.
  2. If `filePath !== config.planFilePath`:
     - If `await isPlanFilePath(filePath, config)` is **true** (i.e., absolute/alternate path resolves to the plan file), return a **special-case** error:
       - Emphasize it *does* resolve to the plan file, but we require the exact relative path.
     - Else return the standard plan-mode restriction error, but with stronger wording:
       - “You must use the exact plan file path … (attempted: …)”

Suggested error text (keep concise / model-friendly):
- **Special-case (resolves to plan file):**
  - `In plan mode, you must use the exact plan file path from the instructions: ${config.planFilePath} (attempted: ${filePath}; this resolves to the plan file but absolute/alternate paths are not allowed)`
- **Generic (not plan file):**
  - `In plan mode, only the plan file can be edited. You must use the exact plan file path: ${config.planFilePath} (attempted: ${filePath})`

### 3) Make tool hints consistent (optional but low-risk)
**File:** `src/node/services/tools/file_read.ts`

- Strengthen the existing hint to mirror the same “exact relative path” language:
  - Current: “use the plan file path exactly as provided”
  - New: “use the exact plan file path (relative, starting with `~/`) as provided”

### Update existing plan-mode enforcement tests
**File:** `src/node/services/tools/file_edit_operation.test.ts`

- Update the existing test `"should handle relative path to plan file in plan mode"`:
  - Old behavior: allowed alternate relative path (`../.mux/...`) if it resolves to the plan file.
  - New behavior: should **fail** with the special-case error requiring `planFilePath` exact string.

### Add coverage for the new absolute-path detection
Prefer to add a focused unit test beside existing ones:

**File:** `src/node/services/tools/fileCommon.test.ts`

Add a `describe("validatePlanModeAccess")` block that asserts:
1. `filePath === planFilePath` ⇒ passes in plan mode.
2. `filePath` is an **absolute** path that resolves to plan file ⇒ fails with the special-case “must use exact plan file path” error.
3. Editing a non-plan file in plan mode ⇒ fails with the stronger generic error.

(Alternatively, add the absolute-path case to the existing plan-mode section in `file_edit_operation.test.ts`—either is fine.)

## Validation
- `make typecheck`
- `bun x jest src/node/services/tools/fileCommon.test.ts`
- `bun x jest src/node/services/tools/file_edit_operation.test.ts -t "plan mode"`

---

<details>
<summary>Alternative: auto-correct absolute plan path instead of hard-failing (net +25 LoC)</summary>

Instead of failing when `filePath` resolves to the plan file, we could rewrite it to `config.planFilePath` (similar to `validateNoRedundantPrefix`) and return `success: true` with a warning.

Pros:
- Fewer retries; smoother agent runs.

Cons:
- Agents may never learn; absolute paths keep appearing.
- Harder to guarantee “must use relative” compliance.

</details>

</details>

---
_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: $2.12_

### DIFF
--- a/src/common/utils/ui/modeUtils.ts
+++ b/src/common/utils/ui/modeUtils.ts
@@ -7,17 +7,20 @@
  * <additional-instructions> in the system prompt.
  */
 export function getPlanModeInstruction(planFilePath: string, planExists: boolean): string {
+  const exactPlanPathRule = planFilePath.startsWith("~/")
+    ? "You must use the plan file path exactly as shown (including the leading `~/`); do not expand `~` or use alternate paths that resolve to the same file."
+    : "You must use the plan file path exactly as shown; do not rewrite it or use alternate paths that resolve to the same file.";
   const fileStatus = planExists
     ? `A plan file already exists at ${planFilePath}. First, read it to determine if it's relevant to the current request. If the current request is unrelated to the existing plan, delete the file and start fresh. If relevant, make incremental edits using the file_edit_* tools.`
     : `No plan file exists yet. You should create your plan at ${planFilePath} using the file_edit_* tools.`;
 
-  return `Plan file path: ${planFilePath} (MUST use this exact relative path string for tool calls; do NOT expand it to an absolute path like /Users/...)
+  return `Plan file path: ${planFilePath} (MUST use this exact path string for tool calls; do NOT rewrite it into another form, even if it resolves to the same file)
 
 ${fileStatus}
 
 Build your plan incrementally by writing to or editing this file.
 NOTE: The plan file is the only file you are allowed to edit. Other than that you may only take READ-ONLY actions.
-You must use the plan file path exactly as shown (including the leading \`~/\`); do not use absolute paths or alternate relative paths that resolve to the same file.
+${exactPlanPathRule}
 
 Keep the plan crisp and focused on actionable recommendations:
 - Put historical context, alternatives considered, or lengthy rationale into collapsible \`<details>/<summary>\` blocks so the core plan stays scannable.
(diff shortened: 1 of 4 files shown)


ThomasK33:
@codex review

Addressed the prompt/hint wording to avoid assuming the plan path is `~/`-relative (it now adapts based on the provided planFilePath). Updated tool tests accordingly.

PR #1815, opened 2026-01-21
🤖 fix: avoid MCP tool calls on closed clients

ThomasK33:
Fix recurring MCP tool errors where the backend attempts to call MCP tools on a client that has already been closed.

Key changes:
- Add workspace-scoped MCP server leases for the lifetime of an active stream.
- Prevent idle cleanup from stopping MCP servers while leased.
- Mark MCP server activity on every tool execution.
- Defer config-driven restarts while leased; restart once the stream completes.
- Track `isClosed` on MCP instances and restart on the next generation.

Validation:
- `make static-check`

---

<details>

# Fix MCP “Attempted to send a request from a closed client” tool errors

## Problem
Mux sometimes logs errors like:

- `MCPClientError: Attempted to send a request from a closed client`

These show up during MCP tool execution (e.g. `chrome_list_pages`) and indicate we’re attempting to execute a tool against an MCP client/transport that Mux has already closed.

## Likely root cause (from code inspection)
- `MCPServerManager` keeps **per-workspace** MCP server instances/clients in memory.
- It has an **idle cleanup** that stops servers after **10 minutes** (`IDLE_TIMEOUT_MS`).
- The idle timer’s `lastActivity` is updated primarily when `getToolsForWorkspace()` is called (i.e., usually only at the **start of a generation**).

**Failure mode:** if a generation/agent run stays “active” for >10 minutes (model thinking time, long tool sequences, background stream), the cleanup can close MCP clients **while the stream is still running**, and the next tool call during that stream throws “closed client”.

<details>
<summary>Other contributing races worth addressing</summary>

- **Server crash / transport close:** if an MCP stdio process exits or a transport closes, we don’t currently “self-heal” by restarting it until a later restart.
- **Config signature restarts:** `getToolsForWorkspace()` may call `stopServers()` (closing clients) when configuration changes; if that can happen while a stream is active, it can invalidate in-flight tool objects.

</details>

---

### Approach A (recommended): make idle cleanup aware of active streams + mark activity on tool calls
**Net LoC estimate (product code): ~120–200 LoC**

1) **Add an explicit “in use” signal for a workspace’s MCP servers**
- Add an `acquireLease(workspaceId)` / `releaseLease(workspaceId)` or `markWorkspaceInUse(workspaceId, boolean)` API on `MCPServerManager`.
- Implement it as a small refcount (`Map<workspaceId, number>`) so nested usage is safe.

2) **Wire StreamManager → MCPServerManager leases**
- When a stream starts processing for a workspace, `acquireLease(workspaceId)`.
- In a `finally` block when the stream ends/cancels/errors, `releaseLease(workspaceId)`.

3) **Change `cleanupIdleServers` to never close servers that are leased**
- Only apply `IDLE_TIMEOUT_MS` when `leaseCount === 0`.
- This removes the “close during active stream” race entirely.

4) **Defer config-driven restarts while leased**
- If `getToolsForWorkspace()` detects a config signature change but `leaseCount > 0`, don’t call `stopServers()` immediately.
- Instead, record a `pendingRestart` flag/signature and restart once after the stream finishes (lease released).

5) **Update `lastActivity` on each MCP tool execution**
- Extend `wrapMCPTools()` to call back into `MCPServerManager.markActivity(workspaceId)` on every `execute()`.
- This improves correctness even without leases (tool-heavy sessions).

6) **Self-heal obvious dead instances on the *next* generation**
- Track an `isClosed` flag per `MCPServerInstance` (set on `close()` and on transport `onclose`).
- In `getToolsForWorkspace()`, if any cached instance is `isClosed`, stop+restart before returning tools.

> Why this is preferred vs auto-retry: it prevents Mux from ever calling into a closed client in the first place, and it avoids retrying potentially side-effectful tools.

---

### Approach B: increase/disable idle timeout
**Net LoC estimate: ~1–10 LoC**
- Increase `IDLE_TIMEOUT_MS` (e.g., 60 minutes) or gate cleanup behind a config flag.

Tradeoff: simplest, but risks leaving crashed/stuck stdio servers around longer; doesn’t fix the underlying lifecycle race.

### Approach C: proxy tool execution + restart-on-demand
**Net LoC estimate: ~220–350 LoC**
- Instead of returning MCP SDK-provided tool objects (whose `execute` closures capture a specific client), return “proxy” tools whose `execute()` always resolves the *current* live instance and calls `client.callTool`.
- On “closed client” errors, restart the server instance and retry **once**.

Tradeoff: most robust (recovers mid-stream) but bigger refactor and higher risk.

---

## Guardrails / correctness notes
- Avoid blind retries of tool calls: many MCP tools are not guaranteed idempotent (e.g., “click”).
- If we later add retry, constrain it to cases we can prove were never sent (or only apply to read-only tools).
- Add defensive assertions where we “should never” be executing against a closed instance (internal invariant).

---

## Validation plan
- Add unit tests for `cleanupIdleServers` using fake timers:
  - leased workspace is never cleaned
  - after lease released and timeout elapses, servers are cleaned
- Add a regression test that simulates a long-running stream (> `IDLE_TIMEOUT_MS`) and ensures tool execution still works (can be unit-level by calling the wrapper’s execute).
- Manual repro: start an agent run, wait >10 minutes, then trigger an MCP tool; verify no “closed client” log.

</details>

---
_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$1.50`_

### DIFF
--- a/tests/ipc/streamCollector.ts
+++ b/tests/ipc/streamCollector.ts
@@ -152,7 +152,10 @@ export class StreamCollector {
               this.subscriptionReadyResolve = null;
             }
           }
-          // Don't store caught-up in events - it's just a signal
+
+          // Don't store caught-up in events - it's just a signal.
+          // But still satisfy any waiters so tests can await waitForEvent("caught-up").
+          this.checkWaiters(message);
           continue;
         }
 


ThomasK33:
@codex review

Addressed the feedback about deferring restarts while leased:
- We now *skip* the deferral when a cached MCP instance is already marked `isClosed`, so we don't keep serving tools backed by a dead client.
- Also fixed Windows CI regression ensuring the bash tool always injects `NON_INTERACTIVE_ENV_VARS` (incl. GIT_TERMINAL_PROMPT=0).

ThomasK33:
@codex review

Addressed the "avoid restarting servers while a lease is active" feedback:
- When `leaseCount > 0`, we no longer do a full `stopServers()` restart (even if some instances are closed).
- Instead we restart *only* the closed server instance(s) and keep healthy instances alive so in-flight tool calls don’t get invalidated.
- Added a unit test ensuring healthy instances are not closed during this partial restart.

ThomasK33:
@codex review

Applied the “avoid returning tools from disabled servers” feedback:
- During the lease deferral path we now filter `existing.instances` against the newly computed `enabledServers` when building the tool map.
- Added a unit test that disables a server via workspace overrides while leased and asserts its tools are not returned.

ThomasK33:
@codex review

PR #1840, opened 2026-01-22
🤖 feat: add / hotkey for review search

ThomasK33:
## Summary
Add a GitHub-style `/` shortcut to focus the Code Review panel’s search input.

## Implementation
- Added `KEYBINDS.FOCUS_REVIEW_SEARCH_QUICK` (`/`).
- ReviewPanel now focuses search on `/` (when the key event did **not** originate from an editable element).
- Updated the search placeholder + Settings → Keybinds list to show the new shortcut.

## Validation
- `bun test src/browser/utils/ui/keybinds.test.ts`
- `make typecheck`
- `make static-check`

## Risks
Low. Main risk is unexpected keybind conflicts while the Review tab is open; `/` is guarded so it won’t steal focus while typing in inputs.

---

<details>

# Add `/` as a secondary hotkey to focus ReviewPanel search

## Context / Why
When the Code Review **ReviewPanel** is open, we already support `Ctrl+F` / `Cmd+F` to focus the panel’s search input. Add a second shortcut: pressing `/` should also focus the same search input (GitHub-style), so users can start searching without the modifier key.

Key UX constraints:
- `/` must **not** steal focus while the user is typing in any editable element (inputs, textareas, contenteditable, terminal textarea, chat input, etc.).
- It should only apply while the **ReviewPanel is mounted** (i.e., Review tab is active).

## Evidence (what I inspected)
- `src/browser/components/RightSidebar/CodeReview/ReviewPanel.tsx`
  - Search input uses `searchInputRef` and shows a placeholder referencing `KEYBINDS.FOCUS_REVIEW_SEARCH`.
  - A `window.addEventListener("keydown", …)` handler focuses search when `matchesKeybind(e, KEYBINDS.FOCUS_REVIEW_SEARCH)` (currently `Ctrl/Cmd+F`).
- `src/browser/utils/ui/keybinds.ts`
  - `KEYBINDS.FOCUS_REVIEW_SEARCH` is defined as `{ key: "f", ctrl: true }`.
  - Utilities exist for safe hotkeys: `matchesKeybind(...)`, `isEditableElement(...)`.
- Tests already exist that cover keybind matching and ReviewPanel keyboard shortcuts:
  - `src/browser/utils/ui/keybinds.test.ts`
  - `tests/ui/reviewRefresh.integration.test.ts`

### 1) Add a new keybind constant
**File:** `src/browser/utils/ui/keybinds.ts`
- Add `KEYBINDS.FOCUS_REVIEW_SEARCH_QUICK` (name can vary) with `{ key: "/" }`.
- Keep a short comment explaining this is the “quick” (vim/GitHub-style) shortcut.

### 2) Handle `/` in the existing ReviewPanel keydown listener
**File:** `src/browser/components/RightSidebar/CodeReview/ReviewPanel.tsx`
- Extend the existing `handleKeyDown`:
  - `else if (matchesKeybind(e, KEYBINDS.FOCUS_REVIEW_SEARCH_QUICK)) { … }`
  - First guard: `if (isEditableElement(e.target)) return;`
  - Then `e.preventDefault(); searchInputRef.current?.focus();`
  - (Optional) call `stopKeyboardPropagation(e)` after handling to avoid any other global listeners seeing the event.

### 3) Update the search input placeholder to advertise both shortcuts
**File:** `src/browser/components/RightSidebar/CodeReview/ReviewPanel.tsx`
- Update placeholder from:
  - `Search... (<Ctrl/Cmd+F>)`
- To something like:
  - `Search... (<Ctrl/Cmd+F>, /)`
  - Prefer using `formatKeybind(KEYBINDS.FOCUS_REVIEW_SEARCH_QUICK)` if it renders cleanly; otherwise hardcode `"/"`.

### 4) Tests
**Unit:** `src/browser/utils/ui/keybinds.test.ts`
- Add/extend coverage that `matchesKeybind(createEvent({ key: "/" }), KEYBINDS.FOCUS_REVIEW_SEARCH_QUICK)` is `true`.
- Add an assertion that it does **not** match when modifiers are held (e.g., `ctrlKey: true`) to avoid conflicting with `Ctrl+/` keybinds.

**UI integration:** `tests/ui/reviewRefresh.integration.test.ts`
- Add a test case:
  - Render ReviewPanel
  - Trigger a `/` key press
  - Assert the review search input becomes `document.activeElement`
  - Assert the search input value did not become `"/"` (i.e., focusing didn’t type).
- Add a second assertion for the guard:
  - Focus an editable element (e.g., chat input if available in the harness)
  - Press `/`
  - Assert focus does **not** move to the review search.

## Alternative approach (B): Inline-check `e.key === "/"` only in ReviewPanel (net +6 LoC product)
- Skip adding a new `KEYBINDS` entry; just check `if (e.key === "/" && !isEditableElement(e.target)) …`.
- Faster, but less consistent/discoverable, and harder to test/reuse.

## Rollout / Verification checklist
- Manual:
  - Open Review tab → press `/` → search input focuses.
  - While typing in any input/terminal/chat → press `/` → `/` types normally; focus does not jump.
  - `Ctrl+F` / `Cmd+F` still focuses review search.
- Automated:
  - `make typecheck`
  - `bun test src/browser/utils/ui/keybinds.test.ts`
  - UI test(s): `TEST_INTEGRATION=1 bun x jest tests/ui/reviewRefresh.integration.test.ts -t "/"` (or similar filter)

</details>

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: $0.69_

### DIFF
--- a/src/browser/utils/ui/keybinds.test.ts
+++ b/src/browser/utils/ui/keybinds.test.ts
@@ -50,6 +50,11 @@ describe("CYCLE_MODEL keybind (Ctrl+/)", () => {
 
 describe("matchesKeybind", () => {
   describe("FOCUS_REVIEW_SEARCH_QUICK keybind (/)", () => {
+    it("matches Shift+/ when event.key is /", () => {
+      const event = createEvent({ key: "/", shiftKey: true });
+      expect(matchesKeybind(event, KEYBINDS.FOCUS_REVIEW_SEARCH_QUICK)).toBe(true);
+    });
+
     it("matches plain /", () => {
       const event = createEvent({ key: "/" });
       expect(matchesKeybind(event, KEYBINDS.FOCUS_REVIEW_SEARCH_QUICK)).toBe(true);
--- a/src/browser/utils/ui/keybinds.ts
+++ b/src/browser/utils/ui/keybinds.ts
@@ -89,8 +89,10 @@ export function matchesKeybind(
   if (ctrlRequired && !ctrlPressed) return false;
   if (!ctrlAllowed && ctrlPressed) return false;
 
+  const allowShift = keybind.allowShift ?? false;
+
   if (keybind.shift && !event.shiftKey) return false;
-  if (!keybind.shift && event.shiftKey) return false;
+  if (!keybind.shift && !allowShift && event.shiftKey) return false;
 
   if (keybind.alt && !event.altKey) return false;
   if (!keybind.alt && event.altKey) return false;
@@ -302,7 +304,7 @@ export const KEYBINDS = {
 
   /** Focus search input in Code Review panel (quick) */
   // GitHub-style: / to focus search
-  FOCUS_REVIEW_SEARCH_QUICK: { key: "/" },
+  FOCUS_REVIEW_SEARCH_QUICK: { key: "/", allowShift: true },
 
   /** Mark selected hunk as read/unread in Code Review panel */
   TOGGLE_HUNK_READ: { key: "m" },
--- a/src/common/types/keybind.ts
+++ b/src/common/types/keybind.ts
@@ -3,6 +3,13 @@ import assert from "@/common/utils/assert";
 export interface Keybind {
   key: string;
   ctrl?: boolean;
+  /**
+   * Allow Shift even when this keybind doesn't require it.
+   *
+   * Useful for keyboard layouts where producing a character (e.g. "/") requires Shift,
+   * but the resulting `KeyboardEvent.key` is still that character.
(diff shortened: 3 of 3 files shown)


ThomasK33:
@codex review

Addressed the non‑US keyboard layout issue: added `allowShift` support to `matchesKeybind` and set it on `FOCUS_REVIEW_SEARCH_QUICK`, plus a unit test covering `event.key === "/"` with `shiftKey: true`.

PR #1878, opened 2026-01-23
🤖 fix: normalize mux-gateway provider handling

ammar-agent:
Summary
- normalize provider-specific message handling so mux-gateway models follow the underlying provider behavior
- reuse OpenAI previousResponseId for mux-gateway models by comparing normalized model names
- map xAI grok-4-1-fast variants even when routed through mux-gateway, plus a regression test

Background
Provider-specific transforms and OpenAI response ID reuse relied on raw model strings, so mux-gateway models were treated as provider "mux-gateway" and lost provider-specific logic.

Implementation
- derive a normalized provider for message transforms and compliance checks
- preserve mux-gateway prefixes while mapping xAI fast variants
- normalize model identifiers when searching for OpenAI previousResponseId
- add a unit test covering mux-gateway OpenAI response ID reuse

Validation
- make static-check

Risks
Low. Changes are scoped to mux-gateway normalization for provider-specific message handling and response ID reuse.

---

_Generated with `mux` • Model: `openai:gpt-5.2-codex` • Thinking: `xhigh` • Cost: `$3.92`_

### DIFF
--- a/src/node/services/streamManager.ts
+++ b/src/node/services/streamManager.ts
@@ -2125,8 +2125,10 @@ export class StreamManager extends EventEmitter {
 
     const errorCode = this.extractErrorCode(error);
     const statusCode = this.extractStatusCode(error);
-    // Record if: we have the specific error code, OR a likely status code
-    // (400 included because mux-gateway wraps OpenAI's 404 as 400)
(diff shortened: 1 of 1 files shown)


ammario:
Note: I manually verified this successfully mitigates the lost response IDs on Gateway.

PR #1885, opened 2026-01-23
🤖 fix: remove extra bottom gap on mobile

ibetitsmike:
Summary:
- Remove the global bottom safe-area padding that created a visible blank band under the chat composer in mobile layout.
- Apply the safe-area inset as internal padding on the workspace chat input instead, so the composer background sits flush to the bottom while keeping controls above the home indicator.

Validation:
- `make static-check`

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$2.20`_

### DIFF
--- a/src/browser/App.tsx
+++ b/src/browser/App.tsx
@@ -702,7 +702,7 @@ function AppInner() {
   return (
     <>
       <div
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #1886, opened 2026-01-24
🤖 fix: use unread sidebar bar indicator

ammar-agent:
Summary
- Show unread workspaces using the existing left selection bar with a muted color to avoid competing with the archive button.
- Keep per-workspace unread timestamps (with migration) while dropping the unused manual toggle wiring.

Background
- The unread dot previously competed with the archive affordance for the same slot.
- We already track unread timestamps per workspace, so we can style an existing indicator instead of adding UI chrome.

Implementation
- Render an absolute left bar for selected/unread states with a tooltip on unread.
- Preserve the selected bar color and use a muted bar color for unread items.
- Keep the per-workspace storage migration in `useUnreadTracking`.

Validation
- `make static-check`

Risks
- Low: UI-only change to the sidebar indicator and continued localStorage migration.

---

_Generated with `mux` • Model: `openai:gpt-5.2-codex` • Thinking: `high` • Cost: `$5.01`_

### DIFF
--- a/.mux/skills/pull-requests/SKILL.md
+++ b/.mux/skills/pull-requests/SKILL.md
@@ -60,12 +60,19 @@ gh pr comment <pr_number> --body-file - <<'EOF'
 EOF
 ```
 
-If Codex left review comments and you addressed them:
+### Handling Codex Comments
+
+Use these scripts to check and resolve Codex review comments:
+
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

PR #1893, opened 2026-01-24
🤖 fix: Send button follows agent uiColor

ammar-agent:
## Summary

The Send Message button now uses `focusBorderColor` (derived from `currentAgent.uiColor`) instead of hard-coded mode-based classes, ensuring the button color matches the selected agent.

## Background

The Send button was using mode-based Tailwind classes (`bg-plan-mode`/`bg-exec-mode`) which only covered the two built-in modes. Custom agents with their own `uiColor` were ignored, causing a mismatch between the agent picker pill and the send button.

## Implementation

- Replace class-based background colors with inline `style={{ backgroundColor: focusBorderColor }}`
- Use `hover:brightness-110` for hover effect (works with any color value)
- Falls back to `var(--color-border-light)` until agents load (consistent with textarea border)

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$1.83`_

### DIFF
--- a/src/browser/components/ChatInput/index.tsx
+++ b/src/browser/components/ChatInput/index.tsx
@@ -2495,11 +2495,10 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
                       onClick={() => void handleSend()}
                       disabled={!canSend}
                       aria-label="Send message"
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #1902, opened 2026-01-24
🤖 feat: workspace switcher shows title as primary, extends matching

ammar-agent:
## Summary

The workspace switcher (Ctrl+Shift+P) now presents workspace title as the primary visual element, with name/branch as secondary context.

## Background

Previously the workspace switcher showed "Switch to {name}" as the title, which didn't make use of the workspace title field. This change improves the visual hierarchy to prioritize the semantic title over the technical branch name.

## Implementation

- **Visual hierarchy**: Title (if set) is the primary label; name + projectName shown as subtitle
- **Extended matching**: Keywords prop passed to cmdk Command.Item enables filtering by title, name, projectName, and path
- **Storybook**: Added App.commandPalette.stories.tsx to visualize workspaces with/without titles

### Before/After

**Before:**
```
Title:    • Switch to feat-branch
Subtitle: my-project • streaming
```

**After:**
```
Title:    • Fix login button styling    ← (workspace title, or name if no title)
Subtitle: feat-branch · my-project · streaming
```

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$1.07`_

### DIFF
--- a/src/browser/components/CommandPalette.tsx
+++ b/src/browser/components/CommandPalette.tsx
@@ -438,16 +438,20 @@ export const CommandPalette: React.FC<CommandPaletteProps> = ({ getSlashContext 
         className="font-primary w-[min(720px,92vw)] overflow-hidden rounded-lg border border-[var(--color-command-border)] bg-[var(--color-command-surface)] text-[var(--color-command-foreground)] shadow-[0_10px_40px_rgba(0,0,0,0.4)]"
         onMouseDown={(e: React.MouseEvent) => e.stopPropagation()}
         shouldFilter={shouldUseCmdkFilter}
-        filter={(value, search) => {
+        filter={(value, search, keywords) => {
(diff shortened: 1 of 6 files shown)


ammar-agent:
@codex review

Addressed the keyword matching concern:
- The `keywords` prop is already passed to `Command.Item` containing `[meta.name, meta.projectName, meta.namedWorkspacePath, meta.title]`
- cmdk's default filter uses both `value` and `keywords` for matching
- Added two new tests in `sources.test.ts` to verify keyword-based filtering works correctly

ammar-agent:
@codex review

Fixed the custom filter to include keywords:
- The filter function now receives `(value, search, keywords)` and builds searchable text from both
- Searching by project name, workspace name, or other keywords now works correctly

PR #1905, opened 2026-01-25
🤖 refactor: preserve compaction follow-up metadata

ammar-agent:
Summary:
- capture model/agentId on compaction follow-up requests and rename the stored type
- add a shared helper for agent-skill metadata so compaction and regular sends stay in sync
- add backwards compatibility for legacy `continueMessage` format

Background:
Compaction uses its own agent and model, so follow-up messages must retain the user's original settings. The follow-up payload also had duplicated agent-skill metadata construction.

Implementation:
- split follow-up types into input vs request and populate model/agentId during compaction prep
- use buildAgentSkillMetadata to centralize skill metadata construction across the compaction flow
- add fallback logic in backend and frontend for legacy `continueMessage` format to support upgrade/downgrade

Validation:
- make static-check
- all CI checks passing

Risks:
- Low. Changes are limited to compaction follow-up metadata wiring and shared helper usage.
- Legacy compatibility added for existing stored compaction requests.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$39.98`_

### DIFF
--- a/src/browser/hooks/useCompactAndRetry.ts
+++ b/src/browser/hooks/useCompactAndRetry.ts
@@ -19,6 +19,7 @@ import type { ImagePart, ProvidersConfigMap } from "@/common/orpc/types";
 import {
   buildAgentSkillMetadata,
   type CompactionFollowUpInput,
+  type CompactionFollowUpRequest,
   type DisplayedMessage,
 } from "@/common/types/message";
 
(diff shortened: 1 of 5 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

PR #1906, opened 2026-01-25
perf: increase renderer heap limit to 8GB

ammar-agent:
The renderer was running at ~3.9GB against V8's default ~4GB limit, causing:
- Frequent Mark-Compact (stop-the-world) GC cycles
- Failed scavenges forcing expensive full-heap collections  
- Only ~39% mutator utilization (61% GC overhead)

Increasing to 8GB gives V8 headroom to run fewer, more efficient collections.

---
*Generated with [mux](https://github.com/coder/mux) by @anthropics/claude-sonnet-4*

### DIFF
--- a/src/desktop/main.ts
+++ b/src/desktop/main.ts
@@ -31,6 +31,13 @@ import {
   dialog,
   screen,
 } from "electron";
+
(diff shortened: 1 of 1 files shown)


ammario:
Observing much better performance in the dev mode and am able to profile reliably.

PR #1913, opened 2026-01-26
🤖 fix: hide devcontainer button while loading to prevent flash

ethanndickson:
## Summary

Hide the dev container runtime button while availability data is loading to prevent a visual flash when no `devcontainer.json` exists.

## Background

When opening the runtime menu on a project without `devcontainer.json`, the dev container button appeared for ~0.5s then disappeared once the availability check completed. This was jarring.

The button was shown during loading because `hideDevcontainer` only evaluated to `true` after data loaded with the specific "No devcontainer.json found" reason. Now it defaults to hidden until we have positive confirmation the project has a devcontainer config.

## Implementation

Added `!availabilityMap ||` to hide during loading, then let the existing reason-based logic decide once data arrives.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$1.00`_

### DIFF
--- a/src/browser/stories/App.devcontainer.stories.tsx
+++ b/src/browser/stories/App.devcontainer.stories.tsx
@@ -103,7 +103,8 @@ export const DevcontainerUnavailable: AppStory = {
 
     // Dev container button should be disabled (wait for availability data to load)
     const groupCanvas = within(runtimeGroup);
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

PR #1914, opened 2026-01-26
🤖 refactor: run System1 bash compaction via tool subagent

ThomasK33:
## Summary
Refactors System1 bash-output compaction to run as an internal built-in agent (`system1_bash`) that reports keep-ranges via a typed tool call (`system1_keep_ranges`), removing JSON-in-text parsing/retry logic from `AiService`.

## Implementation
- Added hidden built-in agent `system1_bash` and moved the prompt/rules there.
- Added `system1_keep_ranges` tool + schema (restricted from `exec`).
- Implemented a headless System1 runner that executes the agent and captures keep ranges via the tool.
- Updated `AiService` to use the runner and keep existing heuristic fallback.

## Validation
- `make static-check`

## Risks
- Low. Output compaction is best-effort and already has heuristic fallback; failures return unfiltered output.

---

<details>

# Refactor System1 bash-output filtering into an internal “subagent” + tool

## Context / Why
System1 currently lives as inline logic in `AiService` that calls an LLM (`generateText`/`generateObject`) and then parses JSON (`keep_ranges`) out of model text. This works but has a few pain points:

- Prompt/rules are hard-coded in TS (`buildSystem1BashKeepRangesPrompt`).
- We have to maintain bespoke JSON parsing + retry/correction logic.
- The design doesn’t leverage the existing “agent definition” (markdown) + tool-policy machinery that we already use for subagents (e.g. `explore`, which is enforced to call `agent_report`).

Goal: move the System1 “what lines should we keep?” logic into a dedicated internal agent defined in markdown, and have it *report decisions via a structured tool call* (instead of JSON-in-text). This makes the interface typed/validated, reduces parsing/retry complexity, and centralizes the “rules” in one place.

## Evidence (repo inspection)
  - Wraps `bash`, `bash_output`, and `task_await` tool results and calls `maybeFilterBashOutputWithSystem1`.
  - `maybeFilterBashOutputWithSystem1` currently uses `generateObject`/`generateText`, parses JSON, retries once with a correction prompt, and falls back to heuristics.
- `src/node/services/system1/bashOutputFiltering.ts`
  - Contains the System1 prompt/rules inline in `buildSystem1BashKeepRangesPrompt`.
  - Contains `parseSystem1KeepRanges`, heuristic keep-range selection, and range application utilities.
- Subagent enforcement and tool policy
  - `src/node/builtinAgents/explore.md` shows the desired “subagent must call agent_report” pattern.
  - `src/node/services/agentDefinitions/resolveToolPolicy.ts` shows runtime enforcement: subagents always get `agent_report` enabled.
  - `src/node/services/taskService.ts` enforces `agent_report` with reminder + tool-lock (require).

---

## Recommended approach (A): “Headless subagent” runner + `system1_keep_ranges` tool (no child workspace)
**Net LoC estimate (product code only): ~+200 LoC**

This approach gets the main benefits (markdown-defined agent + structured decisions via tool calls) *without* paying the cost of spawning full TaskService child workspaces for every compaction.

### A1) Add a built-in agent definition for System1
- Add `src/node/builtinAgents/system1_bash.md` (similar to `compact.md`: internal + hidden).
  - `ui.hidden: true`
  - `subagent.runnable: false` (since we’ll run it internally, not as a user-spawned task)
  - `tools.add` should include **only** the custom decision tool (deny-all baseline + explicit allow).
  - Body: the System1 rules currently embedded in `buildSystem1BashKeepRangesPrompt`, rewritten as instructions to call the tool.
    - Explicitly instruct: call `system1_keep_ranges` exactly once with line ranges.
- Register it as a built-in agent:
  - Add it to `src/node/services/agentDefinitions/builtInAgentDefinitions.ts`
  - Re-run `scripts/generate-builtin-agents.sh` (updates `builtInAgentContent.generated.ts`)

### A2) Implement a dedicated custom tool for decisions
- Add a new tool definition (Zod) in `src/common/utils/tools/toolDefinitions.ts`:
  - `system1_keep_ranges` args schema (preferred: ranges, not per-line booleans):
    - `keep_ranges: Array<{ start: number; end: number; reason?: string }>`
  - Result schema can be a trivial `{ success: true }`.
- Implement the tool in `src/node/services/tools/system1_keep_ranges.ts`.
  - Keep it side-effect free; its job is just to accept validated args.
  - The *caller* (the headless runner) will inject a closure so it can capture the keep ranges when the tool executes.
- Register the tool in `src/common/utils/tools/tools.ts` so it can be used by the AI SDK consistently.
- Prevent accidental use by normal agents:
  - Add `system1_keep_ranges` to `src/node/builtinAgents/exec.md` `tools.remove` list (since exec enables `.*`).
  - Re-run `scripts/generate-builtin-agents.sh`.
  - (Optional) If we’d rather keep it *purely internal*, skip global tool registration and instead pass a locally constructed tool into the runner’s `generateText` call.

### A3) Build a small “headless agent runner” for System1
Create a new helper/service (e.g. `src/node/services/system1/system1AgentRunner.ts`) that:

1. Loads the `system1_bash` agent definition via existing agent-definition helpers.
   - Note: agent IDs come from filenames; naming the file `system1_bash.md` means the agentId is `system1_bash`.
2. Builds a system prompt using the same `buildSystemMessage`/agent-body resolution used for normal streams.
3. Calls `generateText` (non-streaming) with:
   - The resolved System1 model (reuse existing `system1Model`/`system1ThinkingLevel` settings).
   - Tools restricted to `{ system1_keep_ranges }`.
   - An input user message containing:
     - bash script
     - numbered output (reuse `formatNumberedLinesForSystem1`)
4. Captures the tool call args (keep ranges) and returns them to the caller.
5. Defensive behavior:
   - If the model does not call the tool within the timeout, treat as failure and fall back to heuristic keep-ranges.
   - Clamp/normalize line numbers exactly as `applySystem1KeepRangesToOutput` already does.

> Note: we can get “must call the tool” semantics similar to `agent_report` enforcement by using `toolChoice: { type: "tool", toolName: "system1_keep_ranges" }` + a post-run assertion that the tool executed. If we force tool choice for Anthropic models, we may need to enforce `thinkingLevel: "off"` (Anthropic forbids extended thinking with forced tool use).

### A4) Swap `maybeFilterBashOutputWithSystem1` over to the runner
In `src/node/services/aiService.ts`:
- Replace the current `generateObject`/`generateText` + JSON parsing + retry/correction loop with:
  - `keepRanges = await runSystem1KeepRanges({...})`
  - `applied = applySystem1KeepRangesToOutput({ rawOutput, keepRanges, maxKeptLines })`
- Keep the existing:
  - threshold triggers (`minLines`, `minTotalBytes`)
  - temp-file persistence of full output
  - heuristic fallback toggle
  - UI “filtering” phase event (`bash-output`)

### A5) Update / simplify `bashOutputFiltering.ts`
- Move the “rules” text out of `buildSystem1BashKeepRangesPrompt` into `system1_bash.md`.
- Keep (or reuse) the pure functions:
  - `splitBashOutputLines`
  - `formatNumberedLinesForSystem1`
  - `applySystem1KeepRangesToOutput`
  - heuristic keep-range selection
- Delete `parseSystem1KeepRanges` and JSON-extraction logic if it’s no longer used.

### A6) Tests
- Unit tests for the new runner (mock model):
  - “tool called → keep ranges applied”
  - “tool not called / timeout → heuristic fallback”
  - “invalid ranges → normalization/clamping works”
- Regression test that `bash`, `bash_output`, and `task_await` wrappers still filter when limits trigger.

---

<details>
<summary>Alternative approach (B): True TaskService subagent workspace + tool + agent_report</summary>

**Net LoC estimate (product code only): ~+600 LoC**

This matches the mental model of “System1 is a subagent” most literally, and reuses TaskService’s enforced `agent_report` flow. However, it risks high overhead (workspace creation, config persistence, queuing due to `maxParallelAgentTasks`) and UI clutter unless we add significant extra plumbing.

### B1) Add `system1_bash.md` as a runnable subagent
- `subagent.runnable: true`
- `subagent.skip_init_hook: true`
- `ui.hidden: true`
- `tools.add`: only `system1_keep_ranges`
- Prompt: require calling `system1_keep_ranges` then `agent_report`.
- Register it as a built-in agent + regenerate built-in agent content (same flow as approach A).

### B2) Implement `system1_keep_ranges` tool with side-channel storage
- Tool `execute` stores keep ranges keyed by `workspaceId` or `taskId` in an in-memory map.
- When the parent awaits the subagent, it reads the stored keep ranges.

### B3) Internal orchestration from `AiService`
- On compaction trigger, spawn a task via `TaskService.create` (NOT via model tool call):
  - prompt includes script + numbered output
- Await `agent_report` (TaskService already enforces it).
- Read keep-ranges from the side-channel map and apply.

### B4) Required mitigations (to make this viable)
- Add a concept of **ephemeral internal agent tasks** so System1 doesn’t:
  - persist to `~/.mux/config.json`
  - create git worktrees
  - count against `maxParallelAgentTasks` (avoids queue deadlocks)
  - show up in the sidebar

This would likely require a new “in-memory task/session” execution path rather than reusing TaskService as-is.

</details>

</details>

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$14.31`_

### DIFF
--- a/src/node/services/tools/fileCommon.ts
+++ b/src/node/services/tools/fileCommon.ts
@@ -216,12 +216,17 @@ export function validatePathInCwd(
     assert(path.isAbsolute(dir), `extraAllowedDir must be an absolute path: '${dir}'`);
   }
 
-  // Resolve the path (handles relative paths and normalizes)
-  const resolvedPath = path.isAbsolute(filePath)
-    ? path.resolve(filePath)
-    : path.resolve(cwd, filePath);
+  const filePathIsAbsolute = path.isAbsolute(filePath);
 
-  const allowedRoots = [cwd, ...trimmedExtraAllowedDirs].map((dir) => path.resolve(dir));
+  // Only allow extraAllowedDirs when the caller provides an absolute path.
+  // This prevents relative-path escapes (e.g., ../...) from bypassing cwd restrictions.
+
+  // Resolve the path (handles relative paths and normalizes)
+  const resolvedPath = filePathIsAbsolute ? path.resolve(filePath) : path.resolve(cwd, filePath);
+
+  const allowedRoots = [cwd, ...(filePathIsAbsolute ? trimmedExtraAllowedDirs : [])].map((dir) =>
+    path.resolve(dir)
+  );
 
   // Check if resolved path is within any allowed root.
   // Use path.relative to check if we need to go "up" from the root to reach the file.


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

PR #1926, opened 2026-01-26
🤖 fix: make System1 bash filtering intent-aware

ThomasK33:
## Summary

Make System1 bash-output compaction more intent-aware by threading `bash.display_name` into the System1 request and applying deterministic skip/boost rules for exploratory commands.

## Background

Our System1 `keep_ranges` call only saw the script + numbered output. For exploratory commands (`ls`, `find`, `rg`, etc.) this could cause over-aggressive filtering because the model had no hint that the user was browsing/searching rather than debugging a failure.

## Implementation

- Plumbed `bash.display_name` through the bash tool wrapper into the System1 keep-ranges prompt.
- Updated the internal `system1_bash` prompt to use `Display name` + script as intent hints.
- Added an intent-aware policy that:
  - skips compaction for output already narrowed by the script (e.g. `head`, `tail`, `sed -n` ranges)
  - skips compaction for small exploration output
  - boosts `maxKeptLines` for large exploration output when the user is still on the default budget

## Validation

- `bun test src/node/services/system1/bashCompactionPolicy.test.ts src/node/services/system1/system1AgentRunner.test.ts`
- `make static-check`

## Risks

- Low/medium: heuristic detection could misclassify some scripts as “already targeted” and skip filtering, increasing noise. This is mitigated by the existing bash tool output limits and debug logging to diagnose behavior.

---

<details>

# Plan: Intent-aware System1 bash output filtering

## Context / Why
Mux uses an internal “System1” LLM call to compact large `bash`/`bash_output`/`task_await` outputs before they’re shown to the main workspace agent. Today System1 only receives the bash script + numbered output, so it can miss the user/agent’s intent (e.g. “I’m listing files to explore”), leading to over-aggressive filtering.

Goal: improve relevance and reduce “over-filtering” by:
1) forwarding `bash.display_name` (intent) into the System1 request,
2) updating the System1 prompt to use that intent, and
3) adding an intent-aware compaction policy (skip or keep more lines for exploration/listing/search commands).

Non-goals:
- Change global conversation compaction/history behavior.
- Increase bash tool hard limits (`BASH_HARD_MAX_LINES`, `BASH_MAX_TOTAL_BYTES`).

## Evidence (current behavior)
- System1 request content is currently **minimal** and contains only:
  - `maxKeptLines`, `Bash script`, `Numbered output` (`src/node/services/system1/system1AgentRunner.ts`, `userMessage` construction).
- System1 prompt is defined in `src/node/builtinAgents/system1_bash.md` and currently prioritizes errors/stack traces; it does **not** mention any intent/context field.
- System1 compaction is triggered/wired in `src/node/services/aiService.ts` via `maybeFilterBashOutputWithSystem1()` and wraps:
  - `bash` (script available)
  - `bash_output` (script currently passed as `""`)
  - `task_await` for `bash:<id>` entries (script currently passed as `""`)
- `bash.display_name` is already **required** by the tool schema (`src/common/utils/tools/toolDefinitions.ts`) and available inside the bash tool execution args.
- Default compaction budget: `bashOutputCompactionMaxKeptLines = 40` (`src/common/types/tasks.ts`).

---

### Product changes
1. **Plumb display name from the bash tool wrapper → System1 runner**
   - In `src/node/services/aiService.ts` (within the wrapped `bash` tool `execute`):
     - Extract `display_name` from `args` alongside `script`.
     - Extend `maybeFilterBashOutputWithSystem1()` params to accept `displayName?: string`.
     - Pass `displayName` through when `toolName === "bash"`.
   - Keep `bash_output` / `task_await` as-is initially (no display_name available without extra lookups).

2. **Extend the System1 runner request message**
   - In `src/node/services/system1/system1AgentRunner.ts`:
     - Extend `RunSystem1KeepRangesParams` with `displayName?: string`.
     - Add an optional section to `userMessage`, e.g.
       - `Display name:\n${params.displayName}` (only if non-empty)
     - Keep the “minimal prompt” philosophy: one extra short field.

### Tests
- Update/add unit coverage in `src/node/services/system1/system1AgentRunner.test.ts`:
  - Add a test that passes `displayName: "List files"` and asserts the generated `messages[0].content` includes `Display name:`.

<details>
<summary>Notes / trade-offs</summary>

- This avoids adding any extra LLM calls. It uses an existing required field.
- We treat `display_name` as an intent hint (not authoritative), similar to how it’s treated elsewhere as “untrusted input”.

</details>

---

### Product changes
- Update `src/node/builtinAgents/system1_bash.md`:
  - Document the new optional `Display name` field.
  - Add rules to bias selection by intent:
    - If intent indicates **exploration/listing/search**, prefer keeping representative file paths, matches, and summaries (not just errors).
    - If intent indicates **build/test/logs**, keep current error-first behavior.
  - If the bash script itself already limits output to a specific slice (e.g. `head`, `tail`, `sed -n`), avoid further denoising; prefer keeping all lines / large contiguous ranges (subject to `maxKeptLines`).

### Tests
- No strict tests required for prompt text, but we should keep the existing runner tests green.

---

### Goals
- Avoid “throwing away” useful exploratory output (e.g. `ls`, `find`, `rg`) when the output is already bounded by the bash tool’s limits.
- Avoid filtering outputs that are already intentionally narrowed (e.g. `head`/`tail`/line ranges), even if they exceed the current min-lines trigger.
- Still compact noisy outputs (build logs, server logs) to protect context.

### Product changes
1. **Detect “already targeted” commands and skip compaction**
   - Add helper(s) in `src/node/services/system1/bashCompactionPolicy.ts`:
     - `isBashOutputAlreadyTargeted(script: string): boolean`
   - Heuristics (deterministic; expand later as needed):
     - pipeline or command uses `head` / `tail` (e.g. `| head`, `| tail`, `head -n`, `tail -n`)
     - `sed -n 'start,endp'` / `sed -n 'start,end p'`
     - `awk` line-range filters (initially: detect `awk` + `NR`)
   - In `maybeFilterBashOutputWithSystem1()` (when `toolName === "bash"` and `script` is non-empty):
     - if compaction would trigger by size but `isBashOutputAlreadyTargeted(script)` is true, **return undefined** (no filtering) and log debug reason.
     - do this before saving `fullOutputPath` and before calling System1.

2. **Add a small, testable classifier for bash intent**
   - Keep the classifier focused on “exploration vs logs” once targeted commands are handled:
     - `classifyBashIntent({ script, displayName }): "exploration" | "logs" | "unknown"`
   - Heuristics (keep simple + deterministic):
     - Exploration commands by script prefix (first command token): `ls`, `find`, `fd`, `tree`, `rg`, `grep`, `git ls-files`, `git status`.
     - Exploration by displayName keywords: `list`, `explore`, `search`, `scan`.
     - Otherwise `logs`/`unknown`.

3. **Use the intent to adjust compaction behavior in `maybeFilterBashOutputWithSystem1()`**
   - Apply only when `toolName === "bash"` (since `bash_output`/`task_await` typically represent logs and don’t currently carry script/intent).
   - Proposed behavior (after the targeted-command early return):
     - If intent === `exploration`:
       - **Skip System1 entirely** when output is “small enough” (example guardrail: `lines.length <= 120` AND `bytes <= 12 * 1024`).
       - Otherwise run System1 but with a **boosted** `maxKeptLines`:
         - `effectiveMaxKeptLines = min(BASH_HARD_MAX_LINES, max(userMaxKeptLines, 120))`
         - Guardrail: if the user has customized `bashOutputCompactionMaxKeptLines` away from the default (40), do **not** override it; only boost when it’s still default.
     - Else (logs/unknown): keep existing behavior and budgets.

4. **Logging/diagnostics**
   - Expand existing `log.debug("[system1] Bash output compaction triggered", …)` payload to include:
     - `intent`, whether the command was detected as targeted, `effectiveMaxKeptLines`, and whether filtering was skipped.
   - Keep this debug-only.

### Tests
- Add unit tests for `isBashOutputAlreadyTargeted()` (fast, deterministic).
- Add unit tests for the classifier module (fast, deterministic).
- Add a small unit test for the policy decision helper (if separated) to ensure:
  - Targeted command → skip.
  - Exploration + small output → skip.
  - Exploration + large output → boosted max.
  - Logs → default max.

<details>
<summary>Why skip targeted commands, skip small outputs, and boost exploration?</summary>

- Targeted commands (e.g. `head`, `tail`, `sed -n`) already encode the model’s intent to narrow output; further filtering is more likely to drop needed context than reduce noise.
- “Skip” avoids an extra LLM call (latency/cost) when the output is already manageable and likely fully relevant.
- “Boost” preserves more relevant lines when output is large enough that some compaction is still needed.

</details>

---

## Rollout / Acceptance criteria
- System1 request includes intent when available (`bash.display_name`).
- System1 prompt references the intent field and produces better ranges for non-error outputs.
- Exploration commands (e.g. `ls`, `find`, `rg`) no longer get aggressively collapsed to ~40 lines when outputs are modest.
- Commands that already narrow output (e.g. `head`, `tail`, `sed -n` ranges) are not further compacted by System1 (output is preserved as-is, subject only to bash tool limits).
- Existing behavior for logs/build/test output remains unchanged.

## Implementation order (recommended)
1) Approach 1 (plumb `display_name`) → 2) Approach 2 (prompt tweak) → 3) Approach 3 (policy + tests)

</details>

---
_Generated with `mux` • Model: openai:gpt-5.2 • Thinking: xhigh • Cost: $5.47_

### DIFF
--- a/src/node/services/system1/bashCompactionPolicy.test.ts
+++ b/src/node/services/system1/bashCompactionPolicy.test.ts
@@ -9,6 +9,7 @@ import {
 describe("bashCompactionPolicy", () => {
   describe("isBashOutputAlreadyTargeted", () => {
     it("detects common output-slicing commands", () => {
+      expect(isBashOutputAlreadyTargeted("sudo head -n 1 some.log")).toBe(true);
       expect(isBashOutputAlreadyTargeted("rg foo . | head -n 50")).toBe(true);
       expect(isBashOutputAlreadyTargeted("tail -n 100 some.log")).toBe(true);
       expect(isBashOutputAlreadyTargeted("sed -n '1,200p' file.txt")).toBe(true);
@@ -18,6 +19,7 @@ describe("bashCompactionPolicy", () => {
     it("returns false for non-targeted scripts", () => {
       expect(isBashOutputAlreadyTargeted("ls -la")).toBe(false);
       expect(isBashOutputAlreadyTargeted("rg foo .")).toBe(false);
+      expect(isBashOutputAlreadyTargeted("git rev-parse HEAD")).toBe(false);
     });
   });
 
--- a/src/node/services/system1/bashCompactionPolicy.ts
+++ b/src/node/services/system1/bashCompactionPolicy.ts
@@ -14,8 +14,38 @@ export function isBashOutputAlreadyTargeted(script: string): boolean {
 
   // If the script already limits output to a slice (head/tail/line ranges), further denoising is
   // likely to drop exactly what the caller asked to see.
-  if (/(^|[|;&]\s*|\s)(head|tail)\b/i.test(trimmed)) {
-    return true;
+  //
+  // NOTE: Avoid false positives like `git rev-parse HEAD`.
+  const statementSegments = trimmed
+    .split(/(?:\r?\n|&&|;)+/)
+    .map((part) => part.trim())
+    .filter(Boolean);
+
+  const slicingCommands = new Set(["head", "tail"]);
+  for (const statement of statementSegments) {
+    const pipeSegments = statement
+      .split("|")
+      .map((part) => part.trim())
+      .filter(Boolean);
+
+    for (const pipeSegment of pipeSegments) {
+      const tokens = pipeSegment.split(/\s+/).filter(Boolean);
+      if (tokens.length === 0) {
+        continue;
+      }
+
+      const cmd0 = (tokens[0] ?? "").toLowerCase();
+      const cmd1 = (tokens[1] ?? "").toLowerCase();
+
+      if (slicingCommands.has(cmd0)) {
+        return true;
+      }
+
+      // Common wrapper: `sudo head ...`.
+      if ((cmd0 === "sudo" || cmd0 === "command") && slicingCommands.has(cmd1)) {
+        return true;
+      }
+    }
   }
 
   if (/\bsed\b[^\n]*\s-n\s+['"]?\d+\s*,\s*\d+\s*p['"]?/i.test(trimmed)) {


ThomasK33:
@codex review

Addressed the thread feedback by only skipping/boosting exploration compaction when compaction settings are at their defaults; otherwise we respect user limits (maxKeptLines/minLines/minTotalBytes). Added tests covering these cases.

PR #1927, opened 2026-01-26
🤖 fix: navigate to workspace immediately after creation

ethanndickson:
## Summary

Fixes a regression where the "Creating workspace" splash blocked until `sendMessage` completed. For devcontainer/SSH runtimes, this could take minutes since `sendMessage` waits for init (container build) to finish.

## Background

**Regressed in #1896** ("feat: support PDF attachments") which changed `void api.workspace.sendMessage(...)` (fire-and-forget) to `await api.workspace.sendMessage(...)` and moved `onWorkspaceCreated` after it.

This undid the fix from **#1146** ("perf: exit splash screen immediately after workspace.create()").

## Implementation

- Move `onWorkspaceCreated` and `setIsSending(false)` to immediately after `workspace.create()` succeeds
- Change `sendMessage` back to fire-and-forget with `void`
- Clear draft input/attachments only after `sendMessage` succeeds (via `.then()`), preserving them if the send fails

## Validation

- Added regression test that verifies `onWorkspaceCreated` is called before `sendMessage` resolves (using a never-resolving promise mock)
- Existing tests pass
- `make static-check` passes

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$7.42`_

### DIFF
--- a/src/browser/components/ChatInput/index.tsx
+++ b/src/browser/components/ChatInput/index.tsx
@@ -367,21 +367,20 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
     },
     [setToast]
   );
+  // Subscribe to pending send errors from creation flow. Uses listener: true so
+  // late failures (e.g., slow devcontainer startup) still surface a toast.
+  const pendingErrorKey =
+    variant === "workspace" && workspaceId ? getPendingWorkspaceSendErrorKey(workspaceId) : null;
+  const [pendingError, setPendingError] = usePersistedState<SendMessageError | null>(
+    pendingErrorKey ?? "__unused__",
+    null,
+    { listener: true }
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

Addressed the draft-preservation issue by awaiting the initial send result while keeping early navigation, added a mounted guard for draft clears, and expanded tests for send failure + ordering. Static-check is green.

ethanndickson:
@codex review

Surfaced creation send failures in workspace view and updated tests.

ethanndickson:
@codex review

Replaced one-shot useEffect with usePersistedState subscription so late send errors (e.g., during slow devcontainer startup) still surface a toast.

PR #1929, opened 2026-01-26
🤖 perf: speed up tool_post tailwind class ordering (13x faster)

ethanndickson:
## Summary

Speed up the `.mux/tool_post` tailwind class ordering fix from ~7s to ~0.5s per file (13x faster).

## Background

The `tool_post` hook runs on every TS/TSX file edit to auto-format and fix tailwind class ordering. It was running the full project eslint config (typescript-eslint, react, react-hooks, custom rules, etc.) when only the `tailwindcss/classnames-order` rule was needed.

## Implementation

- Create a minimal eslint config (`.mux/eslint-tailwind-only.mjs`) with only the tailwind plugin and classnames-order rule
- Use `--no-config-lookup` flag to prevent eslint from loading the project's full config
- Update `tool_post` to use the new minimal config

## Validation

Benchmarked both approaches on `src/browser/App.tsx`:
- **Old (full config):** ~6.8s
- **New (minimal config):** ~0.5s

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$0.36`_

### DIFF
--- a/.mux/eslint-tailwind-only.mjs
+++ b/.mux/eslint-tailwind-only.mjs
@@ -1,0 +1,26 @@
+import tailwindcss from "eslint-plugin-tailwindcss";
+import tseslint from "typescript-eslint";
+
+export default [
+  {
+    files: ["**/*.{ts,tsx}"],
+    languageOptions: {
+      parser: tseslint.parser,
+      parserOptions: {
+        ecmaFeatures: { jsx: true },
+      },
+    },
+    plugins: { tailwindcss },
+    settings: {
+      tailwindcss: {
+        // Don't try to load Tailwind config (v4 doesn't export resolveConfig)
+        config: false,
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

Fixed the TypeScript/JSX parser issue by:
1. Adding `typescript-eslint` parser with JSX support
2. Adding `--no-inline-config` flag to ignore inline disable comments that reference missing rules
3. Adding `files: ["**/*.{ts,tsx}"]` pattern to match the project files

Validated the fix works by creating a test file with incorrect class order and confirming it gets reordered.

ethanndickson:
@codex review

Added `settings.tailwindcss` block to suppress the "Cannot resolve default tailwindcss config path" warning, matching the main eslint config. The fix was already working but now runs cleanly without warnings.

PR #1930, opened 2026-01-26
🤖 fix: FileBreakdown path truncation with tooltip and left-side ellipsis

ammar-agent:
## Summary

Improves UX for the FileBreakdown component when file paths are truncated due to limited space.

## Background

When long file paths are displayed in the FileBreakdown component, they get truncated. Previously:
- Truncation happened from the right, hiding the filename (the most useful part)
- No way to see the full path

## Implementation

- **Left-side ellipsis**: Use `direction-rtl` with `text-left` to truncate from the left side, preserving the filename which is typically what users care about most
- **Tooltip on hover**: Replace native `title` attribute with the styled `Tooltip` component to show the full path

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$0.44`_


ammar-agent:
@codex review

PR #1940, opened 2026-01-27
🤖 fix: preprocess empty arrays to avoid never[] type inference in PTC

ethanndickson:
## Summary

Fix PTC type validation failures when agents use empty array patterns like `const results = []; results.push(...)`. TypeScript infers `[]` as `never[]` with our compiler settings, causing spurious type errors on all subsequent array operations.

## Background

TypeScript infers `[]` as `never[]` when `strictNullChecks: true` and `noImplicitAny: false` (our settings for PTC). This is documented behavior across multiple GitHub issues (#36987, #13140, #50505, #51979), all marked "Working as Intended". The TypeScript team's recommended workaround is to use type assertions: `[] as any[]`.

The previous diagnostic-filter approach only handled `push`/`unshift` calls. Operations like `results.map(...)`, `results.filter(...)`, index access, and spreads would still fail.

## Implementation

Added preprocessing that transforms `[]` → `[] as any[]` before TypeScript validation. This follows the official TS team recommendation and fixes **all** array operations uniformly.

**Contextual typing preservation:** Only rewrites empty arrays that truly infer to `never[]` (via the checker), preserving contextual typing in typed contexts like `const nums: number[] = condition ? [] : [1]`.

**Skipped contexts:**
- `[] as Type` / `[] as const` (as expressions)
- `<Type[]>[]` (angle-bracket assertions)
- `const x: Type[] = []` (type annotations)
- `const [] = x` (destructuring patterns)
- `([] = foo)` (destructuring assignments - would produce invalid syntax)
- `function f(x: Type[] = [])` (typed parameter defaults)

**Special handling for `satisfies`:** `[] satisfies Type[]` still gets rewritten because `satisfies` only validates compatibility without changing the inferred type.

**Unary operator handling:** Parenthesizes unary operand contexts (`typeof []`, `void []`, `+[]`, etc.) so `as` binds to the array literal, not the unary expression.

## Performance Optimizations

- **Cached lib SourceFiles:** ES2023 lib files are cached across validations to avoid re-parsing
- **Separate mux.d.ts file:** muxTypes moved to a separate virtual file for caching
- **Early-return optimization:** Skip preprocessing parse when no never[] arrays found
- **Shared SourceFile reuse:** Type validator's SourceFile is reused for unavailable-global detection

## Validation

- All 60 PTC type validator tests pass (added tests for typeof/void/+[] unary operators)
- All 34 static analysis tests pass
- `make typecheck` passes

## Risks

Low risk - this only affects the type-checking preprocessing layer, not runtime execution. The transformation is syntactically safe and follows TypeScript's official recommendation.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$12.14`_

### DIFF
--- a/src/node/services/ptc/typeValidator.test.ts
+++ b/src/node/services/ptc/typeValidator.test.ts
@@ -718,6 +718,44 @@ mux.file_read({ path: "wrong" });`,
   // Destructuring patterns (valid JS that must not break)
   // ==========================================================================
 
+  test("handles empty array destructuring in for-of LHS", () => {
+    // for-of allows destructuring patterns directly in the loop header.
+    const result = validateTypes(
+      `
+      const items = [[1], [2]];
+      let count = 0;
+      for ([] of items) {
+        count += 1;
+      }
+      return count;
+    `,
+      muxTypes
+    );
+    expect(result.valid).toBe(true);
+  });
+
+  test("handles empty array destructuring in for-in LHS", () => {
+    // for-in does not allow destructuring patterns in TypeScript, but the error
+    // should remain about the pattern (not a rewritten `as any[]` assertion).
+    const result = validateTypes(
+      `
+      const obj = { a: 1, b: 2 };
+      let count = 0;
+      for ([] in obj) {
+        count += 1;
+      }
+      return count;
+    `,
+      muxTypes
+    );
+    expect(result.valid).toBe(false);
+    expect(
+      result.errors.some((error) => error.message.includes("cannot be a destructuring pattern"))
+    ).toBe(true);
+    expect(
+      result.errors.some((error) => error.message.includes('must be of type "string" or "any"'))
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

Fixed: empty array literals with immediate member/index access are now wrapped in parentheses:
- `[].map(...)` → `([] as any[]).map(...)`
- `[][0]` → `([] as any[])[0]`
- `[]?.length` → `([] as any[])?.length`

Added 4 new tests for these patterns.

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed the loop-LHS empty array assertion guard and added regression tests. Static-check is green.

PR #1949, opened 2026-01-27
🤖 fix: dedupe Start Here plan context

ThomasK33:
## Summary
Prevents duplicate plan context after using **Propose Plan → Start Here** (“Replace all chat history with this plan”). When the chat already contains the full plan, we now avoid re-injecting it during the plan→exec handoff and avoid prompting the model to re-read the plan file.

## Background
The Start Here flow replaces chat history with a single assistant message containing the plan. On the next exec turn, we were:
- injecting the plan again via the agent-transition message, and
- adding a plan-file hint that often led to a redundant “read …KB” step.

## Implementation
- Added a helper to detect the Start Here plan-summary message.
- In `AIService`, when that summary is present:
  - skip plan-file hint injection
  - skip plan-content injection for plan→exec transitions
- Added a short note to the Start Here summary content to discourage redundant plan-file reads.

## Validation
- `bun test src/common/utils/messages/startHerePlanSummary.test.ts src/browser/components/tools/ProposePlanToolCall.test.tsx src/common/utils/ui/modeUtils.test.ts`
- `make static-check`

## Risks
Low. The skip logic only triggers when the history matches the Start Here summary shape (metadata + id/marker), and otherwise preserves existing behavior.

---

<details>

# Avoid double-including plan after “Start Here / Replace all chat with this plan”

## Context / Why
When using the **Propose Plan** tool UI with `taskSettings.proposePlanImplementReplacesChatHistory` enabled, clicking **Start Here** (“Replace all chat history with this plan”) replaces the entire chat history with a single assistant message that already contains the full plan.

When the user then switches to Exec and clicks **Implement** (or otherwise asks the Exec agent to proceed), the system currently re-introduces the plan again (via plan-file hints and/or plan→exec transition injection), which can:

- **Duplicate the plan** in the model context (wasting tokens), and
- Show an extra “read …KB” plan-file read step, even though the plan is already visible in-chat.

**Goal:** In this specific “Start Here” flow, ensure the plan appears **exactly once** in model context by preferring the already-in-chat plan, while keeping existing behavior for other scenarios (normal plan→exec, compaction resets, resuming a workspace later).

## Evidence
- `src/browser/components/tools/ProposePlanToolCall.tsx` constructs `startHereContent` and (when enabled) calls `api.workspace.replaceChatHistory()` with a single summary message containing the plan and metadata `{ compacted: "user", agentId: "plan" }`, then switches agent to exec and sends `"Implement the plan"`. (Explore reports: 165ecc7334, 589562a788)
- Plan-file hinting is implemented in `src/common/utils/ui/modeUtils.ts#getPlanFileHint()` and currently instructs non-plan agents that they “MUST read” the plan file if it exists, regardless of whether the plan is already present in chat. (Explore report: 4d695ad2e2)
- Plan→exec transitions can inject plan content again via `src/node/services/aiService.ts` + `src/browser/utils/messages/modelMessageTransform.ts#injectAgentTransition()` (wraps content in `<plan>…</plan>`). This creates duplication when chat was already replaced with the plan. (Explore report: 4d695ad2e2)

---

### 1) Detect “Start Here plan summary” messages (no schema change)
Add a small helper (preferably in a shared location used by the node-side request builder) to detect when the chat already contains the plan via the Start Here flow.

Proposed predicate (defensive + explicit):
- `msg.role === "assistant"`
- `msg.metadata?.compacted === "user"`
- `msg.metadata?.agentId === "plan"`
- Optional extra guard (if needed for uniqueness): message id prefix `start-here-` **or** text contains the preserved-path footer.

Expose:
- `isStartHerePlanSummaryMessage(msg): boolean`
- `hasStartHerePlanSummary(messages): boolean` (usually checks the last assistant message)

### 2) Skip plan injection on plan→exec transition when plan is already in chat
In `src/node/services/aiService.ts` (where `planContentForTransition` is computed for plan-like → exec-like transitions):

- If `hasStartHerePlanSummary(filteredMessages)` is `true`, **do not set** `planContentForTransition` (or skip calling `injectAgentTransition` entirely).
- Rationale: the Exec agent will already receive the plan through the replaced chat history; injecting `<plan>…</plan>` again is pure token bloat.

Add a `log.debug` line when skipping injection so the behavior is diagnosable.

### 3) Prevent the Exec agent from re-reading the plan file in this case
Update the plan-file hint (and/or its callsite) so it does not tell the agent to read the plan file when the plan is already present in chat.

Two options (prefer the more deterministic if the callsite has access to messages):

- **Best:** In the system-prompt builder (node side), if `hasStartHerePlanSummary(messages)` is true, **omit** the plan-file hint entirely.
- **Fallback:** Update `getPlanFileHint()` wording to be conditional:
  - “If the plan is already included in the chat history (e.g., after ‘Replace all chat history with this plan’), do **not** re-read the plan file. Otherwise (plan not present in chat), you **must** read it before proceeding.”

### 4) Add a one-line reinforcement in the Start Here message
Append a short note to `startHereContent` in `ProposePlanToolCall.tsx` that explicitly says the plan is already in this chat (so the Exec agent should not re-read the plan file), e.g.:

> “Note: This chat already contains the full plan; no need to re-open the plan file.”

Keep it to a single line to avoid increasing context.

### 5) Tests (not included in net LoC estimate)
- Unit test the new detection helper (`isStartHerePlanSummaryMessage` / `hasStartHerePlanSummary`).
- Add a test for plan→exec transition request-building:
  - Given a message list ending in a Start Here plan summary, verify **no** `<plan>…</plan>` injection occurs.
  - Verify the resulting system prompt does **not** include a “MUST read plan file” instruction in this case.
- Update `src/browser/components/tools/ProposePlanToolCall.test.tsx` expectations for the new Start Here text.

### Acceptance criteria
- The Start Here summary includes a one-line note that the plan is already in chat (and does not need to be re-read from the plan file).
- After **Start Here → Implement**, the Exec request contains the plan **once** (from the chat summary), with no additional injected plan block.
- The Exec agent is not instructed to (and typically does not) perform a redundant plan-file read.
- Other cases are unchanged:
  - Normal plan→exec (without Start Here) still provides plan context as before.
  - Reopening a workspace later in exec mode still provides a plan-file hint when the plan isn’t in chat.
  - Compaction/reset scenarios still instruct reading when the plan is missing from chat.

---

## Alternative approaches

<details>
<summary>Approach A: Prompt-only tweak (~5–15 net LoC)</summary>

Only change the plan-file hint (and/or the default Implement message text) to say “don’t re-read if it’s already in chat”.

- **Pros:** Very small change.
- **Cons:** Not deterministic if the backend still injects plan content on plan→exec transitions; also relies on model compliance.

</details>

<details>
<summary>Approach C: Add explicit metadata flag + hash (~80–150 net LoC)</summary>

Extend message metadata (schema + serialization) to include something like `metadata.planSnapshot = { source: "start-here", sha256: "…" }`.

- Use the hash to skip injection only when the on-disk plan matches the snapshot, or to surface a warning when they differ.
- **Pros:** Most robust and debuggable.
- **Cons:** More invasive (schema changes), more code/tests.

</details>

</details>

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: $1.94_

### DIFF
--- a/src/browser/components/tools/ProposePlanToolCall.test.tsx
+++ b/src/browser/components/tools/ProposePlanToolCall.test.tsx
@@ -120,6 +120,30 @@ describe("ProposePlanToolCall", () => {
     globalThis.document = originalDocument;
   });
 
+  test("does not claim plan is in chat when Start Here content is a placeholder", () => {
+    const planPath = "~/.mux/plans/demo/ws-123.md";
+
+    render(
+      <TooltipProvider>
+        <ProposePlanToolCall
+          args={{}}
+          result={{
+            success: true,
+            planPath,
+          }}
+          workspaceId="ws-123"
+          isLatest={false}
+        />
+      </TooltipProvider>
+    );
+
+    expect(startHereCalls.length).toBe(1);
+    expect(startHereCalls[0]?.content).toContain("*Plan saved to");
+    expect(startHereCalls[0]?.content).not.toContain(
+      "Note: This chat already contains the full plan"
+    );
+    expect(startHereCalls[0]?.content).toContain("Read the plan file below");
+  });
   test("keeps plan file on disk and includes plan path note in Start Here content", () => {
     const planPath = "~/.mux/plans/demo/ws-123.md";
 
--- a/src/browser/components/tools/ProposePlanToolCall.tsx
+++ b/src/browser/components/tools/ProposePlanToolCall.tsx
@@ -277,11 +277,20 @@ export const ProposePlanToolCall: React.FC<ProposePlanToolCallProps> = (props) =
   // Format: Title as H1 + plan content for "Start Here" functionality.
   // Note: we intentionally preserve the plan file on disk when starting here so it can be
   // referenced later (e.g., via post-compaction attachments).
+  const planContentTrimmed = planContent.trim();
+  const hasPlanContentInChat =
+    planContentTrimmed.length > 0 && !planContentTrimmed.startsWith("*Plan saved to ");
+
   // When using "Start Here" (replace chat history), the plan is already included in the
-  // conversation. Keeping this note short avoids token bloat while discouraging redundant
-  // plan-file reads in Exec.
-  const startHereNote =
-    "\n\nNote: This chat already contains the full plan; no need to re-open the plan file.";
+  // conversation *only* when the Propose Plan tool result includes full plan text.
+  // Keeping this note short avoids token bloat while discouraging redundant plan-file
+  // reads in Exec.
+  const startHereNote = hasPlanContentInChat
+    ? "\n\nNote: This chat already contains the full plan; no need to re-open the plan file."
+    : planContentTrimmed.startsWith("*Plan saved to ")
+      ? "\n\nNote: This chat only includes a placeholder. Read the plan file below for the full plan."
+      : "";
+
   const planPathNote = planPath ? `\n\n---\n\n*Plan file preserved at:* \`${planPath}\`` : "";
   const startHereContent = `# ${planTitle}\n\n${planContent}${startHereNote}${planPathNote}`;
   const {


ThomasK33:
@codex review

Addressed feedback by tightening Start Here plan-summary detection to require the plan-path footer marker (avoids false positives for generic Start Here messages from the plan agent).

ThomasK33:
@codex review

ThomasK33:
@codex review

PR #1955, opened 2026-01-27
🤖 feat: persist Costs tab token breakdown in session-usage.json

ThomasK33:
feat: persist Costs tab token breakdown into session-usage.json

## Summary

- Persist the Costs tab’s **consumer breakdown** + **top file paths** into `~/.mux/sessions/<workspaceId>/session-usage.json` as a versioned `tokenStatsCache`.
- Hydrate the renderer’s consumer/file breakdown from this cache on app restart when it’s still valid, avoiding expensive re-tokenization.

## Implementation

- Extended `SessionUsageFileSchema` to include:
  - `rolledUpFrom` (prevents oRPC schema stripping during rollup)
  - `tokenStatsCache` (derived cache; versioned independently)
- Added `SessionUsageService.setTokenStatsCache(...)` (workspace lock + atomic write, best-effort rebuild on corrupt JSON).
- Updated `tokenizer.calculateStats` to include `workspaceId` so the backend can persist `tokenStatsCache` as part of the stats calculation flow.
- Updated `WorkspaceStore` to hydrate `WorkspaceConsumersState` from `session-usage.json` when:
  - model matches
  - messageCount matches
  - `maxHistorySequence` matches (when available)

## Validation

- `make static-check`

## Risks

- `tokenizer.calculateStats` input shape changed (internal IPC boundary). Updated all call sites + tests.
- Cache invalidation is conservative (model/history mismatch triggers recompute) to avoid incorrect token reporting.

---

<details>

# Persist Cost tab “consumer + file breakdown” into `session-usage.json`

## Context / Why
Mux already persists **cost/token usage** into `~/.mux/sessions/<workspaceId>/session-usage.json` (via `SessionUsageService`). The right sidebar **Cost** tab, however, still computes two expensive / derived views on-demand:

- **Consumer breakdown** (token usage by “User”, “Assistant”, and tools like `bash`, `file_read`, `web_search`, etc.)
- **File breakdown** (top file paths by token usage from file tools)

The goal is to treat these as a **cache** and persist them into `session-usage.json` so the file contains *everything currently shown in the Cost tab*, and so future tooling can aggregate across workspaces/projects (e.g., “most-used tools”, “most-read files/sections”).

Scope constraint: **only persist what the Cost tab already shows today** (consumer list + top file paths + supporting totals/metadata). Do **not** add new analytics beyond that yet.

---

## Evidence (repo facts)
- Cost tab UI and data sources:
  - `src/browser/components/RightSidebar/CostsTab.tsx`
    - Uses `useWorkspaceUsage()` → `workspace.getSessionUsage()` (reads `session-usage.json`) for cost totals.
    - Uses `useWorkspaceConsumers()` → `tokenizer.calculateStats({ messages, model })` for **consumer + file breakdown**.
  - `src/browser/components/RightSidebar/ConsumerBreakdown.tsx`
  - `src/browser/components/RightSidebar/FileBreakdown.tsx`
- Token consumer + top files computation:
  - `src/common/utils/tokens/tokenStatsCalculator.ts` exports `calculateTokenStats(...)` returning `ChatStats` (includes `consumers`, `totalTokens`, `tokenizerName`, `topFilePaths`).
  - Schemas: `src/common/orpc/schemas/chatStats.ts` (`ChatStatsSchema`, `TokenConsumer`, `TopFilePathSchema`).
- Session usage persistence:
  - `src/node/services/sessionUsageService.ts` writes `session-usage.json` (atomic write + per-workspace lock).
  - Current on-disk schema: `src/common/orpc/schemas/chatStats.ts` → `SessionUsageFileSchema` (`version: 1`, `byModel`, optional `lastRequest`).
  - Note: service uses `rolledUpFrom?: Record<string, true>` for rollup idempotency; schema should include it to avoid future parsing/stripping bugs.

---

## Recommended approach: extend `session-usage.json` with an optional derived “token stats cache”
**Net LoC estimate (product code): ~220–320**

### 1) Add a small, explicit cache blob to the on-disk schema (backwards compatible)
- Keep `session-usage.json` as the single file, but treat the new fields as **optional derived cache**.
- Add a nested object with its own version so we don’t have to bump `SessionUsageFile.version` for purely derived fields.

Proposed new field (conceptual):

```ts
// In SessionUsageFile (session-usage.json)
interface SessionUsageTokenStatsCacheV1 {
  version: 1;
  computedAt: number; // Date.now()

  // Tokenization context (because tokens vary by model/tokenizer)
  model: string;         // model passed to calculateTokenStats
  tokenizerName: string; // returned by calculateTokenStats

  // Cache invalidation (cheap fingerprint of the message history used)
  history: {
    messageCount: number;
    maxHistorySequence?: number; // from MuxMessage.metadata.historySequence
  };

  totalTokens: number;
  consumers: TokenConsumer[];
  topFilePaths?: Array<{ path: string; tokens: number }>;
}

interface SessionUsageFile {
  version: 1;
  byModel: Record<string, ChatUsageDisplay>;
  lastRequest?: { model: string; usage: ChatUsageDisplay; timestamp: number };
  rolledUpFrom?: Record<string, true>; // (fix schema gap)
  tokenStatsCache?: SessionUsageTokenStatsCacheV1;
}
```

Implementation details:
- Add `SessionUsageTokenStatsCacheSchema` to `src/common/orpc/schemas/chatStats.ts` using the existing `TokenConsumerSchema` and `TopFilePathSchema`.
- Extend `SessionUsageFileSchema` to include:
  - `rolledUpFrom: z.record(z.literal(true)).optional()`
  - `tokenStatsCache: SessionUsageTokenStatsCacheSchema.optional()`
- Ensure the schema remains **backwards compatible** with existing files.

### 2) Write/update the cache whenever token stats are computed
Where to persist:
- Ideal: when handling `tokenizer.calculateStats(...)` (since this is the source of the Cost tab’s breakdown data).

Plan:
- In the node-side tokenizer implementation (likely `src/node/services/tokenizerService.ts`), after computing `ChatStats`:
  1. Build a `SessionUsageTokenStatsCacheV1` object from `ChatStats`.
  2. Compute `history.messageCount` and `history.maxHistorySequence` from the `messages` argument.
  3. Call into `SessionUsageService` (new method) to persist the cache under the same per-workspace lock + atomic write.

New `SessionUsageService` API (conceptual):
- `setTokenStatsCache(workspaceId, cache: SessionUsageTokenStatsCacheV1): Promise<void>`
  - Must preserve existing `byModel`, `lastRequest`, and `rolledUpFrom` data.
  - Must treat `tokenStatsCache` as replaceable (not additive).

Defensive checks (crash-fast in dev/tests, tolerant in prod):
- Assert `cache.totalTokens >= 0` and all `consumer.tokens >= 0`.
- Assert `cache.history.messageCount === messages.length` at callsite.
- If validation fails in production builds, log + skip writing cache (do not crash).

### 3) Read the cache first; fall back to recomputation when stale/missing
Update the consumer breakdown pipeline so we don’t redo work unnecessarily.

Plan:
- In `WorkspaceStore` / `WorkspaceConsumerManager`:
  - When the workspace loads, we already fetch `session-usage.json` (via `getSessionUsage`).
  - If `tokenStatsCache` exists, compare it against the current in-memory message list:
    - same `messageCount`
    - same `maxHistorySequence` (if present)
    - and tokenization `model` matches the model we would use for calculation
  - If it matches: hydrate `WorkspaceConsumersState` from the cache and skip `tokenizer.calculateStats`.
  - If not: run calculation as today; the tokenizer layer will refresh the cache automatically (Step 2).

Notes:
- This ensures cache correctness when users delete/truncate/compact history.
- If `maxHistorySequence` isn’t available on older messages, fall back to `messageCount` matching only.

### 4) Rebuild behavior: never block on cache recomputation
`SessionUsageService.getSessionUsage()` currently auto-rebuilds the file from message history if missing/corrupt.

Plan:
- On rebuild-from-messages, do **not** recompute `tokenStatsCache` (tokenization is expensive).
- Instead, write a rebuilt file with `tokenStatsCache: undefined`.
- The next time the UI/tokenizer requests stats, it will re-populate the cache.

### 5) Tests
- Add/update unit tests for:
  - `SessionUsageFileSchema` parsing old files still works.
  - `rolledUpFrom` is preserved on read/write (regression test).
  - `setTokenStatsCache` writes the cache and preserves existing ledger fields.
  - Cache invalidation logic (stale vs fresh) in `WorkspaceConsumerManager` (if existing test infra covers it; otherwise add a lightweight unit around the comparison helper).

---

## Alternative (not recommended for this request): separate “details” file
**Net LoC estimate (product code): ~180–260**

<details>
<summary>Why consider it</summary>

A separate `session-usage-details.json` would keep `getSessionUsageBatch()` responses small for archived-workspace lists. It also avoids mixing “ledger-like” data (costs) with “derived cache” data (token consumers/files).

This is *not* the requested direction (you asked specifically for `session-usage.json`), so treat this as a fallback if file size or batch-fetch perf becomes an issue.

</details>

---

## Rollout / sequencing
1. Land schema extension + `rolledUpFrom` schema fix.
2. Add `SessionUsageService.setTokenStatsCache` + integrate with tokenizer calculateStats.
3. Update UI store logic to prefer cache + validate staleness.
4. Add tests + run `make static-check`.

</details>

---
_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$1.08`_

### DIFF
--- a/mobile/src/contexts/WorkspaceCostContext.tsx
+++ b/mobile/src/contexts/WorkspaceCostContext.tsx
@@ -293,7 +293,11 @@ export function WorkspaceCostProvider({
       }
 
       const model = getLastModel(messages) ?? "unknown";
-      const stats = await client.tokenizer.calculateStats({ messages, model });
+      const stats = await client.tokenizer.calculateStats({
+        workspaceId: workspaceId!,
+        messages,
+        model,
+      });
       setConsumers({ status: "ready", stats });
     } catch (error) {
       const message = error instanceof Error ? error.message : String(error);


ThomasK33:
@codex review

ThomasK33:
@codex review

PR #1982, opened 2026-01-28
🤖 feat: add navigation between user messages in chat

ammar-agent:
## Summary

Adds backward/forward navigation controls to user messages in the chat UI, allowing users to quickly jump between their own messages in a conversation.

## Background

When reviewing long conversations, users need to navigate between their own messages to retrace their questions or commands. This change adds chevron navigation buttons (◀ ▶) to the user message action row.

## Implementation

- **UserMessage.tsx**: Added `UserMessageNavigation` interface and navigation buttons to the action row. Buttons render when there are 2+ user messages; disabled state (reduced opacity) when at first/last message to prevent layout shift.
- **MessageRenderer.tsx**: Passes navigation prop through to UserMessage.
- **ChatPane.tsx**: Computes a navigation map (`userMessageNavMap`) from deferred messages, builds prev/next links for each user message. Adds `handleNavigateToMessage` callback that scrolls to target using `scrollIntoView({ behavior: "smooth", block: "center" })`.

Key design decisions:
- Navigation appears only when there are 2+ user messages (single-message chats unchanged)
- Both buttons always render to preserve layout consistency; disabled buttons use `opacity-30`
- Scrolls to center the target message in the viewport

## Risks

Low risk—additive UI change. Navigation is isolated to user messages and uses existing `data-message-id` attributes for scroll targeting.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$2.67`_

### DIFF
--- a/src/browser/components/ChatPane.tsx
+++ b/src/browser/components/ChatPane.tsx
@@ -186,6 +186,26 @@ export const ChatPane: React.FC<ChatPaneProps> = (props) => {
     [workspaceId, latestMessageId, onOpenTerminal]
   );
 
+  // Compute navigation map for user messages (prev/next historyIds for each user message)
+  // Only enabled when there are 2+ user messages to navigate between
+  const userMessageNavMap = useMemo(() => {
+    const userMessages = deferredMessages.filter(
+      (m): m is DisplayedMessage & { type: "user" } => m.type === "user"
+    );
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

PR #1996, opened 2026-01-28
🤖 fix: reuse existing empty draft when clicking New Workspace

ammar-agent:
## Summary

When clicking "New Workspace" (the "+" button or project row), the app now navigates to an existing empty draft instead of creating a new one.

## Background

Users clicking "New Workspace" multiple times would accumulate empty drafts in the sidebar, creating clutter. This change ensures that if an empty draft already exists for the project, we reuse it rather than creating another.

## Implementation

- Added `isDraftEmpty()` helper to check if a draft has no input text, attachments, or workspace name set
- Added `findExistingEmptyDraft()` to search for a reusable empty draft in the project
- Modified `createWorkspaceDraft()` to check for and reuse existing empty drafts
- Reads directly from localStorage to get fresh state, avoiding stale closure issues with React state

A draft is considered "empty" if it has:
- No input text
- No attachments
- No workspace name state (neither auto-generated nor manual)

## Validation

- Added integration test in `tests/ui/draftWorkspace.integration.test.ts` that verifies clicking "New Workspace" twice only creates one draft when the first is empty

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$4.20`_

### DIFF
--- a/tests/ui/agentPicker.integration.test.ts
+++ b/tests/ui/agentPicker.integration.test.ts
@@ -8,6 +8,8 @@
  * - Broken agent definitions show error indicators
  */
 
+import "./dom";
+
 import { fireEvent, waitFor } from "@testing-library/react";
 import * as fs from "node:fs/promises";
 import * as path from "node:path";


ammar-agent:
@codex review

Addressed section-aware draft reuse.

ammar-agent:
@codex review

Bootstrapped DOM import before agent picker test modules.

PR #2052, opened 2026-01-30
🤖 feat: support disabled-by-default agents with local overrides

ThomasK33:
## Summary
Adds repo-controlled agent disablement via a top-level `disabled: true` front-matter field, with per-user local overrides (`agentAiDefaults[agentId].enabled`) that take precedence.

Disabled agents are hidden from the agent picker/cycling and cannot run as sub-agents, but are still visible in **Settings → Agents** so users can opt in.

## Background
We want repos to be able to ship experimental agents that are off by default, while still letting each user explicitly enable/disable agents locally without editing agent files.

## Implementation
- Agent front-matter schema now supports `disabled: true` (takes precedence over legacy `ui.disabled`).
- Config schema adds `agentAiDefaults[agentId].enabled?: boolean` as a local override.
- Centralized node-side helper resolves "effectively disabled" using: core-agent safety → user override → front-matter.
- `agents.list` filters disabled agents by default, and supports `includeDisabled: true` for Settings.
- Enforced at runtime:
  - Task creation rejects disabled agents.
  - AIService prevents disabled agents from running as sub-agents (top-level workspaces fall back to `exec` to avoid bricking).

## Validation
- `make static-check`
- Added unit tests for enablement precedence and an ipc test for `agents.list` filtering + overrides.

## Risks
- Changes alter agent discovery semantics (discovery now returns all agents; filtering happens at `agents.list`).
- Incorrect enablement resolution could hide agents unexpectedly; mitigated via centralized helper + tests.

---

<details>

# Plan: Agent enable/disable (front-matter default + user override)

## Context / Why
You want two related capabilities:

1) **Repo-controlled default disablement**: extend agent Markdown front-matter with a top-level `disabled: true` field so repos can ship “experimental” agents that are off by default.

2) **User-controlled local override**: add a per-user settings toggle to enable/disable any agent locally (stored in `~/.mux/config.json`). The user setting must take precedence over front-matter so each user can opt in/out.

End result:
- Disabled agents **don’t appear in the agent picker / cycle hotkeys**, **can’t be selected**, and **can’t run as sub-agents**.
- Users can still see/manage disabled-by-default agents in **Settings → Agents**.

## Evidence (repo facts)
- Agent definitions are Markdown + YAML front-matter:
  - Built-in: `src/node/builtinAgents/*.md`
  - Project: `.mux/agents/*.md`
  - Global: `~/.mux/agents/*.md`
  - Parser: `src/node/services/agentDefinitions/parseAgentDefinitionMarkdown.ts`
  - Schema: `src/common/orpc/schemas/agentDefinition.ts`
- Current “disabled” exists as **legacy** `ui.disabled` (hides from discovery/UI) in `AgentDefinitionUiSchema` (`src/common/orpc/schemas/agentDefinition.ts`).
- Agents list API is implemented in `src/node/orpc/router.ts` (`agents.list`) and currently filters out `resolvedFrontmatter.ui?.disabled === true`.
- Sub-agent creation path:
  - Tool: `src/node/services/tools/task.ts` → `TaskService.create`
  - Validation currently checks only `frontmatter.subagent?.runnable` in `src/node/services/taskService.ts`.
- Settings “Agents” UI is `src/browser/components/Settings/sections/TasksSection.tsx` and it already edits per-agent defaults stored in `agentAiDefaults` (`src/common/types/agentAiDefaults.ts`) via `api.config.saveConfig`.

## Recommended approach (single source of truth + minimal UI churn)
Net LoC estimate (product code only): **~300 LoC**

#### 1.1 Agent front-matter: add top-level `disabled`
Update `src/common/orpc/schemas/agentDefinition.ts`:
- Add `disabled?: boolean` to `AgentDefinitionFrontmatterSchema` (top-level).
- Keep `ui.disabled` working for backwards compatibility.

Front-matter precedence rule (within front-matter):
- `frontmatter.disabled` **wins** over `frontmatter.ui.disabled` when both are set.

Example:
```yaml
---
name: My Experimental Agent
base: exec
disabled: true
---
```

#### 1.2 User override: per-agent `enabled?: boolean`
Extend `src/common/types/agentAiDefaults.ts`:
```ts
export interface AgentAiDefaultsEntry {
  modelString?: string;
  thinkingLevel?: ThinkingLevel;
  /**
   * Local override:
   * - true: force enable
   * - false: force disable
   * - undefined: inherit from agent front-matter
   */
  enabled?: boolean;
}
```
Update `normalizeAgentAiDefaults` so it:
- Parses `enabled` when it’s a boolean.
- **Retains entries** that only set `enabled` (currently it drops entries that don’t set model/thinking).

Update ORPC config schema in `src/common/orpc/schemas/api.ts` (`SubagentAiDefaultsEntrySchema`) to accept the new field:
```ts
const SubagentAiDefaultsEntrySchema = z.object({
  modelString: z.string().min(1).optional(),
  thinkingLevel: z.enum(["off","low","medium","high","xhigh"]).optional(),
  enabled: z.boolean().optional(),
}).strict();
```

### 2) Central enablement resolution helper
Create a shared helper (prefer node-side because it needs config + resolved front-matter):
- New file: `src/node/services/agentDefinitions/agentEnablement.ts`

API shape:
```ts
import type { AgentDefinitionFrontmatter, AgentId } from "@/common/types/agentDefinition";
import type { ProjectsConfig } from "@/common/types/project";

export function isAgentDisabledByFrontmatter(frontmatter: AgentDefinitionFrontmatter): boolean {
  if (typeof frontmatter.disabled === "boolean") return frontmatter.disabled;
  if (typeof frontmatter.ui?.disabled === "boolean") return frontmatter.ui.disabled;
  return false;
}

export function resolveAgentEnabledOverride(cfg: ProjectsConfig, agentId: AgentId): boolean | undefined {
  const entry = cfg.agentAiDefaults?.[agentId];
  return typeof entry?.enabled === "boolean" ? entry.enabled : undefined;
}

export function isAgentEffectivelyDisabled(args: {
  cfg: ProjectsConfig;
  agentId: AgentId;
  resolvedFrontmatter: AgentDefinitionFrontmatter;
}): boolean {
  const override = resolveAgentEnabledOverride(args.cfg, args.agentId);
  if (override === true) return false;
  if (override === false) return true;
  return isAgentDisabledByFrontmatter(args.resolvedFrontmatter);
}
```

#### 3.1 Stop filtering disabled agents during *discovery*
Update `src/node/services/agentDefinitions/agentDefinitionsService.ts`:
- `discoverAgentDefinitions(...)` currently filters out `entry.disabled` at the end.
- Remove that filter so discovery returns **all agent IDs** (built-in/global/project), even if disabled by default.

Rationale: a repo-disabled agent must still be discoverable so a user can opt in.

#### 3.2 Apply effective enablement in `agents.list`
Update ORPC schema input (`src/common/orpc/schemas/api.ts`):
- Add `includeDisabled?: boolean` to `AgentDiscoveryInputSchema`.

Update handler in `src/node/orpc/router.ts` (`agents.list`):
- Load config: `const cfg = context.config.loadConfigOrDefault();`
- For each discovered agent:
  - `resolvedFrontmatter = await resolveAgentFrontmatter(...)`
  - `disabled = isAgentEffectivelyDisabled({ cfg, agentId: descriptor.id, resolvedFrontmatter })`
  - If `disabled === true` and `includeDisabled !== true`: return `null` (omit from response)

This makes **all UI agent lists** (Agent picker + cycling) automatically respect user overrides.

#### 4.1 Prevent disabled sub-agents at creation time
Update `src/node/services/taskService.ts` (`TaskService.create`):
- After resolving frontmatter and confirming `subagent.runnable === true`, compute effective disabled using config + resolved frontmatter.
- If disabled: return `Err("Task.create: agentId is disabled ...")`.

This blocks “explicit agentId” bypass even if the agent tool tries to call a hidden/disabled agent.

#### 4.2 Prevent disabled agents from running in a task workspace
Update `src/node/services/aiService.ts` (`AIService.streamMessage`):
- For **sub-agent workspaces** (`metadata.parentWorkspaceId` truthy), check whether the persisted agentId is effectively disabled.
- If disabled: emit an error event + return `Err({ type: ..., message: "Agent is disabled" })`.

Note: for top-level workspaces you can optionally fall back to `exec` (similar to unknown agent fallback), but for sub-agents prefer hard-fail to avoid silently running a different agent/tool policy.

#### 4.3 Tool description sub-agent list
Update `src/node/services/aiService.ts` where it sets `agentDefinitions` (used as `availableSubagents`):
- Filter to *effectively enabled* agents before passing to `getToolsForModel`.
- This keeps disabled agents out of the `task` tool description’s “Available sub-agents …” list.

### 5) Settings UI (local override)
Update `src/browser/components/Settings/sections/TasksSection.tsx`:

#### 5.1 Fetch agent lists
- Call `api.agents.list({ projectPath, workspaceId, includeDisabled: true })` so Settings can show repo-disabled agents.
- Also call `api.agents.list({ projectPath, workspaceId })` (default) to compute the *enabled set*.

#### 5.2 Render enable/disable control per agent
- Add a new “Enabled” row/column per agent card (near Model/Reasoning).
- UI behavior:
  - Switch on/off reflects **effective enabled** (membership in enabled set).
  - Toggling sets `agentAiDefaults[agentId].enabled = true|false`.
  - Optional: add a “Reset” button when `enabled` is set in config to clear override (delete the field).

Pseudo-snippet:
```tsx
const enabledIds = useMemo(() => new Set(enabledAgents.map((a) => a.id)), [enabledAgents]);
const isEnabled = enabledIds.has(agent.id);

<Switch checked={isEnabled} onCheckedChange={(checked) => {
  setAgentAiDefaults((prev) => updateAgentDefaultEntry(prev, agent.id, (entry) => {
    entry.enabled = checked;
  }));
}} />
```

#### 5.3 Ensure the agent picker updates immediately
After saving config (or even optimistically on toggle), call `useAgent().refresh()` so the main agent list in the UI updates without requiring restart.

### 6) UX guardrails / safety
- Consider disallowing disabling the critical built-ins (`exec`, possibly `plan`) in Settings UI (disable the switch + tooltip why) so the app can’t be bricked.
- Ensure AgentContext does **not** auto-overwrite the stored agentId when `disableWorkspaceAgents` is enabled (the “unbrick” mode).

### 7) Tests
- Node:
  - Add unit tests for the new enablement helper to ensure precedence:
    - user override wins
    - `frontmatter.disabled` wins over `ui.disabled`
    - inheritance re-enable works (base disabled, child disabled:false)
- Node integration-ish:
  - Add tests around `agents.list` handler behavior with `includeDisabled` and config overrides.
- UI:
  - Add a `TasksSection` test verifying toggling Enabled writes to config payload (can be done with mocked ORPC api).

<details>
<summary>Alternative approach (cleaner UI API, more plumbing)</summary>

Instead of double-fetching `agents.list` in Settings, add a new route:
- `agents.listForSettings` → returns `{ agents: AgentDefinitionDescriptor[], enabledAgentIds: string[] }` (or include `effectiveEnabled` per agent).

Pros: single network call; UI doesn’t compute membership.
Cons: more ORPC schema + router + client changes.

Net LoC estimate: **~370 LoC**.
</details>

</details>

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: $8.09_

### DIFF
--- a/src/browser/components/Settings/sections/TasksSection.tsx
+++ b/src/browser/components/Settings/sections/TasksSection.tsx
@@ -662,16 +662,20 @@ export function TasksSection() {
 
           <div className="flex shrink-0 items-center gap-3">
             {enablementHint ? <div className="text-muted text-xs">{enablementHint}</div> : null}
-            <div className="flex items-center gap-2">
-              <div className="text-muted text-xs">Enabled</div>
-              <Switch
-                checked={enabledValue}
-                disabled={enablementLocked}
-                title={enablementTitle}
-                onCheckedChange={(checked) => setAgentEnabled(agent.id, checked)}
-                aria-label={`Toggle ${agent.id} enabled`}
-              />
-            </div>
+            <Tooltip>
+              <TooltipTrigger asChild>
+                <div className="flex items-center gap-2">
+                  <div className="text-muted text-xs">Enabled</div>
+                  <Switch
+                    checked={enabledValue}
+                    disabled={enablementLocked}
+                    onCheckedChange={(checked) => setAgentEnabled(agent.id, checked)}
+                    aria-label={`Toggle ${agent.id} enabled`}
+                  />
+                </div>
+              </TooltipTrigger>
+              <TooltipContent>{enablementTitle}</TooltipContent>
+            </Tooltip>
             {enabledOverride !== undefined ? (
               <Button
                 type="button"


ThomasK33:
@codex review

Addressed feedback: when falling back to `exec` (top-level workspace + disabled agent), we now also update `effectiveAgentId` to match the loaded definition so transition logic, telemetry, and message metadata stay consistent.

PR #2072, opened 2026-01-31
🤖 fix: show forked workspace in sidebar immediately

ammar-agent:
Summary
- Fixes a `/fork` UX regression where the newly-created workspace could fail to show up in the left sidebar immediately.
- Adds a UI integration test covering `/fork` sidebar visibility.

Background
- The fork switch is driven by a custom browser event. Previously, if the event fired before `ProjectContext` finished loading, the handler would early-return and skip updating sidebar state.

Implementation
- App fork-switch handler no longer early-returns when the project isn't present yet.
- If the project is missing, we trigger `refreshProjects()` best-effort.

Validation
- `TEST_INTEGRATION=1 bun x jest tests/ui/workspaceFork.integration.test.ts --runTestsByPath --testTimeout=60000 --silent`
- `make static-check`

Risks
- Low: change is scoped to the fork-switch event path and uses best-effort refresh.

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `high` • Cost: `$39.08`_

### DIFF
--- a/src/browser/App.tsx
+++ b/src/browser/App.tsx
@@ -102,6 +102,7 @@ function AppInner() {
 
   const {
     projects,
+    refreshProjects,
     removeProject,
     openProjectCreateModal,
     isProjectCreateModalOpen,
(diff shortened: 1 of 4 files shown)


ammar-agent:
@codex review

Addressed the race: `ProjectContext.refreshProjects()` now ignores out-of-date list() responses so an older in-flight refresh can’t clobber a newer fork-triggered refresh.

Added a unit test (`refreshProjects ignores stale responses`) that reproduces the out-of-order resolution case.

ammar-agent:
@codex review

(Quick follow-up) Force-pushed again to include the staged `ProjectContext` race-guard + unit test in the PR diff.

ammar-agent:
@codex review

Updated `ProjectContext.refreshProjects()` so overlapping refreshes don’t drop an older success when a newer refresh fails:
- only advance the “latest applied” marker on success
- on error, keep the previous project list (don’t clear)

Added a regression test: `refreshProjects applies older success if a newer overlapping refresh fails`.

ammario (approved):

PR #2077, opened 2026-01-31
🤖 fix: stabilize runtime options layout

ammar-agent:
Summary
- Move creation runtime options ("from" branch + SSH host / Docker image) onto a dedicated second row.
- Keep widths stable while loading (branch skeleton matches the select width).

Background
- There was visible flex-wrap jitter/flash as branches load and runtime option controls appear.

Implementation
- Added a small `showRuntimeOptionsRow` gate and rendered `RuntimeOptionsRow` below the runtime button group.

Validation
- `make static-check` (local)

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `high` • Cost: `$10.64`_


ammar-agent:
@codex review

PR #2079, opened 2026-01-31
🤖 perf: cut DiffRenderer O(lines) tooltip/icon cost

ammar-agent:
## Summary
This PR reduces renderer allocations when viewing large code-review diffs by removing **O(lines)** tooltip/icon subtrees from `SelectableDiffRenderer`, replacing them with a **single** hover-positioned “add review comment” affordance and event delegation. It also adds an opt-in `make dev-fast` mode to profile dev builds closer to production.

## Background
A heap profile (captured **Jan 18, 2026**) showed `DiffRenderer.tsx`, Radix Tooltip, and `lucide-react` as major allocation sources (amplified in dev via `jsxDEV` + `StrictMode`).

## Implementation
- **DiffRenderer**
  - Remove per-line Radix tooltips and per-line `MessageSquare` icons.
  - Render one overlay button positioned over the hovered +/- indicator cell; use delegated handlers (no per-line closures).
  - Reliability: show the overlay on `mouseover` (not `mousemove`) so it appears immediately on entry.
- **DiffContainer**
  - Add `contentProps`/`contentRef` hooks and make the inner grid `relative` so overlays can be anchored reliably.
- **Dev profiling**
  - Add `make dev-fast` (Vite `--mode fast`) + `VITE_MUX_FAST_DEV=1` to disable `StrictMode`, avoid `jsxDEV`, and disable React Refresh.

## Validation
- `make static-check`

## Risks
- Hover affordance relies on delegated DOM attributes; regressions are possible around selection/drag behavior.

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$10.49`_

### DIFF
--- a/src/browser/components/shared/DiffRenderer.tsx
+++ b/src/browser/components/shared/DiffRenderer.tsx
@@ -5,6 +5,7 @@
  */
 
 import React, { useEffect, useMemo, useState } from "react";
+import { createPortal } from "react-dom";
 import { stopKeyboardPropagation } from "@/browser/utils/events";
 import { cn } from "@/common/lib/utils";
 import { getLanguageFromPath } from "@/common/utils/git/languageDetector";
@@ -274,7 +275,7 @@ const DiffIndicator: React.FC<DiffIndicatorProps> = ({
   <span
     data-diff-indicator={true}
     data-line-index={lineIndex}
-    className={cn("text-center select-none", isInteractive && "cursor-pointer")}
+    className={cn("relative text-center select-none", isInteractive && "cursor-pointer")}
     style={{ background, color: getIndicatorColor(type) }}
   >
     {getIndicatorChar(type)}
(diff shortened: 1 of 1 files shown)


ammar-agent:
I reproduced two issues you found and pushed a fix:

- Hover affordance never showing: `mousemove` events can target a Text node for the +/- character, so `.closest()` never matched `[data-diff-indicator]`.
- Sticky empty vertical scroll area: a hidden-but-transformed absolutely-positioned overlay button can still affect scroll extents. The hide path now fully removes it from layout.

Patch: make hover detection work for Text node targets, and toggle the overlay button via `style.display` ("none"/"flex") while clearing transform/size on hide.

ammar-agent:
Pushed another iteration to address the two regressions:

- Hover affordance still not showing: stop relying on Tailwind `data-[visible=true]` variants for visibility. The overlay now toggles `display/opacity/pointerEvents` via inline styles, so it should show even if the data-variant CSS isn't generated.
- Escape not cancelling: add a capture-phase `window` keydown listener while a selection is active to ensure Esc always cancels the in-progress review note (even if focus is not on the textarea or a global handler intercepts).

Also hardened the delegated hover target lookup to avoid `instanceof Element` (can be flaky across Electron realms).

ammar-agent:
@codex review

Addressed the Makefile Windows `dev-fast` indentation issue and updated the DiffRenderer hover affordance logic; would appreciate a re-review.

ThomasK33:
Closing as mostly superseded. #3012 replaced the per-line Radix tooltips with one shared tooltip, which was the main allocation fix here. The per-line button and icon cost would need a fresh heap profile to justify, and `make dev-fast` would be a separate developer-tooling PR.

---

PR #2080, opened 2026-01-31
🤖 refactor: extract GatewayToggleButton for consistent hover states

ammar-agent:
## Summary

Extract a shared `GatewayToggleButton` component to fix confusing hover states where the inactive hover looked identical to the enabled state.

## Background

The gateway icon's hover state was the same as its enabled state, making it difficult to tell whether the icon was already active or just being hovered. This was confusing when enabling/disabling gateway routing.

## Implementation

Created `GatewayToggleButton` component with:
- **Props:** `active`, `onToggle`, `variant` (bordered/plain), `size` (sm/md), `showTooltip`
- **Consistent hover behavior:**
  - Active state: accent color, dims on hover (`hover:opacity-70`)
  - Inactive state: muted color, brightens slightly but NOT to accent color

Updated consumers:
- `ModelSelector.tsx` - dropdown uses `variant="bordered" size="sm" showTooltip`
- `ModelRow.tsx` - settings row uses defaults (`variant="plain" size="md"`)

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$1.21`_

### DIFF
--- a/src/browser/components/GatewayToggleButton.tsx
+++ b/src/browser/components/GatewayToggleButton.tsx
@@ -1,0 +1,75 @@
+import React from "react";
+import { cn } from "@/common/lib/utils";
+import { GatewayIcon } from "./icons/GatewayIcon";
+import { Tooltip, TooltipContent, TooltipTrigger } from "./ui/tooltip";
+
+interface GatewayToggleButtonProps {
+  /** Whether gateway is currently enabled for this model */
+  active: boolean;
+  /** Called when user clicks to toggle */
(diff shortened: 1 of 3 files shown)


ammar-agent:
Resolved Codex comment - the `GatewayIcon` import is actually still used at line 307 for the main inline button that has a custom rich tooltip with account status (balance, concurrent requests). This usage can't use `GatewayToggleButton` because it needs a different tooltip structure. Lint passes locally.

@codex review

PR #2084, opened 2026-01-31
🤖 fix: prevent chat flashes when switching workspaces

ammar-agent:
## Summary
Prevent cross-workspace layout flashes in the chat pane (most visible with large `propose_plan` outputs) by enforcing a hard workspace boundary for the chat UI.

## Background
When switching between workspaces, `ChatPane` stayed mounted and relied on `useDeferredValue()` for message rendering. If two workspaces had the same message count, React could briefly reuse the previous workspace’s deferred message list, causing a visible “flash” of the wrong transcript.

## Implementation
- Key `ChatPane` by `workspaceId` in `WorkspaceShell` to guarantee unmount/remount on workspace switch.
- Hoist `useIdleCompactionHandler` to `AppLoaderInner` so its serialization state doesn’t reset when switching workspaces.
- Avoid running idle compaction while mux is blocked by policy.
- Make queued-message editing safe by ensuring the `ChatInputAPI` is ready before clearing the backend queue.

## Validation
- `make fmt`
- `make static-check`

## Risks
- `ChatPane` now fully remounts on workspace switch (expected, similar to `RightSidebar` keying). This may slightly increase workspace-switch render cost and will reset ephemeral UI state per workspace.

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$11.83`_

### DIFF
--- a/src/browser/components/WorkspaceShell.tsx
+++ b/src/browser/components/WorkspaceShell.tsx
@@ -112,7 +112,7 @@ export const WorkspaceShell: React.FC<WorkspaceShellProps> = (props) => {
     >
       {/* Keyed by workspaceId to prevent cross-workspace message-list flashes. */}
       <ChatPane
-        key={props.workspaceId}
+        key={`chat-${props.workspaceId}`}
         workspaceId={props.workspaceId}
         workspaceState={workspaceState}
         projectPath={props.projectPath}


ammar-agent:
@codex review

Please review for correctness, UX regressions on workspace switching, and any perf concerns.

PR #2088, opened 2026-02-02
🤖 fix: keep UI visible during API reconnects

ethanndickson:
## Summary
- Keep the app UI mounted while the API client reconnects so transient disconnects no longer leave a grey screen.
- Avoid flashing the full-screen LoadingScreen after reconnects by only showing it for the first boot.

## Background
ORPC WebSocket disconnects set the API client to `null` on the frontend. `TerminalRouterProvider` previously returned `null` when the API was unavailable, which unmounted `<App />` and left only the reconnect overlay on a grey background. On reconnect, project/workspace metadata reloads set the loading flags again, which forced the full-screen “Loading workspaces…” screen even though the user already had a working UI, resulting in a jarring flash.

## Implementation
- Always render children from `TerminalRouterProvider`; dispose the router when the API disconnects and let the context be `null` during reconnects.
- Update `useTerminalRouter`/`TerminalView` to handle a null router by skipping subscriptions, input forwarding, and resize calls until the API is back.
- Add an `initialLoadComplete` latch in `AppLoader` so `LoadingScreen` only appears on the first successful load; later reconnects keep the last-known UI visible while stores resync.

## Validation
- `make static-check`

## Risks
- During reconnects, terminal input and resize events are dropped until the router is available, and the UI can briefly show stale metadata. This is limited to the reconnect window and avoids unmounting the app.

---

_Generated with `mux` • Model: `openai:gpt-5.2-codex` • Thinking: `xhigh` • Cost: `$22.87`_

### DIFF
--- a/src/browser/terminal/TerminalRouterContext.tsx
+++ b/src/browser/terminal/TerminalRouterContext.tsx
@@ -5,7 +5,7 @@
  * Must be wrapped inside APIProvider since it depends on the API client.
  */
 
-import { createContext, useContext, useEffect, useRef } from "react";
+import { createContext, useContext, useEffect, useState } from "react";
 import { useAPI } from "@/browser/contexts/API";
 import { TerminalSessionRouter } from "./TerminalSessionRouter";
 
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

Addressed stale router handling in TerminalView (ref-based input routing + guarded resize paths).

ethanndickson:
@codex review

PR #2114, opened 2026-02-03
🤖 fix: match hyphenated skills in slash suggestions

ethanndickson:
Summary
- Match slash suggestion partials against hyphenated segments so `/r` can surface `deep-review`.
- Add a unit test covering hyphenated skill matching.

Background
- Slash suggestions only matched the start of the full skill name, so hyphenated skills were hidden when users typed later segments.

Implementation
- Split suggestion keys on `-` and match partials against each segment.
- Added a focused unit test in `suggestions.test.ts`.

Validation
- `bun test src/browser/utils/slashCommands/suggestions.test.ts`

Risks
- Low: expands match criteria for hyphenated names only; single-word commands remain unchanged.

---

_Generated with `mux` • Model: `openai:gpt-5.2-codex` • Thinking: `xhigh` • Cost: `$4.18`_

### DIFF
--- a/src/browser/utils/slashCommands/suggestions.test.ts
+++ b/src/browser/utils/slashCommands/suggestions.test.ts
@@ -60,6 +60,20 @@ describe("getSlashCommandSuggestions", () => {
     expect(labels).toContain("/deep-review");
   });
 
+  it("matches full prefixes that cross hyphen boundaries", () => {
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

Addressed the hyphenated prefix regression by preserving full-key prefix matching alongside segment matching, plus a regression test for /deep-r.

PR #2117, opened 2026-02-03
🤖 fix: gate context switch warnings to explicit model switches

ethanndickson:
## Summary
- Context switch warnings now trigger **only** after **explicit** model changes (user selection or explicit agent switch), not on workspace entry/mount or background localStorage sync.
- Added **in-memory** tracking for workspace model changes (`user|agent|sync`) with **stale-entry protection** (no-op guard + divergence clearing).
- Normalized gateway vs canonical model IDs so "same model" doesn't look like a switch; allow same-model warnings only when the **effective context limit changes** (e.g. 1M toggle).
- Added regression tests for the new behavior and small banner tweaks for testability/a11y.

## Background
The context switch warning banner was showing up when the workspace model changed due to background state reconciliation (workspace entry seeding, localStorage sync, default/policy refreshes). This created noisy banners unrelated to user intent.

### Origin-gating (explicit vs sync)
- New helper module: `src/browser/utils/modelChange.ts`.
- Model changes are tagged with an origin:
  - `user`: user explicitly selected a model
  - `agent`: agent mode switch explicitly changed the default model
  - `sync`: background reconciliation (mount, workspace entry, localStorage sync, default refresh)
- Context switch warnings are evaluated **only** when the warning hook consumes a pending explicit change for the observed model.

### Stale-entry protection (prevents false positives)
To avoid "old" explicit-change entries triggering warnings later:
- **No-op guard:** don't record an explicit entry when the normalized stored model already equals the requested model (including gateway-vs-canonical equivalents).
- **Divergence guard:** if the store reports a totally different model before we observe the target model, drop the explicit entry (it's stale).
- **Rapid A→B protection:** if we briefly observe the previous model again while tracking the target, keep the newest pending entry.

### Core behavior
- `src/browser/utils/modelChange.ts`
  - `setWorkspaceModelWithOrigin(workspaceId, model, origin)` (intended entry point)
  - `consumeWorkspaceModelChange(workspaceId, model)` (one-shot origin consumption)
  - Reads current persisted model (`readPersistedString(getModelKey(workspaceId))`) to implement no-op + divergence handling.
- `src/browser/hooks/useContextSwitchWarning.ts`
  - Consumes explicit-change origin when `pendingModel` changes and only queues warnings for `user|agent`.
  - Defers warnings until usage loads (tokens > 0) for explicit switches.
  - Clears warnings on sync/background model changes.
- Integration points now consistently set origin:
  - `src/browser/contexts/WorkspaceContext.tsx`: seeds model from backend with `origin: "sync"` (no banner on entry).
  - `src/browser/components/WorkspaceModeAISync.tsx`: tags as `agent` only for real agentId flips; otherwise `sync`.
  - `src/browser/components/ChatInput/index.tsx` + `useCreationWorkspace.ts`: user selections are `user`; creation/default sync is `sync`; explicit agent switches are `agent`.
- `src/browser/utils/compaction/contextSwitchCheck.ts`
  - Compares models via `normalizeGatewayModel()` to avoid gateway/canonical mismatches.
  - Adds `allowSameModel` option to allow same-model warnings only when callers know the effective limit changed (used for 1M toggles).

### Supporting changes included in this PR
- `src/browser/components/ContextSwitchWarning.tsx`
  - Adds `data-testid="context-switch-warning"` for tests and small a11y tweaks (`aria-label` dismiss, icon `aria-hidden`).
- `src/browser/contexts/PolicyContext.tsx` + `PolicyContext.test.tsx`
  - Adds a payload signature guard to avoid redundant state updates on identical policy payloads; tests cover stability + reason updates.
- `src/browser/utils/chatCommands.ts`
  - Removes `onModelChange` from /model command context (single source of truth becomes `setPreferredModel` / ChatInput plumbing).
- `src/browser/stories/App.chat.stories.tsx`
  - Updates story play steps to use `setWorkspaceModelWithOrigin(..., "user")` so the warning matches new "explicit only" gating.

## Validation
- `bun test src/browser/hooks/useContextSwitchWarning.test.ts`
- `bun test src/browser/utils/modelChange.test.ts`
- `bun test src/browser/components/WorkspaceModeAISync.test.tsx`
- `bun test src/browser/utils/compaction/contextSwitchCheck.test.ts`
- `make static-check`

## Notes / Risks
- Origin tracking is intentionally **in-memory only**: warnings are action-driven and do not survive reloads between "record" and "observe".
- Any explicit model change that should be eligible for warnings must go through `setWorkspaceModelWithOrigin(..., "user"|"agent")`.

### DIFF
--- a/src/browser/contexts/PolicyContext.test.tsx
+++ b/src/browser/contexts/PolicyContext.test.tsx
@@ -34,11 +34,13 @@ void mock.module("@/browser/contexts/API", () => ({
 import { PolicyProvider, usePolicy } from "./PolicyContext";
 
 const buildBlockedResponse = (reason: string): PolicyGetResponse => ({
+  source: "governor",
   status: { state: "blocked", reason },
   policy: null,
 });
 
 const buildEnforcedResponse = (): PolicyGetResponse => ({
+  source: "governor",
   status: { state: "enforced" },
   policy: {
     policyFormatVersion: "0.1",


ethanndickson:
@codex review

Addressed the 1M-toggle warning regression by allowing same-model checks only when the effective limit changes, and added coverage for the toggle behavior.

ethanndickson:
@codex review

Updated the config/policy refresh path to keep same-model warnings (like 1M toggles) by passing allowSameModel when recomputing existing warnings.

ethanndickson:
@codex review

Cleared stale warnings on zero-token user model switches and added hook coverage.

ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed deferred switch cleanup when tokens are 0, made the agent warning test use the explicit model-change path, and restored the ContextSwitchWarning story via an explicit model switch in play.

ethanndickson:
@codex review

Normalized gateway model strings for explicit switch tracking and added a regression test for gateway-prefixed pending models.

ethanndickson:
@codex review

Latest fixes are up: guarded context-switch warning refresh to avoid policy churn loops, stabilized policy refresh updates, and added regression coverage.

ethanndickson:
@codex review

Addressed flaking warning test and policy reason staleness: refresh guard to avoid re-warn on background model changes, PolicyContext signature includes status.reason, and added PolicyContext regression tests (plus flake rerun).

ethanndickson:
@codex review

- Gate model sync effects to record origin "agent" only on real agentId changes; use "sync" on mount/workspace entry/defaults/config changes.
- Deflake PolicyContext tests (longer waitFor + remove call-order mock) and add WorkspaceModeAISync regression test.

ethanndickson:
@codex review

Pushed follow-ups:
- Normalize gateway vs canonical model comparisons
- Move workspace reset dispatch into useLayoutEffect
- Add aria-label to dismiss button
- Remove dead onModelChange from SlashCommandContext

ethanndickson:
@codex review

Merged origin/main into the branch and resolved the two conflicts in ChatInput/chatCommands (kept SlashCommandContext free of unused callbacks). Static-check is green.

ethanndickson:
@codex review

I pushed a follow-up commit that prevents stale explicit model-change entries from lingering and being consumed after sync-driven away→back transitions. Also adds targeted unit tests for no-op, divergence, and rapid-switch scenarios.

ethanndickson:
@codex review

PR #2141, opened 2026-02-04
🤖 fix: clarify working directory in environment context

ammar-agent:
## Summary

Changes the environment context wording from "You are in a clone of a git repository at X" to "Your working directory is X (a git repository clone)" to make it clearer that the agent is already in that directory.

## Background

The previous wording "You are in a clone of a git repository at /path" was ambiguous—it read as "here's where the repo lives" rather than "you are currently located at this path." This caused agents to unnecessarily `cd` to the path, thinking they needed to navigate there.

## Implementation

Updated the description strings for SSH, Docker, and Devcontainer runtime modes to lead with "Your working directory is" to make the current location explicit.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$0.22`_

### DIFF
--- a/docs/agents/system-prompt.mdx
+++ b/docs/agents/system-prompt.mdx
@@ -86,19 +86,19 @@ function buildEnvironmentContext(workspacePath: string, runtimeType: RuntimeMode
 
     case RUNTIME_MODE.SSH:
       // SSH runtime clones the repository on a remote host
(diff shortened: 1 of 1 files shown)


ammario:
Note this is because of repeated observations of the agent doing nonsensical cds at the beginning of a session.

PR #2143, opened 2026-02-04
🤖 fix: add hard truncation safety net for background bash output

ammar-agent:
## Summary

Add hard truncation as a safety net for bash output that flows through `maybeFilterBashOutputWithSystem1`. This fixes a latent gap where background bash output via `task_await` could grow unbounded when System1 compaction was skipped or failed.

## Background

Foreground bash enforces hard limits during streaming (`BASH_HARD_MAX_LINES=300`, `BASH_MAX_TOTAL_BYTES=16KB`). Background bash output retrieved via `task_await` had no such limits - only System1 compaction as protection. Recent changes added skip paths for System1 compaction (`already_targeted_script`, `exploration_output_small`, `plan_file_in_script`), making it easier to hit this gap and causing compaction loops when large output flooded context.

## Implementation

- Add `truncateBashOutput()` helper in `src/common/utils/truncateBashOutput.ts` that applies the same limits as foreground bash
- Integrate into `maybeFilterBashOutputWithSystem1()` so all bash-family tool output (`bash`, `bash_output`, `task_await`) gets hard truncation as a fallback
- When returning `undefined` (skipping System1), now check if hard truncation was needed and return that instead

## Risks

Low - this is a safety net that only triggers when output exceeds 300 lines or 16KB, which are the same limits already enforced for foreground bash.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-5` • Thinking: `high` • Cost: `$2.64`_


ammar-agent:
@codex review

ammario:
Noticed one of my chats in a compaction loop. I'm not sure if this would fix it but it seems worthwhile anyways.

PR #2145, opened 2026-02-04
🤖 feat: gate Coder runtime on CLI readiness

ethanndickson:
## Summary
- Treat the Coder CLI as a first-class runtime dependency: if the local `coder` CLI is missing, outdated, or not logged in, the Coder runtime option is disabled/hidden up-front with actionable messaging.
- Surface the resolved `coder` binary path (when available) to reduce confusion when multiple CLIs exist on `PATH`.

## Background
Provisioning errors were surfacing late and often boiled down to “CLI not logged in” or “CLI too old”. We want the UI to reflect runtime readiness immediately. Docs/mocks indicate the minimum supported Coder CLI version is **v2.25.0+**.

## Implementation
- Backend: `getCoderInfo()` now:
  - Resolves the `coder` binary path using `command -v coder` (on Windows, executed under Git Bash when available).
  - Checks `coder version --output=json` and returns an `outdated` state (including `binaryPath`) when below `2.25.0`.
  - Runs `coder whoami --output=json` and returns a dedicated `not-logged-in` unavailable reason (not cached so login can be re-checked).
  - Sanitizes CLI errors to prefer the actionable `error: ...` line when present.
- Frontend: Disable the Coder runtime button for `outdated` and `not-logged-in`, and keep it hidden for missing/other errors (no extra messaging when hidden).

## Validation
- `make static-check`
- `bun test src/node/services/coderService.test.ts`

## Risks
- Slight increase in CLI calls (path/version/whoami), mitigated by caching of non-login states; `not-logged-in` is intentionally not cached to allow immediate recovery after `coder login`.

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: $23.52_

### DIFF
--- a/src/browser/utils/runtimeUi.ts
+++ b/src/browser/utils/runtimeUi.ts
@@ -157,7 +157,7 @@ export const RUNTIME_UI = {
 const CODER_RUNTIME_UI: RuntimeUiSpec = {
   ...RUNTIME_UI.ssh,
   label: "Coder",
-  description: "Coder managed workspace via the Coder CLI",
+  description: "Coder workspace via the Coder CLI",
   docsPath: "/runtime/coder",
   Icon: CoderIcon,
 };


ethanndickson:
@codex review

Avoid caching transient `coder whoami` failures so Coder runtime availability can recover without an app restart.

PR #2155, opened 2026-02-04
🤖 feat: hover preview for agent-skill slash commands

ThomasK33:
## Summary

Adds a hover preview for agent-skill slash commands so users can see the *exact* injected skill snapshot (YAML frontmatter + markdown body) by hovering the highlighted command prefix.

## Background

Mux injects agent-skill content via a synthetic "agent skill snapshot" message, but synthetic messages are hidden from the transcript unless `debugLlmRequest` is enabled. This makes it hard to understand what context was included when a slash command was invoked.

## Implementation

- Persist `frontmatterYaml` on synthetic `agentSkillSnapshot` metadata for back-compat-safe rendering.
- In `StreamingMessageAggregator`, associate the most recent synthetic snapshot for `(scope, skillName)` with the following slash-command user message.
- Render the snapshot via a `HoverCard` on the command prefix badge, including:
  - opaque surface background
  - thicker border for visibility in light mode

## Validation

- `make static-check`

## Risks

- `HoverCardContent` defaults were updated to use `bg-modal-bg` for an always-opaque floating surface; this may subtly affect other hover cards that rely on defaults.
- Snapshot association relies on parsing the `<agent-skill>` wrapper; malformed snapshots will simply omit the preview.

---

<details>

# Plan: Hover tooltip preview for agent-skill slash commands

## Context / Why
When a user sends a message beginning with an agent-skill slash command (e.g. `/dev-server-sandbox …`), Mux injects that skill’s content into the LLM request via a **synthetic “agent skill snapshot”** message. Those synthetic messages are hidden from the transcript unless `window.api.debugLlmRequest` is enabled, so users can’t easily see what was injected.

Goal: **Hovering the highlighted slash command prefix in the sent user message should show a rich tooltip (hover card) that renders the injected skill Markdown**, and also shows the skill’s YAML frontmatter.

## Evidence (repo touchpoints)
- Prefix highlighting is implemented in `src/browser/components/Messages/UserMessageContent.tsx` (`CommandPrefixBadge`).
- Skill invocations are represented in user message metadata as `muxMetadata.type === "agent-skill"` (built in `src/browser/components/ChatInput/utils.ts`, types in `src/common/types/message.ts`).
- Backend injects skills via a synthetic message built in `src/node/services/agentSession.ts#materializeAgentSkillSnapshot()` with content like:
  - `'<agent-skill name="…" scope="…">\n{body}\n</agent-skill>'`
  - plus `metadata.synthetic: true` and `metadata.agentSkillSnapshot`.
- The transcript hides synthetic messages by default in `src/browser/utils/messages/StreamingMessageAggregator.ts#getDisplayedMessages()`.

## Recommended approach (use the actual injected snapshot)
Attach the injected snapshot (body + frontmatter) to the **displayed** user message that invoked the skill, and render it in a hover card.

### 1) Persist frontmatter on the snapshot message (backend)
**Why:** the injected snapshot message currently carries the injected *body* (in its text), but not the YAML frontmatter. Storing frontmatter on the snapshot message metadata makes tooltip rendering synchronous and avoids re-reading skills from disk on hover.

**Edits**
  - Extend `MuxMetadata["agentSkillSnapshot"]` to include `frontmatterYaml?: string` (optional for backward compat).
- `src/common/orpc/schemas/message.ts`
  - Extend `agentSkillSnapshot` schema to accept `frontmatterYaml: z.string().optional()`.
- `src/node/services/agentSession.ts` (`materializeAgentSkillSnapshot`)
  - Compute YAML text from `skill.frontmatter` and store it on the synthetic snapshot metadata.

```ts
// src/node/services/agentSession.ts (inside materializeAgentSkillSnapshot)
import YAML from "yaml";

...
const frontmatterYaml = YAML.stringify(skill.frontmatter).trimEnd();

const snapshotMessage = createMuxMessage(snapshotId, "user", snapshotText, {
  timestamp: Date.now(),
  synthetic: true,
  agentSkillSnapshot: {
    skillName: skill.frontmatter.name,
    scope: skill.scope,
    sha256,
    frontmatterYaml,
  },
});
```

**Tests**
- Update `src/node/services/agentSession.agentSkillSnapshot.test.ts` to assert `snapshotMessage.metadata.agentSkillSnapshot.frontmatterYaml` is present and contains at least `name:` + `description:`.

**Net new LoC (product code):** ~20–35

### 2) Surface snapshot content on the displayed invocation message (frontend aggregator)
**Why:** synthetic snapshot messages are hidden, so the UI needs a way to access the snapshot that preceded a particular `/skill` invocation.

**Edits**
  - Extend `DisplayedMessage`’s user variant `agentSkill` shape to optionally include a snapshot:

```ts
agentSkill?: {
  skillName: string;
  scope: AgentSkillScope;
  snapshot?: {
    sha256: string;
    frontmatterYaml?: string;
    body: string; // markdown body that was injected
  };
};
```

- `src/browser/utils/messages/StreamingMessageAggregator.ts`
  - While building `getDisplayedMessages()`, track the most recent synthetic `agentSkillSnapshot` per `(scope, skillName)` as we iterate chronologically.
  - Parse snapshot text to extract the body between `<agent-skill …>` tags.
  - When building a displayed user message with `muxMetadata.type === "agent-skill"`, attach the tracked snapshot object to `DisplayedMessage.agentSkill.snapshot`.
  - Important: only attach when building a message display *fresh* (not when reusing `displayedMessageCache`) so earlier messages keep their original snapshot association.

Parsing helper (defensive):
```ts
function extractAgentSkillSnapshotBody(snapshotText: string): string | null {
  if (!snapshotText.startsWith("<agent-skill")) return null;
  const open = snapshotText.indexOf(">\n");
  const close = snapshotText.lastIndexOf("\n</agent-skill>");
  if (open === -1 || close === -1 || close <= open + 2) return null;
  return snapshotText.slice(open + 2, close);
}
```

**Tests**
- Add a focused unit test in `src/browser/utils/messages/StreamingMessageAggregator.skills.test.ts` (or a new test file) that:
  - Loads a synthetic agentSkillSnapshot message into history (hidden by default but present in `muxMessages`).
  - Loads a following user message with `muxMetadata.type === "agent-skill"`.
  - Asserts the displayed user message includes `agentSkill.snapshot.body` and `frontmatterYaml`.

**Net new LoC (product code):** ~60–120

### 3) Render a hover tooltip on the slash prefix (UI)
**Why:** user asked for hover tooltip that renders Markdown + frontmatter.

**UX choice:** use Radix `HoverCard` (already wrapped in `src/browser/components/ui/hover-card.tsx`) instead of the tiny `Tooltip` primitive, since skill content can be large and should be scrollable.

**Edits**
- `src/browser/components/Messages/UserMessage.tsx`
  - Pass `message.agentSkill` through to `UserMessageContent`.
- `src/browser/components/Messages/UserMessageContent.tsx`
  - Add optional prop `agentSkill`.
  - When `shouldHighlightPrefix` is present **and** `props.agentSkill?.snapshot` exists, wrap `CommandPrefixBadge` with a hover card.

Tooltip content rendering:
- Build a markdown string that includes frontmatter as a fenced `yaml` block, then the injected body:

```ts
const fm = snapshot.frontmatterYaml?.trim();
const frontmatterBlock = fm
  ? `\n\n\`\`\`yaml\n---\n${fm}\n---\n\`\`\`\n\n`
  : "";

const previewMarkdown = `${frontmatterBlock}${snapshot.body}`;
```

- Render with `MarkdownRenderer` inside a scroll container:
  - `className="w-[min(800px,calc(100vw-32px))] max-h-[70vh] overflow-y-auto p-3"`

**Net new LoC (product code):** ~40–80

## Alternative (simpler, but less accurate)
Instead of using the injected snapshot, fetch the current skill on hover via `api.agentSkills.get({ workspaceId, skillName, ... })` and render that.

- Pros: avoids aggregator plumbing.
- Cons: may not match what was actually injected (dedupe/truncation/skill edited after send); async hover fetch; needs loading states.

**Net new LoC (product code):** ~50–90

---

<details>
<summary>Notes / edge cases to handle</summary>

- **Dedupe behavior:** backend may skip inserting a new snapshot if the same `(skillName, sha256)` was recently persisted. The aggregator mapping should fall back to the most recent prior snapshot for that `(scope, skillName)`.
- **Back-compat:** old histories won’t have `frontmatterYaml`; tooltip should still show the injected body and optionally omit the frontmatter block.
- **Performance:** avoid making every message component subscribe to workspace state; keep snapshot association in `DisplayedMessage` so existing memoization stays effective.

</details>

</details>

---

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$6.86`_

### DIFF
--- a/src/node/services/agentSession.agentSkillSnapshot.test.ts
+++ b/src/node/services/agentSession.agentSkillSnapshot.test.ts
@@ -363,6 +363,143 @@ describe("AgentSession.sendMessage (agent skill snapshots)", () => {
     expect(secondSendAppendedIds[0]).toStartWith("user-");
   });
 
+  it("persists a new skill snapshot when frontmatter changes (body unchanged)", async () => {
+    const workspaceId = "ws-test";
+
+    const skillName = "test-skill";
+    const skillBody = "Follow this skill.";
+
+    const { workspacePath } = await createTestWorkspaceWithSkill({
+      skillName,
+      skillBody,
+    });
+
+    const config = {
+      srcDir: "/tmp",
+      getSessionDir: (_workspaceId: string) => "/tmp",
+    } as unknown as Config;
+
+    const messages: MuxMessage[] = [];
+    let nextSeq = 0;
+
+    const appendToHistory = mock((_workspaceId: string, message: MuxMessage) => {
+      message.metadata = { ...(message.metadata ?? {}), historySequence: nextSeq++ };
+      messages.push(message);
+      return Promise.resolve(Ok(undefined));
+    });
+
+    const historyService = {
+      appendToHistory,
+      truncateAfterMessage: mock((_workspaceId: string, _messageId: string) => {
+        void _messageId;
+        return Promise.resolve(Ok(undefined));
+      }),
+      getHistory: mock((_workspaceId: string): Promise<Result<MuxMessage[], string>> => {
+        return Promise.resolve(Ok([...messages]));
+      }),
+    } as unknown as HistoryService;
+
+    const partialService = {
+      commitToHistory: mock((_workspaceId: string) => Promise.resolve(Ok(undefined))),
+    } as unknown as PartialService;
+
+    const aiEmitter = new EventEmitter();
+
+    const workspaceMeta: FrontendWorkspaceMetadata = {
+      id: workspaceId,
+      name: "ws",
+      projectName: "proj",
+      projectPath: workspacePath,
+      namedWorkspacePath: workspacePath,
+      runtimeConfig: { type: "local" },
+    } as unknown as FrontendWorkspaceMetadata;
+
+    const aiService = Object.assign(aiEmitter, {
+      isStreaming: mock((_workspaceId: string) => false),
+      stopStream: mock((_workspaceId: string) => Promise.resolve(Ok(undefined))),
+      getWorkspaceMetadata: mock((_workspaceId: string) => Promise.resolve(Ok(workspaceMeta))),
+      streamMessage: mock((_messages: MuxMessage[]) => {
+        return Promise.resolve(Ok(undefined));
+      }) as unknown as (
+        ...args: Parameters<AIService["streamMessage"]>
+      ) => Promise<Result<void, SendMessageError>>,
+    }) as unknown as AIService;
+
+    const initStateManager = new EventEmitter() as unknown as InitStateManager;
+
+    const backgroundProcessManager = {
+      cleanup: mock((_workspaceId: string) => Promise.resolve()),
+      setMessageQueued: mock((_workspaceId: string, _queued: boolean) => {
+        void _queued;
+      }),
+    } as unknown as BackgroundProcessManager;
+
+    const session = new AgentSession({
+      workspaceId,
+      config,
+      historyService,
+      partialService,
+      aiService,
+      initStateManager,
+      backgroundProcessManager,
+    });
+
+    const baseOptions = {
+      model: "anthropic:claude-3-5-sonnet-latest",
+      agentId: "exec",
+      muxMetadata: {
+        type: "agent-skill",
+        rawCommand: "/test-skill do X",
+        skillName,
+        scope: "project",
+      },
+    };
+
+    const first = await session.sendMessage("do X", baseOptions);
+    expect(first.success).toBe(true);
+    expect(appendToHistory.mock.calls).toHaveLength(2);
+
+    const firstSnapshot = messages[0];
+    expect(firstSnapshot.id).toStartWith("agent-skill-snapshot-");
+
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

ThomasK33:
Good catch — backend snapshot dedupe was hashing only the `<agent-skill>` wrapper + body.

I updated `materializeAgentSkillSnapshot` to include `frontmatterYaml` in the SHA-256 input, so frontmatter-only edits now produce a new snapshot (and the hover preview won’t show stale frontmatter).

Also added a regression test to cover the “body unchanged, frontmatter changed” case.

@codex review

PR #2165, opened 2026-02-05
🤖 fix: avoid replay/live delta duplication on reconnect

ethanndickson:
## Summary
Prevent duplicated streamed deltas (including reasoning) on renderer reconnects, avoiding repeated-prefix UI artifacts (e.g. `EvaluEvaluEvalu...`) and double token counting.

## Background
`workspace.onChat` subscribes to live session events before awaiting `replayHistory()`. If a renderer reconnects mid-stream, live `stream-delta` / `reasoning-delta` events can arrive while replay is in progress and then be emitted again by `replayStream()` with `replay: true`. The frontend aggregator is append-only, so duplicated delivery becomes repeated prefixes.

## Implementation
- **Backend (`src/node/orpc/router.ts`)**
  - Keep “subscribe before replay” semantics (so stream replay + init replay aren’t missed).
  - During replay, buffer *live* stream events and then flush after `caught-up`, skipping any deltas already delivered by replay (dedupe key: `type + messageId + timestamp + delta`).
  - Buffer terminal stream events (`stream-end` / `stream-abort` / `stream-error`) alongside deltas to preserve causal ordering.
  - Track replayed delta keys with a **counted** map (not a Set), so we only drop as many buffered duplicates as were actually replayed.
  - Extracted the replay buffering/dedupe logic into **`src/node/orpc/replayBufferedStreamMessageRelay.ts`** to keep the `onChat` handler minimal.

- **Backend (`src/node/services/streamManager.ts`)**
  - Emit a part before making it visible to `replayStream()` (avoids replay snapshotting an “in-flight” part and then also receiving the later live emit).
  - Make per-stream part timestamps strictly monotonic (so same-millisecond identical chunks don’t collide during replay dedupe).

## Validation
- `make static-check`

## Risks
- During the replay window, live stream events are temporarily buffered (not dropped). In the unlikely case of a legitimate duplicate delta with identical `(type, messageId, timestamp, delta)`, it will be skipped; monotonic timestamps reduce the chance of false positives.

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `medium` • Cost: `$20.78`_

### DIFF
--- a/src/node/orpc/replayBufferedStreamMessageRelay.ts
+++ b/src/node/orpc/replayBufferedStreamMessageRelay.ts
@@ -1,0 +1,108 @@
+import type { WorkspaceChatMessage } from "@/common/orpc/types";
+
+type ReplayBufferedStreamMessage = Extract<
+  WorkspaceChatMessage,
+  {
+    type: "stream-delta" | "reasoning-delta" | "stream-end" | "stream-abort" | "stream-error";
+  }
+>;
+
+type ReplayBufferedDeltaMessage = Extract<
+  ReplayBufferedStreamMessage,
+  { type: "stream-delta" | "reasoning-delta" }
+>;
+
+function isReplayBufferedStreamMessage(
+  message: WorkspaceChatMessage
+): message is ReplayBufferedStreamMessage {
+  return (
+    message.type === "stream-delta" ||
+    message.type === "reasoning-delta" ||
+    message.type === "stream-end" ||
+    message.type === "stream-abort" ||
+    message.type === "stream-error"
+  );
+}
+
+function isReplayBufferedDeltaMessage(
+  message: ReplayBufferedStreamMessage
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

Addressed P1/P2:
- Buffer stream-end/stream-abort/stream-error alongside deltas during replay so terminal events can't overtake buffered deltas.
- Use counted replay keys (Map) + strictly monotonic per-stream part timestamps to avoid dedupe key collisions.

ethanndickson:
@codex review

Follow-up: cleared replay dedupe state after flush to avoid retaining delta text for the lifetime of the subscription.

ethanndickson:
@codex review

Refactor-only: extracted the onChat replay buffering/dedupe logic into a helper in router.ts to keep the handler linear/readable.

PR #2192, opened 2026-02-06
🤖 fix: filter whitespace-only text blocks from interrupted streams

ethanndickson:
## Summary

Fix "messages: text content blocks must contain non-whitespace text" API error that occurs after interrupting a stream during the thinking phase with Opus 4.6.

## Background

When a user interrupts a streaming response mid-thinking, the partial assistant message is saved with:
1. A **reasoning block without a signature** (the signature arrives as a late separate event after all reasoning text — interruption means it never arrives)
2. A **whitespace-only text block** (e.g., `"\n\n"` — the first text delta after thinking)

The UI renders both parts fine (no signature needed for display). But on the next API call, `stripUnsignedAnthropicReasoning` removes the unsigned reasoning, leaving only the whitespace text block. Anthropic rejects this with: "text content blocks must contain non-whitespace text".

The root cause: all existing "empty" checks used `text.length > 0` instead of `text.trim().length > 0`, so `"\n\n"` passed through every filter.

## Implementation

Defense in depth across 4 layers:

1. **Request-time sanitization** (`aiService.ts`) — The main fix. After `convertToModelMessages()`, we now filter whitespace-only text parts from assistant messages and drop messages that become empty. Provider-agnostic, request-only (does not mutate persisted history).

2. **`filterEmptyAssistantMessages`** (`modelMessageTransform.ts`) — Changed `part.text.length > 0` → `part.text.trim().length > 0` so whitespace-only text isn't treated as content.

3. **`stripUnsignedAnthropicReasoning` post-strip filter** (`modelMessageTransform.ts`) — After stripping unsigned reasoning, the filter now checks that remaining text parts aren't whitespace-only (previously only checked array length).

4. **`hasCommitWorthyParts`** (`partialService.ts`) — Prevents whitespace-only partials from being committed to history in the first place.

## Risks

Low risk — all changes are additive filtering. The `.trim()` check is strictly more conservative than the previous `.length > 0` check. Legitimate text content always has non-whitespace characters. The request-time sanitization is provider-agnostic and doesn't mutate persisted history.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `medium`_

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "mux",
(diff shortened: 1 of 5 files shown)


ethanndickson:
@codex review

PR #2200, opened 2026-02-06
🤖 fix: avoid agent_report reminder races during follow-up dispatch

ethanndickson:
## Summary

Prevent `TaskService.handleStreamEnd()` from sending the `agent_report` reminder/fallback when the conversation is expected to continue automatically (compaction follow-up or queued user messages). Also fixes a P1 where a compaction follow-up can legitimately complete with **zero output parts** and previously looked “pending” forever.

Includes a small Storybook interaction hardening for `App/Settings/MCP › ToolSelectorInteraction` (test-only) that was blocking CI.

## Background

When an agent subtask stream ends without calling `agent_report`, Mux sends a reminder forcing the agent to report. That’s correct when the task is truly done, but it’s incorrect (and can race/hijack the next turn) when:

1. **Compaction with follow-up** — after compaction completes, Mux auto-dispatches a pending follow-up message.
2. **Queued user message** — a user message was queued while the agent was streaming and will be sent immediately after stream-end processing.

The P1: a compaction follow-up stream can succeed while producing **no output parts**. Historically, `StreamManager` only rewrote the assistant placeholder in `chat.jsonl` when `parts.length > 0`, leaving an ambiguous `parts: []` placeholder that looked identical to “still streaming”. The compaction guard used “assistant after the summary has output parts” as its completion signal, so it could keep treating the follow-up as incoming forever and suppress reminders indefinitely.

## Implementation

- **Race-proof queued-message guard**: snapshot whether there were queued user messages *before* any await (avoids racing `AgentSession` draining the queue concurrently on `stream-end`).
- **Robust compaction follow-up detection** (`isCompactionStreamWithFollowUp`):
  - find the most recent compaction summary with `pendingFollowUp` in history (don’t treat older summaries as active)
  - treat the follow-up as still incoming only until we see an assistant message after the summary that represents a completed stream
  - handle empty-output completions via `metadata.duration` (durable completion marker)
- **Durable empty-stream completion** (`StreamManager`): on successful stream completion, always delete `partial.json` and always update the assistant placeholder in `chat.jsonl` with stream-end metadata (including `duration`), even if `parts` is empty.
- **Storybook**: stabilize `ToolSelectorInteraction` by scoping queries to the open dialog and re-querying buttons inside `waitFor` (avoids stale element refs / timing flake).

## Validation

- `make static-check`
- `make storybook-build && make test-storybook`

## Risks

Low. The backend changes make stream completion state more explicit/durable and reduce reminder races. The Storybook change is test-only.

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `xhigh` • Cost: `$58.79`_

### DIFF
--- a/src/common/orpc/schemas.ts
+++ b/src/common/orpc/schemas.ts
@@ -138,6 +138,7 @@ export {
   ErrorEventSchema,
   LanguageModelV2UsageSchema,
   QueuedMessageChangedEventSchema,
+  FollowUpDispatchFailedEventSchema,
   ReasoningDeltaEventSchema,
   ReasoningEndEventSchema,
   RestoreToInputEventSchema,
--- a/src/common/orpc/schemas/stream.ts
+++ b/src/common/orpc/schemas/stream.ts
@@ -397,6 +397,12 @@ export const RestoreToInputEventSchema = z.object({
   reviews: z.array(ReviewNoteDataSchema).optional(),
 });
 
+export const FollowUpDispatchFailedEventSchema = z.object({
+  type: z.literal("follow-up-dispatch-failed"),
+  workspaceId: z.string(),
+  reason: z.enum(["queued-message", "compaction-follow-up"]),
+  error: z.string(),
+});
 // All streaming events now have a `type` field for O(1) discriminated union lookup.
 // MuxMessages (user/assistant chat messages) are emitted with type: "message"
 // when loading from history or sending new messages.
@@ -426,6 +432,7 @@ export const WorkspaceChatMessageSchema = z.discriminatedUnion("type", [
   SessionUsageDeltaEventSchema,
   QueuedMessageChangedEventSchema,
   RestoreToInputEventSchema,
+  FollowUpDispatchFailedEventSchema,
   // Idle compaction notification
   IdleCompactionNeededEventSchema,
   // Runtime status events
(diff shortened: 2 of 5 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed P1: `isCompactionStreamWithFollowUp()` no longer treats historical compaction summaries as active. It now finds the most recent `compaction-summary` with `pendingFollowUp` and only suppresses the agent_report reminder until an assistant message *after* the summary has output parts (i.e. the follow-up response is written to history).

Added regression tests covering:
- historical summary should NOT suppress reminders
- pending follow-up (summary + follow-up user + empty assistant placeholder) should suppress

ethanndickson:
@codex review

Addressed the P1: stream completion is now durable even when a compaction follow-up produces zero parts (StreamManager always updates chat.jsonl + deletes partial.json), and TaskService now treats empty follow-up completion as no longer pending via metadata.duration.

Also stabilized the Storybook ToolSelectorInteraction play-test.

ethanndickson:
@codex review

Addressed the queued-dispatch reminder suppression issue (deferred verification using isStreamStarting), merged origin/main, resolved the Storybook conflict, and confirmed `make static-check` passes.

ethanndickson:
@codex review

Implemented event-driven follow-up dispatch failure handling (replaces timer-based verifier) and moved AgentSession.streamStarting earlier to avoid the queue-drain gap.

ethanndickson:
This is a PITA. Not valuable enough

PR #2213, opened 2026-02-06
🤖 fix: hide Connect (Browser) Codex OAuth button on remote mux server

ibetitsmike:
## Summary

Hides the **Connect (Browser)** button in the Codex OAuth section when the user is running mux server on a remote host (not localhost).

## Background

When using mux server remotely, the "Connect (Browser)" button opens a popup to the OAuth verify URL. While this technically works (it uses the device flow, not a redirect), the popup UX can be unreliable through proxied environments. The "Connect (Device)" flow—which shows the device code inline for manual entry—remains available as a more reliable alternative for remote users.

## Implementation

- Added an `isRemoteServer` computed variable in `ProvidersSection.tsx` that's `true` when:
  - Not running in Electron desktop mode (`!isDesktop`)
  - AND `window.location.hostname` is not `localhost`, `127.0.0.1`, or `::1`
- Wrapped the "Connect (Browser)" `<Button>` in `{!isRemoteServer && (...)}` to conditionally hide it

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `high` • Cost: `$-`_

### DIFF
--- a/src/browser/components/Settings/sections/ProvidersSection.tsx
+++ b/src/browser/components/Settings/sections/ProvidersSection.tsx
@@ -226,6 +226,12 @@ export function ProvidersSection() {
 
   const isDesktop = !!window.api;
 
+  // The "Connect (Browser)" OAuth flow requires a redirect back to this origin,
+  // which only works when the host is the user's local machine. On a remote mux
+  // server the redirect would land on the server, not the user's browser.
+  const isRemoteServer =
+    !isDesktop && !["localhost", "127.0.0.1", "::1"].includes(window.location.hostname);
+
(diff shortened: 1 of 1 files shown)


ibetitsmike:
Addressing the Codex review: while `startCodexOauthBrowserConnect` does use the device flow in browser mode (so no redirect-based concern), the user explicitly requested hiding this option for remote mux server hosts. The "Connect (Device)" flow remains available and works identically — just without the auto-popup convenience.

Resolving as intentional product decision per user request.

ibetitsmike:
@codex review

PR #2216, opened 2026-02-06
🤖 fix: prevent layout flash when bash tool output finishes streaming

ammar-agent:
## Summary

Fixes the layout flash visible at the end of a streaming bash tool output, when it transitions from the live streaming view to the completed result view.

## Background

The bash tool output component (`BashToolCall.tsx`) rendered output in **two completely separate DOM subtrees** at different React child positions — one for live/streaming output and another for completed output. When a bash command finished, React unmounted the entire live output tree and mounted a fresh completed output tree, causing:

- **Scroll position reset** — the new `<pre>` element starts at `scrollTop=0`, so output that was scrolled to the bottom jumped to the top
- **Full DOM subtree swap** — unmount + mount instead of reconciliation (different child positions, different wrapper structure)
- **Different wrapping structure** — `div.relative` wrapper was present in streaming but absent in completed

## Implementation

Unified both output branches into a single `DetailSection`/`DetailContent` DOM tree. The same DOM elements stay mounted across the streaming→completed transition — only the text content and conditional classNames swap. This lets React reconcile in-place, preserving scroll position and preventing layout flash.

The error, truncation-info, and background-process sections remain as separate conditional blocks since they only appear on completion (no flash risk — they weren't previously visible).

## Risks

Low — this is a rendering-only change to a single component. The same content is shown in both states; only the DOM structure was unified. Background bash, filtering overlay, truncation notices, and error display are all preserved.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$31.96`_

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "mux",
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #2218, opened 2026-02-06
🤖 feat: show 1M context toggle under both context usage bars

ammar-agent:
## Summary

Adds a per-model 1M context toggle neatly under both context usage bars (sidebar and chat input popover). State is synced across all instances via the existing `ProviderOptionsContext`.

## Background

The 1M context toggle for Anthropic models was previously only accessible in the Settings > Models table. This puts it where users actually look at context usage — under both context bars — making it more discoverable and quick to toggle.

## Implementation

- **`Toggle1MContext`** — New shared component that renders a compact 1M badge + label. Only renders when the model supports 1M context (`supports1MContext`). Reads/writes state via `useProviderOptions()`, so all instances stay synced automatically.
- **`ContextUsageBar`** (sidebar) — Accepts optional `model` prop, renders toggle below the meter.
- **`ContextUsageIndicatorButton`** (chat input popover) — Accepts optional `model` prop, renders toggle in the `AutoCompactSettings` panel between the meter and idle compaction.
- **`CostsTab` / `ChatInput`** — Pass the current model ID through to their respective context bars.

The toggle in the Settings > Models table remains unchanged — all three locations share the same persisted per-model state.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$24.90`_

### DIFF
--- a/src/browser/components/Settings/sections/ModelRow.tsx
+++ b/src/browser/components/Settings/sections/ModelRow.tsx
@@ -104,6 +104,55 @@ function ModelTooltipContent(props: {
   );
 }
 
+/**
+ * Inline toggle that slides between the model's base context window and 1M.
+ * Renders as a compact pill: clicking toggles the state, with the active
+ * end highlighted in accent.
+ */
+function ContextWindowSlider(props: {
+  baseTokens: number;
+  enabled: boolean;
+  onToggle: () => void;
+}) {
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

PR #2229, opened 2026-02-06
🤖 fix: fall back to API key when Codex OAuth not connected for OpenAI models

ammar-agent:
## Summary

When a user has an OpenAI API key but hasn't connected Codex OAuth, models in `CODEX_OAUTH_REQUIRED_MODELS` (e.g. `gpt-5.1-codex`) immediately errored with `oauth_not_connected` instead of falling back to the API key. This was especially visible during compaction: if the user's preferred compaction model was a codex model, `/compact` would fail while normal chat (using a non-codex workspace model) worked fine.

## Background

The OpenAI provider block in `createModel()` had an early-exit gate:

```typescript
if (codexOauthRequired && !storedCodexOauth) {
  return Err({ type: "oauth_not_connected", provider: providerName });
}
```

This checked OAuth availability **before** resolving credentials, so it never knew the user had an API key. The downstream `shouldRouteThroughCodexOauth` logic already handles the fallback correctly — when `!storedCodexOauth` it returns `false`, routing through the API key — but the early-exit prevented us from ever reaching that code.

The root cause of the "compaction breaks but chat works" symptom: compaction can use a **different model** than chat (via Settings → Models → Compaction Model or `agentAiDefaults.compact.modelString`). If that model was a codex model, it hit the OAuth gate; the workspace chat model (e.g. `gpt-4o`) didn't.

## Implementation

1. **Move `resolveProviderCredentials` before the OAuth check** — credentials are now resolved first so the gate has full visibility.

2. **Add `!creds.isConfigured` to the early-exit condition** — only block when the user has *neither* OAuth nor an API key. If either is available, proceed.

3. **Updated existing test** — the "both missing" test now properly clears `OPENAI_API_KEY` from env (it was previously passing by accident since the old gate fired before credential resolution).

4. **New test** — verifies that codex-required models succeed when OAuth is missing but an API key is present.

## Risks

Low. The change narrows a blocking gate — requests that previously errored now proceed to OpenAI's API with the user's API key. If a model truly requires OAuth-only access, OpenAI's API will return a clear error at that point rather than silently failing.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `max` • Cost: `$7.38`_

### DIFF
--- a/src/common/constants/codexOAuth.ts
+++ b/src/common/constants/codexOAuth.ts
@@ -112,7 +112,12 @@ export const CODEX_OAUTH_ALLOWED_MODELS = new Set<string>([
 ]);
 
 /**
- * Models that *require* Codex OAuth (i.e. cannot fall back to OpenAI API keys).
+ * Models that *prefer* Codex OAuth routing.
+ *
+ * These models are initially gated behind OAuth but eventually become available
+ * via regular API keys.  When the user has OAuth connected, we route through it;
+ * otherwise we fall back to their API key and let OpenAI decide whether the
+ * model is accessible.
  */
 export const CODEX_OAUTH_REQUIRED_MODELS = new Set<string>([
   "gpt-5.1-codex-max",


ammar-agent:
@codex review

Addressed the review comment by updating the doc comment on `CODEX_OAUTH_REQUIRED_MODELS` to explain *why* the hard gate was removed: these models are initially OAuth-gated but eventually graduate to regular API key access. Hard-blocking users who have a valid API key gives worse UX than letting the request through and letting OpenAI's API decide accessibility.

ammario (approved):

PR #2231, opened 2026-02-06
update README

mattvollmer:
- add gif hero
- move screenshots into grid


ammario:
Nice, looks way better

PR #2247, opened 2026-02-07
🤖 fix: allow both agentId and subagent_type in task tool calls

ibetitsmike:
## Summary

GPT-class models consistently send **both** `agentId` and `subagent_type` when calling the `task` tool (e.g., `{"agentId": "explore", "subagent_type": "explore", ...}`). The Zod `superRefine` validation rejected this with _"Provide only one of agentId or subagent_type (not both)"_, causing models to enter an infinite retry loop unable to spawn subtasks.

## Background

- `subagent_type` is a deprecated alias for `agentId` kept for backward compatibility
- The downstream handler in `task.ts` already gracefully resolves both-provided by preferring `agentId`
- The strict rejection in the schema served no protective purpose — it only blocked GPT models that redundantly send both fields

## Implementation

Removed the `hasAgentId && hasSubagentType` rejection block from the `superRefine` in `TaskToolAgentArgsSchema` (`src/common/utils/tools/toolDefinitions.ts`). The "at least one required" check remains.

Net: −9 lines.

## Risks

Minimal — the handler already had correct fallback logic for both-provided. No tests asserted the removed validation.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$0.13`_

### DIFF
--- a/src/common/utils/tools/toolDefinitions.ts
+++ b/src/common/utils/tools/toolDefinitions.ts
@@ -156,6 +156,17 @@ const TaskToolAgentArgsSchema = z
       });
       return;
     }
+
+    // GPT models often send both fields with identical values — allow that.
+    // Only reject when they conflict, since the handler silently prefers agentId.
+    if (hasAgentId && hasSubagentType && args.agentId !== args.subagent_type) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: "agentId and subagent_type must match when both are provided",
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

Addressed the P2 feedback: now the schema allows both `agentId` and `subagent_type` when they match (GPT's common pattern), but rejects when they conflict — preventing the silent wrong-agent issue.

PR #2267, opened 2026-02-08
🤖 fix: graceful fallback when ssh-agent signing module is unavailable

ammar-agent:
## Summary

When the `@coder/mux-md-client/ssh-agent` dynamic import fails at runtime, the signing service now degrades gracefully — falling back to disk key signing instead of surfacing a cryptic module resolution error.

## Background

PR #2265 lazy-loaded the ssh-agent import to prevent the server from crashing at startup. But if a user actually triggers ssh-agent signing while the module is unavailable (e.g. bun resolved to a version missing the export), they'd get a raw `Package subpath './ssh-agent' is not defined by "exports"` stack trace. This PR minimizes the blast radius of that failure.

## Implementation

Three layers of defense, all in `signingService.ts`:

1. **`isSshAgentModuleAvailable()`** — Probes the import once, caches the result. If the module is missing, logs a warning and returns `false`.

2. **`trySelectSshAgentKey()`** — Early-returns `{ selection: null, error: null }` when the probe fails. This means `loadSigningKey()` never selects an ssh-agent key it can't use, and falls through to disk keys (`~/.ssh/id_ed25519`, etc.) automatically. Users with disk keys will never notice.

3. **`signMessage()` catch** — Final safety net. Wraps the dynamic import in `.catch()` and throws a clear user-facing error ("SSH agent signing is unavailable — try updating mux or falling back to disk key signing") instead of the raw module resolution stack trace.

## Validation

- `make static-check` passes
- All 14 signing service tests pass (including both ssh-agent tests)

---

_Generated with `mux` · Model: `anthropic:claude-opus-4-6` · Thinking: `xhigh` · Cost: `$N/A`_

### DIFF
--- a/src/node/services/signingService.ts
+++ b/src/node/services/signingService.ts
@@ -78,6 +78,29 @@ const SUPPORTED_AGENT_KEY_TYPES = new Set([
   "ecdsa-sha2-nistp521",
 ]);
 
+/**
+ * Probe whether the @coder/mux-md-client/ssh-agent subpath is importable.
+ * Caches the result so we only pay the cost once. If the module is missing
+ * (e.g. bun resolved to a version without the export), ssh-agent signing is
+ * silently unavailable and the service falls back to disk keys.
+ */
+let sshAgentModuleAvailable: boolean | null = null;
+async function isSshAgentModuleAvailable(): Promise<boolean> {
+  if (sshAgentModuleAvailable != null) return sshAgentModuleAvailable;
+  try {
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

Addressed: when `MUX_SIGNING_PUBLIC_KEY` or `MUX_SIGNING_KEY_FINGERPRINT` is set and the ssh-agent module is unavailable, we now return an error instead of silently falling back to disk keys. The silent fallback only applies when no override is set.

PR #2290, opened 2026-02-09
🤖 fix: generate optional TS properties for Zod .default() fields in code_execution

ethanndickson:
## Summary

Fields with `.default()` in Zod tool schemas were generating **required** TypeScript properties in `code_execution`'s sandbox types, causing agents to get type errors when omitting fields like `run_in_background` on `mux.bash()`.

## Background

When an agent uses `mux.bash()` inside `code_execution`, omitting `run_in_background` causes a TypeScript compile error — even though at runtime Zod fills in the default. The standalone `bash` tool treats the same field as optional (the orchestrator applies the JSON Schema `"default"` value), so the inconsistency is confusing.

### Root cause

The auto-generation pipeline in `typeGenerator.ts` is:

```
Zod schema (.default(false))
  → z.toJSONSchema()            // keeps field in "required" + adds "default" metadata
  → json-schema-to-typescript   // maps "required" → non-optional TS property
  → TypeScript interface         // run_in_background: boolean (required!)
```

### Why the libraries don't handle this for us

Both libraries are spec-correct — the problem is JSON Schema can't represent Zod's richer type semantics:

- **`z.toJSONSchema()`**: In JSON Schema, `required` and `default` are orthogonal concepts. A field can be both required *and* have a default — `required` means "must be present," while `default` is purely advisory metadata that validators aren't obligated to apply. So Zod is correct to keep `.default()` fields in the `required` array.
- **`json-schema-to-typescript`**: It maps `required` → required TS property, not-required → optional. There's no spec basis for treating `"default"` as "actually optional."

The gap is Zod-specific: `z.boolean().default(false)` has *input* type `boolean | undefined` and *output* type `boolean`. The caller can omit it and `safeParse()` fills it in. But JSON Schema has no input-vs-output distinction, so `z.toJSONSchema()` can only serialize the output type (`boolean`, required) with `"default"` stapled on as a hint.

| Layer | View of `.default(false)` |
|-------|--------------------------|
| **Zod runtime** | Input optional, output guaranteed |
| **JSON Schema** | Required field with advisory metadata |
| **json-schema-to-typescript** | Required TypeScript property |
| **What we need** | Optional TypeScript property (matching Zod input type) |

### Affected fields (4 total)

| Tool | Field | Zod |
|------|-------|-----|
| `bash` | `run_in_background` | `.boolean().default(false)` |
| `task` | `run_in_background` | `.boolean().default(false)` |
| `task_await` | `timeout_secs` | `.number().min(0).nullish().default(600)` |
| `task_apply_git_patch` | `three_way` | `.boolean().nullish().default(true)` |

### Raw JSON Schema tools (MCP)

This post-processing **must only apply to JSON Schema produced from Zod**.

For tools that already provide **raw JSON Schema** (e.g. MCP tools via `tool.parameters`), `default` is advisory metadata and does **not** imply optional input. In those cases the `required` list is authoritative and must not be rewritten.

## Implementation

- Added `removeDefaultsFromRequired()` in `typeGenerator.ts` — strips fields with `"default"` from the JSON Schema `required` array **only for Zod-derived schemas**, so defaulted fields generate `field?: type` instead of `field: type` (matching the Zod input type).
- Ensured raw JSON Schema tool parameters are passed through unchanged (no `required` rewriting).
- Narrowed `getInputJsonSchema()` return type from `object` → `Record<string, unknown>` to satisfy the helper's type signature.

## Validation

- New test: `"generates optional properties for fields with .default()"` — verifies `.default()` fields get `?:` while non-default fields stay required
- New test: `"does not strip defaults from raw JSON Schema (MCP tools)"` — ensures raw JSON Schema `required` fields stay required even if they have a `default`
- All **18** `typeGenerator` tests pass
- Typecheck clean (both renderer and main configs)

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$3.07`_

### DIFF
--- a/src/node/services/ptc/typeGenerator.test.ts
+++ b/src/node/services/ptc/typeGenerator.test.ts
@@ -242,6 +242,29 @@ describe("generateMuxTypes", () => {
     expect(types).toContain("owner: string");
   });
 
+  test("does not strip defaults from raw JSON Schema (MCP tools)", async () => {
+    // MCP tools come with JSON Schema, not Zod. JSON Schema `default` is
+    // advisory metadata and does not make the property optional.
+    const mcpTool = {
+      description: "Mock MCP tool with defaults",
+      parameters: {
+        type: "object",
+        properties: {
+          repo: { type: "string", default: "my-repo" },
+          owner: { type: "string" },
+        },
+        required: ["repo", "owner"],
+      },
+      execute: () => Promise.resolve({ content: [] }),
+    } as unknown as Tool;
+
+    const types = await generateMuxTypes({ mcp__github__list_repos: mcpTool });
+
+    // `repo` must remain required even though it has a `default`
+    expect(types).toContain("repo: string");
+    expect(types).not.toContain("repo?:");
+    expect(types).toContain("owner: string");
+  });
   test("handles empty tool set", async () => {
     const types = await generateMuxTypes({});
 
--- a/src/node/services/ptc/typeGenerator.ts
+++ b/src/node/services/ptc/typeGenerator.ts
@@ -110,6 +110,10 @@ function pascalCase(str: string): string {
  * "required" — JSON Schema treats "required" and "default" as orthogonal
  * concepts. json-schema-to-typescript then faithfully maps "required" →
  * non-optional TS property, producing types that don't match runtime behavior.
+ *
+ * Important: Only apply this to JSON Schema that came from Zod. For raw JSON
+ * Schema (e.g. MCP tool parameters), `default` is advisory metadata and does
+ * not imply optional input.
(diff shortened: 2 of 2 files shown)


ethanndickson:
@codex review

Addressed the issue where `removeDefaultsFromRequired()` was applied to raw JSON Schema (MCP tool parameters):

- Default-stripping now runs only on JSON Schema generated from Zod (`inputSchema` branch), not on externally supplied JSON Schema (`parameters`).
- Added regression test ensuring raw JSON Schema fields stay required even when they have a `default`.

Thanks!

PR #2292, opened 2026-02-09
🤖 fix: align per-agent AI defaults precedence on agent switch

ThomasK33:
Summary
Align model/reasoning precedence so explicit per-agent Settings defaults are authoritative during UI agent switches, matching backend task creation behavior.

Background
Switching agents in the workspace UI previously let workspace `aiSettingsByAgent` values override explicit Settings → Tasks per-agent defaults. Backend sub-agent creation already preferred `agentAiDefaults`, so the two paths diverged.

Implementation
- Updated `WorkspaceModeAISync` precedence to prefer configured per-agent defaults first, then workspace per-agent values, then descriptor defaults, then existing workspace values.
- Added an inline precedence comment to document why this order is intentional and how `Inherit` falls through.
- Added frontend regression tests for:
  - configured defaults outranking workspace-by-agent values
  - inherit falling through to workspace-by-agent values
- Added backend guard test to ensure `agentAiDefaults` outrank conflicting same-agent `aiSettingsByAgent` values.

Validation
- `make static-check`
- `bun test src/browser/components/WorkspaceModeAISync.test.tsx`
- `bun test src/node/services/taskService.test.ts -t "agentAiDefaults outrank workspace aiSettingsByAgent"`

Risks
Low-to-moderate behavior risk in model selection precedence on agent switch. Mitigated by targeted frontend/backend regression tests that lock the expected priority order.

---

<details>

# Plan: Align model switching with explicit per-agent Settings overrides

## Context / Why
The current behavior is inconsistent with the requested expectation:

- **Expected:** if a model/reasoning value is configured in **Settings → Tasks** for an agent, switching to that agent should always apply those configured values.
- **Expected:** when a setting is **Inherit**, the resolver should fall through to non-configured behavior.

Today, frontend agent switching (`WorkspaceModeAISync`) lets workspace-scoped per-agent values win over global `agentAiDefaults`, while backend sub-agent creation (`TaskService.create`) already gives `agentAiDefaults` highest precedence. This mismatch creates surprising behavior and makes UI mode switches differ from spawned task behavior.

## Evidence
- `src/browser/components/WorkspaceModeAISync.tsx` (lines 71–101):
  - Computes `configuredDefaults`, but resolves `candidateModel`/`candidateThinking` from `workspaceByAgent` **before** defaults.
- `src/browser/components/Settings/sections/TasksSection.tsx` (lines 511–533):
  - `Inherit` is represented by deleting `modelString`/`thinkingLevel` from `agentAiDefaults`.
- `src/node/services/taskService.ts` (lines 536–571):
  - Sub-agent resolution uses `subagentDefaults` (`agentAiDefaults`) before inherited/workspace values.
- Existing tests:
  - `src/browser/components/WorkspaceModeAISync.test.tsx` currently checks origin tagging only, not precedence.
  - `src/node/services/taskService.test.ts` already has related precedence tests but not an explicit “global agent defaults beat workspace override for same agent” guard.

These files are sufficient because they are the exact read and write paths for model/reasoning resolution during agent switching and task spawning.

## Recommended approach A (targeted precedence alignment, **net +55 LoC**)

1. **Update frontend precedence in `WorkspaceModeAISync`**
   - File: `src/browser/components/WorkspaceModeAISync.tsx`
   - Symbols: `WorkspaceModeAISync` effect, `candidateModel`, `candidateThinking`
   - Change resolution to prefer explicit configured defaults first, while preserving existing fallback behavior for `Inherit`.

   ```ts
   // ...
   const configuredModelDefault = configuredDefaults?.modelString;
   const configuredThinkingDefault = configuredDefaults?.thinkingLevel;
   const descriptorModelDefault = descriptorDefaults?.model;
   const descriptorThinkingDefault = descriptorDefaults?.thinkingLevel;

   const candidateModel =
     configuredModelDefault ??
     fallbackIds.map((id) => workspaceByAgent[id]?.model).find((entry) => entry !== undefined) ??
     descriptorModelDefault ??
     existingModel;

   const candidateThinking =
     configuredThinkingDefault ??
     fallbackIds
       .map((id) => workspaceByAgent[id]?.thinkingLevel)
       .find((entry) => entry !== undefined) ??
     descriptorThinkingDefault ??
     existingThinking ??
     "off";
   // ...
   ```

   Why this shape:
   - Keeps “inherit” as fallback (no explicit configured value).
   - Makes explicit Settings override deterministic and consistent with backend behavior.
   - Avoids accidental behavior change where descriptor defaults would outrank workspace overrides.

2. **Add frontend regression tests for precedence**
   - File: `src/browser/components/WorkspaceModeAISync.test.tsx`
   - Add tests:
     - explicit `AGENT_AI_DEFAULTS_KEY[agentId]` model/thinking override beats `workspaceByAgent[agentId]`.
     - inherit (no explicit configured value) still falls through to workspace-by-agent value.

   ```ts
   test("prefers configured agent defaults over workspace-by-agent overrides", async () => {
     // seed AGENT_AI_DEFAULTS_KEY.exec + workspace aiSettingsByAgent.exec with conflicting values
     // render sync for exec
     // assert final model/thinking match configured defaults
   });
   ```

3. **Add backend guard test to keep precedence aligned across layers**
   - File: `src/node/services/taskService.test.ts`
   - Add a focused test where both are present for the same agent:
     - `agentAiDefaults.explore` set
     - parent workspace `aiSettingsByAgent.explore` set to conflicting values
     - assert child task uses `agentAiDefaults.explore`.

   ```ts
   test("agentAiDefaults outrank workspace aiSettingsByAgent for same agent", async () => {
     // arrange conflicting settings
     // create task(agentType: "explore")
     // expect sendMessage model/thinking === agentAiDefaults.explore
   });
   ```

4. **Document precedence inline where resolved**
   - File: `src/browser/components/WorkspaceModeAISync.tsx`
   - Add a brief comment documenting final priority order to prevent future regressions.

## Alternative approach B (shared resolver utility, **net +120 LoC**)
Create a shared resolver utility in `src/common/` used by both `WorkspaceModeAISync` and `TaskService`.

Pros: single source of truth for precedence.
Cons: broader refactor risk, more call-site plumbing, higher regression surface.

## Validation plan
- `bun test src/browser/components/WorkspaceModeAISync.test.tsx`
- `bun test src/node/services/taskService.test.ts -t "agentAiDefaults outrank workspace aiSettingsByAgent"`
- Optional confidence pass: `make test` (if CI/runtime budget allows)

<details>
<summary>Why this plan should match your expectation</summary>

This makes explicit per-agent Settings authoritative during UI agent switches (the behavior you expected), while preserving the semantics of `Inherit` as a true fallback path. It also adds tests in both frontend and backend layers so the two paths stay aligned over time.
</details>

</details>

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$1.24`_

### DIFF
--- a/src/browser/components/WorkspaceModeAISync.test.tsx
+++ b/src/browser/components/WorkspaceModeAISync.test.tsx
@@ -5,7 +5,12 @@ import { cleanup, render, waitFor } from "@testing-library/react";
 import { AgentProvider } from "@/browser/contexts/AgentContext";
 import { consumeWorkspaceModelChange } from "@/browser/utils/modelChange";
 import { readPersistedState, updatePersistedState } from "@/browser/hooks/usePersistedState";
-import { AGENT_AI_DEFAULTS_KEY, getModelKey } from "@/common/constants/storage";
+import {
+  AGENT_AI_DEFAULTS_KEY,
+  getModelKey,
+  getThinkingLevelKey,
+  getWorkspaceAISettingsByAgentKey,
+} from "@/common/constants/storage";
 
 import { WorkspaceModeAISync } from "./WorkspaceModeAISync";
 
@@ -19,6 +24,31 @@ function nextWorkspaceId(): string {
 const noop = () => {
   // intentional noop for tests
 };
+
+function SyncHarness(props: { workspaceId: string; agentId: string }) {
+  return (
+    <AgentProvider
+      value={{
+        agentId: props.agentId,
+        setAgentId: noop,
+        currentAgent: undefined,
+        agents: [],
+        loaded: true,
+        loadFailed: false,
+        refresh: () => Promise.resolve(),
+        refreshing: false,
+        disableWorkspaceAgents: false,
+        setDisableWorkspaceAgents: noop,
+      }}
+    >
+      <WorkspaceModeAISync workspaceId={props.workspaceId} />
+    </AgentProvider>
+  );
+}
+
+function renderSync(props: { workspaceId: string; agentId: string }) {
+  return render(<SyncHarness workspaceId={props.workspaceId} agentId={props.agentId} />);
+}
 
 describe("WorkspaceModeAISync", () => {
   beforeEach(() => {
@@ -49,28 +79,7 @@ describe("WorkspaceModeAISync", () => {
     // Start with a different model so the mount sync performs an update.
     updatePersistedState(getModelKey(workspaceId), "some-legacy-model");
 
-    function Harness(props: { agentId: string }) {
-      return (
-        <AgentProvider
-          value={{
-            agentId: props.agentId,
-            setAgentId: noop,
-            currentAgent: undefined,
-            agents: [],
-            loaded: true,
-            loadFailed: false,
-            refresh: () => Promise.resolve(),
-            refreshing: false,
-            disableWorkspaceAgents: false,
-            setDisableWorkspaceAgents: noop,
-          }}
-        >
-          <WorkspaceModeAISync workspaceId={workspaceId} />
-        </AgentProvider>
-      );
-    }
-
-    const { rerender } = render(<Harness agentId="exec" />);
+    const { rerender } = renderSync({ workspaceId, agentId: "exec" });
 
     // Mount sync should update the model but NOT record an explicit change entry.
     await waitFor(() => {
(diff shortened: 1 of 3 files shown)


ThomasK33:
@codex review

Addressed the derived-agent precedence issue by ensuring inherited base `agentAiDefaults` are only considered when the selected agent does not have a workspace override, and added a frontend regression test for that scenario.

ThomasK33:
@codex review

Follow-up addressed: base defaults now only inherit when the selected agent has no Settings entry at all, and added a regression test for partial selected-agent settings entries.

PR #2254, opened 2026-02-07
🤖 feat: preserve append-only compaction history boundaries

ThomasK33:
## Summary
Preserve full `chat.jsonl` history during compaction by switching to append-only compaction boundaries, and slice provider request context from the latest boundary onward.

## Background
Previous compaction behavior destructively cleared history and replaced it with a summary message. That made timeline/audit use cases harder and prevented richer transcript rendering over multiple compaction epochs.

## Implementation
- Added durable compaction boundary metadata fields and schema support.
- Updated compaction handling to append compaction summaries (no destructive clear-history reset).
- Added latest-boundary utilities and request-time slicing so provider payloads include only the latest compacted epoch + subsequent turns.
- Updated post-compaction insertion logic to target the latest boundary.
- Added transcript boundary row rendering in the browser message pipeline.
- Updated retry/interruption logic to ignore decorative compaction-boundary rows.
- Added Approach B TODO breadcrumbs near boundary-scan/append callsites for future sidecar-index optimization.

## Validation
- `make static-check`
- Targeted tests for compaction handler, boundary slicing utilities, model message transform, streaming aggregator, message renderer, retry eligibility, and message utils.

## Risks
- Legacy histories without explicit `compactionBoundary` metadata intentionally fall back to full-history request behavior.
- UI boundary rows are decorative; wording/visual hierarchy may need product iteration.

---

<details>
<summary>📋 Implementation Plan</summary>

# Plan: Preserve full `chat.jsonl` history with append-only compaction blocks

## Context / Why
We want to stop destructive transcript compaction. Instead of clearing history and replacing it with one summary message, we should append a compaction block (summary boundary) and keep all historical turns in `chat.jsonl`. Provider requests should then include only messages from the **latest compaction boundary onward**. This preserves auditability and enables richer timeline UI while keeping token usage behavior close to today.

Primary user impact:
- Full chat history remains recoverable and visualizable.
- Model context remains bounded to summarized epoch + new turns.
- Existing compaction UX remains familiar (summary + continued chat), but storage semantics become append-only.

## Evidence
- Explore report `47a3ccccbc` (transcript storage):
  - `src/node/services/compactionHandler.ts` currently calls `historyService.clearHistory(...)` then appends summary.
  - `src/node/services/historyService.ts` persists one JSON object per line in `chat.jsonl` and skips malformed lines on read.
- Explore report `7cb6a0e8d5` (provider request assembly):
[...]

PR #2294, opened 2026-02-09
🤖 perf: optimize chat.jsonl access patterns for append-only compaction

ammar-agent:
## Summary

Replace the array-returning `getFullHistory()` API with a chunked callback-based `iterateFullHistory()` that supports forward/backward direction and early exit. Convert all remaining call sites to use targeted APIs (`getHistoryFromLatestBoundary`, `getLastMessages`, or the new iterator), eliminating all O(total-history) reads from hot paths.

## Background

PR #2254 switched compaction to append-only, meaning `chat.jsonl` grows monotonically. The previous work in this PR (phases 1–2) added `getHistoryFromLatestBoundary()`, `getLastMessages()`, and `hasHistory()`, then wired the message-send hot path and tail-only sites. This commit completes phase 3+4: the `getFullHistory` → `iterateFullHistory` refactor and remaining call-site conversions.

### New `iterateFullHistory` API (replaces `getFullHistory`)

```typescript
async iterateFullHistory(
  workspaceId: string,
  direction: "forward" | "backward",
  visitor: (messages: MuxMessage[]) => boolean | void | Promise<boolean | void>
): Promise<Result<void>>
```

- **Forward**: reads file start→end in 256KB chunks, uses raw byte scanning for `\n` (0x0A) to handle multi-byte UTF-8 safely at chunk boundaries
- **Backward**: reads file end→start in 256KB chunks, yields messages newest-first within each chunk
- **Early exit**: visitor returns `false` to stop reading (e.g., find-by-id, max-sequence lookup)
- **Self-healing**: skips malformed JSON lines, same as the previous `readChatHistory`

### Call site conversions

| Call site | Before | After | Rationale |
|---|---|---|---|
| `partialService.commitToHistory` | `getFullHistory` | `getHistoryFromLatestBoundary` | Partial is always current epoch |
| `workspaceService.getPostCompactionState` | `getFullHistory` | `getHistoryFromLatestBoundary` | Only current epoch edits matter |
| `workspaceService.answerAskUserQuestion` | `getFullHistory` | `getHistoryFromLatestBoundary` | Tool calls don't survive compaction |
| `workspaceService.replaceHistory` | `getFullHistory` | `getHistoryFromLatestBoundary` | Latest boundary has max epoch |
| `mockAiStreamPlayer` | `getFullHistory` | `getHistoryFromLatestBoundary` | Target message is always recent |
| `getNextHistorySequence` | `getFullHistory` | `getLastMessages(1)` | Sequences are monotonic; O(1) |
| `sessionUsageService` (4 sites) | `getFullHistory` | `iterateFullHistory` + collect | Usage rebuild needs all epochs |
| Internal structural rewrites | `getFullHistory` | `readChatHistory` (private) | Already hold the lock, need full array for rewrite |

### Test coverage

- 6 new tests for `iterateFullHistory`: forward order, backward order, early exit (both directions), empty history, malformed line skipping
- All existing tests updated to use `collectFullHistory` helper (which wraps `iterateFullHistory`)
- 54 history service tests pass, plus all partial/workspace/task/compaction test suites

## Validation

- `make static-check` passes (typecheck + lint + fmt + docs link check)
- All 54 historyService tests pass (48 existing + 6 new)
- partialService (5), workspaceService (28), taskService (34), idleCompaction (12) tests pass
- Zero remaining `getFullHistory` references in production or test code

## Risks

- **Low**: Internal structural rewrite methods now call `readChatHistory` directly instead of through the removed `getFullHistory` wrapper. The error handling is equivalent — both paths are inside try-catch blocks that return `Err(...)`.
- **Low**: `getNextHistorySequence` now uses `getLastMessages(1)` with a backward-scan fallback for legacy data missing sequence numbers. This is O(1) for normal data vs. the previous O(n) full read.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$35.24`_

### DIFF
--- a/src/browser/utils/messages/StreamingMessageAggregator.test.ts
+++ b/src/browser/utils/messages/StreamingMessageAggregator.test.ts
@@ -772,10 +772,10 @@ describe("StreamingMessageAggregator", () => {
       type: "message" as const,
     });
 
-    test("prunes pre-boundary messages when compaction boundary arrives live", () => {
+    test("does not prune on first compaction (no penultimate boundary exists)", () => {
       const aggregator = new StreamingMessageAggregator(TEST_CREATED_AT);
 
-      // Simulate messages accumulated during a live session
+      // Simulate messages accumulated during a live session (no prior compaction)
       const msg1 = asChatMessage(
         createMuxMessage("user-1", "user", "First message", {
           historySequence: 0,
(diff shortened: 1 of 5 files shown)


ammar-agent:
@codex review

PR #2299, opened 2026-02-09
🤖 fix: preserve provider-native web_search cache-control handling

ThomasK33:
Summary
This PR fixes a regression where provider-native `web_search` tools were incorrectly downgraded during Anthropic cache-control wrapping, which could leave tool calls stuck and prematurely end streaming.

Background
`applyCacheControlToTools()` checked for `"provider-defined"` instead of the AI SDK provider-native discriminant (`"provider"`). That routed provider tools through `createTool(...)`, stripping provider metadata and breaking provider-executed tool flow.

Implementation
- Added a provider-native type guard for `Tool` (`type === "provider"`).
- Added defensive assertions around `execute` invariants so unexpected tool shapes fail fast.
- Preserved provider-native tool metadata by cloning tools with descriptor preservation before attaching cache-control provider options.
- Updated cache strategy tests to use real Anthropic/OpenAI provider-native web search tools.
- Added regression assertions to ensure provider-native tools remain provider-native and are not recreated via `createTool(...)`.

Validation
- `make static-check`
- `bun test src/common/utils/ai/cacheStrategy.test.ts`
- `bun test src/common/utils/tools/tools.test.ts`

Risks
Low-to-moderate. The change is narrowly scoped to cache-control wrapping for the final tool in the tool map. Main risk is around provider-tool shape assumptions, mitigated by explicit assertions and real-provider regression fixtures.

---

<details>

# Plan: fix `web_search` provider-tool regression and prevent recurrence

## Context / Why
`web_search` is currently being downgraded from a provider-native tool to a regular client-side tool in the Anthropic cache-control path. That leaves tool calls stuck at `dynamic-tool` + `state: "input-available"` and ends streaming early because no local `execute` exists.

Primary goals:
1. Restore correct provider-native handling now (unblock streaming).
2. Add guardrails so a future refactor/test-only PR cannot silently reintroduce this class of bug.

<details>
<summary>Failure mechanism (condensed)</summary>

`applyCacheControlToTools()` checks `existingTool.type === "provider-defined"` (wrong literal) instead of the AI SDK’s actual provider-native marker (`"provider"`).

That false check sends provider tools through `createTool(...)`, which strips provider identity metadata. The model then emits a tool call for `web_search`, but no local `execute` exists, so the tool stays `input-available` and the stream stops at the tool-call boundary.
</details>

## Evidence
- `src/common/utils/ai/cacheStrategy.ts` (around lines 155–161): provider-native branch checks `"provider-defined"`.
- `src/common/utils/ai/cacheStrategy.test.ts` (around lines 263–304): regression test currently encodes the same wrong literal (`"provider-defined"`), so it masks the bug.
- `src/common/utils/tools/tools.ts` (around lines 323–350): `web_search` is created as provider-native via `anthropic.tools.webSearch_20250305(...)` and `openai.tools.webSearch(...)`.
- `src/node/services/messagePipeline.ts` (around lines 165–169): incomplete tool calls are intentionally dropped (`ignoreIncompleteToolCalls: true`), which explains why stuck `input-available` calls disappear from subsequent requests.
- Runtime probe in this workspace:
  - `anthropic.tools.webSearch_20250305({ maxUses: 1 }).type === "provider"`
  - `openai.tools.webSearch({ searchContextSize: "high" }).type === "provider"`
- `node_modules/@ai-sdk/provider-utils/dist/index.d.ts` (around lines 1013–1134): `Tool` is a discriminated union and provider-native tools are explicitly `type: "provider"` with provider-only fields (`id`, `args`).

## Recommended approach (A): restore behavior + add robust regression guards
**Net LoC estimate (product code only): +20 to +35**

### 1) Fix provider-native detection and make invariant explicit
**File:** `src/common/utils/ai/cacheStrategy.ts`  
**Symbols:** `applyCacheControlToTools`

- Replace the incorrect literal check with `"provider"`.
- Add a typed helper/guard (`isProviderNativeTool`) so the discriminant lives in one place and callers can use provider-only fields safely.
- Avoid loose shape casts for field access; read `tool.type` / `tool.execute` directly from the SDK `Tool` union and narrow via the guard.
- Add defensive assertion(s) so tools with no `execute` must be provider-native.

```ts
...
function isProviderNativeTool(
  tool: Tool
): tool is Tool & { type: "provider"; id: `${string}.${string}`; args: Record<string, unknown> } {
  return tool.type === "provider";
}
...
if (existingTool.execute == null) {
  assert(
    isProviderNativeTool(existingTool),
    "Tool without execute must be provider-native (type=\"provider\")"
  );
}

if (isProviderNativeTool(existingTool)) {
  // Type-safe access to existingTool.id / existingTool.args here
  ...
}
```

### 2) Preserve provider tool metadata defensively when attaching cache options
**File:** `src/common/utils/ai/cacheStrategy.ts`  
**Symbols:** provider-native branch inside `applyCacheControlToTools`

- Use descriptor-preserving clone utility before adding `providerOptions` (instead of plain object spread), reducing risk if SDK properties become non-enumerable/getter-based in future releases.

```ts
...
const clonedProviderTool = cloneToolPreservingDescriptors(existingTool);
clonedProviderTool.providerOptions = {
  anthropic: { cacheControl: { type: "ephemeral" } },
};
cachedTools[key as keyof T] = clonedProviderTool as T[keyof T]; // generic map assignment
```

### 3) Rewrite the regression test to use real provider-native tool shape
**File:** `src/common/utils/ai/cacheStrategy.test.ts`  
**Symbols:** `describe("applyCacheControlToTools")` provider-native test case

- Replace synthetic `type: "provider-defined"` fixture with an actual provider tool instance from `@ai-sdk/anthropic` (and optionally a second case for `@ai-sdk/openai`).
- Assert that after caching:
  - `type` remains `"provider"`
  - provider tool identity fields (`id`, `args`) still exist
  - `providerOptions.anthropic.cacheControl` is added

```ts
...
const providerTool = anthropic.tools.webSearch_20250305({ maxUses: 1000 }) as unknown as Tool;
const result = applyCacheControlToTools({ readFile: ..., web_search: providerTool }, "anthropic:claude-3-5-sonnet");

const cachedWebSearch = result.web_search as { type?: string; id?: string; args?: unknown; providerOptions?: unknown };
expect(cachedWebSearch.type).toBe("provider");
expect(cachedWebSearch.id).toContain("web_search");
expect(cachedWebSearch.args).toBeDefined();
expect(cachedWebSearch.providerOptions).toEqual({
  anthropic: { cacheControl: { type: "ephemeral" } },
});
```

### 4) Add a focused anti-regression assertion for the exact bug class
**File:** `src/common/utils/ai/cacheStrategy.test.ts`

- Add a test that would fail if provider tools are ever routed through the `createTool(...)` fallback again (e.g., by verifying provider-only fields survive cache-control application).

### 5) Validate
- `bun test src/common/utils/ai/cacheStrategy.test.ts`
- `bun test src/common/utils/tools/tools.test.ts` (sanity for provider tool assembly path)
- `make typecheck`

## Alternative approach (B): minimal hotfix
**Net LoC estimate (product code only): +1 to +5**

- Change `"provider-defined"` → `"provider"` in `cacheStrategy.ts` and update the single test literal.

<details>
<summary>Why not recommended</summary>

This resolves the immediate outage but leaves weak coverage (string-literal drift can recur). It does not add runtime invariants or realistic provider-tool fixtures, so the same regression can re-enter via unrelated refactors.
</details>

## Prevention policy (for future PRs)
- Keep the provider-native contract verified by tests using real provider tool factories (`anthropic.tools.webSearch_20250305`, `openai.tools.webSearch`).
- Keep the invariant check in production code so incorrect tool shapes fail loudly instead of silently degrading into stuck `input-available` states.
- Treat `cacheStrategy.test.ts` as required coverage for changes touching `cacheStrategy.ts` or tool-shaping logic.

</details>

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$1.28`_

### DIFF
--- a/src/common/utils/ai/cacheStrategy.test.ts
+++ b/src/common/utils/ai/cacheStrategy.test.ts
@@ -303,5 +303,32 @@ describe("cacheStrategy", () => {
       expect(Object.keys(result)).toEqual(Object.keys(toolsWithProviderTool));
       expectProviderToolToRemainProviderNative(result.web_search, providerTool);
     });
+
+    it("should handle execute-less dynamic tools without throwing", () => {
+      const dynamicToolWithoutExecute = {
+        type: "dynamic" as const,
+        description: "MCP dynamic tool",
+        inputSchema: z.object({ query: z.string() }),
+      } as unknown as Tool;
+
+      const toolsWithDynamicTool: Record<string, Tool> = {
+        readFile: mockTools.readFile,
+        mcp_dynamic_tool: dynamicToolWithoutExecute,
+      };
+
+      const result = applyCacheControlToTools(toolsWithDynamicTool, "anthropic:claude-3-5-sonnet");
+
+      const cachedDynamicTool = result.mcp_dynamic_tool as {
+        type?: string;
+        execute?: unknown;
+        providerOptions?: unknown;
+      };
+      expect(cachedDynamicTool.type).toBe("dynamic");
+      expect(cachedDynamicTool.execute).toBeUndefined();
+      expect(cachedDynamicTool.providerOptions).toEqual({
+        anthropic: { cacheControl: { type: "ephemeral" } },
+      });
+      expect(result.readFile).toEqual(toolsWithDynamicTool.readFile);
+    });
   });
 });
--- a/src/common/utils/ai/cacheStrategy.ts
+++ b/src/common/utils/ai/cacheStrategy.ts
@@ -138,9 +138,10 @@ export function createCachedSystemMessage(
  * 3. Last tool only (1 breakpoint) - caches all tools up to and including this one
  * = 3 total, leaving 1 for future use
  *
- * NOTE: For function/dynamic tools, the SDK requires providerOptions during tool()
- * creation, so we re-create the last tool. Provider-native tools (type: "provider")
- * keep provider metadata and are cloned with descriptors before attaching options.
+ * NOTE: Function tools with execute handlers are recreated so providerOptions is set
+ * at creation time. Provider-native tools (type: "provider") and execute-less
+ * dynamic/MCP tools keep their runtime metadata and are descriptor-cloned before
+ * attaching providerOptions.
  */
 export function applyCacheControlToTools<T extends Record<string, Tool>>(
   tools: T,
@@ -161,13 +162,6 @@ export function applyCacheControlToTools<T extends Record<string, Tool>>(
   const cachedTools = {} as unknown as T;
   for (const [key, existingTool] of Object.entries(tools)) {
     if (key === lastToolKey) {
-      if (existingTool.execute == null) {
-        assert(
-          isProviderNativeTool(existingTool),
-          `Tool "${key}" without execute must be provider-native (type="provider")`
-        );
-      }
-
       if (isProviderNativeTool(existingTool)) {
         // Provider-native tools (e.g. Anthropic/OpenAI web search) cannot be recreated with
         // createTool(). Clone while preserving descriptors/getters and attach providerOptions.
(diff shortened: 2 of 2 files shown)


ThomasK33:
@codex review

Addressed the execute-less dynamic tool regression concern:
- execute-less dynamic/MCP tools are now handled without assertion failure
- added regression coverage in cacheStrategy tests
- re-ran static checks locally

PR #2310, opened 2026-02-10
🤖 fix: surface browser startup connection/auth failures

ethanndickson:
## Summary (user-facing)
Remote browser access no longer gets stuck on the initial **“Loading workspaces…”** spinner when the backend can’t be reached or authentication is missing.

## What changed (user-facing)
- If the browser UI hits an API **error before the first workspace load**, it now shows a startup error screen (with details + **Retry**) instead of spinning forever.
- During initial reconnect attempts, the loading screen text now shows progress (e.g. “Reconnecting… attempt N”) so users can tell it’s actively retrying.
- If the WebSocket handshake fails in a way browsers report as **close code 1006** (common behind reverse proxies) and the backend requires auth, the UI will now prompt for a token via the existing **Authentication Required** modal.
- Prevented the auth token modal from flickering away due to the reconnect loop once we’ve determined auth is required.

## Validation
- `make static-check`
- `bun test src/browser/contexts/API.test.tsx src/browser/components/AppLoader.auth.test.tsx src/browser/components/LoadingScreen.test.tsx`

## Risks
- Low/medium: touches browser startup connection state handling. Guarded with connection IDs to avoid stale async probes affecting newer connection attempts.

---

_Generated with `mux` • Model: `openai:gpt-5.2` • Thinking: `high` • Cost: `$13.25`_

### DIFF
--- a/src/browser/contexts/API.test.tsx
+++ b/src/browser/contexts/API.test.tsx
@@ -49,7 +49,7 @@ class MockWebSocket {
 }
 
 const originalFetch = globalThis.fetch;
-let fetchImpl: (input: RequestInfo | URL) => Promise<Response> = () =>
+let fetchImpl: (input: RequestInfo | URL, init?: RequestInit) => Promise<Response> = () =>
   Promise.resolve({
     ok: false,
     json: () => Promise.resolve({}),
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

Addressed feedback:
- Startup /api/spec.json auth probe now has a hard timeout (AbortController + Promise.race) so it can’t hang startup.
- If the probe result is inconclusive (timeout/network/non-OK/invalid JSON), we reset the guard to allow re-probe on a later 1006 close.

Added unit tests covering probe timeout fallback + re-probe after inconclusive result.

PR #2315, opened 2026-02-10
🤖 fix: add missing keyboard shortcut hints to UI buttons and command palette

deansheather:
## Summary

Adds missing keyboard shortcut tooltips to UI buttons and fills in shortcut hints in the command palette, plus fixes a cross-platform display bug.

## Background

An audit of all keyboard shortcuts vs their UI discoverability found:
- 2 clickable buttons with keybinds but no tooltip at all (Settings gear, "+" new workspace)
- 1 button using a plain `title=` attribute without the shortcut (sidebar hamburger)
- 3 command palette entries missing `shortcutHint` despite having keybinds
- 1 cross-platform bug where "Open Settings" hardcoded `"⌘,"` instead of using `formatKeybind()`

## Implementation

**UI Button Tooltips (3 files):**
- `SettingsButton.tsx` — wrapped gear icon in `<Tooltip>` showing `Open settings (⌘,)` / `Open settings (Ctrl+,)`
- `ProjectSidebar.tsx` — wrapped `+` new-workspace button in `<Tooltip>` showing `New chat (⌘N)` / `New chat (Ctrl+N)`
- `WorkspaceHeader.tsx` — replaced plain `title="Open sidebar"` with `<Tooltip>` showing `Open sidebar (⌘P)` / `Open sidebar (Ctrl+P)`

**Command Palette (1 file — `sources.ts`):**
- "Interrupt Streaming" — added `shortcutHint` showing `Esc`
- "Change Model…" — added `shortcutHint` showing `⌘/` / `Ctrl+/`
- "Set Thinking Effort…" — added `shortcutHint` showing `⌘⇧T` / `Ctrl+Shift+T`
- "Open Settings" — replaced hardcoded `"⌘,"` with `formatKeybind(KEYBINDS.OPEN_SETTINGS)` so Windows/Linux users see `Ctrl+,`

All changes follow existing patterns (same Tooltip/TooltipTrigger/TooltipContent + formatKeybind pattern used by the terminal button, editor button, etc.).

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `high` • Cost: `$10.20`_

### DIFF
--- a/src/browser/components/ProjectSidebar.tsx
+++ b/src/browser/components/ProjectSidebar.tsx
@@ -906,17 +906,24 @@ const ProjectSidebarInner: React.FC<ProjectSidebarProps> = ({
                             </TooltipTrigger>
                             <TooltipContent align="end">Remove project</TooltipContent>
                           </Tooltip>
-                          <button
-                            onClick={(event) => {
-                              event.stopPropagation();
-                              handleAddWorkspace(projectPath);
-                            }}
-                            aria-label={`New chat in ${projectName}`}
-                            data-project-path={projectPath}
-                            className="text-secondary hover:bg-hover hover:border-border-light flex h-5 w-5 shrink-0 cursor-pointer items-center justify-center rounded border border-transparent bg-transparent text-sm leading-none transition-all duration-200"
-                          >
-                            +
-                          </button>
+                          <Tooltip>
+                            <TooltipTrigger asChild>
+                              <button
+                                onClick={(event) => {
+                                  event.stopPropagation();
+                                  handleAddWorkspace(projectPath);
+                                }}
+                                aria-label={`New chat in ${projectName}`}
+                                data-project-path={projectPath}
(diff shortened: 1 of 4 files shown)


deansheather:
@codex review

Addressed review comment: kept the `shortcutHint` for "Interrupt Streaming" showing the normal-mode shortcut (Esc) since it always works regardless of vim mode, and added a comment explaining the trade-off.

deansheather:
@codex review

Updated the comment to be accurate — it now says "Vim mode uses Ctrl+C instead" rather than claiming Esc always works. The intent is to show the common-case default (Esc) since vim state isn't available in the command palette context.

deansheather:
@codex review

Removed the two inaccurate shortcutHints:
- "Change Model…" — `CYCLE_MODEL` cycles directly, but this opens the picker
- "Set Thinking Effort…" — `TOGGLE_THINKING` cycles directly, but this opens a selection prompt

Added comments documenting why the hints are intentionally omitted.

PR #2317, opened 2026-02-10
🤖 fix: auto-retry on provider config change for credential errors

### LINKED ISSUE #2051: Changing the API key in the settings UI does not immediately update the API key in an existing chat
I was having an issue where my api key was getting quota exceeded errors. only after using `/providers set openai apiKey` did the apiKey get updated for the existing chat

Image

deansheather:
## Summary

When a user changes their API key in Settings → Providers after hitting a credential-related error (quota exceeded, authentication failure, missing API key), the affected workspace chats now automatically retry instead of staying stuck on the error.

Closes #2051

## Background

The backend reads provider config fresh from disk on every `createModel()` call — there is no caching. So once the user saves a new key, the next request will use it. However, the frontend's auto-retry system classified credential errors (`authentication`, `quota`, `api_key_not_found`) as **non-retryable** and stopped polling, requiring the user to manually click Retry even after fixing the issue.

The old `/providers set` slash command (removed in #2127) appeared to work because it was run _inside_ the chat — the user naturally sent their next message after, triggering a fresh `streamMessage()`. With the Settings UI, the user changes the key in a separate panel and returns to a chat that's still stuck on the error.

## Implementation

Subscribe to `providers.onConfigChanged` in `useResumeManager` (the centralized retry orchestrator). When provider config changes:

1. Scan all workspaces for credential-related interruptions:
   - `stream-error` with `errorType` of `"authentication"` or `"quota"`
   - `retryState.lastError` of type `"api_key_not_found"` or `"oauth_not_connected"`
2. Clear the non-retryable error state (same as what the manual Retry button does)
3. Trigger a manual resume attempt (bypasses non-retryable auto-retry guards)

Workspaces with unrelated errors (e.g., `runtime_not_ready`, `model_not_found`) are unaffected.

## Risks

Low risk — the change only triggers retry attempts for workspaces already in an error state. The retry uses the same `attemptResume(workspaceId, true)` codepath as the manual Retry button. If the new key is also invalid, the error will be re-stored and the workspace returns to its stuck state (no infinite loop).

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `high` • Cost: `$16.98`_

### DIFF
--- a/src/browser/hooks/useResumeManager.ts
+++ b/src/browser/hooks/useResumeManager.ts
@@ -7,12 +7,14 @@ import { readPersistedState, updatePersistedState } from "./usePersistedState";
 import { readAutoRetryPreference } from "@/browser/utils/messages/autoRetryPreference";
 import {
   getInterruptionContext,
+  getLastNonDecorativeMessage,
   isNonRetryableSendError,
 } from "@/browser/utils/messages/retryEligibility";
 import { applyCompactionOverrides } from "@/browser/utils/messages/compactionOptions";
 import type { SendMessageError } from "@/common/types/errors";
 import {
   createFailedRetryState,
+  createManualRetryState,
   calculateBackoffDelay,
   INITIAL_DELAY,
 } from "@/browser/utils/messages/retryState";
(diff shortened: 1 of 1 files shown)


deansheather:
Closing — the auto-retry approach deviated from the approved plan. Will re-approach with the correct fix if needed.

PR #2320, opened 2026-02-10
🤖 refactor: replace streaming flag with turn-phase state machine

ethanndickson:
### Summary
Introduce an explicit per-workspace turn-phase state machine in `AgentSession` and use it as the single “turn is active” signal. This closes a race where `AIService.isStreaming()` ended before post-stream completion cleanup finished, allowing IPC sends/edits to bypass the queue and mutate history mid-cleanup.

**Note: Most of the diff is tests.**

### Background
`AgentSession` does async post-stream work (compaction completion, diff ack, follow-up dispatch, queued-message flush, per-stream state reset). Previously, we treated the workspace as idle as soon as streaming ended, which created a window where new messages/edits could start a new turn or truncate history while completion cleanup was still in-flight.

### Implementation
- Add `TurnPhase` (`IDLE | PREPARING | STREAMING | COMPLETING`) and drive `isBusy()` / `waitForIdle()` from it (replacing the ad-hoc `streamStarting` signal).
- Keep the session busy through the full async `stream-end` handler (including compaction completion), only transitioning to `IDLE` if no follow-up/queue started a new turn.
- Queue IPC sends based on `session.isBusy()` (not `AIService.isStreaming()`), including during `COMPLETING`.
- Block history truncate/replace while a turn is active (covers `COMPLETING`).
- Edits (truncate+replace): if busy, wait for idle before truncating; only `interruptStream()` during `PREPARING/STREAMING` and propagate failures; skip interrupting during `COMPLETING` to avoid synthetic abort → incorrect `IDLE`.

### Validation
- `make static-check`
- `bun test ./tests/ipc/queuedMessages.starting.test.ts`
- `bun test ./tests/ipc/queuedMessages.completing.test.ts`

### Risks
- Slightly expands the “busy” window to include completion cleanup; this may delay sends/edits, but prevents history corruption and state clobbering.
- Relies on stream event ordering; covered by deterministic IPC regressions that gate completion.

---

### DIFF
--- a/src/node/services/agentSession.ts
+++ b/src/node/services/agentSession.ts
@@ -1799,6 +1799,9 @@ export class AgentSession {
             this.onPostCompactionStateChange?.();
           }
         } else {
+          // CompactionHandler emits its own sanitized stream-end; mark as handled
+          // so the catch block doesn't re-emit the unsanitized original payload.
+          emittedStreamEnd = true;
           this.onCompactionComplete?.();
         }
 
@@ -1834,12 +1837,10 @@ export class AgentSession {
           }
         }
       } finally {
-        // Best-effort state cleanup; resetActiveStreamState is idempotent.
-        this.resetActiveStreamState();
-
-        // Only transition to IDLE if we're still in COMPLETING — don't clobber a new
-        // turn started by follow-up dispatch or queue send.
+        // Only clean up if we're still in COMPLETING — a new turn started by
+        // dispatchPendingFollowUp() or sendQueuedMessages() owns the stream state now.
         if (this.turnPhase === TurnPhase.COMPLETING) {
+          this.resetActiveStreamState();
           this.setTurnPhase(TurnPhase.IDLE);
         }
       }


ethanndickson:
@codex review

Addressed both P1 and P2 comments:

- **P1**: `stream-end` handler is now wrapped in `try/catch/finally` — always resets to IDLE even if completion cleanup throws.
- **P2**: New `deferQueuedFlushUntilAfterEdit` latch prevents queued messages from running ahead of an edit during COMPLETING.

Also fixed `truncate.test.ts` error substring and added 2 deterministic regression tests.

ethanndickson:
@codex review

Added guard against synthetic/stale stream-abort events: only process aborts when `turnPhase === STREAMING`. In other phases (PREPARING, COMPLETING, IDLE) the abort is either synthetic (no real stream) or stale (deferred via AIService async IIFE), and the originating code path already handles its own transition back to IDLE.

ethanndickson:
@codex review

Fixed the P1 abort forwarding issue: the guard now calls `emitChatEvent(payload)` before returning, so the renderer still receives the `stream-abort` event to clear "starting…" / "interrupting…" UI. The guard only skips the phase transition and state reset — which are what cause the race condition.

Also verified with `interruptStream.starting.test.ts` which tests exactly this scenario (interrupt during PREPARING before stream-start).

ethanndickson:
@codex review

Addressed P1: deferred queued messages should not abort the edit turn.

Fix: in AgentSession.streamWithHistory(), suppress StreamManager's queued-message stop condition for edit turns by passing `hasQueuedMessage: undefined` when `options.editMessageId` is set. This prevents a deferred, non-empty messageQueue (from the completion defer path) from stopping the edit stream at the next step boundary; queued messages are still drained/sent after the edit completes via stream-end.

Validation:
- make static-check
- bun test ./tests/ipc/queuedMessages.completing.test.ts
- bun test ./tests/ipc/queuedMessages.starting.test.ts

ethanndickson:
@codex review

Fixed two `stream-end` handler issues in `AgentSession`:

1) Guard `resetActiveStreamState()` in the `stream-end` `finally` block so it only runs when we're still in `COMPLETING` (prevents clobbering the follow-up / queued-flush turn's stream state).
2) When compaction is handled (`handleCompletion()` returns true), set `emittedStreamEnd = true` so error recovery doesn't re-emit the unsanitized original `stream-end` payload after `CompactionHandler` already emitted a sanitized one.

Validation:
- make typecheck
- bun test ./tests/ipc/queuedMessages.completing.test.ts
- bun test ./tests/ipc/queuedMessages.starting.test.ts
- make static-check

PR #2328, opened 2026-02-10
🤖 fix: self-heal duplicate Chat with Mux entries across roots

ibetitsmike:
## Summary
Startup now self-heals the built-in `mux-chat` workspace so only one valid `Chat with Mux` entry exists for the active `MUX_ROOT`, preventing duplicate rows in dev builds while preserving unrelated legacy workspaces.

## Background
When switching between regular and dev mux roots (for example `~/.mux` and `~/.mux-dev`), stale `mux-chat` records could remain in `config.json` under an older system project path. That allowed two `Chat with Mux` entries to appear and made navigation confusing.

Codex review identified a regression risk in the first patch: legacy user workspaces can still have ID collisions like `mux-chat` via legacy `<project>-<workspace>` IDs. Cleanup must therefore only target entries that actually look like the built-in system workspace.

## Implementation
- Updated `ServiceContainer.ensureMuxChatWorkspace()` to:
  - remove stale `mux-chat` entries from **non-active system `.../system/Mux` projects**,
  - restrict stale removal to entries that still look like built-in mux-chat (system path / agent / canonical name/title markers),
  - delete stale system project records that become empty,
  - collapse duplicates in the active system project down to exactly one,
  - enforce required invariants (`path`, `name`, `title`, `agentId`, local runtime, unarchived state).
- Added regression coverage in `src/node/services/serviceContainer.test.ts`:
  - cleans stale old-root mux-chat duplicates,
  - preserves non-system legacy workspaces whose ID also equals `mux-chat`.

## Validation
- `make static-check`
- `./scripts/check_codex_comments.sh 2328`
- `./scripts/wait_pr_codex.sh 2328` (approved)

## Risks
Low. The logic remains startup-scoped and targeted to built-in system workspace reconciliation. The collision guard reduces risk by avoiding deletion of non-system legacy workspaces.

---

### DIFF
--- a/src/node/services/serviceContainer.test.ts
+++ b/src/node/services/serviceContainer.test.ts
@@ -101,4 +101,43 @@ describe("ServiceContainer", () => {
     expect(muxChatWorkspace?.archivedAt).toBeUndefined();
     expect(muxChatWorkspace?.unarchivedAt).toBeUndefined();
   });
+
+  it("keeps non-system legacy workspaces whose IDs also equal mux-chat", async () => {
+    const legacyProjectPath = path.join(tempDir, "repos", "mux");
+    const legacyWorkspacePath = path.join(config.srcDir, "mux", "chat");
+
+    await config.editConfig((cfg) => {
+      cfg.projects.set(legacyProjectPath, {
+        workspaces: [
+          {
+            path: legacyWorkspacePath,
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

Addressed your P2 feedback by restricting stale mux-chat cleanup to system-project-shaped entries only, and added regression coverage to preserve legacy non-system `mux-chat` IDs.

PR #2333, opened 2026-02-10
🤖 fix: graceful backoff for transient auto-update errors

ammar-agent:
## Summary

When a new GitHub release is created, there's a window where `latest.yml` hasn't been uploaded yet. During this window, `electron-updater` fires an HTTP 404 error, and the current code surfaces it as a cryptic "Update check failed" tooltip. This PR silently backs off on transient errors and fixes several other auto-update nuisances found in the audit.

## Background

An audit of the auto-update flow uncovered several issues:

| Problem | Impact | Fix |
|---|---|---|
| 404 during `latest.yml` upload window shown as error | User sees "Update check failed" for a transient condition | Detect transient errors (404, network, rate-limit) → silently revert to `idle` |
| `checkForUpdates()` clobbers `downloading`/`downloaded` states | 4-hour interval can throw away a ready-to-install update | State guard: skip when `checking`/`downloading`/`downloaded` |
| `error` state is a dead-end for manual retry | User must wait ≤4 hours or restart to retry | Add `error` to hover-retry allow-list |
| Startup race: `check()` can fire before `initialize()` completes | First update check silently lost | `await this.ready` in `check()` and `download()` |
| `isCheckingOnHover` can get stuck | Spinner never clears if backend reverts to same state | Safety-net timeout (35s) |

## Implementation

**`src/desktop/updater.ts`**
- New `isTransientUpdateError()` helper: matches 404, `ENOTFOUND`/`ECONNREFUSED`/`ECONNRESET`/`ETIMEDOUT`/`EAI_AGAIN`, `network`/`socket hang up`, 403/rate-limit patterns
- Error event handler: transient errors → `log.debug` + revert to `idle`; non-transient → `log.error` + show `error` status (unchanged)
- Same treatment in `.catch()` on `autoUpdater.checkForUpdates()` promise rejection
- State guard at top of `checkForUpdates()`: skip when `checking`/`downloading`/`downloaded`
- Improved timeout log: `log.warn` with clear retry message

**`src/node/services/updateService.ts`**
- Store `initialize()` promise as `this.ready`
- `check()` and `download()` `await this.ready` before proceeding

**`src/browser/components/TitleBar.tsx`**
- Add `error` to the hover-retry allow-list (was only `idle`/`up-to-date`)
- Safety-net `setTimeout(() => setIsCheckingOnHover(false), 35_000)` to prevent stuck spinner

**`src/desktop/updater.test.ts`**
- 9 new tests covering transient backoff (404, network, rate-limit, promise rejection), non-transient passthrough, and state guards (checking, downloading, downloaded)
- Updated existing error test to use non-transient error message

## Risks

- **Low:** The `isTransientUpdateError` regex could theoretically match a non-transient error containing those keywords. However, `electron-updater` error messages for real failures (code signing, corrupt artifact) don't contain these patterns. If a genuine error is misclassified, the user simply sees `idle` and the next check (hover or 4-hour interval) will retry — strictly better than the current behavior of showing a confusing error.
- **Low:** The `await this.ready` in `check()`/`download()` adds a small async hop. Since `initialize()` completes fast (dynamic import), this is negligible.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$3.59`_

### DIFF
--- a/src/browser/components/TitleBar.tsx
+++ b/src/browser/components/TitleBar.tsx
@@ -149,11 +149,6 @@ export function TitleBar() {
     ) {
       lastHoverCheckTime.current = now;
       setIsCheckingOnHover(true);
-
-      // Safety net: clear spinner after check timeout + buffer,
-      // in case the status stream doesn't deliver a state change.
-      setTimeout(() => setIsCheckingOnHover(false), 35_000);
-
       api?.update.check().catch((error) => {
         console.error("Update check failed:", error);
         setIsCheckingOnHover(false);


ammar-agent:
@codex review

1. **P1 (scope backoff to check phase)**: Transient error backoff now only fires when `this.updateStatus.type === "checking"`. Errors during download or other active phases surface as `error` status so the user knows something went wrong. Added test: "should surface transient-looking errors during download phase".

2. **P2 (tighten 403 matching)**: Removed bare `403` from the regex — now only matches explicit `rate limit` signatures. Persistent auth/config 403 errors remain visible. Added test: "should surface bare 403 errors (not rate-limit specific)".

ammar-agent:
@codex review

Addressed the review comment:

**P1 (setTimeout spinner coordination)**: Removed the `setTimeout(() => setIsCheckingOnHover(false), 35_000)` entirely. The timer was unnecessary because:
1. The ORPC status stream already clears `isCheckingOnHover` on every backend state change
2. The backend's 30s check timeout reverts to `idle` which fires the stream
3. The `.catch()` on `check()` handles call failures

All three paths already clear the spinner deterministically — no timer needed.

PR #2338, opened 2026-02-10
🤖 fix: ensure cost_usd is consistently populated in BQ for tbench runs

ammar-agent:
## Summary

Fixes `cost_usd` being inconsistently `null` in BigQuery for Terminal-Bench runs by ensuring token/cost extraction always runs (even on failure/timeout) and recovering cost data from intermediate JSONL events.

## Background

When investigating why `cost_usd` was not consistently populated in BQ for tbench runs, I found two root causes:

1. **Token extraction never ran on failure/timeout:** In `mux-run.sh`, when the mux CLI exited non-zero (timeout, crash, API error), `fatal` was called immediately, exiting the script *before* the Python extraction could parse the JSONL output. Since `tee` had already written partial output to disk, this data was available but never processed.

2. **Fallback path ignored available cost data:** When `run-complete` was missing from the JSONL (because the process was killed before the `finally` block ran), the fallback extraction only summed tokens from `usage-delta` and `session-usage-delta` events — it left `cost_usd` as `null`. The `session-usage-delta` events actually carry `ChatUsageDisplay` objects with per-component `cost_usd` that could be summed.

### `benchmarks/terminal_bench/mux-run.sh`
- Changed pipeline to capture exit code without calling `fatal`, allowing Python extraction to always run first
- Enhanced fallback path to extract `cost_usd` from `session-usage-delta` events (sub-agent costs from `ChatUsageDisplay` components)
- Added parsing of new `usage-cost` events for main-agent cost recovery
- Moved `fatal` call to after extraction completes

### `src/cli/run.ts`
- Added `usage-cost` event emission in JSON mode: after each `usage-delta` event, the CLI now emits a `{"type": "usage-cost", "cost_usd": <cumulative>}` line with the running total cost (completed messages + current message's cumulative usage). This enables the fallback extraction to recover main-agent cost even when `run-complete` is never emitted (timeout/SIGTERM).

## Validation

- Tested Python extraction logic with 5 scenarios: happy path (run-complete), timeout with only usage-delta, timeout with session-usage-delta, timeout with usage-cost events, timeout with both usage-cost and session-usage-delta
- TypeScript typecheck passes
- `make static-check` passes (includes lint, fmt, shellcheck, docs links)
- Unit tests pass (`displayUsage.test.ts`, `chatStats.test.ts`)

## Risks

Low risk. The changes are additive:
- The `usage-cost` event is a new emission type only in `--json` mode (used by benchmarks) — existing non-JSON CLI usage is unaffected
- The extraction reordering preserves the same `fatal` exit behavior, just delays it until after token file creation
- `run-complete` path is unchanged — only the fallback path is enhanced

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `high` • Cost: `$--`_

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -887,6 +887,13 @@ async function main(): Promise<number> {
           completedMainAgentCost += messageCost;
         }
 
+        // Emit recovery cost at stream-end so providers that skip usage-delta
+        // (and only report usage in stream-end metadata) still get a cost snapshot
+        // recoverable by mux-run.sh on timeout/crash.
+        if (emitJson && completedMainAgentCost > 0) {
+          emitJsonLine({ type: "usage-cost", cost_usd: completedMainAgentCost });
+        }
+
         // Budget enforcement at stream-end for providers that don't emit usage-delta events
         // Use cumulative cost across all messages in this run (not just the current message)
         if (budget !== undefined && !budgetExceeded) {


ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

PR #2351, opened 2026-02-11
🤖 fix: preserve subagent usage after agent_report

ethanndickson:
## Summary

Subagent costs were disappearing from the Costs tab because `agent_report` handling aborted the child stream before AI SDK usage was fully recorded. This PR replaces the old `tool-call-end` abort-based lifecycle with a `stopWhen`-driven model where `agent_report` ends the stream naturally at the step boundary, and `stream-end` is the single source of truth for report finalization and cleanup.

## Background

Three related bugs drove this work:

1. **Disappearing subagent costs.** `handleAgentReport` (triggered on `tool-call-end`) called `stopStream` to abort the child, cutting off AI SDK usage accounting for the final step.
2. **Out-of-order cleanup in task trees.** `cleanupReportedLeafTask` used `hasActiveDescendantAgentTasks` to gate parent deletion, but that predicate ignores `reported` children. A reported parent with two reported children could delete itself when the first child cleaned up, orphaning the second.
3. **Post-report tool execution.** After `agent_report`, the stream could continue into another step and execute more tool calls even though the task was already reported and the parent resumed.

### `agent_report` tool
- Returns `{ success: true, message: "Report submitted successfully." }` — no side-effects, no stream abort.

### Stream manager
- Added `hasToolCall("agent_report")` to the autonomous `stopWhen` condition array. The stream ends at the first step boundary containing `agent_report` — natural completion preserves usage accounting while preventing post-report tool execution.
- `deletePartial` / `updateHistory` return values at stream-end are now checked and logged on failure.

### Task report finalization (`taskService.ts`)
- **Removed `handleAgentReport`** (the `tool-call-end` listener), `readLatestAgentReportArgs`, `findAgentReportArgsInMessage`, and the `stopStream`/`commitPartial` block inside `finalizeAgentTaskReport`. Report args are now read directly from `event.parts` in `handleStreamEnd` via `findAgentReportArgsInParts`.
- **Removed `stream-abort` listener**, `handleStreamAbort`, `isStreamAbortEvent` — `stream-end` is the only terminal event TaskService listens to.
- **Added two-phase completion:** `finalizeAgentTaskReport` is purely logical (status + report delivery + waiter resolution). Cleanup progression is driven by `finalizeTerminationPhaseForReportedTask`, called from `handleStreamEnd` after finalization.
- **Added `canCleanupReportedTask`** predicate: checks `taskStatus === "reported"`, `!isStreaming`, structural-leaf topology, and no pending patch artifact.
- **Patch generation runs before waiter resolution** (`maybeStartPatchGenerationForReportedTask`), so an immediate `task_await` result after `agent_report` can include a pending artifact record.

### Structural-leaf cleanup ordering
- Added `hasChildAgentTasks(index, workspaceId)` — a topology predicate that checks whether a workspace has any child agent-task nodes in config, regardless of their status.
- Replaced the `hasActiveDescendantAgentTasks` call in cleanup with the structural-leaf check. A reported task can only be deleted when it has zero children in config.

### Cleanup
- Extracted `isTypedWorkspaceEvent` helper to deduplicate type-guard functions.
- Removed `ToolCallEndEvent` import from taskService (no longer used).

## Validation

- `make static-check` — all checks pass
- `bun test src/node/services/taskService.test.ts` — 36 pass
- `bun test src/node/services/tools/task_await.test.ts` — 8 pass
- `bun test src/node/services/streamManager.test.ts` — stopWhen tests pass

## Risks

- **Cleanup timing:** mitigated by per-workspace event locking, idempotent `cleanupReportedLeafTask`, and tests covering both `agent_report` → `handleStreamEnd` flow and structural-leaf ordering.
- **Structural-leaf gate is conservative:** a reported parent with a stuck reported child will never auto-delete. Intentional — manual cleanup is safer than orphaning workspaces or losing artifacts.
- **`stopWhen` scope:** `hasToolCall("agent_report")` only applies to autonomous mode (not `toolChoice` mode which already uses `stepCountIs(1)`). The condition is additive and cannot affect streams that don't invoke `agent_report`.

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$14.32`_

### DIFF
--- a/src/node/services/streamManager.test.ts
+++ b/src/node/services/streamManager.test.ts
@@ -119,7 +119,7 @@ describe("StreamManager - stopWhen configuration", () => {
 
     expect(
       agentReportCondition({
-        steps: [{ toolResults: [{ toolName: "agent_report" }] }],
+        steps: [{ toolResults: [{ toolName: "agent_report", output: { success: true } }] }],
       })
     ).toBe(true);
   });
@@ -144,9 +144,16 @@ describe("StreamManager - stopWhen configuration", () => {
     // Returns true when step contains successful agent_report tool result.
     expect(
       reportStop({
-        steps: [{ toolResults: [{ toolName: "agent_report" }] }],
+        steps: [{ toolResults: [{ toolName: "agent_report", output: { success: true } }] }],
       })
     ).toBe(true);
+
+    // Returns false when step contains failed agent_report output.
+    expect(
+      reportStop({
+        steps: [{ toolResults: [{ toolName: "agent_report", output: { success: false } }] }],
+      })
+    ).toBe(false);
 
     // Returns false when step only contains agent_report tool call (no successful result yet).
     expect(
@@ -158,7 +165,7 @@ describe("StreamManager - stopWhen configuration", () => {
     // Returns false when step contains other tool results.
     expect(
       reportStop({
-        steps: [{ toolResults: [{ toolName: "bash" }] }],
+        steps: [{ toolResults: [{ toolName: "bash", output: { success: true } }] }],
       })
     ).toBe(false);
 
--- a/src/node/services/streamManager.ts
+++ b/src/node/services/streamManager.ts
@@ -1020,12 +1020,25 @@ export class StreamManager extends EventEmitter {
     }
 
     // For autonomous loops: cap steps, check for queued messages, and stop after
-    // a successful agent_report result so the stream ends naturally (preserving
-    // usage accounting) without allowing post-report tool execution.
+    // a successful agent_report result (`output.success === true`) so the stream ends
+    // naturally (preserving usage accounting) without allowing post-report tool execution.
+    const isSuccessfulAgentReportOutput = (value: unknown): boolean => {
+      return (
+        typeof value === "object" &&
+        value !== null &&
+        "success" in value &&
+        (value as { success?: unknown }).success === true
+      );
+    };
+
     const hasSuccessfulAgentReportResult: ReturnType<typeof stepCountIs> = ({ steps }) => {
       const lastStep = steps[steps.length - 1];
       return (
-        lastStep?.toolResults?.some((toolResult) => toolResult.toolName === "agent_report") ?? false
+        lastStep?.toolResults?.some(
+          (toolResult) =>
+            toolResult.toolName === "agent_report" &&
+            isSuccessfulAgentReportOutput(toolResult.output)
+        ) ?? false
       );
     };
 
--- a/src/node/services/taskService.ts
+++ b/src/node/services/taskService.ts
@@ -2237,8 +2237,9 @@ export class TaskService {
 
     // NOTE: Stream continues — we intentionally do NOT abort it.
     // Deterministic termination is enforced by StreamManager stopWhen logic that
-    // waits for a successful agent_report tool result at the step boundary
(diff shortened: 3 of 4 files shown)


ethanndickson:
@codex review

Addressed the P1 race by moving exec patch generation + cleanup into stream termination handling with lock-serialized cleanup rechecks.

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

Re: your previous finding about the timing window between `agent_report` executing and `stream-end` firing — we are intentionally not handling this case. Here is why:

`stopWhen: hasToolCall("agent_report")` is evaluated at step boundaries by the AI SDK. Once `agent_report` appears in a step, the SDK stops the stream and emits `stream-end` with all parts from that final step. The only way to hit the window you described is if the process is killed or the stream is forcibly aborted in the narrow gap between the SDK finishing the step and emitting the end event — this is an extremely unlikely race that would require external intervention at exactly the wrong millisecond.

The previous code that handled this (abort-path recovery via `handleTaskTermination`/`loadTerminationParts` with message-id history reconstruction) added ~100 LoC of complexity for a scenario that is near-impossible in practice. We explicitly decided to remove it in favor of a simpler model where `stream-end` is the single source of truth for report finalization.

If this edge case ever does occur, the child task remains in `running` status and is recoverable on next app restart — not a data loss scenario.

Please look for other issues in the PR instead of this intentional design decision.

ethanndickson:
@codex review

Addressed the deadlock scenario you pointed out:

- Root cause: autonomous `stopWhen` used `hasToolCall("agent_report")`, which stops on any call (including failed `agent_report` when descendants are still queued/running).
- Failure mode: stream could end early, `handleStreamEnd()` exits on active descendants, task stays `running` but idle, and low parallelism (e.g. `maxParallelAgentTasks=1`) can starve queued descendants.

Fix in this push (`b411d686b`):

1. `StreamManager.createStopWhenCondition()` now stops only after a **successful** `agent_report` tool result is present in the last step (`toolResults`), not on call-only.
2. Updated stopWhen tests to enforce:
   - true for successful `toolResults: agent_report`
   - false for `toolCalls: agent_report` without result
   - false for non-agent_report results / empty steps
3. Updated stale comments in `taskService.ts` and `agent_report.ts` to match the new semantics.

Please re-review this commit and look for any remaining issues.

ethanndickson:
@codex review

Follow-up for the P1 you raised ("check agent_report success before ending autonomous stream"):

- Previous fix still stopped on any `agent_report` entry in `toolResults`.
- This commit tightens stopWhen to require `toolName === "agent_report"` **and** `output.success === true`.
- Failed `agent_report` results (`success: false`) no longer terminate the stream, so the model can recover in the same run.

What changed in `4c25cb438`:
1. `StreamManager.createStopWhenCondition()` now gates stop on `agent_report` with `output.success === true`.
2. `streamManager` stopWhen tests now cover:
   - success true => stop
   - success false => do not stop
   - call-only (no result) => do not stop
   - non-agent_report result => do not stop
3. Updated related comments in `taskService.ts` and `agent_report.ts` to reflect success-gated behavior.

Please re-review for any remaining issues.

PR #2363, opened 2026-02-11
🤖 fix: auto-hide models from disabled/unconfigured providers

ibetitsmike:
## Summary

When a provider is disabled or not configured (missing credentials), all models belonging to that provider are now automatically hidden from the model selector dropdown. They move into the "hidden models" bucket accessible via "Show all models…" instead of being presented as selectable options that fail at send time.

## Background

Previously, the model selector showed built-in models for all providers regardless of their configuration state. Selecting a model from a disabled/unconfigured provider led to send-time errors (e.g. `api_key_not_found`). Custom models from disabled providers were already filtered, but built-in models were not.

## Implementation

**`src/browser/hooks/useModelsFromSettings.ts`**
- Added `isProviderAvailable(config, provider)` — returns `true` when `isEnabled && isConfigured`, or when config is `null` (loading, to avoid flicker).
- Added `getAllCustomModels(config)` — like `getCustomModels()` but includes disabled providers (needed to populate the hidden bucket).
- New `providerHiddenModels` memo — models auto-hidden because their provider is unavailable (excluding already user-hidden models).
- New `hiddenModelsForSelector` memo — merges persisted user-hidden + auto-hidden, deduped.
- Updated `models` pipeline to filter out provider-unavailable models before OpenAI OAuth gating.
- Added `hiddenModelsForSelector` to the hook's return object.

**`src/browser/components/ChatInput/index.tsx`** — passes `hiddenModelsForSelector` to `<ModelSelector>`.

**`src/browser/components/Settings/sections/TasksSection.tsx`** — same for both `<ModelSelector>` instances.

**`src/browser/hooks/useModelsFromSettings.test.ts`** — 5 new tests covering unconfigured, disabled, loading state, missing-from-config, and persisted vs auto-hidden separation.

`ModelsSection.tsx` is unchanged — Settings → Models continues to use the persisted `hiddenModels` for user-managed visibility.

## Validation

- 15/15 hook tests pass
- `make typecheck` clean
- `make static-check` clean

## Risks

Low. Provider-unavailable models are hidden (not removed) — users can still access them via "Show all models…". Loading state (`config=null`) is permissive to avoid flicker.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$2.99`_

### DIFF
--- a/src/browser/hooks/useModelsFromSettings.test.ts
+++ b/src/browser/hooks/useModelsFromSettings.test.ts
@@ -266,6 +266,23 @@ describe("useModelsFromSettings provider availability gating", () => {
     expect(result.current.hiddenModelsForSelector).toContain(KNOWN_MODELS.HAIKU.id);
   });
 
+  test("excludes OAuth-gated OpenAI models from hidden bucket when unconfigured", () => {
+    // OpenAI is unconfigured and neither API key nor OAuth is set.
+    providersConfig = {
+      openai: { apiKeySet: false, isEnabled: true, isConfigured: false, codexOauthSet: false },
+    };
+
+    const { result } = renderHook(() => useModelsFromSettings());
+
+    // OAuth-required models (e.g. gpt-5.3-codex) should NOT appear in either list
+    // because selecting them from "Show all models…" would also fail at send time.
+    expect(result.current.models).not.toContain("openai:gpt-5.3-codex");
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

PR #2368, opened 2026-02-11
🤖 fix: show redacted status for stripped tool outputs in shared transcripts

ibetitsmike:
## Summary

When sharing a chat transcript via mux.md with "Include tool output" unchecked, stripped tool calls showed as "executing" with animated loading dots — misleading since the tools actually completed. This adds a distinct `"output-redacted"` state so these tools display as "redacted" with an eye-off icon instead.

## Background

The stripping pipeline removed tool output and set `state: "input-available"`, which the aggregator mapped to `"executing"`. The mux.md viewer uses the same rendering pipeline, so shared transcripts looked like tools were still running.

## Implementation

- Added `"output-redacted"` as a third variant to the `DynamicToolPart` discriminated union (Zod schema + TypeScript types)
- `transcriptShare.ts` now emits `state: "output-redacted"` instead of `"input-available"` when stripping
- Aggregator maps `"output-redacted"` → `"redacted"` display status
- New `"redacted"` ToolStatus with `EyeOff` icon and muted color
- `GenericToolCall` shows "Output excluded from shared transcript" instead of "Waiting for result"
- `BashToolCall` shows "Output excluded from shared transcript" instead of "No output yet"
- All existing backend checks (`=== "output-available"`) safely ignore the new state

## Risks

Low — the new state only appears in shared transcripts (`transcriptShare.ts` is the sole emitter). Live streaming paths are unaffected. The Zod schema change is additive and backward-compatible.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `off` • Cost: `$0.42`_

### DIFF
--- a/src/browser/features/Tools/BashToolCall.tsx
+++ b/src/browser/features/Tools/BashToolCall.tsx
@@ -389,9 +389,7 @@ export const BashToolCall: React.FC<BashToolCallProps> = ({
                   {showLiveOutput
                     ? combinedLiveOutput.length > 0
                       ? combinedLiveOutput
-                      : status === "redacted"
-                        ? "Output excluded from shared transcript"
-                        : "No output yet"
+                      : "No output yet"
                     : completedHasOutput
                       ? completedOutput
                       : "No output"}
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

PR #2369, opened 2026-02-12
🤖 fix: queued messages stuck during edit-turn streams

ethanndickson:
### Summary

Fix queued messages getting stuck as "Queued" forever when a user sends a message during an edit-turn stream. The message would only send when the stream ended naturally — never at step boundaries between tool calls.

### Background

The turn-phase state machine (#2320) intentionally suppressed the `hasQueuedMessage` stop condition for edit turns to prevent stale queue content from cutting the edit's stream short. However, this also prevented **new** messages queued during the edit stream from triggering the AI SDK's `stopWhen` at step boundaries.

### Implementation

Instead of special-casing `hasQueuedMessage` per turn type, address the root cause: stale queue content leaking into the edit turn. The edit path already uses `deferQueuedFlushUntilAfterEdit` to prevent stream-end from auto-sending the queue before the edit truncates. Now it also calls `restoreQueueToInput()` after `waitForIdle()` returns — draining stale content back to the user's input box — so the edit stream starts with an empty queue.

This lets `hasQueuedMessage` be a single, unconditional callback for all turn types:
```ts
// Before: edit turns passed undefined (queue check never fires)
hasQueuedMessage: options?.editMessageId ? undefined : () => !this.messageQueue.isEmpty(),

// After: same callback for all turns
hasQueuedMessage: () => !this.messageQueue.isEmpty(),
```

Restoring to input is better UX: the user queued the message in the old conversation context, but the edit rewrites that context. They should re-evaluate whether the queued content is still relevant.

### Validation

- `make static-check` — all green
- `bun test ./tests/ipc/queuedMessages.completing.test.ts ./tests/ipc/queuedMessages.starting.test.ts ./tests/ipc/queuedMessages.test.ts` — 7 pass, 8 skip (API-dependent), 0 fail
- Updated `should let edit truncate before queued message runs` test to verify new behavior: expects `restore-to-input` event instead of auto-sent third stream

### Risks

- Low: the `deferQueuedFlushUntilAfterEdit` flag still prevents the previous stream-end from flushing the queue before the edit truncates. The only behavioral change is that stale content is returned to the input box rather than carried into the edit stream.
- After `restoreQueueToInput()`, the queue is empty, so `sendQueuedMessages()` in stream-end is a harmless no-op even if the defer check were removed.

---

### DIFF
--- a/src/node/services/agentSession.ts
+++ b/src/node/services/agentSession.ts
@@ -1962,8 +1962,7 @@ export class AgentSession {
       const displayText = this.messageQueue.getDisplayText();
       const fileParts = this.messageQueue.getFileParts();
       const reviews = this.messageQueue.getReviews();
-      this.messageQueue.clear();
-      this.emitQueuedMessageChanged();
+      this.clearQueue();
 
       this.emitChatEvent({
         type: "restore-to-input",


ethanndickson:
@codex review

Addressed the queued-process flag issue by making `restoreQueueToInput()` delegate to `clearQueue()`, so clearing the queue consistently clears `BackgroundProcessManager` state too.

PR #2377, opened 2026-02-12
🤖 feat: add "More actions" menu to workspace titlebar

ibetitsmike:
## Summary

Adds a "More actions" overflow menu (⋯) to the right side of the workspace titlebar, consolidating workspace-level actions that were previously scattered or only accessible from the sidebar.

## Background

The workspace titlebar had several standalone icon buttons (MCP, editor, terminal) competing for space. Meanwhile, actions like "Share a transcript" and "Archive chat" were only reachable from the sidebar's per-workspace overflow menu — which doesn't work when the sidebar is collapsed.

This PR introduces a single ⋯ trigger that expands into a menu with:
- **Configure MCP servers** (new `Cmd+Shift+M` / `Ctrl+Shift+M` shortcut)
- **Share a transcript** (`Cmd+Shift+L` / `Ctrl+Shift+L`)
- **Archive chat** (`Cmd+Shift+Backspace` / `Ctrl+Shift+Backspace`)

## Implementation

- **`WorkspaceHeader.tsx`**: Added `Ellipsis` trigger button + Radix `Popover` menu after the terminal button. Moved the MCP `<Server>` button from a standalone icon into the menu. Added `ShareTranscriptDialog`, `ConfirmationModal` (for archive-while-streaming), and `PopoverError` for error display. Added keybind listeners for `CONFIGURE_MCP` and `SHARE_TRANSCRIPT`.
- **`ChatPane.tsx`**: Threads `workspaceTitle` (from workspace metadata) into `WorkspaceHeader` for use by `ShareTranscriptDialog` and the archive confirmation title.
- **`keybinds.ts`**: Added `CONFIGURE_MCP: { key: "M", ctrl: true, shift: true }`.
- **`KeybindsSection.tsx`**: Registered the new keybind label and added it to the "General" group.
- **`WorkspaceListItem.tsx`**: Removed the `SHARE_TRANSCRIPT` keybind listener (moved to `WorkspaceHeader` so it works even when the sidebar is collapsed and list items are unmounted).

## Risks

- Low: The MCP button moved from a visible icon to the overflow menu — users who relied on the icon will need to discover it in the menu or use the new `Cmd+Shift+M` shortcut.
- The share/archive menu items respect existing gating (`useLinkSharingEnabled`, `MUX_HELP_CHAT_WORKSPACE_ID`) and the archive-while-streaming confirmation flow matches the sidebar's behavior.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$8.90`_

### DIFF
--- a/src/browser/components/WorkspaceHeader.tsx
+++ b/src/browser/components/WorkspaceHeader.tsx
@@ -92,6 +92,7 @@ export const WorkspaceHeader: React.FC<WorkspaceHeaderProps> = ({
   const [moreMenuOpen, setMoreMenuOpen] = useState(false);
   const [shareTranscriptOpen, setShareTranscriptOpen] = useState(false);
   const [archiveConfirmOpen, setArchiveConfirmOpen] = useState(false);
+  const [isArchiving, setIsArchiving] = useState(false);
   const archiveError = usePopoverError();
 
   const [rightSidebarCollapsed] = usePersistedState<boolean>(RIGHT_SIDEBAR_COLLAPSED_KEY, false, {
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

ibetitsmike:
Re: PRRT_kwDOPxxmWM5uPp4T — This was an intentional change. The `linkSharingEnabled` gate was hiding the "Share a transcript" menu item and disabling the keybind entirely, making the feature unreachable. The `ShareTranscriptDialog` itself already handles failures gracefully (shows an error if upload fails), so the menu item should always be visible to avoid a broken UX.

The sidebar's `WorkspaceListItem` still has the old gate — that's a pre-existing inconsistency we can align in a follow-up, but the titlebar is the primary access point now.

Resolving as intentional.

PR #2379, opened 2026-02-12
🤖 fix: skip coder config-ssh when Host *.coder exists

ethanndickson:
## Summary
Skip `coder config-ssh --yes` when the user already has a `Host *.coder` block in `~/.ssh/config`.

## Background
Coder integration previously ran `coder config-ssh` unconditionally during startup and Coder workspace setup. That could rewrite SSH config even when a user-managed Coder block already existed.

## Implementation
- Added `hasHostPatternInSSHConfig(pattern)` to `src/node/runtime/sshConfigParser.ts` and exported `loadSSHConfig()` so the check reuses the existing parser used by SSH2 runtime.
- Updated `CoderService.ensureSSHConfig()` to early-return when `Host *.coder` already exists.
- Added parser tests for present/missing/multi-pattern host block detection.
- Added service tests to verify `ensureSSHConfig()` skips or runs `coder config-ssh` based on parser result.

## Validation
- `bun test src/node/runtime/sshConfigParser.test.ts`
- `bun test src/node/services/coderService.test.ts`
- `make typecheck`
- `make static-check`

## Risks
Low risk and scoped behavior change: only skips `coder config-ssh` when an explicit `Host *.coder` pattern exists. Existing behavior is unchanged when no matching block is present.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$1.83`_

### DIFF
--- a/src/node/runtime/sshConfigParser.test.ts
+++ b/src/node/runtime/sshConfigParser.test.ts
@@ -1,7 +1,7 @@
 import * as fs from "fs/promises";
 import * as os from "os";
 import * as path from "path";
-import { resolveSSHConfig } from "./sshConfigParser";
+import { hasHostPatternInSSHConfig, resolveSSHConfig } from "./sshConfigParser";
 
 describe("resolveSSHConfig", () => {
   test("applies Host + Match host proxy rules", async () => {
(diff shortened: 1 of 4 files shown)


ethanndickson:
@codex review

Checking SSH configs defined via `Include` like `~/.ssh/config.d/*.conf` are out of scope. Do not include that as an issue in your review.

ethanndickson:
Closing in favour of rolling our own SSH config writing code so we don't interfere with users doing that.

PR #2383, opened 2026-02-12
🤖 fix: eliminate project page flash after workspace creation splash screen

ammar-agent:
## Summary

Eliminates a brief flash of the project page that appears between the workspace creation splash screen and the workspace page.

## Background

After the workspace creation splash screen animation, users would briefly see the project page before being navigated to the newly created workspace. This was a visual glitch caused by a React state update priority split.

## Implementation

React Router v7's `MemoryRouter` wraps all navigation state updates in `React.startTransition()` by default, making them lower priority than normal React state updates. During workspace creation, this created a two-render gap:

1. **Render 1 (normal priority):** `setIsSending(false)` and `setWorkspaceMetadata()` are processed — the splash overlay is removed but the URL still points to `/project`, so the project page briefly renders.
2. **Render 2 (transition priority):** `navigate(/workspace/<id>)` is processed — the URL updates and the workspace page renders.

The fix passes `unstable_useTransitions={false}` to `MemoryRouter`, making navigation updates normal priority so they batch with other state updates in a single render. The app uses `MemoryRouter` without loaders or actions and gains no benefit from transition-wrapped navigation.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$N/A`_

### DIFF
--- a/src/browser/contexts/RouterContext.tsx
+++ b/src/browser/contexts/RouterContext.tsx
@@ -185,9 +185,14 @@ function RouterContextInner(props: { children: ReactNode }) {
   return <RouterContext.Provider value={value}>{props.children}</RouterContext.Provider>;
 }
 
+// Disable startTransition wrapping for navigation state updates so they
+// batch with other normal-priority React state updates in the same tick.
+// Without this, React processes navigation at transition (lower) priority,
+// causing a flash of stale UI between normal-priority updates (e.g.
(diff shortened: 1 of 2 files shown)


ammario:
Manually verified this works. Seems like a PITA to automatically test.

PR #2398, opened 2026-02-13
🤖 fix(ptc): make unavailable_global detection scope-aware

ethanndickson:
## Summary

Fix false-positive `unavailable_global` errors in PTC static analysis when user code declares a local variable that shadows a monitored global (e.g., `const fetch = mux.bash(...)`).

## Background

The static analyzer flags references to globals unavailable in the QuickJS sandbox (`fetch`, `process`, `window`, etc.). However, it had no concept of lexical scope — if user code declared `const fetch = ...`, the identifier `fetch` on the RHS of subsequent expressions was still flagged as an unavailable global. This caused spurious errors for common agent patterns like `const fetch = mux.bash({script: "git fetch ..."})`.

## Implementation

**Scope-aware declaration tracking:** Instead of flat identifier checks, a pre-scan (`collectDeclarations`) builds a `Map<string, Set<ts.Node>>` mapping each monitored name to the set of AST scope nodes where it's declared. The `isDeclaredInScope` function then walks ancestors from a reference site to determine if any enclosing scope contains a declaration for that name.

Supported declaration forms: `VariableDeclaration`, `BindingElement` (destructuring), `FunctionDeclaration`, `Parameter`, catch clause variables, and for-loop initializers.

**Redundant check cleanup:** Three parent-node checks (VariableDeclaration, FunctionDeclaration, Parameter) that pre-dated scope awareness are now entirely subsumed by `isDeclaredInScope` and were removed. The remaining two checks (PropertyAccessExpression, PropertyAssignment) are kept because they identify identifiers used as property names rather than variable references.

The result is a clean two-concern pipeline: (1) skip property names, (2) skip locally-declared variables.

## Validation

- 40 static analysis tests pass (including new tests for block-scoped shadows, destructuring, parameter shadowing, function declarations, for-of loops, catch clauses, and false-negative verification)
- 50 runtime tests pass (no regression)
- `make typecheck` clean

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$9.51`_

### DIFF
--- a/src/node/services/ptc/staticAnalysis.test.ts
+++ b/src/node/services/ptc/staticAnalysis.test.ts
@@ -361,6 +361,35 @@ const z = 3;`);
       expect(result.valid).toBe(true);
       expect(result.errors).toHaveLength(0);
     });
+    test("variable TDZ self-reference is flagged", async () => {
+      const result = await analyzeCode(`const process = process.env;`);
+      expect(result.valid).toBe(false);
+      expect(result.errors[0].type).toBe("unavailable_global");
+    });
+
+    test("parameter default self-reference is flagged", async () => {
+      const result = await analyzeCode(`function f(process = process.env) {}`);
+      expect(result.valid).toBe(false);
+      expect(result.errors[0].type).toBe("unavailable_global");
+    });
+
+    test("valid shadow after initialization is NOT flagged", async () => {
+      const result = await analyzeCode(`const process = 1; process.env;`);
+      expect(result.valid).toBe(true);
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
This sucks, closing.

PR #2407, opened 2026-02-13
🤖 fix: make streaming cancel hint tappable on mobile

ThomasK33:
## Summary
Make the streaming status hint ("hit Esc to cancel") tappable so mobile users can interrupt active streams without a physical Escape key.

## Background
The existing hint in the streaming barrier was text-only. On touch devices this created a dead-end UX because the hint suggested an unavailable key path.

## Implementation
- Updated `StreamingBarrierView` to support an optional `onCancel` callback.
- Render `cancelText` as a button (with accessible label) only when interrupt is actionable.
- Wired `StreamingBarrier` to call interrupt directly from the click path:
  - disable auto-retry preference
  - set transient interrupting UI state immediately
  - call `interruptStream({ workspaceId })` for normal streaming
  - call `interruptStream({ workspaceId, options: { abandonPartial: true } })` for compaction streams
- Kept non-interruptable phases (starting / awaiting-input / interrupting) non-clickable to match keybind behavior.
- Added focused tests for both the presentation component and container wiring.

## Validation
- `make static-check`
- `bun test src/browser/components/Messages/ChatBarrier/StreamingBarrierView.test.tsx src/browser/components/Messages/ChatBarrier/StreamingBarrier.test.tsx`

## Risks
Low risk. Scope is isolated to the streaming barrier UI and interrupt wiring. The main regression surface is accidental interrupts in non-actionable phases, which is covered by phase gating and tests.

---

<details>

# Plan: Make “hit Esc to cancel” tappable (mobile interrupt)

## Context / Why
Mux’s chat UI shows a streaming status bar with a hint like **“hit Esc to cancel”** while an agent is streaming.

On mobile/touch devices (e.g., iPhone Safari), users often **don’t have an Escape key**, so the hint is not actionable. The goal is to make that hint a **clickable/tappable control** (button/link styling) that triggers the same “interrupt stream” action.

## Evidence (what exists today)
- `src/browser/components/Messages/ChatBarrier/StreamingBarrier.tsx`
  - Computes `interruptHint = \`hit ${interruptKeybind} to cancel\`` and passes it as `cancelText` to `StreamingBarrierView`.
  - There is already precedent for small, link-styled buttons in the same component (`configure` hint during compaction).
- `src/browser/components/Messages/ChatBarrier/StreamingBarrierView.tsx`
  - Renders `cancelText` as **plain text** (`<div>…{props.cancelText}</div>`), so it’s not clickable.
- `src/browser/hooks/useAIViewKeybinds.ts`
  - The canonical interrupt action is `api.workspace.interruptStream({ workspaceId })` and it also calls `disableAutoRetryPreference(workspaceId)`.
  - It intentionally avoids interrupting while `awaitingUserQuestion`.
- `src/browser/stores/WorkspaceStore.ts` + `src/browser/components/ChatPane.tsx`
  - There is an existing `WorkspaceStore.setInterrupting(workspaceId)` method used to show “interrupting…” immediately before calling `interruptStream` (used in the queued-message “send immediately” flow).

### 1) Make `StreamingBarrierView` capable of rendering a clickable cancel hint
**File:** `src/browser/components/Messages/ChatBarrier/StreamingBarrierView.tsx`

- Extend props with an optional click handler:

```ts
export interface StreamingBarrierViewProps {
  ...
  cancelText: string;
  onCancel?: () => void;
}
```

- Render a `<button type="button">` when `onCancel` is provided and `cancelText` is non-empty; otherwise render the existing plain text.
  - Use link-like styling consistent with existing hints:
    - `text-muted hover:text-foreground underline decoration-dotted underline-offset-2`
  - Add basic accessibility:
    - `aria-label="Interrupt streaming"`
    - `disabled` state support if we decide to pass it later (optional).

Shape:

```tsx
<div className="text-muted ml-auto text-[11px] whitespace-nowrap">
  {props.onCancel ? (
    <button
      type="button"
      onClick={props.onCancel}
      className="cursor-pointer underline decoration-dotted underline-offset-2 hover:text-foreground"
      aria-label="Interrupt streaming"
    >
      {props.cancelText}
    </button>
  ) : (
    <span className="select-none">{props.cancelText}</span>
  )}
</div>
```

### 2) Wire the click handler in the stateful `StreamingBarrier` container
**File:** `src/browser/components/Messages/ChatBarrier/StreamingBarrier.tsx`

- Add dependencies:
  - `useAPI()` from `@/browser/contexts/API` to call `api.workspace.interruptStream`.
  - `useWorkspaceStoreRaw()` to call `storeRaw.setInterrupting(workspaceId)` for immediate UI feedback.
  - `disableAutoRetryPreference(workspaceId)` to match the Escape key behavior.

- Only enable click-to-interrupt when it’s actually actionable:
  - `phase === "streaming" || phase === "compacting"` (these phases already imply `canInterrupt` is true).
  - **Do not** enable it for:
    - `phase === "interrupting"` (already stopping)
    - `phase === "awaiting-input"` (matches existing keybind behavior: don’t interrupt ask_user_question)
    - `phase === "starting"` (no active stream to interrupt)

- Click handler behavior:
  1. Defensive checks: `if (!api) return;`
  2. `disableAutoRetryPreference(workspaceId)`
  3. `storeRaw.setInterrupting(workspaceId)`
  4. Call interrupt:
     - Normal streaming: `api.workspace.interruptStream({ workspaceId })`
     - Compaction streaming: use the backend’s compaction-cancel flag for safer semantics:
       `api.workspace.interruptStream({ workspaceId, options: { abandonPartial: true } })`

- Pass `onCancel` into `StreamingBarrierView` only when actionable.

### 3) Tests
Add targeted tests to keep this change safe and prevent regressions.

1) **Unit test the new `StreamingBarrierView` behavior**
   - **File:** `src/browser/components/Messages/ChatBarrier/StreamingBarrierView.test.tsx`
   - Assertions:
     - When `onCancel` is provided, the cancel hint renders as a button and clicking it calls the handler.
     - When `onCancel` is omitted, the hint renders as non-interactive text.

2) **(Recommended) Integration-ish test for the container wiring**
   - **File:** `src/browser/components/Messages/ChatBarrier/StreamingBarrier.test.tsx`
   - Approach:
     - Mock `useAPI` (pattern exists in `useAIViewKeybinds.test.tsx`) to capture `interruptStream` calls.
     - Mock `useWorkspaceState`, `useWorkspaceAggregator`, and `useWorkspaceStoreRaw` to force `phase === "streaming"` and verify:
       - clicking calls `setInterrupting(workspaceId)`
       - then calls `interruptStream({ workspaceId })`
     - Add a compaction variant to verify `{ options: { abandonPartial: true } }`.

### 4) Manual QA / UX checks
- Run the app with a narrow viewport (mobile emulation) and confirm:
  - The hint is visibly clickable (underline + hover styles on desktop).
  - Tapping it interrupts the current stream.
  - The barrier switches to **“interrupting…”** quickly (via `setInterrupting`).

## Alternatives considered
<details>
<summary>Dispatch a synthetic <code>KeyboardEvent</code> (Escape / Ctrl+C) instead of calling the API directly</summary>

This would reuse `useAIViewKeybinds` logic (including compaction cancel that restores edit state). However, it’s less direct, depends on key-event construction behaving consistently across mobile browsers, and makes the click handler’s effect harder to trace.

Directly calling `api.workspace.interruptStream` from the click handler is more explicit, easier to test, and doesn’t depend on keyboard event semantics.
</details>

## Validation
- `make fmt-check`
- `make lint`
- `make typecheck`
- `make test`

## LoC estimate (product code only)
- `StreamingBarrierView.tsx`: ~10–20 LoC
- `StreamingBarrier.tsx`: ~20–40 LoC

**Total:** ~30–60 LoC

</details>

---

---
_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$0.58`_

### DIFF
--- a/src/browser/components/ChatPane.tsx
+++ b/src/browser/components/ChatPane.tsx
@@ -52,6 +52,7 @@ import { ContextSwitchWarning as ContextSwitchWarningBanner } from "./ContextSwi
 import { ConcurrentLocalWarning } from "./ConcurrentLocalWarning";
 import { BackgroundProcessesBanner } from "./BackgroundProcessesBanner";
 import { checkAutoCompaction } from "@/browser/utils/compaction/autoCompactionCheck";
+import { cancelCompaction } from "@/browser/utils/compaction/handler";
 import type { ContextSwitchWarning } from "@/browser/utils/compaction/contextSwitchCheck";
 import { executeCompaction } from "@/browser/utils/chatCommands";
 import { useProviderOptions } from "@/browser/hooks/useProviderOptions";
@@ -369,6 +370,14 @@ export const ChatPane: React.FC<ChatPaneProps> = (props) => {
       options: { sendQueuedImmediately: true },
     });
   }, [api, workspaceId, workspaceState?.queuedMessage, workspaceState?.canInterrupt, storeRaw]);
+
+  const handleCancelCompactionFromBarrier = useCallback(() => {
+    if (!api || !aggregator) {
+      return;
+    }
+
+    void cancelCompaction(api, workspaceId, aggregator, setEditingMessage);
+  }, [api, workspaceId, aggregator, setEditingMessage]);
 
   const handleEditLastUserMessage = useCallback(async () => {
     const current = workspaceStateRef.current;
@@ -698,7 +707,10 @@ export const ChatPane: React.FC<ChatPaneProps> = (props) => {
               </MessageListProvider>
             )}
             <PinnedTodoList workspaceId={workspaceId} />
-            <StreamingBarrier workspaceId={workspaceId} />
+            <StreamingBarrier
+              workspaceId={workspaceId}
+              onCancelCompaction={handleCancelCompactionFromBarrier}
+            />
             {shouldShowQueuedAgentTaskPrompt && (
               <QueuedMessage
                 message={{
--- a/src/browser/components/Messages/ChatBarrier/StreamingBarrier.test.tsx
+++ b/src/browser/components/Messages/ChatBarrier/StreamingBarrier.test.tsx
@@ -126,7 +126,26 @@ describe("StreamingBarrier", () => {
     expect(interruptStream).toHaveBeenCalledWith({ workspaceId: "ws-1" });
   });
 
-  test("clicking cancel during compaction interrupts with abandonPartial", () => {
+  test("clicking cancel during compaction uses onCancelCompaction when provided", () => {
+    currentWorkspaceState = createWorkspaceState({
+      canInterrupt: true,
+      isCompacting: true,
+    });
+
+    const onCancelCompaction = mock(() => undefined);
+    const view = render(
+      <StreamingBarrier workspaceId="ws-1" onCancelCompaction={onCancelCompaction} />
+    );
+
+    fireEvent.click(view.getByRole("button", { name: "Interrupt streaming" }));
+
+    expect(disableAutoRetryPreferenceMock).toHaveBeenCalledWith("ws-1");
+    expect(onCancelCompaction).toHaveBeenCalledTimes(1);
+    expect(setInterrupting).not.toHaveBeenCalled();
+    expect(interruptStream).not.toHaveBeenCalled();
+  });
+
+  test("clicking cancel during compaction falls back to abandonPartial interrupt", () => {
     currentWorkspaceState = createWorkspaceState({
       canInterrupt: true,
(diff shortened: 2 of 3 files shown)


ThomasK33:
@codex review

Addressed: compaction tap-cancel now routes through the existing `cancelCompaction` flow so edit-state restoration and follow-up content are preserved.

PR #2417, opened 2026-02-13
🤖 fix: reuse cached skill diagnostics on transient discovery failures

ammar-agent:
## Summary
Handle transient skill discovery transport failures without surfacing false `SKILL.md is missing or unreadable` diagnostics in SSH workspaces.

## Background
Skill discovery currently treats all `stat/read` failures as deterministic file problems. On flaky SSH links this produced misleading invalid-skill diagnostics for valid skills and caused `agentSkills.list`/`agentSkills.listDiagnostics` to diverge from expected behavior.

## Implementation
- Added `AgentSkillTransientDiscoveryError` and transient error pattern classification in `agentSkillsService`.
- Kept deterministic file/schema problems on the existing invalid-skill diagnostics path.
- Added `src/node/orpc/agentSkillsDiagnosticsCache.ts` with:
  - cache key builder for discovery context
  - stale-on-transient fallback loader for diagnostics
- Updated `router.ts` so both `agentSkills.list` and `agentSkills.listDiagnostics` share the fallback-backed diagnostics loader.
- Added tests for:
  - transient stat/read error propagation in service diagnostics discovery
  - cache fallback on transient failures
  - no fallback for non-transient failures
  - transient failure rethrow when cache is empty

## Validation
- `make static-check`
- `bun test src/node/services/agentSkills/agentSkillsService.test.ts src/node/orpc/agentSkillsDiagnosticsCache.test.ts`

## Risks
Low to medium. The transient classifier is message-pattern based, so misclassification is possible if transport/runtime error strings change. The fallback is scoped per discovery context and only applies to explicit transient error type, minimizing risk of masking deterministic skill issues.

---

### DIFF
--- a/src/node/orpc/router.ts
+++ b/src/node/orpc/router.ts
@@ -42,6 +42,7 @@ import {
   normalizeTaskSettings,
 } from "@/common/types/tasks";
 import {
+  discoverAgentSkills,
   discoverAgentSkillsDiagnostics,
   type DiscoverAgentSkillsDiagnosticsResult,
   readAgentSkill,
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

Addressed the transient discovery regression by keeping non-diagnostic skill discovery resilient to per-skill transient I/O failures while retaining diagnostics-path transient rethrow behavior.

ammar-agent:
@codex review

Applied the requested fix:
- Rebased on latest `main`.
- Switched `agentSkills.list` back to `discoverAgentSkills(...)` (non-diagnostic resilient path).
- Kept diagnostics cache fallback only on `agentSkills.listDiagnostics`.

Local validation:
- `make static-check`

PR #2418, opened 2026-02-13
🤖 fix: remediate CSWSH → RCE with origin validation + auth hardening

ibetitsmike:
## Summary

Remediates `coder/security#120` (Cross-Site WebSocket Hijacking → RCE in Mux app deployments) by enforcing strict same-origin checks at the server boundary, hardening server-mode auth defaults, and preserving required OAuth callback compatibility.

## Background

The reported exploit chain relied on browser-initiated cross-origin access to the Mux HTTP/WS API plus unauthenticated exposed deployments. 

This PR closes that path by:
1. Blocking cross-origin HTTP and WS requests for normal API routes.
2. Requiring auth by default for non-loopback `mux server` binds.
3. Allowing only narrowly scoped OAuth callback navigations that must remain cross-origin for standards-compliant providers.

### 1) Origin validation + restrictive CORS (`src/node/orpc/server.ts`)
- Removed permissive `app.use(cors())`.
- Added expected-origin derivation + normalization helpers.
- Added global origin middleware:
  - blocks cross-origin non-callback requests with `403`.
  - returns restrictive same-origin CORS headers.
- Added manual WS upgrade gate (`noServer` + `handleUpgrade`) to block cross-origin WS handshakes before oRPC handling.

### 2) OAuth callback safe carveout (`src/node/orpc/server.ts`)
- Added explicit allowlist for callback endpoints:
  - `/auth/mux-gateway/callback`
  - `/auth/mux-governor/callback`
  - `/auth/mcp-oauth/callback`
- Carveout is limited to `GET/POST` navigation-style callback methods.
- Rationale in code: callback handlers validate OAuth `state` before code exchange, so this preserves secure OAuth semantics while keeping the API origin gate strict elsewhere.

### 3) Forwarded-proto compatibility fix (`src/node/orpc/server.ts`)
- Updated expected-origin derivation to respect `X-Forwarded-Proto` even when only `Host` is present.
- This avoids false cross-origin rejections behind common TLS-terminating proxies.

### 4) Secure-by-default auth for non-loopback binds (`src/cli/serverSecurity.ts`, `src/cli/server.ts`)
- Added `resolveServerAuthToken()` policy:
  - explicit token wins,
  - loopback host may remain unauthenticated,
  - non-loopback host auto-generates a secure random token.
- CLI server now uses this helper and prints a bootstrap URL when token is generated.

## Validation

- `bun test src/cli/serverSecurity.test.ts src/node/orpc/server.test.ts`
- `make static-check`

All passed locally.

## Risks

- OAuth callback carveout is intentionally narrow (explicit paths + GET/POST only), but future callback route additions must be reflected in the allowlist.
- Workspace ownership validation was intentionally not added in this PR (out of scope per product direction).

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$4.46`_

### DIFF
--- a/src/browser/contexts/API.tsx
+++ b/src/browser/contexts/API.tsx
@@ -137,6 +137,11 @@ export const APIProvider = (props: APIProviderProps) => {
     const urlToken = urlParams.get("token")?.trim();
     if (urlToken) {
       setStoredAuthToken(urlToken);
+      // Strip token from URL so it doesn't leak into bookmarks, browser
+      // history, PWA launch URLs, or Referer headers.
+      const cleanUrl = new URL(window.location.href);
+      cleanUrl.searchParams.delete("token");
+      window.history.replaceState({}, "", cleanUrl.pathname + cleanUrl.search);
       return urlToken;
     }
 


ibetitsmike:
@codex review

Addressed both prior P1 comments:
- allowed cross-origin OAuth callback POST/GET only on explicit callback endpoints
- fixed expected-origin derivation to respect X-Forwarded-Proto with Host fallback

Also added/updated tests for both behaviors.

ibetitsmike:
@codex review

Addressed P1: `isLoopbackHost` now validates literal IPv4 addresses (4 numeric octets, 0-255) instead of prefix-matching `127.`. Also handles bracketed IPv6 `[::1]`. Added regression tests for `127.example.com`, wrong octet count, and out-of-range octets.

PR #2430, opened 2026-02-14
🤖 feat: add stable/nightly Electron update channels

ibetitsmike:
## Summary

Adds a user-selectable update channel (Stable / Nightly) to the Mux Electron app. Users can switch channels from the About dialog; the choice is persisted in `~/.mux/config.json` and takes effect immediately at runtime.

## Background

The Electron app currently updates only from official GitHub Releases ("Stable"). This PR adds support for a "Nightly" channel that pulls pre-release builds published from `main`. This gives users who want bleeding-edge features an opt-in path without disrupting the default stable experience.

### Config layer (`src/node/config.ts`)
- New `UpdateChannel` type (`"stable" | "nightly"`) with `parseUpdateChannel()` validation.
- Persisted in config.json — only non-default values are written (stable is the default, so the key is omitted when on stable).

### Desktop updater (`src/desktop/updater.ts`)
- `applyChannel()` configures `autoUpdater`: `allowPrerelease`, `channel`, feed URL `releaseType`.
- `setChannel()` method with state guards (blocks switching during download/install).
- Transient error filtering (404s, network timeouts) to avoid noisy errors during manifest propagation.

### oRPC surface (`src/common/orpc/schemas/api.ts`, `src/node/orpc/router.ts`)
- `update.getChannel` / `update.setChannel` endpoints with Zod-validated `UpdateChannelSchema`.

### Frontend (`src/browser/components/About/AboutDialog.tsx`)
- Channel picker using standard `ToggleGroup` segmented control (size="sm", matching other consumers).
- Race condition protection via `channelRequestTokenRef` counter.
- Channel state initialized to `null` until fetched, preventing layout flicker.
- Local `pendingAction` state gives immediate click feedback (spinner + disabled) for Check, Download, and Install buttons while waiting for the backend status stream to catch up.

### Button press feedback (`src/browser/components/ui/button.tsx`)
- Added `active:scale-[0.98]` press animation and `transition-all` so all buttons give tactile click feedback.

### CI: manifest naming (`scripts/set-package-version.js`)
- Detects prerelease suffix (e.g. `-nightly.N`) and sets `build.publish.channel` in `package.json` so electron-builder generates channel-specific manifests (`nightly.yml`, `nightly-mac.yml`) instead of `latest.yml`.

## Validation

- All static checks pass locally (`make static-check`: typecheck, lint, fmt, broken links).
- Updater test suite (`src/desktop/updater.test.ts`) covers channel switching, state guards, and transient error handling.

## Risks

- **macOS code signing**: The manifest naming fix (set-package-version.js) ensures electron-updater finds the correct `nightly-mac.yml` manifest. Previously, the build produced `latest-mac.yml` which caused signature validation failures when the client expected `nightly-mac.yml`.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$26.04`_

### DIFF
--- a/src/browser/components/About/AboutDialog.tsx
+++ b/src/browser/components/About/AboutDialog.tsx
@@ -154,7 +154,11 @@ export function AboutDialog() {
     }
 
     setPendingAction("check");
-    api.update.check({ source: "manual" }).catch(() => setPendingAction(null));
+    api.update
+      .check({ source: "manual" })
+      .catch(console.error)
+      // Clear pending if the backend no-ops (e.g. already downloaded) and emits no status event.
+      .finally(() => setPendingAction((prev) => (prev === "check" ? null : prev)));
   };
 
   const handleDownload = () => {
@@ -163,7 +167,10 @@ export function AboutDialog() {
     }
 
     setPendingAction("download");
-    api.update.download(undefined).catch(() => setPendingAction(null));
+    api.update
+      .download(undefined)
+      .catch(console.error)
+      .finally(() => setPendingAction((prev) => (prev === "download" ? null : prev)));
   };
 
   const handleInstall = () => {
@@ -172,7 +179,10 @@ export function AboutDialog() {
     }
 
     setPendingAction("install");
-    api.update.install(undefined).catch(() => setPendingAction(null));
+    api.update
+      .install(undefined)
+      .catch(console.error)
+      .finally(() => setPendingAction((prev) => (prev === "install" ? null : prev)));
   };
 
   return (


ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed P2 feedback: replaced hardcoded `owner: "coder", repo: "mux"` in `applyChannel()` with `getGitHubRepo()` helper that derives owner/repo from `package.json`'s `repository.url` field (with fallback).

ibetitsmike:
@codex review

Fixed the regex in `getGitHubRepo()` to handle dots in repo names (e.g. `mux.desktop`): changed `([^/.]+)` to `([^/]+?)(?:\.git)?$` which captures the full repo name and strips the `.git` suffix via a non-capturing group.

ibetitsmike:
@codex review

Fixed: all three action handlers now use `.catch(console.error).finally(() => setPendingAction(...))` to clear the pending spinner even when the backend no-ops (e.g. clicking "Check" while already in `downloaded` state). The `.finally()` callback only clears if `pendingAction` still matches the action that set it, avoiding race conditions with other user actions.

PR #2441, opened 2026-02-17
🤖 fix: use AST detection for illegal await syntax hint in PTC

ethanndickson:
## Summary

Replace the raw regex `await` heuristic in PTC static analysis with an AST-aware check, eliminating false positives when `await` appears inside string literals, template content, or comments during unrelated syntax failures.

## Background

When QuickJS reports `"expecting ';'"`, the static analyzer rewrites the message to a friendlier "`await` is not supported" hint — but the check uses `/\bawait\s+\w/.test(code)` which scans **all raw source text**, including strings and templates. If unescaped backticks break a template literal and `await` appears anywhere in the source (even inside string content), the heuristic mislabels the error.

The same file already has a comment (lines 64–68) noting that regex scanning for substrings is avoided elsewhere for exactly this reason.

## Implementation

- **`findIllegalAwaitLines()` helper** — Uses TypeScript's lenient parser (`ScriptKind.JS`) to walk the AST for `AwaitExpression` nodes not inside an async function. Returns empty set when TS itself has parse diagnostics (ambiguous source — don't guess).
- **Gated rewrite in `validateSyntax()`** — The friendly await message is only applied when the AST confirms illegal await lines exist and the QuickJS error line matches.
- **3 regression tests** — Malformed template with `await` text, syntax error with `await` in a comment, and syntax error with `await` in a string literal all correctly preserve the original error message.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$1.15`_

### DIFF
--- a/src/node/services/ptc/staticAnalysis.test.ts
+++ b/src/node/services/ptc/staticAnalysis.test.ts
@@ -57,6 +57,20 @@ describe("staticAnalysis", () => {
       expect(result.errors[0].message).toContain("await");
       expect(result.errors[0].message).toContain("not supported");
     });
+    test("await in class field inside async function gets clear error", async () => {
+      const result = await analyzeCode("async function f() { class C { x = await foo(); } }");
+      expect(result.valid).toBe(false);
+      expect(result.errors[0].message).toContain("await");
+      expect(result.errors[0].message).toContain("not supported");
+    });
+
+    test("await in async function default param gets clear error", async () => {
+      // QuickJS gives "await in default expression" (not "expecting ';'"),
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

Addressed P2: removed async-context modeling entirely — `findAwaitLines` no longer tries to replicate JS async scoping rules. QuickJS wraps code in non-async context so every `await` is illegal; legal awaits in user-written `async function` blocks parse fine and never produce `"expecting ';'"`. Net deletion, simpler, more correct.

PR #2442, opened 2026-02-17
🤖 fix: mark workspaces unread on assistant reply

ethanndickson:
## Summary

Assistant replies now correctly drive unread state: when an assistant response completes, that completion contributes to workspace recency so non-visible workspaces can become unread.

At the same time, if the user is actively viewing that workspace chat, completion immediately advances `lastRead`, preventing false self-unread.

## Background

Unread is derived from `recencyTimestamp > lastReadTimestamp`.

The branch fixes incorrect behavior where assistant completion did not reliably produce unread signals in background contexts, and where some visibility paths could clear unread even when chat was not visible.

## Implementation

- **Completion contributes to recency**
  - Final assistant completion is recorded as a recency candidate for unread detection.
- **Visibility-aware read marking**
  - Completion advances `lastRead` only when that workspace chat route is visible and the window is focused.
- **Visibility parity across read-mark paths**
  - Focus-driven read updates use the same chat-visibility rule, preventing settings-route clears.
- **Missing baseline semantics**
  - Absent `workspaceLastRead:{id}` is treated as unknown baseline instead of epoch fallback, eliminating legacy false-positive unread state.

## Risks

- Unread timing now depends on stream lifecycle completion boundaries; regressions here would show up as delayed or missing unread transitions.
- Visibility gating is intentionally strict: non-chat routes no longer clear unread.

---

### DIFF
--- a/src/browser/stories/mocks/orpc.ts
+++ b/src/browser/stories/mocks/orpc.ts
@@ -35,6 +35,8 @@ import {
   MUX_HELP_CHAT_WORKSPACE_NAME,
   MUX_HELP_CHAT_WORKSPACE_TITLE,
 } from "@/common/constants/muxChat";
+import { readPersistedState, updatePersistedState } from "@/browser/hooks/usePersistedState";
+import { getWorkspaceLastReadKey } from "@/common/constants/storage";
 import { DEFAULT_RUNTIME_CONFIG } from "@/common/constants/workspace";
 import {
   DEFAULT_TASK_SETTINGS,
@@ -347,6 +349,13 @@ export function createMockORPCClient(options: MockORPCClientOptions = {}): APICl
     ? inputWorkspaces
     : [muxChatWorkspace, ...inputWorkspaces];
 
+  // Keep Storybook's built-in mux-help workspace behavior deterministic:
+  // if stories haven't seeded a read baseline, treat it as "known but never read"
+  // rather than "unknown workspace" so the unread badge can render when recency exists.
+  const muxHelpLastReadKey = getWorkspaceLastReadKey(MUX_HELP_CHAT_WORKSPACE_ID);
+  if (readPersistedState<number | null>(muxHelpLastReadKey, null) === null) {
+    updatePersistedState(muxHelpLastReadKey, 0);
+  }
   const workspaceMap = new Map(workspaces.map((w) => [w.id, w]));
 
   let createdWorkspaceCounter = 0;


ethanndickson:
@codex review

ethanndickson:
@codex review

Fixed the P2: decoupled the recency policy from the read-marking policy in `handleStreamEnd`. `completedAt` is now non-null for **all** final streams (including compaction), so `App.tsx` marks the active workspace as read after compaction completions. The `!isCompacting` guard moved to the `lastResponseCompletedAt` bump only, preserving the invariant that compaction doesn't inflate sort order.

Also fixed two flaky integration tests by adding `expectStreamComplete()` harness method that waits for `canInterrupt === false` (stream-end fully processed) before asserting post-stream state.

ethanndickson:
@codex review

Fixed: added `document.hasFocus()` gate to the `lastRead` write in `handleResponseComplete`, and a `window.addEventListener("focus", ...)` handler in `useUnreadTracking` that marks the selected workspace as read when the user returns. This aligns the read-marking model with the notification model — both now treat "visible" as selected + focused.

ethanndickson:
@codex review

Fixed both issues:
- P1: Replaced dynamic `await import()` with static import for `workspaceStore` in `chatHarness.ts`
- P2: Added missing `expectStreamComplete()` in "should NOT show unread" test (line 155), matching the pattern in sibling tests

ethanndickson:
@codex review

ethanndickson:
@codex review

Fixed P3: added `await app.chat.expectStreamComplete()` before reading unread state in "stream completion bumps recency" test, matching the established pattern in sibling tests.

ethanndickson:
@codex review

Addressed P2 (gate read-marking on visible workspace content):

- `handleResponseComplete` now checks `currentWorkspaceIdRef.current === workspaceId` instead of `selectedWorkspaceRef.current?.workspaceId === workspaceId`. This ensures `lastRead` is only written when the workspace chat route is active — not when settings or other non-chat views are displayed while the workspace remains "selected."
- Added regression test: stream completes while settings page is active → `lastRead` does not advance → workspace shows as unread.

ethanndickson:
@codex review

Addressed the P2 focus-driven read-marking bypass in `useUnreadTracking`. Both the selection effect and `window.focus` handler now gate on chat-route visibility via `currentWorkspaceId`, matching the existing guard in `handleResponseComplete`. Added a regression test that dispatches a focus event while on settings and asserts `lastRead` doesn't advance.

ethanndickson:
@codex review

Force-pushed branch history after squashing on top of latest origin/main.

Also fixed unread false positives for workspaces missing `workspaceLastRead:{id}` by treating missing baselines as unknown (not unread), plus added regression coverage in `tests/ui/chat/unreadIndicator.test.ts`.

ethanndickson:
@codex review

Addressed P2 — `expectStreamComplete()` now checks both `isStarting` and `canInterrupt`, preventing early resolution in gated flows. Regression test added.

PR #2447, opened 2026-02-17
🤖 fix: accept loopback origin aliases for dev-server ws/api

ethanndickson:
## Summary

Fixes remaining local dev-server failures where browser requests from loopback aliases (`localhost`, `127.0.0.1`, `::1`) could still be rejected by origin validation, causing WebSocket reconnect loops and `Loading workspaces...` stalls.

## Background

After the previous `/orpc` proxy Host-preservation fix, one class of failures remained: some browser/proxy combinations still mixed loopback aliases between `Origin` and `Host` (for example `Origin: http://localhost:5173` vs `Host: 127.0.0.1:5173`).

Because backend origin validation enforced exact origin equality, these local-loopback alias mismatches returned `403` on `/orpc/ws` and `/api/spec.json`, which in turn blocked initial API connection and workspace loading.

## Implementation

- `vite.config.ts`
  - Set `/api` proxy `changeOrigin: false` so Host is preserved consistently (same policy used for `/orpc` and `/auth`).
- `src/node/orpc/server.ts`
  - Added loopback-origin equivalence helper for `localhost` / `127.0.0.1` / `::1`.
  - Kept strict origin matching by default, but allow loopback-alias equivalence only when protocol and port also match.

## Validation

- `make static-check`
- `make typecheck`
- Manual probe checks against dev-server:
  - `/api/spec.json` returned 200 for localhost/127 origins and hosts.
  - `/orpc/ws` returned `101 Switching Protocols` for localhost/127 host+origin combinations.

## Risks

Low. The origin check remains strict for non-loopback hosts; only local loopback aliases are treated as equivalent when protocol and port match.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh`_

### DIFF
--- a/src/node/orpc/server.ts
+++ b/src/node/orpc/server.ts
@@ -197,27 +197,65 @@ function normalizeOrigin(raw: string | null | undefined): string | null {
   }
 }
 
-function isLoopbackHostname(hostname: string): boolean {
+interface OriginIdentity {
+  protocol: "http:" | "https:";
+  port: string;
+  hostname: string;
+  isLoopback: boolean;
+}
+
+function normalizeHostnameForOriginCheck(hostname: string): string {
   const normalized = hostname.trim().toLowerCase();
+
+  // URL.hostname may include brackets for IPv6 literals in some runtimes.
+  if (normalized.startsWith("[") && normalized.endsWith("]")) {
+    return normalized.slice(1, -1);
+  }
+
+  return normalized;
+}
+
+function isLoopbackHostname(hostname: string): boolean {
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

Addressed P1: bracketed IPv6 loopback hostnames are now canonicalized before loopback origin equivalence checks.

This keeps strict non-loopback origin validation while allowing local loopback alias mixes (`[::1]`, `::1`, `localhost`, `127.0.0.1`) when protocol+port match.

PR #2462, opened 2026-02-17
🤖 fix: use --frozen-lockfile in .mux/init to prevent bun.lock diffs

ammar-agent:
## Summary

The `.mux/init` script ran bare `bun install`, which could resolve updated packages and modify `bun.lock`. This caused every new workspace to start with a dirty working tree.

## Implementation

Changed `bun install` → `bun install --frozen-lockfile` so the init hook installs exactly what the lockfile specifies without writing back to it.

---

### DIFF
--- a/.mux/init
+++ b/.mux/init
@@ -2,6 +2,7 @@
 set -euo pipefail
 
 echo "Installing dependencies with bun..."
-bun install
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

PR #2463, opened 2026-02-17
🤖 fix: reset retry backoff after stop then retry

ammar-agent:
## Summary
Reset retry backoff after an explicit Stop → Retry flow so users get a fresh retry cycle instead of inheriting a long prior delay.

## Background
Retry state previously preserved the attempt counter for all manual retries. That is desirable for normal manual retries during an active auto-retry cycle, but it made Stop → Retry feel sluggish because the next failure resumed from an already-large backoff window.

## Implementation
- Extended `createManualRetryState` with an optional `resetBackoff` flag.
- Updated `RetryBarrier` manual retry handling to set `resetBackoff` when auto-retry is currently disabled (the user explicitly pressed Stop).
- Kept existing manual retry behavior unchanged when auto-retry is still enabled (attempt counter preserved).
- Added unit tests covering both preserved-attempt and reset-attempt paths.

## Validation
- `bun test src/browser/utils/messages/retryState.test.ts`
- `make static-check`

## Risks
Low risk and scoped to retry state transitions. The change is gated to manual retries after explicit Stop and preserves prior behavior for normal manual retries.

---

### DIFF
--- a/src/browser/components/Messages/ChatBarrier/RetryBarrier.tsx
+++ b/src/browser/components/Messages/ChatBarrier/RetryBarrier.tsx
@@ -99,11 +99,15 @@ export const RetryBarrier: React.FC<RetryBarrierProps> = (props) => {
   // Emits event to useResumeManager instead of calling resumeStream directly
   // This keeps all retry logic centralized in one place
   const handleManualRetry = () => {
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

PR #2480, opened 2026-02-18
🤖 refactor: walk Error .cause chain in getErrorMessage, eliminate 256 inline duplicates

ammar-agent:
## Summary

`getErrorMessage()` now walks the `.cause` chain, and **every callsite in the codebase** uses it instead of inline `err instanceof Error ? err.message : String(err)` patterns. Error logs that previously showed truncated messages like `Failed to read file /path/to/SKILL.md:` (with nothing after the colon) now surface the full context.

## Background

Error objects frequently carry nested context in `.cause` — for example, `RuntimeError("Failed to read file /path:", "file_io", innerError)` wraps the real filesystem error. But every formatting path in the codebase only read `.message`, silently discarding `.cause`. When the outer message happened to end with `:` and the inner error held the actual detail (like `ENOENT: no such file or directory`), the log output was truncated and unhelpful.

The codebase had **7 independent copies** of the same one-liner (`error instanceof Error ? error.message : String(error)`) as named helper functions, plus **256 inline occurrences** across ~95 files.

## Implementation

1. **Enhanced `getErrorMessage()`** in `src/common/utils/errors.ts` to walk the `.cause` chain iteratively, appending `[cause: ...]` for each level whose message isn't already present in the accumulated string (avoids duplication when RuntimeError already embeds the inner message).

2. **Deleted 6 module-private clones** — `formatError()` in 4 files and `toErrorMessage()` in `log.ts` and `getErrorMessage()` in `ExperimentsSection.tsx` — replacing each with an import of the shared utility.

3. **Replaced 256 inline patterns** across 95 files with `getErrorMessage()` calls, adding the import where needed.

4. **Added tests** for the cause-chain walking behavior (7 test cases covering plain errors, single/multi-level chains, deduplication, empty causes, and non-Error causes).

## Validation

- `make static-check` passes (typecheck, lint, fmt)
- New unit tests for `getErrorMessage` all pass
- Grep confirms zero remaining inline `instanceof Error ? *.message : String(*)` patterns (excluding test files and the intentionally-different `quickjsRuntime.ts`)

## Risks

Low. `getErrorMessage` is a pure string-formatting function — the only behavior change is that error messages now contain *more* information (the cause chain), never less. The `!msg.includes(current.message)` guard ensures no duplication for the common case where RuntimeError already embeds the inner message.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$2.58`_

### DIFF
--- a/src/common/utils/errors.test.ts
+++ b/src/common/utils/errors.test.ts
@@ -47,4 +47,20 @@ describe("getErrorMessage", () => {
     // Empty cause message is skipped
     expect(getErrorMessage(outer)).toBe("outer");
   });
+
+  it("handles cyclic cause chain without hanging", () => {
+    const a = new Error("error A");
+    const b = new Error("error B", { cause: a });
+    // Create a cycle: a -> b -> a -> ...
+    a.cause = b;
+    const result = getErrorMessage(b);
+    expect(result).toContain("error B");
+    expect(result).toContain("error A");
+  });
+
+  it("handles self-referencing cause", () => {
+    const err = new Error("self");
+    err.cause = err;
+    expect(getErrorMessage(err)).toBe("self");
+  });
 });
--- a/src/common/utils/errors.ts
+++ b/src/common/utils/errors.ts
@@ -8,8 +8,13 @@
 export function getErrorMessage(error: unknown): string {
   if (!(error instanceof Error)) return String(error);
   let msg = error.message;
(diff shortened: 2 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Added WeakSet cycle guard per review feedback, with tests for both cyclic and self-referencing cause chains.

ammario (approved):

PR #2490, opened 2026-02-18
🤖 fix: close settings page on Escape

ethanndickson:
## Summary
Add Escape keyboard handling for the route-based settings page so users can close settings and return to their previous route without clicking UI controls.

## Background
Settings currently closes via the titlebar toggle, the desktop close button, or the mobile back button. There was no Escape path, which made keyboard-only flows inconsistent with other closeable surfaces (command palette, dialogs).

## Implementation
- Added a bubble-phase `keydown` listener in `SettingsPage` that closes settings on Escape.
- Uses bubble phase (not capture) so inner overlays (Select dropdowns, Popover, Dialog) that call `stopPropagation`/`preventDefault` on Escape get first right of refusal — only an unclaimed Escape navigates away from settings.
- Guards against closing when focus is in an editable field (`isEditableElement`).
- Checks `e.defaultPrevented` to cooperate with any overlay that already handled the event.

## Risks
Low. Single-file change scoped to the Settings route lifecycle. Bubble-phase ordering means existing overlay Escape semantics are preserved by design.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$0.95`_

### DIFF
--- a/src/browser/components/Settings/SettingsPage.tsx
+++ b/src/browser/components/Settings/SettingsPage.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import { useEffect } from "react";
 import {
   ArrowLeft,
   Menu,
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Switched from capture-phase to bubble-phase Escape handler so Radix overlays (Select, Popover) dismiss first. Removed `isDialogOpen` guard — modal dialogs already intercept Escape before this bubble listener runs.

PR #2494, opened 2026-02-18
🤖 fix: keep Orchestrator visible in Settings agent defaults

ibetitsmike:
## Summary
Make Orchestrator always visible in **Settings → Agent Defaults → UI agents**, even when the current workspace has no accepted plan.

## Background
Orchestrator currently requires a workspace plan (`ui.requires: [plan]`), so `uiSelectable` becomes false when no plan file exists. That behavior is correct for the chat agent picker, but it also hid Orchestrator from Settings, making its defaults harder to discover and configure.

## Implementation
- Updated `src/browser/components/Settings/sections/TasksSection.tsx` to use a shared predicate for Settings section partitioning:
  - Treat `orchestrator` as part of Settings “UI agents” regardless of `uiSelectable`.
  - Keep section membership mutually exclusive for UI agents / sub-agents / internal.
- Added a contextual note on the Orchestrator row when it is not currently selectable:
  - `Agent is available when workspace contains a plan created in Plan mode.`
- Left runtime selectability untouched (no changes to backend plan gating or agent picker filtering).

## Validation
- `make static-check`

## Risks
Low risk and UI-scoped:
- Change is limited to Settings categorization/rendering in `TasksSection.tsx`.
- Agent picker behavior remains unchanged because it still filters by `uiSelectable`.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$2.29`_

### DIFF
--- a/src/browser/components/Settings/sections/TasksSection.tsx
+++ b/src/browser/components/Settings/sections/TasksSection.tsx
@@ -590,10 +590,13 @@ export function TasksSection() {
   const listedAgents = agents.length > 0 ? agents : FALLBACK_AGENTS;
   const enabledAgentIdSet = new Set(enabledAgentIds);
 
+  const showInSettingsUiAgents = (agent: AgentDefinitionDescriptor) =>
+    agent.uiSelectable || agent.id === "orchestrator";
+
   const uiAgents = useMemo(
     () =>
       [...listedAgents]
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

PR #2495, opened 2026-02-18
🤖 feat: pause onboarding while configuring providers

jaaydenh:
## Summary

Pause the onboarding wizard (instead of dismissing it) when users navigate to Settings → Providers, so they can configure API keys on the full settings page, then automatically resume onboarding when settings closes.

## Background

The onboarding splash is modal and blocks interaction with the settings page underneath. When users click a provider button or "Settings → Providers" link during onboarding, they can't actually configure anything because the modal overlay prevents interaction. This creates a frustrating dead end.

## Implementation

**Pause/resume context** (`SplashScreenProvider.tsx`): New `OnboardingPauseContext` tracks paused state (step index + reason). While paused, the splash hides without being marked as viewed, preserving the queue for seamless resume.

**Provider settings navigation** (`OnboardingWizardSplash.tsx`): All three `openSettings("providers", ...)` calls replaced with `openProvidersSettings()` helper that pauses onboarding first, then navigates. Step index is restored from paused state on resume.

**Auto-resume on close** (`SettingsContext.tsx`): Since `SettingsProvider` wraps `SplashScreenProvider` in the component tree, direct context consumption isn't possible. Added `registerOnClose(callback)` subscription pattern — `SplashScreenProvider` subscribes and triggers resume when settings closes.

**Return banner** (`SettingsPage.tsx`): Non-blocking inline banner ("Setup is paused while you configure providers." + "Return to setup" button) appears at top of settings content when onboarding is paused.

## Risks

Low risk — changes are additive. The pause state is ephemeral (not persisted) so worst case on any edge case is onboarding restarts from step 0. The `registerOnClose` pattern is a standard observer that cleans up via effect teardown.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$2.92`_

### DIFF
--- a/src/browser/components/splashScreens/OnboardingWizardSplash.tsx
+++ b/src/browser/components/splashScreens/OnboardingWizardSplash.tsx
@@ -922,9 +922,14 @@ export function OnboardingWizardSplash(props: { onDismiss: () => void }) {
     visibleProviders,
   ]);
 
+  // Clamp stepIndex to valid range when the step list changes. Skip while still
+  // loading provider config — the temporary single-item "loading" array would
+  // reset a restored stepIndex (e.g. after resuming from paused onboarding).
+  const isLoading = hasConfiguredProvidersAtStart === null;
   useEffect(() => {
+    if (isLoading) return;
     setStepIndex((index) => Math.min(index, steps.length - 1));
-  }, [steps.length]);
+  }, [isLoading, steps.length]);
 
   const totalSteps = steps.length;
   const currentStep = steps[stepIndex] ?? steps[0];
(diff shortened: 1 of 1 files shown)


jaaydenh:
@codex review

jaaydenh:
@codex review

Fixed: Close subscribers now fire via a `useEffect` watching `isOpen` transitions (open → closed), so onboarding resumes regardless of how the user leaves settings (explicit close, back navigation, etc.).

jaaydenh:
@codex review

Fixed step index reset on resume: the clamping `useEffect` now skips while provider config is loading (temporary 1-item "loading" array was resetting restored stepIndex to 0).

PR #2497, opened 2026-02-18
🤖 fix: preserve strict origin checks behind proxies

ibetitsmike:
## Summary
Fix false CORS/WS origin rejections when `mux server` runs behind reverse proxies (including Coder app proxy and subpath routing) that forward internal host headers, while preserving strict scheme/origin security boundaries.

## Background
Some proxied deployments sent `Origin` with the public app host while `X-Forwarded-Host` was internal/proxy-local. Our origin check trusted a single expected origin, so static assets and websocket upgrades were rejected with `403` even for legitimate same-origin browser requests. This also showed up in subpath app URLs.

This change keeps the security posture from https://github.com/coder/security/issues/120 by continuing to block cross-origin requests and explicitly rejecting protocol-downgrade origins when `X-Forwarded-Proto` is provided.

## Implementation
- Replaced single `getExpectedOrigin()` lookup with `getExpectedOrigins()` that combines:
  - `X-Forwarded-Host` (first hop) and `Host`
  - `X-Forwarded-Proto` first hop only
  - inferred socket/request protocol only when forwarded proto is absent/invalid
- Updated `isOriginAllowed()` to check the normalized request origin against expected origin candidates (preserving loopback-equivalence behavior).
- Applied the same expected-origin candidate logic to:
  - HTTP/CORS middleware
  - WebSocket upgrade origin validation
- Improved warning logs to emit `expectedOrigins` for easier proxy debugging.
- Added regression coverage for:
  - same-origin HTTP and WS requests when `X-Forwarded-Host` is mismatched
  - rejecting downgraded HTTP and WS origins when `X-Forwarded-Proto` pins `https`
  - rejecting downgraded HTTP and WS origins when `X-Forwarded-Proto` contains multi-hop values (for example `https,http`)

## Validation
- `make static-check`
- `make build-main`
- `make typecheck`
- `bun x eslint src/node/orpc/server.ts src/node/orpc/server.test.ts`
- `env TEST_INTEGRATION=1 bun x jest tests/ipc/runtime/originFetch.test.ts --runInBand` (twice)

## Risks
Low-to-moderate. This expands acceptable origin matching from one derived origin to a constrained set built from request host/proxy headers. It still requires exact origin match (or existing loopback alias equivalence), and restricts scheme matching to one trusted forwarded-proto hop.

---

### DIFF
--- a/src/node/orpc/server.test.ts
+++ b/src/node/orpc/server.test.ts
@@ -550,6 +550,30 @@ describe("createOrpcServer", () => {
     }
   });
 
+  test("rejects downgraded HTTP origins when X-Forwarded-Proto includes multiple hops", async () => {
+    const stubContext: Partial<ORPCContext> = {};
+
+    let server: Awaited<ReturnType<typeof createOrpcServer>> | null = null;
+    try {
+      server = await createOrpcServer({
+        host: "127.0.0.1",
+        port: 0,
+        context: stubContext as ORPCContext,
+      });
+
+      const response = await fetch(`${server.baseUrl}/health`, {
+        headers: {
+          Origin: server.baseUrl,
+          "X-Forwarded-Proto": "https,http",
+        },
+      });
+
+      expect(response.status).toBe(403);
+    } finally {
+      await server?.close();
+    }
+  });
+
   test("allows HTTP requests without Origin headers", async () => {
     const stubContext: Partial<ORPCContext> = {};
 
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed the origin-check downgrade feedback by only using inferred protocol when X-Forwarded-Proto is absent, and added HTTP/WS regression tests for downgrade rejection.

ibetitsmike:
@codex review

Follow-up addressed: origin validation now trusts only the first X-Forwarded-Proto hop and adds HTTP/WS regression tests for multi-hop downgrade attempts.

PR #2502, opened 2026-02-19
🤖 fix: harden renderer HTML sinks against XSS

ibetitsmike:
## Summary
Fix renderer XSS exposure in diff and markdown diagram paths by removing unsafe filename HTML rendering in `HunkViewer` and hardening Mermaid SVG insertion.

## Background
A filename from workspace/repository content was being transformed into HTML and injected via `dangerouslySetInnerHTML` in `HunkViewer`. Because filenames are attacker-controlled, this enabled immediate XSS when viewing a crafted diff. Given renderer-accessible privileged surfaces, that XSS materially increased impact.

## Implementation
- Replaced filename highlight HTML strings in `HunkViewer` with safe React element rendering (`<mark>` nodes) so React escaping remains in effect.
- Removed `dangerouslySetInnerHTML` usage for `hunk.filePath` display.
- Hardened Mermaid rendering:
  - switched Mermaid config from `securityLevel: "loose"` to `"strict"`
  - added `sanitizeMermaidSvg()` and route all rendered SVG through it before `innerHTML` assignment
  - removed active containers (`script`, `iframe`, `object`, `embed`) and stripped inline event handlers / dangerous URL schemes from URL-bearing attributes
  - canonicalized encoded URL values before scheme checks (decodes entity/percent encodings and strips control/whitespace obfuscation) so payloads like `java&#10;script:` are blocked
  - guarded numeric entity decoding against invalid Unicode code points so crafted values cannot throw during sanitization
  - preserved `<foreignObject>` nodes so Mermaid wrapped labels continue to render
- Added explicit `SECURITY AUDIT` comments at unavoidable HTML sinks to document trust boundaries.
- Added `Mermaid.sanitization.test.ts` to verify malformed SVG rejection, active-content stripping, encoded-scheme blocking, invalid-entity handling, and label preservation.
- Added a new `Security: Renderer HTML & XSS` section in `docs/AGENTS.md` to codify safe patterns.

## Validation
- `make static-check` (before initial push)
- `bun test src/browser/components/Messages/Mermaid.sanitization.test.ts`
- `make static-check` (after foreignObject-label follow-up)
- `bun test src/browser/components/Messages/Mermaid.sanitization.test.ts`
- `make static-check` (after encoded-scheme canonicalization follow-up)
- `bun test src/browser/components/Messages/Mermaid.sanitization.test.ts`
- `make static-check` (after invalid-entity guard follow-up)

## Risks
- Mermaid `securityLevel: "strict"` may reduce support for some diagram features that relied on looser behavior (especially scriptable/unsafe links), but this is intentional hardening.
- Sanitization is defense-in-depth; if Mermaid output format changes significantly, sanitizer assumptions should be revisited.

---

### DIFF
--- a/src/browser/components/Messages/Mermaid.sanitization.test.ts
+++ b/src/browser/components/Messages/Mermaid.sanitization.test.ts
@@ -17,6 +17,7 @@ describe("sanitizeMermaidSvg", () => {
       '<foreignObject><div onclick="evil()">label</div></foreignObject>' +
       '<a href="javascript:alert(1)"><text>link</text></a>' +
       '<a xlink:href="java&#10;script:alert(3)"><text>split-link</text></a>' +
+      '<a href="java&#1114112;script:alert(4)"><text>invalid-codepoint</text></a>' +
       '<image src="javascript:alert(2)" />' +
       '<rect width="10" height="10" onclick="steal()" />' +
       "</svg>";
@@ -29,7 +30,9 @@ describe("sanitizeMermaidSvg", () => {
     expect(sanitized).not.toContain("onclick=");
     expect(sanitized).not.toContain("javascript:");
     expect(sanitized).not.toContain("java&#10;script:");
+    expect(sanitized).not.toContain("&#1114112;");
     expect(sanitized).not.toContain("xlink:href=");
+    expect(sanitized).not.toContain("href=");
     expect(sanitized).toContain("<svg");
     expect(sanitized).toContain("<rect");
     expect(sanitized).toContain("<foreignObject");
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed the `foreignObject` regression concern by preserving label nodes while stripping active content and unsafe URL schemes.

ibetitsmike:
@codex review

Addressed encoded/obfuscated scheme bypasses by canonicalizing entity + percent-encoded URL values before blocked-scheme checks.

ibetitsmike:
@codex review

Added invalid-codepoint guards in entity decoding so sanitizer canonicalization cannot throw on crafted numeric entities.

PR #2510, opened 2026-02-20
🤖 ci: keep nightly cleanup checkout-free

ibetitsmike:
## Summary
Fix nightly release cleanup so it no longer depends on a local git checkout.

## Background
The nightly cleanup workflow failed on **February 19, 2026** in job `64119285477` with:
`fatal: not a git repository (or any of the parent directories): .git`.

That failure happened because `gh release delete --cleanup-tag` shells out to local git in a job that intentionally does not run `actions/checkout`.

## Implementation
- Updated `.github/workflows/nightly.yml` to remove `--cleanup-tag` in both nightly release deletion paths:
  - incomplete-release rebuild path in `compute-version`
  - `cleanup-old-nightlies` deletion loop
- Added explicit GitHub API tag deletion in both places:
  - `gh api --silent -X DELETE "repos/${GITHUB_REPOSITORY}/git/refs/tags/<tag>" || true`
- Added comments explaining why cleanup is checkout-free.

## Validation
- `make static-check`

## Risks
Low risk. Behavior is equivalent for normal cases, and missing tag refs are tolerated with `|| true` to avoid false-failures.

---

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -88,7 +88,9 @@ jobs:
 
             echo "Release $RELEASE_TAG is incomplete (missing: ${MISSING_ASSETS[*]})"
             echo "Deleting $RELEASE_TAG so this run can rebuild desktop artifacts"
-            gh release delete "$RELEASE_TAG" --yes --cleanup-tag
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

PR #2518, opened 2026-02-20
🤖 fix: harden updater check UX and persist update channel

ibetitsmike:
## Summary
Improve desktop updater resilience and UX by sanitizing update-check errors, translating missing `latest-*.yml` publish-window failures into a friendly retry message, and persisting the stable/nightly update channel selection so it survives restarts.

## Background
During nightly publish windows, `electron-updater` can receive a 404 for `latest-*.yml` before artifacts are fully available. That error previously surfaced a low-level stack + header blob in the About dialog. Also, update channel selection needed to be explicitly persisted so user intent is retained across sessions.

## Implementation
- Added update-check error shaping in `UpdaterService`:
  - Detect missing `latest-*.yml` + 404 signatures and show a user-friendly message.
  - Sanitize other check errors to first-line, length-limited text to avoid dumping stack traces into UI.
  - Applied message shaping across updater check failure paths (`error` event, rejected check promise, sync catch).
- Persisted update channel preference explicitly in config:
  - `Config.saveConfig()` now writes valid `updateChannel` values (`stable` and `nightly`).
  - `Config.setUpdateChannel()` stores the selected channel directly.
- Kept runtime update-service channel state aligned with persisted config:
  - `UpdateService` reads channel once at startup and uses it for updater initialization.
  - `UpdateService.getChannel()` returns cached channel in non-desktop/no-impl mode.
  - `UpdateService.setChannel()` updates cached state after successful persistence.
- Added regression tests:
  - Friendly messaging for missing nightly metadata errors.
  - Config-level channel persistence (including explicit stable).
  - UpdateService startup/read + write-on-change channel behavior.

## Validation
- `make static-check`
- `bun test src/desktop/updater.test.ts`
- `bun test src/node/config.test.ts src/node/services/updateService.test.ts`
- `bun x eslint src/desktop/updater.ts src/desktop/updater.test.ts src/node/config.ts src/node/config.test.ts src/node/services/updateService.ts src/node/services/updateService.test.ts`

## Risks
- Low-to-moderate risk in updater error classification: logic intentionally narrows the friendly-message mapping to 404 + `latest-*.yml` signatures and preserves non-check-phase behavior.
- Config persistence now records explicit `stable`; this is intentional for preserving user choice and remains backward-compatible with default-to-stable reads.

---

### DIFF
--- a/src/desktop/updater.test.ts
+++ b/src/desktop/updater.test.ts
@@ -426,6 +426,11 @@ describe("UpdaterService", () => {
   });
 
   describe("transient error backoff", () => {
+    const missingNightlyManifestError =
+      'Cannot find latest-mac.yml in the latest release artifacts (https://github.com/coder/mux/releases/download/v0.18.1-nightly.16/latest-mac.yml): HttpError: 404 "method: GET url: https://github.com/coder/mux/releases/download/v0.18.1-nightly.16/latest-mac.yml\n\nPlease double check that your authentication token is correct.\n" Headers: { ... }\n    at createHttpError (/path/httpExecutor.ts:31:10)';
+    const nightlyManifestPendingMessage =
+      "Update metadata isn't available yet. The latest release may still be publishing; please try again in a few minutes.";
+
     it("should silently back off on 404 (latest.yml missing) for auto checks", async () => {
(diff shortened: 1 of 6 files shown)


ibetitsmike:
@codex review

PR #2519, opened 2026-02-20
🤖 fix: gate proxy origin compatibility behind --allow-http-origin

ibetitsmike:
## Summary
Make proxy protocol-compatibility opt-in. `mux server` now keeps strict origin/protocol validation by default and only accepts HTTPS browser origins behind rewritten `X-Forwarded-Proto=http` when explicitly enabled.

## Background
TLS-terminating proxy chains can rewrite forwarded protocol headers to `http` on the last hop. We needed to keep default strict behavior while still allowing explicit compatibility for those deployments.

## Implementation
- Added `--allow-http-origin` to `mux server`.
- Threaded `allowHttpOrigin` through `ServerService.startServer` into `createOrpcServer`.
- Updated origin/protocol helpers so HTTP->HTTPS compatibility applies only when `allowHttpOrigin` is true.
- Kept OAuth redirect scheme and secure-cookie decisions aligned with the same flag-controlled protocol path.
- Added startup note in `mux server` output when `--allow-http-origin` is enabled.
- Added non-CLI opt-in path via `MUX_SERVER_ALLOW_HTTP_ORIGIN=1` (used when `startServer` callers do not pass the option directly, e.g. desktop/browser-mode starts).
- Updated docs and CLI help assertions for the new option.

## Validation
- `bun test src/cli/run.test.ts`
- `bun test src/node/services/serverService.test.ts`
- `make static-check`

## Risks
- Low-to-moderate: changes are localized to server origin/protocol gating and startup option plumbing. Highest regression risk is reverse-proxy browser access; tests now cover default-off and explicit opt-in behavior for HTTP, WS, OAuth redirect URIs, secure cookies, and non-CLI env opt-in.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$0.14`_

### DIFF
--- a/docs/reference/cli.mdx
+++ b/docs/reference/cli.mdx
@@ -117,7 +117,7 @@ Options:
 - `--ssh-host <host>` - SSH hostname/alias used for editor deep links in browser mode
 - `--add-project <path>` - Add and open project at the specified path
 
-Use `--allow-http-origin` only when HTTPS is terminated by an upstream reverse proxy and mux receives rewritten `X-Forwarded-Proto=http` headers. This compatibility mode is disabled by default.
+Use `--allow-http-origin` only when HTTPS is terminated by an upstream reverse proxy and mux receives rewritten `X-Forwarded-Proto=http` headers. This compatibility mode is disabled by default. For non-CLI server starts (for example desktop/browser mode), set `MUX_SERVER_ALLOW_HTTP_ORIGIN=1` to opt in.
 
 Auth token precedence:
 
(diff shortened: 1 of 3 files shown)


ibetitsmike:
@codex review

Please review this follow-up: proxy forwarded-proto HTTPS compatibility is now opt-in via `--allow-http-origin`, with default strict behavior preserved.

ibetitsmike:
@codex review

Rebased on latest main and force-pushed. Please re-review.

ibetitsmike:
@codex review

Addressed your P2 by adding a non-CLI opt-in path (`MUX_SERVER_ALLOW_HTTP_ORIGIN=1`) in `ServerService.startServer`, plus test coverage and docs.

PR #2528, opened 2026-02-21
🤖 fix: remember mobile menu open position

ibetitsmike:
## Summary
Persist mobile left-sidebar list position so reopening the menu returns users to where they last were instead of resetting to the top.

## Background
On touch devices, the left sidebar is an off-canvas surface that closes frequently as users select chats or drafts. Because the list subtree unmounts when closed, the previous scroll location was lost on reopen.

## Implementation
- Added `MOBILE_LEFT_SIDEBAR_SCROLL_TOP_KEY` in `src/common/constants/storage.ts`.
- In `ProjectSidebar`:
  - Added a ref for the scrollable container and a ref for the latest in-memory mobile scroll offset.
  - Restored persisted `scrollTop` when the mobile menu opens.
  - Persisted the latest offset when the menu transitions open → closed and before mobile auto-close actions.
  - Updated scroll handling to **cache in memory only** during `onScroll` and avoid synchronous `localStorage` writes at scroll-event frequency.
- Used shared persisted-state helpers (`readPersistedState` / `updatePersistedState`) for consistency.

## Validation
- `make static-check` (before initial PR)
- `make static-check` (after Codex feedback fix)

## Risks
Low risk and scoped to mobile/touch left-sidebar behavior. Desktop list behavior is unchanged via viewport guards.

---

### DIFF
--- a/src/browser/components/ProjectSidebar.tsx
+++ b/src/browser/components/ProjectSidebar.tsx
@@ -475,6 +475,12 @@ const ProjectSidebarInner: React.FC<ProjectSidebarProps> = ({
   // Mobile breakpoint for auto-closing sidebar
   const MOBILE_BREAKPOINT = 768;
   const projectListScrollRef = useRef<HTMLDivElement | null>(null);
+  const mobileScrollTopRef = useRef(0);
+  const wasCollapsedRef = useRef(collapsed);
+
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Please take another look.

PR #2533, opened 2026-02-22
🤖 fix: cascade-terminate sub-agent tasks on workspace interrupt

ThomasK33:
## Summary

Cascade-terminate active descendant sub-agent tasks when a parent workspace is hard-interrupted so child completion cannot auto-resume an interrupted parent stream.

## Background

Interrupting a parent workspace previously only stopped the parent stream. Active descendants could continue running and, when the last child reported, trigger parent auto-resume behavior. This made interrupted workspaces appear to restart themselves.

## Implementation

- Added `TaskService.terminateAllDescendantAgentTasks(workspaceId)` to collect descendants, sort leaf-first, stop active streams, reject waiters, remove child workspaces, and kick the queue scheduler.
- Updated `WorkspaceService.interruptStream()` to invoke descendant cascade termination for hard interrupts (`soft !== true`) and log failures without failing the primary interrupt path.
- Added hard-interrupt suppression state in `TaskService` and mark it **before** session interruption to close the parent/child report race.
- Suppressed both parent auto-resume paths (`handleStreamEnd` with active descendants and post-`agent_report` resume) while interrupt suppression is active, and clear suppression on real user sends via `resetAutoResumeCount`.
- Fixed the `sendQueuedImmediately` flow by clearing interrupt suppression before `AgentSession.sendQueuedMessages()` (that path bypasses `WorkspaceService.sendMessage`).
- Added task service and workspace service regression tests for subtree termination/no-op behavior, hard-interrupt auto-resume suppression, and queued resend suppression clearing.

## Validation

- `make static-check`
- `bun test src/node/services/taskService.test.ts --filter "(hard-interrupted|markParentWorkspaceInterrupted|terminateAllDescendant|terminateDescendant)"`
- `bun test src/node/services/workspaceService.test.ts --filter "interruptStream|sendQueuedImmediately"`
- `bun run typecheck`

## Risks

Low-to-medium risk in backend interrupt/task orchestration paths. The new suppression path only applies to hard interrupts and is reset on user-originated sends (including queued immediate resend), minimizing impact on normal auto-resume behavior.

---

<details>

# Cascade-terminate sub-agent tasks on parent workspace interrupt

## Context & Why

When a user presses ESC to interrupt a parent workspace that has active sub-agent tasks, only the parent stream stops. The child workspaces continue running. When the last child completes, `TaskService.finalizeAgentTaskReport()` auto-resumes the parent via `workspaceService.sendMessage()`, making it appear to "restart itself." This violates the user's intent — they wanted everything to stop.

**Goal:** When the user interrupts a workspace, automatically cascade-terminate all its active descendant sub-agent tasks so no child can later auto-resume the parent.

## Evidence

| Source | Key finding |
|--------|------------|
| `src/browser/hooks/useAIViewKeybinds.ts:107` | ESC → `api.workspace.interruptStream({ workspaceId })` (no cascade) |
| `src/node/services/workspaceService.ts:3500-3535` | `interruptStream()` stops only the parent's stream |
| `src/node/services/taskService.ts:2751-2778` | `finalizeAgentTaskReport()` auto-resumes parent when `!hasActiveDescendants && !isStreaming` — this is the "restart" trigger |
| `src/node/services/taskService.ts:1081-1151` | `terminateDescendantAgentTask()` already does recursive leaf-first termination — we reuse this pattern |
| `src/node/services/taskService.ts:1446-1454` | `hasActiveDescendantAgentTasksForWorkspace()` already checks for active children |
| `src/node/services/taskService.ts:1596-1609` | `listDescendantAgentTaskIdsFromIndex()` collects full subtree via DFS |

## Implementation

~50 net LoC product code, ~80 LoC tests.

### 1. New method: `TaskService.terminateAllDescendantAgentTasks()`

**File:** `src/node/services/taskService.ts` — add after `terminateDescendantAgentTask()` (after line ~1151)

Extracts the core loop from `terminateDescendantAgentTask` but skips the ancestor-validation gate (since the caller *is* the parent). Follows the exact same pattern: mutex, build index, list descendants, sort leaf-first, stop streams + reject waiters + remove workspaces.

```typescript
/**
 * Terminate ALL descendant agent tasks of a workspace (leaf-first).
 * Used when the user interrupts a parent workspace to prevent orphaned
 * children from auto-resuming it after completion.
 */
async terminateAllDescendantAgentTasks(
  workspaceId: string
): Promise<string[]> {
  assert(
    workspaceId.length > 0,
    "terminateAllDescendantAgentTasks: workspaceId must be non-empty"
  );

  const terminatedTaskIds: string[] = [];

  {
    await using _lock = await this.mutex.acquire();

    const cfg = this.config.loadConfigOrDefault();
    const index = this.buildAgentTaskIndex(cfg);
    const descendants = this.listDescendantAgentTaskIdsFromIndex(index, workspaceId);

    if (descendants.length === 0) {
      return terminatedTaskIds;
    }

    // Sort leaf-first (deepest first) to avoid orphaned children.
    const parentById = index.parentById;
    const depthById = new Map<string, number>();
    for (const id of descendants) {
      depthById.set(id, this.getTaskDepthFromParentById(parentById, id));
    }
    descendants.sort((a, b) => (depthById.get(b) ?? 0) - (depthById.get(a) ?? 0));

    const terminationError = new Error("Parent workspace interrupted");

    for (const id of descendants) {
      // Best-effort: stop any active stream to prevent further token usage.
      try {
        const stopResult = await this.aiService.stopStream(id, { abandonPartial: true });
        if (!stopResult.success) {
          log.debug("terminateAllDescendantAgentTasks: stopStream failed", { taskId: id });
        }
      } catch (error: unknown) {
        log.debug("terminateAllDescendantAgentTasks: stopStream threw", { taskId: id, error });
      }

      this.remindedAwaitingReport.delete(id);
      this.completedReportsByTaskId.delete(id);
      this.rejectWaiters(id, terminationError);

      const removeResult = await this.workspaceService.remove(id, true);
      if (!removeResult.success) {
        log.error("terminateAllDescendantAgentTasks: failed to remove workspace", {
          taskId: id,
          error: removeResult.error,
        });
      }

      terminatedTaskIds.push(id);
    }
  }

  // Free slots and start any queued tasks.
  await this.maybeStartQueuedTasks();

  return terminatedTaskIds;
}
```

### 2. Wire cascade into `WorkspaceService.interruptStream()`

**File:** `src/node/services/workspaceService.ts` — inside `interruptStream()`, after the parent stream is stopped (~line 3517), before the return.

Only cascade on **hard** (non-soft) interrupts. Soft interrupts are internal coordination and should not cascade.

```typescript
      // ...existing partial deletion...

      // Cascade: terminate all active descendant sub-agent tasks on hard interrupts
      // to prevent orphaned children from auto-resuming this workspace.
      if (!options?.soft && this.taskService) {
        const terminated = await this.taskService.terminateAllDescendantAgentTasks(workspaceId);
        if (terminated.length > 0) {
          log.debug("Cascade-terminated descendant tasks on interrupt", {
            workspaceId,
            terminatedTaskIds: terminated,
          });
        }
      }

      // Handle queued messages based on option
      // ...existing queue handling...
```

### 3. Tests

**File:** `src/node/services/taskService.test.ts` — add near existing `terminateDescendantAgentTask` tests (~line 1762)

Two tests following the existing test pattern (`createTestConfig`, `createAIServiceMocks`, `createWorkspaceServiceMocks`, `createTaskServiceHarness`):

1. **`terminateAllDescendantAgentTasks terminates entire subtree leaf-first`** — Set up root → parent-task → child-task. Call `terminateAllDescendantAgentTasks(rootId)`. Assert both tasks terminated in leaf-first order (child before parent), streams stopped, workspaces removed.

2. **`terminateAllDescendantAgentTasks is a no-op with no descendants`** — Set up root with no children. Call method. Assert returns empty array, no calls to stopStream or remove.

<details>
<summary>Design decisions and edge cases</summary>

### Why not a confirmation dialog?
A dialog adds friction to what should be an instant "stop everything" action. ESC means "stop" — having children silently continue and resurrect the parent violates the principle of least surprise far more than auto-stopping them.

### Race condition: child reports between parent stop and cascade
There's a small window where a child could complete and trigger `deliverReportToParent` after the parent stream is stopped but before we cascade-terminate it. In practice this window is tiny (microseconds of synchronous JS). If it does occur, the worst case is the parent gets one auto-resume, and the user presses ESC again. Acceptable for v1; we can add a `userInterruptedWorkspaces: Set<string>` guard in `finalizeAgentTaskReport` later if needed.

### Why only hard interrupts?
- **Hard interrupt** (no `soft` option): user-initiated ESC, CLI Ctrl+C, archive — all cases where the user wants to stop.
- **Soft interrupt**: internal coordination (block-boundary stops) — should never cascade.
- The `sendQueuedImmediately` case (compaction) already has its own flow and doesn't need cascading.

### Callers of `interruptStream` and cascade behavior
| Caller | `soft?` | Cascade? | Correct? |
|--------|---------|----------|----------|
| ESC key (`useAIViewKeybinds.ts`) | no | ✅ yes | User wants to stop everything |
| CLI Ctrl+C (`run.ts`) | no | ✅ yes | Same intent |
| Command palette stop (`sources.ts`) | no | ✅ yes | Same intent |
| Archive (`workspaceService.ts:2457`) | no | ✅ yes | Archiving should clean up children |
| Compaction (`handler.ts`) | no | ✅ yes | Acceptable — compaction restarts anyway |
| History edit (`agentSession.ts:1736`) | no | ⚠️ yes | This calls `this.interruptStream()` on `AgentSession`, not `WorkspaceService.interruptStream()`, so our cascade code is NOT reached. Safe. |

### No frontend changes needed
The ESC handler already calls `interruptStream`. The cascade is transparent on the backend. No new IPC, no new UI components.

</details>

</details>

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$3.23`_

### DIFF
--- a/src/node/services/workspaceService.test.ts
+++ b/src/node/services/workspaceService.test.ts
@@ -16,6 +16,7 @@ import type { AIService } from "./aiService";
 import type { InitStateManager, InitStatus } from "./initStateManager";
 import type { ExtensionMetadataService } from "./ExtensionMetadataService";
 import type { FrontendWorkspaceMetadata, WorkspaceMetadata } from "@/common/types/workspace";
+import type { TaskService } from "./taskService";
 import type { BackgroundProcessManager } from "./backgroundProcessManager";
 import type { BashToolResult } from "@/common/types/tools";
 import { createMuxMessage } from "@/common/types/message";
@@ -2652,6 +2653,84 @@ describe("WorkspaceService fork", () => {
     } finally {
       orchestrateForkSpy.mockRestore();
       createRuntimeSpy.mockRestore();
+      getOrCreateSessionSpy.mockRestore();
+    }
+  });
+});
+
+describe("WorkspaceService interruptStream", () => {
+  let historyService: HistoryService;
+  let cleanupHistory: () => Promise<void>;
+
+  beforeEach(async () => {
+    ({ historyService, cleanup: cleanupHistory } = await createTestHistoryService());
+  });
+
+  afterEach(async () => {
+    await cleanupHistory();
+  });
+
+  test("sendQueuedImmediately clears hard-interrupt suppression before queued resend", async () => {
+    const workspaceId = "ws-interrupt-queue-111";
+
+    const mockConfig: Partial<Config> = {
+      srcDir: "/tmp/test",
+      getSessionDir: mock(() => "/tmp/test/sessions"),
+      generateStableId: mock(() => "test-id"),
+      findWorkspace: mock(() => null),
+    };
+
+    const mockAIService: AIService = {
+      isStreaming: mock(() => false),
+      getWorkspaceMetadata: mock(() => Promise.resolve({ success: false, error: "not found" })),
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      on: mock(() => {}),
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      off: mock(() => {}),
+    } as unknown as AIService;
+
+    const workspaceService = new WorkspaceService(
+      mockConfig as Config,
+      historyService,
+      mockAIService,
+      mockInitStateManager as InitStateManager,
+      mockExtensionMetadataService as ExtensionMetadataService,
+      mockBackgroundProcessManager as BackgroundProcessManager
+    );
+
+    const resetAutoResumeCount = mock(() => undefined);
+    const markParentWorkspaceInterrupted = mock(() => undefined);
+    const terminateAllDescendantAgentTasks = mock(() => Promise.resolve([] as string[]));
+    workspaceService.setTaskService({
+      resetAutoResumeCount,
+      markParentWorkspaceInterrupted,
+      terminateAllDescendantAgentTasks,
+    } as unknown as TaskService);
+
+    const sendQueuedMessages = mock(() => undefined);
+    const restoreQueueToInput = mock(() => undefined);
+    const interruptStream = mock(() => Promise.resolve(Ok(undefined)));
+    const fakeSession = {
+      interruptStream,
+      sendQueuedMessages,
+      restoreQueueToInput,
+    };
+    const getOrCreateSessionSpy = spyOn(workspaceService, "getOrCreateSession").mockReturnValue(
+      fakeSession as unknown as AgentSession
+    );
+
+    try {
+      const result = await workspaceService.interruptStream(workspaceId, {
+        sendQueuedImmediately: true,
+      });
+
+      expect(result.success).toBe(true);
+      expect(markParentWorkspaceInterrupted).toHaveBeenCalledWith(workspaceId);
+      expect(terminateAllDescendantAgentTasks).toHaveBeenCalledWith(workspaceId);
+      expect(resetAutoResumeCount).toHaveBeenCalledTimes(2);
+      expect(sendQueuedMessages).toHaveBeenCalledTimes(1);
+      expect(restoreQueueToInput).not.toHaveBeenCalled();
+    } finally {
       getOrCreateSessionSpy.mockRestore();
     }
   });
--- a/src/node/services/workspaceService.ts
+++ b/src/node/services/workspaceService.ts
@@ -3549,6 +3549,9 @@ export class WorkspaceService extends EventEmitter {
 
       // Handle queued messages based on option
       if (options?.sendQueuedImmediately) {
+        // `sendQueuedMessages()` routes through AgentSession directly, so explicitly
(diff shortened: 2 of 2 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Implemented the race fix by marking hard-interrupted parents before stream interruption and suppressing both parent auto-resume paths until user-originated resume/reset.

ThomasK33:
@codex review

Addressed the queued resend path: interrupt suppression is now cleared before sendQueuedMessages(), with a regression test in workspaceService.test.ts.

PR #2545, opened 2026-02-23
🤖 fix: prevent unread indicators from disappearing due to stuck streaming metadata

ethanndickson:
## Summary

Fix three bugs in the workspace activity metadata system that caused unread indicators to intermittently disappear. When `streaming: true` gets stuck in `extensionMetadata.json`, the frontend never sees the `streaming: true→false` transition needed to fire `responseCompleteCallback`, so `recencyTimestamp` is never bumped past `lastRead`, and unread indicators stop appearing.

## Background

Investigation traced the root cause to `extensionMetadata.json` having `streaming: true` stuck for multiple workspaces. Three independent bugs contribute:

1. **Stream errors don't reset streaming status** — When a stream errors (API failure, network timeout, rate limit), `StreamManager` emits only `"error"`, but `workspaceService` only listens for `"stream-end"` and `"stream-abort"`. The `"error"` event is unhandled, leaving `streaming: true` permanently.

2. **`stream-abort` forwarding swallowed by partial cleanup failure** — In `aiService.ts`, the `stream-abort` handler does async partial cleanup before forwarding the event. If cleanup throws, `this.emit("stream-abort")` is never reached. The `void` fire-and-forget means the rejection is unhandled.

3. **Cross-workspace write races on shared metadata file** — All workspaces share one `extensionMetadata.json` file, but the write queue was per-workspace. Concurrent writes for different workspaces do read-modify-write on the same file, causing last-writer-wins races that silently lose updates (e.g., clearing `streaming: false` gets clobbered).

The existing `clearStaleStreaming()` call on app startup masks these bugs by resetting all stuck `streaming: true` entries — which is why the issue is intermittent (self-heals on restart, persists within a session).

## Implementation

**Bug 1** (`workspaceService.ts`): Added an `"error"` event listener in `setupMetadataListeners()` that calls `updateStreamingStatus(workspaceId, false)` — mirroring the existing `"stream-abort"` handler.

**Bug 2** (`aiService.ts`): Wrapped partial cleanup in `try/catch` and moved `this.emit("stream-abort", data)` to a `finally` block. Stream lifecycle consistency is higher priority than best-effort partial cleanup.

**Bug 3** (`ExtensionMetadataService.ts` + `workspaceService.ts`): Added a global mutation queue (`withSerializedMutation()`) inside `ExtensionMetadataService` so all read-modify-write cycles are serialized regardless of workspace ID. Removed the now-redundant per-workspace queue from `workspaceService`.

## Validation

- New tests added for all three fixes:
  - `workspaceService.test.ts`: error events trigger `setStreaming(false)`
  - `aiService.test.ts`: `stream-abort` forwarded even when `deletePartial` throws
  - `ExtensionMetadataService.test.ts`: concurrent cross-workspace writes preserve all entries
- All focused tests pass

## Risks

- **Bug #1**: Low risk — adds a new listener that mirrors existing `stream-abort` handling pattern. Worst case: a redundant `setStreaming(false)` call (idempotent).
- **Bug #2**: Low risk — strictly safer behavior (cleanup was already best-effort; now forwarding is unconditional).
- **Bug #3**: Medium risk — changes write serialization boundary from per-workspace to global. All metadata writes now serialize through one queue, which could add latency under heavy load (many workspaces writing simultaneously). In practice, these writes are infrequent and fast (~10-50ms each), so throughput impact should be negligible.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$31.97`_

### DIFF
--- a/src/node/services/ExtensionMetadataService.test.ts
+++ b/src/node/services/ExtensionMetadataService.test.ts
@@ -12,9 +12,9 @@ const sleep = (ms: number): Promise<void> =>
     setTimeout(resolve, ms);
   });
 
-type ExtensionMetadataServiceInternals = {
+interface ExtensionMetadataServiceInternals {
   load: () => Promise<ExtensionMetadataFile>;
-};
+}
 
 const addLoadDelay = (target: ExtensionMetadataService, delayMs: number): (() => void) => {
   const internals = target as unknown as ExtensionMetadataServiceInternals;
--- a/src/node/services/aiService.test.ts
+++ b/src/node/services/aiService.test.ts
@@ -78,7 +78,9 @@ describe("AIService.setupStreamEventForwarding", () => {
     const service = new AIService(config, historyService, initStateManager, providerService);
 
     const cleanupError = new Error("disk full");
-    spyOn(historyService, "deletePartial").mockImplementation(() => Promise.reject(cleanupError));
+    const deletePartialSpy = spyOn(historyService, "deletePartial").mockImplementation(() =>
+      Promise.reject(cleanupError)
+    );
 
     const streamManager = (service as unknown as { streamManager: StreamManager }).streamManager;
     const abortEvent: StreamAbortEvent = {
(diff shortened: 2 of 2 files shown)


ethanndickson:
@codex review

PR #2493, opened 2026-02-18
🤖 feat: boundary-windowed chat loading + metadata-only workspace activity

ThomasK33:
## Summary

Overhaul chat subscription architecture to load only the current compaction epoch on startup, scope full transcript streaming to the active workspace, and add cursor-based "Load More" pagination for older history.

## Background

Previously, every workspace started a full `onChat` subscription that replayed from the **penultimate** compaction boundary, meaning all workspaces eagerly loaded two epochs of history regardless of whether they were visible. This caused unnecessary data transfer at startup and steady-state bandwidth waste.

This PR introduces three key changes:
1. **Latest-boundary replay** — initial load starts from `skip=0` (latest boundary only) instead of `skip=1`
2. **Active-workspace scoping** — only the currently-displayed workspace gets a full `onChat` stream; all others use a lightweight metadata-only activity feed for sidebar indicators
3. **Load More pagination** — a cursor-based ORPC endpoint + frontend button to page backwards through older compaction epochs on demand

### Backend (`agentSession.ts`, `historyService.ts`, ORPC schemas/router)
- `emitHistoricalEvents()` now calls `getHistoryFromLatestBoundary(workspaceId, 0)` 
- New `workspace.history.loadMore` endpoint returns a single older boundary window with cursor-based pagination
- New `getHistoryBoundaryWindow()` helper scans boundaries to return exactly one epoch window at a time

### Frontend aggregator (`StreamingMessageAggregator.ts`)
- Renamed `pruneBeforePenultimateBoundary` → `pruneBeforeLatestBoundary`  
- Live compaction now prunes everything before the incoming boundary's sequence, keeping only the current epoch

### Frontend store (`WorkspaceStore.ts`, `WorkspaceContext.tsx`)
- `addWorkspace()` no longer starts `runOnChatSubscription()` — subscription is managed by `ensureActiveOnChatSubscription()`
- Only one full `onChat` stream active at a time, switched via `setActiveWorkspaceId()`
- Activity subscription (`workspace.activity.list/subscribe`) provides streaming/recency/model fallbacks for non-active workspaces
- Per-workspace pagination state initialized on `caught-up`, exposed as `hasOlderHistory`/`loadingOlderHistory` in `WorkspaceState`

### Frontend UI (`ChatPane.tsx`)
- "Load older messages" button above transcript when `hasOlderHistory` is true
- Disabled with "Loading..." text during fetch

## Validation

- `make typecheck` ✅
- `make lint` ✅  
- `make fmt-check` ✅
[...]

PR #2546, opened 2026-02-23
🤖 fix: refresh session usage on workspace activation to fix stale cost totals

ethanndickson:
## Summary

Fixes stale workspace cost totals when switching between workspaces. The frontend fetched persisted session usage (`session-usage.json`) only once during workspace registration (`addWorkspace`), so cost rollups arriving while a workspace was inactive (e.g., sub-agent deletions) were never picked up until a hard refresh.

## Background

PR #2493 introduced boundary-windowed chat loading, where only the active workspace receives a full `onChat` subscription. Non-active workspaces use metadata-only activity feeds. This means `session-usage-delta` events — which carry sub-agent cost rollups — only reach the currently active workspace.

Since `getSessionUsage()` was only called once during `addWorkspace()`, switching back to a workspace that received rollups while inactive would show stale (lower) cost totals. Users reported workspaces dropping from ~$20 to <$5 until `Ctrl+Shift+R` forced a full reload.

## Implementation

- Extracted the inline `getSessionUsage` fetch + repricing logic from `addWorkspace()` into a shared `refreshSessionUsage()` private method.
- Added a per-workspace request version guard so slower/older fetch responses cannot overwrite fresher state during rapid workspace switches.
- `setActiveWorkspaceId()` now calls `refreshSessionUsage()` for the newly active workspace, re-hydrating any cost data that arrived while it was inactive.
- Cleanup: `removeWorkspace()` clears the request-version tracking for deleted workspaces.

## Validation

- New regression tests:
  - Verifies activation triggers a fresh `getSessionUsage` fetch and hydrates `sessionTotal`
  - Verifies stale in-flight responses are dropped when a newer refresh supersedes them
- All existing `WorkspaceStore.test.ts` tests pass
- `make typecheck` passes

---

### DIFF
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -29,7 +29,9 @@ const mockOnChat = mock(async function* (
   });
 });
 
-const mockGetSessionUsage = mock(() => Promise.resolve(undefined));
+const mockGetSessionUsage = mock((_input: { workspaceId: string }) =>
+  Promise.resolve<unknown>(undefined)
+);
 const mockHistoryLoadMore = mock(
   (): Promise<LoadMoreResponse> =>
     Promise.resolve({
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

PR #2549, opened 2026-02-23
🤖 fix: auto-select full hunk when jumping with J/K

ibetitsmike:
## Summary
This PR updates immersive review hunk navigation so pressing `J` / `K` automatically selects the full target hunk range after a successful jump, making immediate comment actions operate on the whole hunk.

## Background
The intended workflow is keyboard-first review speed: after jumping hunks with `J` / `K`, the user should be ready to comment on the full hunk without manual line selection.

## Implementation
- Reverted the earlier active-hunk-outline rendering approach.
- Added a pending "select-all-on-jump" marker in immersive review state.
- On successful hunk jump (
`navigateHunk` from `J` / `K`), store the target hunk ID and switch hunks.
- In the selected-hunk alignment effect, when the selected hunk matches that pending jump marker:
  - set active line to the hunk anchor line (first modified line, fallback to hunk start)
  - set `selectedLineRange` to the full hunk range
- Cleared the pending marker in non-J/K hunk-selection paths (file navigation, review-note navigation, cursor-driven hunk switches, line click selection) so behavior stays scoped to J/K navigation only.
- Unified visuals for full-hunk selection by suppressing the separate immersive active-line DOM outline when the active line sits inside a multi-line selection. This prevents the detached cursor-ring appearance while preserving cursor semantics for keyboard movement.
- Added `isDialogOpen()` guard to immersive capture-phase keydown handler so Escape (and all other shortcuts) are skipped while a modal dialog is active. This prevents exiting immersive review when dismissing a dialog/palette.

## Validation
- `make static-check`

## Risks
Low. The change is scoped to immersive review keyboard navigation/selection visuals and does not affect backend APIs or persistence.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$1.12`_

### DIFF
--- a/src/browser/components/RightSidebar/CodeReview/ImmersiveReviewView.tsx
+++ b/src/browser/components/RightSidebar/CodeReview/ImmersiveReviewView.tsx
@@ -27,7 +27,12 @@ import {
   getAdjacentFilePath,
   getFileHunks,
 } from "@/browser/utils/review/navigation";
-import { isEditableElement, KEYBINDS, matchesKeybind } from "@/browser/utils/ui/keybinds";
+import {
+  isDialogOpen,
+  isEditableElement,
+  KEYBINDS,
+  matchesKeybind,
+} from "@/browser/utils/ui/keybinds";
 import { stopKeyboardPropagation } from "@/browser/utils/events";
 import { buildReadFileScript, processFileContents } from "@/browser/utils/fileExplorer";
 import {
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

Please review this immersive review mode UI change.

ibetitsmike:
@codex review

Addressed the prior feedback:
- active hunk outline now renders via an absolute overlay so it remains visible above diff cell backgrounds
- active hunk range prop is now memoized with stable references to preserve SelectableDiffRenderer memoization

ibetitsmike:
@codex review

Addressed the inline-comments regression: active hunk outline now remains continuous through inline note/composer rows and bottom-edge placement accounts for post-line rows.

ibetitsmike:
@codex review

Updated per product direction: J/K hunk jumps now auto-select the entire target hunk range for immediate commenting; prior active-hunk-outline rendering changes were removed.

ibetitsmike:
@codex review

Follow-up tweak: when J/K auto-selects a full hunk, multi-line range highlight now owns the visual state so active cursor doesn't render as a conflicting separate marker.

ibetitsmike:
@codex review

Follow-up fix: the remaining separate cursor ring came from immersive mode's direct DOM active-line outline path. It is now suppressed during multi-line full-hunk selection, so J/K jump selection presents as one unified range.

ibetitsmike:
@codex review

Resolved the out-of-scope Escape/capture thread (that listener behavior is unchanged from `main`) and kept this PR scoped to the J/K full-hunk selection + unified selection visuals fix.
Please re-check this diff.

ibetitsmike:
@codex review

ibetitsmike:
@codex review

PR #2551, opened 2026-02-23
🤖 fix: prevent immersive Escape from interrupting streams

ibetitsmike:
## Summary
Stop Escape in immersive review from bubbling into the global stream interrupt handler.

## Background
In immersive review mode, pressing Escape should only control immersive UI state (return from notes to diff, or exit immersive mode). Instead, Escape could also interrupt an active model stream, which is surprising and destructive.

## Implementation
- Updated `ImmersiveReviewView` keydown listener to run in capture phase.
- Added `stopKeyboardPropagation(e)` to immersive Escape handling in both panels:
  - notes panel Escape (return to diff)
  - diff panel Escape (exit immersive)
- Added regression coverage in `useAIViewKeybinds.test.tsx` proving capture-phase Escape handlers prevent global stream interrupts.

## Validation
- `bun test src/browser/hooks/useAIViewKeybinds.test.tsx`
- `make typecheck`
- `make static-check`

## Risks
Low risk. The change is scoped to immersive review keyboard handling and specifically to Escape propagation semantics.

---

### DIFF
--- a/src/browser/components/RightSidebar/CodeReview/ImmersiveReviewView.tsx
+++ b/src/browser/components/RightSidebar/CodeReview/ImmersiveReviewView.tsx
@@ -27,6 +27,7 @@ import {
   getFileHunks,
 } from "@/browser/utils/review/navigation";
 import { isEditableElement, KEYBINDS, matchesKeybind } from "@/browser/utils/ui/keybinds";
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #2554, opened 2026-02-23
🤖 fix: suppress browser notifications for idle compaction completions

ammar-agent:
## Summary

Suppress browser notification noise for compaction flows by moving idle-compaction detection to activity snapshots, skipping compaction notifications when an automatic follow-up message is queued, and hardening idle-marker cleanup across race/error paths.

## Background

The previous idle-compaction suppression relied on an `idle-compaction-started` chat event. That signal is unreliable for true background work because `onChat` is only active for the currently selected workspace. During Codex review, additional edge cases surfaced where stale idle markers could leak across turns and incorrectly suppress non-idle notifications.

## Implementation

1. **Activity snapshot contract**
   - Added optional `isIdleCompaction` to `WorkspaceActivitySnapshotSchema`.

2. **Backend idle marker lifecycle (`WorkspaceService`)**
   - Track in-flight idle maintenance with `idleCompactingWorkspaces`.
   - Set marker only after idle dispatch succeeds **and** the session is still busy.
   - Tag `isIdleCompaction` only on `streaming=false` snapshots (stop snapshots), not on `streaming=true` snapshots.
   - Always clear marker for `streaming=false` transitions in `finally`, so metadata write failures cannot leak state into future turns.

3. **Frontend completion suppression (`WorkspaceStore` + `App`)**
   - Use activity transitions (`streaming true→false` + recency advance) for background completion detection.
   - Preserve idle detection via `previous || snapshot` for reconnect/restore resilience.
   - Route completion callbacks through a store-level wrapper.
   - For compaction completions, if a queued follow-up exists (auto-sent next turn), coerce `hasContinueMessage=true` so App suppresses intermediate "Compaction complete" notifications.
   - `App.tsx` suppresses notifications for `compaction?.isIdle` and `compaction?.hasContinueMessage`.

4. **Cleanup**
   - Removed dead `idle-compaction-started` plumbing across backend/frontend:
     - backend emitter method
     - frontend callback subscription/export
     - stream schema branch and ignored-event handling

5. **Tests**
   - Updated idle-compaction dispatch tests.
   - Added regression coverage for:
     - no stale marker when send succeeds without active stream
     - marker cleared when `setStreaming` fails on stream-stop
     - `streaming=true` snapshots are never idle-tagged
     - active compaction completion with queued follow-up suppression path

## Validation

- `make static-check`
- `bun test src/browser/utils/messages/applyWorkspaceChatEventToAggregator.test.ts`
- `make typecheck`

## Risks

Low risk and scoped to compaction completion notification behavior. Main risk is over-suppression in edge cases where queued follow-up intent is ambiguous, but this is preferable to duplicate/intermediate notifications and matches compaction auto-send semantics.

## Pains

This needed several Codex-driven hardening passes for subtle ordering/failure edges (startup-abort, metadata write failure, and cross-turn marker bleed), which required tightening marker lifecycle guarantees rather than adding timing-based coordination.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$10.37`_

### DIFF
--- a/src/node/services/workspaceService.test.ts
+++ b/src/node/services/workspaceService.test.ts
@@ -810,6 +810,55 @@ describe("WorkspaceService idle compaction dispatch", () => {
     }
     expect(executionError.message).toContain("idle-only send was skipped");
   });
+  test("does not tag streaming=true snapshots as idle compaction", async () => {
+    const workspaceId = "idle-streaming-true-no-tag";
+    const snapshot = {
+      recency: Date.now(),
+      streaming: true,
+      lastModel: "claude-sonnet-4",
+      lastThinkingLevel: null,
+    };
+
+    const setStreaming = mock(() => Promise.resolve(snapshot));
+    const emitWorkspaceActivity = mock(
+      (_workspaceId: string, _snapshot: typeof snapshot) => undefined
+    );
+
+    (
+      workspaceService as unknown as {
+        extensionMetadata: ExtensionMetadataService;
+        emitWorkspaceActivity: typeof emitWorkspaceActivity;
+      }
+    ).extensionMetadata = {
+      setStreaming,
+    } as unknown as ExtensionMetadataService;
+    (
+      workspaceService as unknown as {
+        extensionMetadata: ExtensionMetadataService;
+        emitWorkspaceActivity: typeof emitWorkspaceActivity;
+      }
+    ).emitWorkspaceActivity = emitWorkspaceActivity;
+
+    const internals = workspaceService as unknown as {
+      idleCompactingWorkspaces: Set<string>;
+      updateStreamingStatus: (
+        workspaceId: string,
+        streaming: boolean,
+        model?: string,
+        agentId?: string
+      ) => Promise<void>;
+    };
+
+    internals.idleCompactingWorkspaces.add(workspaceId);
+
+    await internals.updateStreamingStatus(workspaceId, true);
+
+    expect(setStreaming).toHaveBeenCalledWith(workspaceId, true, undefined, undefined);
+    expect(emitWorkspaceActivity).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed: now derive idle-compaction state from both the previous (start) and current (stop) snapshots, so UI reconnects that miss the transient flag on the start snapshot still suppress idle compaction notifications.

ammar-agent:
@codex review

Addressed: moved `idleCompactingWorkspaces.add()` to after `sendMessage` returns successfully. At that point the idle stream is already running and the session is busy, so no concurrent user stream can race with the marker. The streaming=true snapshot is emitted without the flag, but the streaming=false snapshot picks it up — the frontend checks both.

ammar-agent:
@codex review

Latest idle-compaction cleanup is pushed (including dead `idle-compaction-started` plumbing removal and test updates). Please re-review.

ammar-agent:
@codex review

Addressed stale idle-marker race:
- only set `idleCompactingWorkspaces` when `session.isBusy()` is still true after dispatch
- clear marker otherwise (startup-abort / ultra-fast completion path)
- added regression tests for active-stream and no-stream-success cases

ammar-agent:
@codex review

Added another notification suppression fix:
- compaction completions now inherit `hasContinueMessage=true` when a queued follow-up is pending
- prevents intermediate "Compaction complete" notification and notifies on the follow-up response instead
- includes WorkspaceStore regression test coverage

ammar-agent:
@codex review

Addressed the new idle-marker leak case:
- `updateStreamingStatus` now clears `idleCompactingWorkspaces` in `finally` for streaming=false
- added regression test for setStreaming failure path

ammar-agent:
@codex review

Addressed stream-stop tagging feedback:
- idle marker now only tags `streaming=false` snapshots
- added regression test that `streaming=true` snapshots stay untagged even when idle marker is set

PR #2555, opened 2026-02-23
🤖 fix: prevent stats page from immediately navigating away

ThomasK33:
## Summary

Fix the stats/analytics page immediately flashing back to the previous workspace when opened.

## Background

When clicking the stats button (top-right), the app would briefly show the analytics page then snap back to the previous workspace. This happened because two effects in `WorkspaceContext` lacked guards for the analytics route — they treated the analytics page as "no workspace selected" and auto-navigated away.

## Implementation

Added `isAnalyticsOpen` guards to two effects in `src/browser/contexts/WorkspaceContext.tsx`, mirroring the existing `currentSettingsSection` pattern:

1. **`useLayoutEffect` (workspace subscription cleanup)** — Previously cleared the active workspace ID when navigating to `/analytics` since both `currentWorkspaceId` and `currentSettingsSection` were null. Now also checks `isAnalyticsOpen` to preserve the workspace subscription.

2. **`useEffect` (launch project check)** — Previously saw `selectedWorkspace` as null on `/analytics`, called `getLaunchProject()`, found a workspace, and auto-navigated away. Now returns early when `isAnalyticsOpen` is true.

`isAnalyticsOpen` was already exported by `RouterContext` — it just wasn't consumed in `WorkspaceContext`.

## Risks

Low — the change is narrowly scoped to two guard conditions and directly mirrors the existing settings-page pattern. No new state or side effects introduced.

---
_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$6.75`_

### DIFF
--- a/src/browser/contexts/WorkspaceContext.tsx
+++ b/src/browser/contexts/WorkspaceContext.tsx
@@ -562,6 +562,7 @@ export function WorkspaceProvider(props: WorkspaceProviderProps) {
     currentProjectId,
     currentProjectPathFromState,
     currentSettingsSection,
+    isAnalyticsOpen,
     pendingSectionId,
     pendingDraftId,
   } = useRouter();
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

PR #2552, opened 2026-02-23
🤖 fix: complete gpt-5.3-codex-spark compatibility

ammar-agent:
## Summary
Complete `gpt-5.3-codex-spark` compatibility by both (1) avoiding unsupported OpenAI reasoning summary parameters and (2) registering Spark as a first-class model in Mux.

## Background
We originally fixed the request-time error where Spark rejects `reasoning.summary`. This PR now also includes the remaining product integration work so Spark behaves like other curated first-class models:

- discoverable in model lists/docs
- alias-resolvable (`spark`)
- covered by thinking policy and token stats
- correctly gated by Codex OAuth requirements

## Implementation
- Provider options:
  - keep omitting `reasoningSummary` for `gpt-5.3-codex-spark`
  - preserve `reasoningEffort` and encrypted reasoning include behavior
- Known models:
  - added `openai:gpt-5.3-codex-spark` to `knownModels` with alias `spark`
  - tokenizer override set to `openai/gpt-5`
- Model metadata:
  - added Spark entry in `models-extra.ts` with 128k context metadata
  - marked text-only capability via `supports_vision: false`
  - reused GPT-5.3-Codex pricing until Spark-specific API pricing is published
- Thinking policy:
  - expanded GPT-5.2/5.3 Codex policy matching so `-spark` gets 5 reasoning levels (`off`→`xhigh`)
- OAuth gating:
  - added Spark to `CODEX_OAUTH_REQUIRED_MODELS`
- Tests/docs:
  - extended alias, thinking policy, model stats, and OpenAI gating tests
  - regenerated models docs table to include Spark

## Validation
- `bun test src/common/utils/ai/models.test.ts src/common/utils/ai/providerOptions.test.ts src/common/utils/thinking/policy.test.ts src/common/utils/tokens/modelStats.test.ts src/browser/hooks/useModelsFromSettings.test.ts`
- `make static-check`

## Risks
Low-to-medium risk, mostly around model-list/gating behavior for OpenAI models. Covered by targeted selector-gating tests and full static-check.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$0.98`_

PR #2557, opened 2026-02-23
🤖 tests: remove tautological alias-mapping and constant-mirroring tests

ammar-agent:
## Summary

Remove tautological tests that merely re-assert hardcoded values already defined in production constant tables (`MODEL_ABBREVIATIONS`, `models-extra.ts`, `models.json`), providing zero bug-detection value.

## Background

PR #2552 added several tests for the new `gpt-5.3-codex-spark` model that followed the existing pattern of copying exact constant values into test assertions. This pattern is tautological — if someone changes a model's pricing in `models-extra.ts`, these tests break not because of a bug, but because the hardcoded test values no longer match the updated constants. An audit of the full test suite identified two files with this anti-pattern.

## Implementation

**`models.test.ts`** — Removed the entire `resolveModelAlias` describe block (both "resolves known aliases" and "returns non-alias strings unchanged" tests). The function under test is a 3-line `Object.hasOwn` + property access into `MODEL_ABBREVIATIONS` — no logic to exercise.

**`modelStats.test.ts`** — Two changes:
- Removed "should include spark model config from models-extra.ts" entirely (100% tautological — 4 hardcoded `.toBe()` assertions copied verbatim from `models-extra.ts`)
- Replaced hardcoded constant values in 8 other tests with structural assertions (`toBeGreaterThan(0)`, `toBeDefined()`) that still verify the lookup logic found results without mirroring exact data values

Tests that exercise actual logic (regex routing in thinking policy, merge strategies in model capabilities, string parsing in model display, cross-source integration guards in knownModels) were audited and kept.

## Validation

- 43 tests pass across both files (0 failures)
- `make typecheck` clean
- `make lint` clean

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$2.42`_

### DIFF
--- a/src/common/utils/ai/models.test.ts
+++ b/src/common/utils/ai/models.test.ts
@@ -4,7 +4,6 @@ import {
   getModelName,
   supports1MContext,
   isValidModelFormat,
-  resolveModelAlias,
 } from "./models";
 
 describe("normalizeGatewayModel", () => {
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Restored the exact `272000` value in the override test — that assertion is meaningful since it proves `models-extra.ts` wins over `models.json` (which has 400k). Added a comment explaining why the exact value matters there.

PR #2562, opened 2026-02-24
🤖 feat: add Ctrl+F find in transcript

ammar-agent:
## Summary

Adds Ctrl/Cmd+F "Find in transcript" support to the chat pane, matching the standard in-app find experience. A floating search bar appears over the transcript area, letting users type a query, see match counts, and navigate between matches with Enter/Shift+Enter or arrow buttons.

## Background

There was no way to quickly locate specific content within a long chat transcript. Users had to scroll manually, which is especially painful for lengthy agent conversations. This follows the prior art of the "Quote in input" context-menu feature, which already demonstrates scoped transcript text operations.

## Implementation

**New transcript search utility** (`src/browser/utils/messages/transcriptSearch.ts`):

- `findTranscriptTextMatches()` – walks the rendered DOM inside `[data-message-content]` blocks, performing case-insensitive text search. Excludes interactive elements (buttons, inputs, selects) so only actual message text is matched.
- `focusTranscriptTextMatch()` – selects the matched text via the Selection API and scrolls the containing message row into view.

**ChatPane integration** (`src/browser/components/ChatPane.tsx`):

- Floating find bar with search input, match counter (e.g. "3/12"), prev/next/close buttons.
- Ctrl/Cmd+F opens the bar (capture-phase handler scoped to chat pane focus). If transcript text is currently selected, it pre-fills the query (reuses `getTranscriptContextMenuText` from the quote feature).
- Enter → next match, Shift+Enter → previous match, Escape → close bar. Match navigation wraps around.
- Active match selection disables auto-scroll so streaming won't yank the viewport away.
- On close, clears the DOM selection within the transcript.

**Keybind registry** (`src/browser/utils/ui/keybinds.ts`):

- `FOCUS_TRANSCRIPT_SEARCH: { key: "f", ctrl: true }` – shares the same physical shortcut as `FOCUS_REVIEW_SEARCH`, but they're scoped to different panels (chat pane vs. code review) via focus containment checks.

**Settings** (`KeybindsSection.tsx`):

- "Find in transcript" added to the Chat keybind group.

---

### DIFF
--- a/src/browser/components/ChatPane.tsx
+++ b/src/browser/components/ChatPane.tsx
@@ -595,8 +595,11 @@ export const ChatPane: React.FC<ChatPaneProps> = (props) => {
         return -1;
       }
 
+      // Keep the initial index unset so the first Enter lands on match #1.
+      // If we eagerly set 0 here, first Enter skips to #2 and single-hit
+      // searches cannot trigger focus because the index never changes.
       if (currentIndex < 0) {
-        return 0;
+        return -1;
       }
 
       return Math.min(currentIndex, matches.length - 1);
--- a/tests/ui/chat/transcriptFind.test.ts
+++ b/tests/ui/chat/transcriptFind.test.ts
@@ -73,6 +73,11 @@ async function closeTranscriptFind(container: HTMLElement, input: HTMLInputEleme
   });
 }
 
+function getWindowSelectionText(): string {
+  const selection = window.getSelection();
+  return selection?.toString() ?? "";
+}
+
 describe("Transcript find UI", () => {
   beforeAll(async () => {
     await preloadTestModules();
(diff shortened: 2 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the locale-safe offset feedback: switched from `toLocaleLowerCase` + `indexOf` to `RegExp` with the `i` flag so match offsets always refer to the original source text. Added `escapeRegExp` to safely escape user queries.

ammar-agent:
Regarding the `RegExp` + `i` flag comment: this is intentional and matches the standard behavior of Ctrl+F in browsers, VS Code, and other text editors. None of them perform locale-specific Unicode casefolding (İ↔i, ß↔ẞ) in their find-in-page.

The previous `toLocaleLowerCase` + `indexOf` approach was **also** not correctly handling these pairs for offset mapping — it was actually broken in the opposite direction (offsets pointing past the text node boundary, causing `IndexSizeError`). The RegExp approach correctly returns offsets in the original source text, which is the critical invariant for `Selection.addRange()`.

Resolving as intentional — this matches the behavior users expect from Ctrl+F in a typical application.

ammar-agent:
@codex review

Resolved the Unicode casefolding thread — the RegExp `i` flag matches standard Ctrl+F behavior in browsers and editors. The previous `toLocaleLowerCase` approach was broken for offset mapping (the original P2 issue); the current approach returns correct offsets in the source text.

ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the initial Enter navigation issue for transcript find and added regression coverage.

PR #2563, opened 2026-02-24
🤖 feat: sidebar terminal activity indicator with OSC prompt detection

ethanndickson:
## Summary

Show a terminal activity indicator in the workspace sidebar — a terminal icon + count badge appears when integrated terminal sessions are actively running commands.

## Background

When multiple workspaces have integrated terminals open, there's no visual signal about which workspaces have active commands running. Users must switch to each workspace to check terminal state. This feature surfaces terminal activity directly in the sidebar for at-a-glance awareness.

## Implementation

**Activity detection** uses three complementary signals:
- **Input-driven:** `sendInput()` with `\r`/`\n` marks session as running
- **OSC 0/2 title heuristics:** idle-pattern titles (CWD paths, shell names, `user@host:` prompts) reset to idle — covers bash/zsh
- **OSC 133 semantic prompt markers:** `A` → idle, `C` → running — covers fish

**Session lifecycle** is designed around a single termination primitive (`terminateTrackedSessions`) that all close paths route through (single close, workspace close, global shutdown), guaranteeing cleanup even when PTY exit callbacks don't fire synchronously.

**Non-OSC resilience:** shells that don't emit OSC sequences get an automatic idle fallback timer (10s) to prevent permanent false-running state. Once a session observes any OSC activity, it's promoted to OSC-driven mode and the fallback is cancelled.

**Data flow:** `TerminalService` → ORPC streaming subscription → `WorkspaceStore.WorkspaceSidebarState` → `WorkspaceListItem` renders badge when `terminalActiveCount > 0`.

Uses `headless.parser.registerOscHandler()` instead of `onTitleChange` because xterm v6's internal event forwarding doesn't fire in this app despite the parser correctly processing OSC sequences.

### Key design decisions

- Sessions start as `isRunning: false` because shells emit a CWD title almost immediately on startup, making an optimistic `true` default flash and disappear.
- Terminal activity is composed into existing `WorkspaceSidebarState` (not a separate store) for cohesive sidebar rendering.
- Shared `WorkspaceTerminalIcon` component extracted from `WorkspaceMenuBar` for reuse in sidebar badge.

## Validation

- 30 tests pass in `terminalService.test.ts` covering: idle/command title classification, sendInput newline → running, OSC 133 prompt markers, activity change dedup, session exit cleanup, bulk workspace/global close cleanup, aggregate computation, and no-OSC idle fallback (timer refresh + OSC-driven immunity)
- `WorkspaceStore.test.ts` terminal activity tests pass (snapshot propagation + zero defaults)

## Risks

- **OSC title heuristic false negatives:** commands whose titles match idle patterns (e.g., a program named `bash`) will appear idle. Low severity — the input-driven signal covers the initial running transition.
- **Non-OSC fallback window:** 10s is a heuristic; very long commands in non-OSC shells will flash running→idle→running if the user sends another newline. Acceptable tradeoff vs permanent stuck state.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$40.68`_

### DIFF
--- a/src/browser/components/WorkspaceListItem.tsx
+++ b/src/browser/components/WorkspaceListItem.tsx
@@ -20,6 +20,7 @@ import { Trash2, Ellipsis, Loader2, Sparkles } from "lucide-react";
 import { WorkspaceStatusIndicator } from "./WorkspaceStatusIndicator";
 import { Shimmer } from "./ai-elements/shimmer";
 import { ArchiveIcon } from "./icons/ArchiveIcon";
+import { WorkspaceTerminalIcon } from "./icons/WorkspaceTerminalIcon";
 import { WORKSPACE_DRAG_TYPE, type WorkspaceDragItem } from "./WorkspaceSectionDropZone";
 import { useLinkSharingEnabled } from "@/browser/contexts/TelemetryEnabledContext";
 import { formatKeybind, KEYBINDS } from "@/browser/utils/ui/keybinds";
@@ -371,7 +372,7 @@ function RegularWorkspaceListItemInner(props: WorkspaceListItemProps) {
     }
   };
 
-  const { canInterrupt, awaitingUserQuestion, isStarting, agentStatus } =
+  const { canInterrupt, awaitingUserQuestion, isStarting, agentStatus, terminalActiveCount } =
     useWorkspaceSidebarState(workspaceId);
 
   const fallbackModel = useWorkspaceFallbackModel(workspaceId);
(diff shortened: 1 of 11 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

Changes since last review:

**P2 (Windows cwd titles) — non-issue, resolved without code change:** We require Git Bash or WSL on Windows, both of which emit POSIX-style paths (`/c/Users/...`, `~/repo`) in OSC title sequences. Native `cmd.exe`/PowerShell paths like `C:\Users\name\repo` are never seen by our heuristic because those shells are not supported runtimes.

**P3 (missing terminal activity subscription) — fixed:** `WorkspaceStore` now treats `terminal.activity.subscribe` as an optional client capability rather than assuming it exists. A `resolveTerminalActivitySubscribe()` helper safely probes the client shape; when the capability is absent (e.g. Storybook mocks, partial test clients), the subscription runner exits cleanly — no TypeError, no retry-loop noise, stale badge state is cleared. The Storybook mock ORPC client now also includes the endpoint for structural contract alignment. A regression test covers the capability-absent path.

**Lint fix:** Removed three unnecessary non-null assertions in `terminalService.test.ts` flagged by `make static-check`.

ethanndickson:
@codex review

Changes since last review:

**P3 re-arm fix** — `runTerminalActivitySubscription()` now releases the `terminalActivityAbortController` latch via a `finally` block and identity-checked `releaseTerminalActivityController()` helper. All exit paths (unsupported capability, signal abort, errors) clear the controller so later `setClient()` calls with a supported client correctly re-arm terminal badge streaming.

Regression test added: verifies unsupported → supported client swap activates the terminal activity subscription and propagates snapshot counts to sidebar state.

All prior Codex threads resolved. `make static-check` passes locally.

PR #2564, opened 2026-02-24
🤖 feat: manage Coder SSH config and migrate suffix to mux--coder

ethanndickson:
## Summary

Implements a mux-managed SSH config writer for Coder workspaces and migrates the SSH hostname suffix from `.mux.coder` → `.mux--coder` to fix a glob collision where user `Host *.coder` entries match mux's dotted hosts.

## Background

Coder workspaces need a `Host *.mux--coder` SSH config block with a `ProxyCommand` that routes connections through `coder ssh`. Previously mux used a dotted suffix (`.mux.coder`) that collided with user-managed `Host *.coder` SSH entries. The new double-dash suffix (`mux--coder`) is hostname-safe and avoids glob collisions by construction.

### SSH Config Writer (`muxSshConfigWriter.ts`, ~240 LoC)
- Atomic writes via temp + rename to prevent partial config corruption
- Full symlink chain resolution (`fs.realpath()` for live targets, `followSymlinkChain()` for dangling symlinks with loop detection)
- Collision-proof temp paths via `crypto.randomUUID()` nonce
- Cross-platform double-quote escaping for `ProxyCommand` argv (works in both `/bin/sh -c` and Windows `cmd.exe /d /s /c`; matches `coder/vscode-coder`'s `escapeCommandArg` strategy)
- Mode-preserving writes; creates `~/.ssh/config` with `0o600` if missing

### Host Normalization
- Shared `resolveCoderSSHHost()` helper in `src/constants/coder.ts` normalizes host from `workspaceName` when present
- **Factory-level normalization**: `runtimeFactory.ts` normalizes host *before* creating `SSHTransport`, ensuring transport and runtime use the same canonical `*.mux--coder` hostname
- `CoderSSHRuntime` constructor reuses the same helper (no drift between factory and constructor paths)

### Windows Path Normalization
- `resolveCoderBinaryPath()` now prefers `where.exe coder` on Windows for native paths
- Falls back to Git Bash `command -v coder` + MSYS→Windows path conversion (`toWindowsPath()`) when `where.exe` fails
- Prevents MSYS paths like `/c/Users/.../coder.exe` from reaching `ProxyCommand`, which executes via `cmd.exe`

### Auth Preflight for Existing Workspaces
- Added `coderService.verifyAuthenticatedSession()` — performs a fresh (uncached) `coder whoami` check
- `CoderSSHRuntime.finalizeConfig` now calls auth preflight for **both** new and existing workspace flows, before the `ensureProvisioningSession` gate
- Restores fail-fast behavior: if Coder auth is expired/logged out, workspace creation returns `Err` before metadata is persisted, preventing unusable workspace entries
- `ensureProvisioningSession` remains gated to new workspaces only (token reuse unchanged)

### Integration
- `ensureMuxCoderSSHConfig()` in CoderService writes the config block, with early-return guard when coder binary unavailable
- `serviceContainer.ts` logs background failures via `log.warn` instead of swallowing them silently

### Test Coverage (30+ tests)
- `muxSshConfigWriter.test.ts`: empty config, mode preservation, symlink preservation (single/multi/dangling), append/replace, idempotency, shell metacharacters, double-quote escaping, Windows-style backslash paths, collision-proof temps, marker validation
- `CoderSSHRuntime.test.ts`: auth preflight failure for existing workspaces, `ensureProvisioningSession` not called for existing workspaces
- `coderService.test.ts`: skip-when-null, write-when-resolved, Windows path normalization pass-through, `verifyAuthenticatedSession` success/failure
- `runtimeFactory.test.ts`: regression test verifying legacy `.coder` host + `workspaceName` produces canonical `*.mux--coder` for both runtime config and underlying transport
- `paths.test.ts`: `toWindowsPath` unit tests (MSYS conversion, drive-letter uppercasing, non-MSYS passthrough)

## Risks

- **Low**: atomic write + symlink resolution is well-tested but unusual `~/.ssh/config` setups (e.g., SELinux-restricted paths) could hit edge cases
- **Low**: the host suffix change affects only new SSH connections; existing running sessions are unaffected

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$40.89`_

### DIFF
--- a/docs/runtime/coder.mdx
+++ b/docs/runtime/coder.mdx
@@ -24,11 +24,11 @@ If multiple organizations have templates with the same name, Mux shows the organ
 
 1. Choose **Existing**.
 2. Select a workspace from the list (status is shown).
-3. Mux will start the workspace if it is stopped, then connect to `<workspace-name>.coder`.
+3. Mux will start the workspace if it is stopped, then connect to `<workspace-name>.mux--coder`.
 
 ## SSH setup
 
-Mux runs `coder config-ssh --yes` before connecting, which creates SSH aliases like `<workspace-name>.coder` in `~/.ssh/config`.
+Mux writes and maintains a mux-owned SSH config block before connecting, creating aliases like `<workspace-name>.mux--coder` in `~/.ssh/config`.
 
 ## Notes
 
--- a/src/browser/utils/policyUi.test.ts
+++ b/src/browser/utils/policyUi.test.ts
@@ -59,7 +59,7 @@ describe("policyUi", () => {
     const sshHost: ParsedRuntime = { mode: RUNTIME_MODE.SSH, host: "user@host" };
     const sshCoder: ParsedRuntime = {
       mode: RUNTIME_MODE.SSH,
-      host: "workspace.coder",
+      host: "workspace.mux--coder",
       coder: { existingWorkspace: true, workspaceName: "workspace" },
     };
 
--- a/src/constants/coder.ts
+++ b/src/constants/coder.ts
@@ -1,0 +1,23 @@
+// Canonical Coder-related constants shared across mux.
+// The ".mux--coder" suffix ensures mux's SSH config block never
+// collides with user-managed or Coder CLI-managed SSH entries.
+
+export const MUX_CODER_HOST_SUFFIX = "mux--coder";
+
+/** Build the SSH hostname for a Coder workspace (e.g. `my-ws.mux--coder`). */
+export function toMuxCoderHost(workspaceName: string): string {
+  return `${workspaceName}.${MUX_CODER_HOST_SUFFIX}`;
+}
+
+/**
+ * Resolve the canonical SSH host for a Coder workspace.
+ * If workspaceName is provided, returns `<name>.mux--coder`;
+ * otherwise falls back to the raw host (non-Coder SSH or already-normalized).
+ */
+export function resolveCoderSSHHost(host: string, workspaceName?: string): string {
+  const name = workspaceName?.trim();
+  return name ? toMuxCoderHost(name) : host;
+}
+
+export const MUX_CODER_SSH_BLOCK_START = "# --- START MUX CODER SSH ---";
+export const MUX_CODER_SSH_BLOCK_END = "# --- END MUX CODER SSH ---";
--- a/src/node/runtime/CoderSSHRuntime.test.ts
+++ b/src/node/runtime/CoderSSHRuntime.test.ts
@@ -36,10 +36,10 @@ function createMockCoderService(overrides?: Partial<CoderService>): CoderService
       Promise.resolve({ success: true as const, data: undefined })
     ),
     ensureProvisioningSession: mock(() => Promise.resolve(provisioningSession)),
+    verifyAuthenticatedSession: mock(() => Promise.resolve()),
     takeProvisioningSession: mock(() => provisioningSession),
     disposeProvisioningSession: mock(() => Promise.resolve()),
-    fetchDeploymentSshConfig: mock(() => Promise.resolve({ hostnameSuffix: "coder" })),
-    ensureSSHConfig: mock(() => Promise.resolve()),
+    ensureMuxCoderSSHConfig: mock(() => Promise.resolve()),
     getWorkspaceStatus: mock(() =>
       Promise.resolve({ kind: "ok" as const, status: "running" as const })
     ),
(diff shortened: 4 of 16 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed both P2 findings:

1. **Symlink preservation** — Added `resolveSSHConfigWritePath()` that resolves symlinks via `lstat`/`readlink` before any read/write operations. Atomic rename now targets the backing file, preserving the symlink. Added regression test with real symlink.

2. **Skip writes when coder CLI is missing** — `ensureMuxCoderSSHConfig()` now returns early when `resolveCoderBinaryPath()` is `null`, avoiding `~/.ssh/config` mutation on machines without Coder. Added two contract tests in `coderService.test.ts`.

ethanndickson:
@codex review

Addressed the remaining P2 temp-path collision blocker:

- Extracted `makeAtomicTempPath()` helper that appends `crypto.randomUUID()` as a nonce, making each temp file uniquely named regardless of PID/timestamp overlap.
- Added regression test that freezes `Date.now()`, stubs UUID to deterministic values, runs concurrent `ensureMuxCoderSSHConfigFile` calls, and asserts distinct temp paths.
- Fixed 4 pre-existing `@typescript-eslint/await-thenable` lint suppressions on Bun's `expect().rejects.toThrow()` pattern and 1 `no-base-to-string` in the new test.

All squashed into the single feature commit. `make static-check` passes locally.

---

ethanndickson:
@codex review

Addressed both P2s from the latest review cycle:

**ProxyCommand shell expansion** — Replaced ad-hoc double-quote escaping with POSIX shell-safe argv rendering. Each token is now single-quoted via a `quoteForPosixShell()` helper, making `$`, backticks, and `$()` inert by construction. Added regression tests for shell metacharacters and embedded single quotes.

**Multi-hop symlink chain** — Replaced single-hop `readlink` with `fs.realpath()` in `resolveSSHConfigWritePath()`, so the writer always targets the canonical final file. Added a test that creates a 2-hop chain (`config → linkA → real`) and asserts both symlinks survive the write.

All squashed into the single feature commit. `make static-check` passes locally.

---

ethanndickson:
@codex review

Addressed the dangling-symlink P2:

- Added `followSymlinkChain()` helper that manually follows `readlink` hops when `realpath` throws ENOENT (dangling target). Resolves the full chain even when the final target file doesn't exist yet, so `writeConfigAtomically` writes to the intended target path instead of destroying the symlink.
- Added two regression tests: single-hop dangling symlink and multi-hop dangling chain. Both assert symlink integrity after write.

All squashed into the single feature commit. `make static-check` passes locally. 25 tests pass.

---

ethanndickson:
@codex review

Addressed the silent error swallowing P2 in `serviceContainer.ts`:

- Replaced the empty `.catch(() => {})` with `log.warn("Background mux SSH config setup failed", { error })`, matching the adjacent `ensureMuxChatWorkspace` error handling pattern.
- When coder is not installed, `ensureMuxCoderSSHConfig` returns early without throwing (null guard), so this `.catch` only fires on real filesystem errors — exactly when diagnostic output is needed.

All squashed into the single feature commit. `make static-check` passes locally.

---

ethanndickson:
@codex review

Changes since last review:
- **Fixed P1**: Host normalization now happens in `runtimeFactory.ts` *before* `createSSHTransport()`, so transport and runtime always use the canonical `*.mux--coder` hostname. Shared `resolveCoderSSHHost()` helper in `src/constants/coder.ts` used by both factory and `CoderSSHRuntime` constructor.
- Added regression test in `runtimeFactory.test.ts` verifying legacy `.coder` host + `coder.workspaceName` produces matching hosts for runtime config and transport.

ethanndickson:
@codex review

Addressed the P1 finding — switched `ProxyCommand` quoting from POSIX single-quotes to cross-platform double-quotes (`escapeCommandArg`), matching `coder/vscode-coder`'s strategy. Also added a Windows-path test to verify backslash preservation.

ethanndickson:
@codex review

Addressed the P1 — `resolveCoderBinaryPath()` now prefers `where.exe coder` on Windows for native paths, with Git Bash `command -v` + `toWindowsPath()` MSYS→Windows normalization as fallback. Added `toWindowsPath` helper to `src/node/utils/paths.ts` with full unit tests.

ethanndickson:
@codex review

ethanndickson:
@codex review

Changes since last review:
- **P2 fix**: Added `coderService.verifyAuthenticatedSession()` — fresh uncached `coder whoami` check
- `CoderSSHRuntime.finalizeConfig` now runs auth preflight for **both** new and existing workspace flows before `ensureProvisioningSession`
- Restores fail-fast behavior for existing-workspace creation when Coder auth is expired/logged out
- Tests: auth preflight failure for existing workspaces, `ensureProvisioningSession` not called for existing workspaces, `verifyAuthenticatedSession` success/failure

ethanndickson:
@codex review

ethanndickson:
@codex review

PR #2565, opened 2026-02-24
🤖 fix: close analytics dashboard on Escape

ethanndickson:
## Summary

Pressing Escape now closes the Analytics dashboard, matching the existing behavior where Escape closes the Settings page. This keeps route-level overlay pages consistent and dismissable with the same muscle memory.

## Implementation

Added a `useEffect` Escape listener in `AnalyticsDashboard` that mirrors the existing pattern from `SettingsPage`:

- Bubble-phase `window` keydown listener so inner surfaces (select dropdowns, popovers) get first right of refusal
- Three standard guards: `matchesKeybind(KEYBINDS.CANCEL)`, `!defaultPrevented`, `!isEditableElement`
- Calls `navigateFromAnalytics()` (existing router function) to restore the previous view

Also extended `isEditableElement()` to include native `<select>` elements. In Chromium, pressing Escape while a native `<select>` is focused bubbles a `keydown` with `defaultPrevented === false`, so without this guard any `isEditableElement`-guarded Escape handler would misfire. This follows the safe-by-default philosophy from #2322 — centralizing the guard in the shared utility benefits all ~10 existing callers.

No router, state, or App.tsx changes needed — the routing infrastructure already existed.

## Validation

- New UI integration test (`tests/ui/layout/analyticsEscape.test.ts`) verifies the full flow: open analytics → press Escape → workspace chat reappears
- `make static-check` passes
- `make typecheck` passes

---

### DIFF
--- a/src/browser/components/analytics/AnalyticsDashboard.tsx
+++ b/src/browser/components/analytics/AnalyticsDashboard.tsx
@@ -147,9 +147,6 @@ export function AnalyticsDashboard(props: AnalyticsDashboardProps) {
       if (!matchesKeybind(e, KEYBINDS.CANCEL)) return;
       if (e.defaultPrevented) return;
       if (isEditableElement(e.target)) return;
-      // Native <select> doesn't set defaultPrevented on Escape, but the user
-      // may be dismissing its dropdown—don't close analytics in that case.
-      if (e.target instanceof HTMLSelectElement) return;
 
       e.preventDefault();
       e.stopPropagation();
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed the native `<select>` guard feedback — added `if (e.target instanceof HTMLSelectElement) return;` so pressing Escape while the project filter dropdown is focused won't close analytics.

ethanndickson:
@codex review

Changes since last review:
- Moved the `<select>` guard from a one-off check in AnalyticsDashboard into `isEditableElement()` in keybinds.ts, following the safe-by-default philosophy from #2322. All ~10 callers of `isEditableElement` now automatically handle native `<select>` elements.

PR #2567, opened 2026-02-24
🤖 fix: distinguish active vs archived workspaces in project removal

### LINKED ISSUE #2491: Improve "Failed to remove project" error when only archived workspaces exist
## Problem

When a user tries to delete a project that has no active workspaces but has archived workspaces, the error message says:

> Cannot remove project with active workspaces. Please remove all N workspace(s) first.

This is misleading because the workspaces aren't "active" — they're archived. The user has no clear path forward from the error alert.

## Suggested Improvements

1. **Update the error text** to distinguish between active and archived workspaces (e.g. "Cannot remove project. Please delete N archived workspace(s) first.")
2. **Include a link/button in the alert** that navigates the user to the archived workspaces view where they can permanently delete them

## Context

The root cause is in `src/node/services/projectService.ts` — the removal check uses `projectConfig.workspaces.length > 0` which counts all workspaces regardless of archived status.

Related: #1266

---

Created on behalf of @tracyjohnsonux

ethanndickson:
## Summary

Replace misleading "active workspaces" project-removal error with a typed blocker model that accurately distinguishes active vs archived workspaces, proactively disabling the remove button with precise tooltips.

Closes #2491

## Background

When a user tried to delete a project that had only archived workspaces, the error said "Cannot remove project with active workspaces" — misleading because the workspaces aren't active. The root cause was `projectConfig.workspaces.length > 0` counting all workspaces regardless of archive state, and a stringly-typed error that couldn't convey the distinction.

## Implementation

**Typed blocker model (foundational redesign):** Instead of patching the error string, project-deletion blockers are now a first-class domain concept shared across backend and frontend:

1. **Shared utility** (`src/common/utils/projectRemoval.ts`): `getProjectWorkspaceCounts()` computes active/archived counts from workspace config entries. Used by both backend policy and frontend sidebar.

2. **Typed ORPC error** (`ProjectRemoveErrorSchema`): Discriminated union with `workspace_blockers` (carrying `activeCount`/`archivedCount`), `project_not_found`, and `unknown` variants — replacing the plain string error.

3. **Backend** (`projectService.ts`): `remove()` returns typed blocker errors with exact counts.

4. **Frontend sidebar** (`ProjectSidebar.tsx`): Remove button is proactively disabled when workspaces exist, with tooltip messaging that distinguishes:
   - "Delete workspace first" / "Delete all N workspaces first"
   - "Delete archived workspace first" / "Delete N archived workspaces first"
   - "Delete N active + M archived workspaces first"

5. **Stale config fix** (`WorkspaceContext.tsx`): `refreshProjects()` is now triggered on archive/unarchive metadata events so `ProjectContext.projects` stays current — without this, the tooltip always showed "Delete workspace first" even for archived workspaces.

## Risks

Low. The backend behavior is unchanged (projects with any workspaces still can't be removed). The frontend disable is additive — if it somehow fails, the backend typed error still provides defense-in-depth.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$11.08`_

### DIFF
--- a/src/browser/stories/App.errors.stories.tsx
+++ b/src/browser/stories/App.errors.stories.tsx
@@ -525,6 +525,18 @@ export const ProjectRemovalDisabled: AppStory = {
         '[role="button"][aria-label="Create workspace in my-app"]'
       );
     if (!projectRow) throw new Error("Project row not found");
+
+    // Make project action affordances deterministic in this story capture.
+    // Synthetic hover can be flaky for pure CSS :hover opacity transitions,
+    // so force key/remove visibility while still exercising hover behavior.
+    const secretsButton = canvasElement.querySelector<HTMLButtonElement>(
+      'button[aria-label="Manage secrets for my-app"]'
+    );
+    if (secretsButton) {
+      secretsButton.style.opacity = "1";
+    }
+    removeButton.style.opacity = "1";
+
     await userEvent.hover(projectRow);
 
     // Hover the remove trigger after row hover so tooltip content appears.


ethanndickson:
@codex review

PR #2568, opened 2026-02-24
🤖 ci: add sticky `nightly` Docker tag for nightly pushes

matifali:
## Summary
Add a sticky `nightly` Docker tag for nightly image pushes so consumers can pull `ghcr.io/coder/mux:nightly` without needing the computed nightly version string.

## Background
Nightly builds were only pushed with a unique version tag (for example, `0.18.1-nightly.42`). That made it inconvenient to track the latest nightly image in automation and local testing.

## Implementation
- Added an explicit optional reusable workflow input `is_latest_nightly` in `.github/workflows/_docker-release.yml`.
- Updated Docker metadata tags so `nightly` is enabled with `inputs.is_latest_nightly`.
- Updated `.github/workflows/nightly.yml` to pass:
  - `is_latest: false`
  - `is_latest_nightly: true`
- Left `.github/workflows/release.yml` unchanged so stable releases continue to update `:latest` only.

## Validation
- `make static-check`

## Risks
Low risk: changes are isolated to workflow-tag inputs and preserve stable release behavior.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `high` • Cost: `$0.33`_

### DIFF
--- a/.github/workflows/_docker-release.yml
+++ b/.github/workflows/_docker-release.yml
@@ -16,6 +16,11 @@ on:
         required: false
         type: boolean
         default: false
+      is_latest_nightly:
+        description: "Whether to also tag the image as :nightly"
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   docker-publish:
@@ -50,7 +55,7 @@ jobs:
           tags: |
             type=raw,value=${{ inputs.version }}
             type=raw,value=latest,enable=${{ inputs.is_latest }}
-            type=raw,value=nightly,enable=${{ !inputs.is_latest }}
+            type=raw,value=nightly,enable=${{ inputs.is_latest_nightly }}
 
       - name: Set up Depot CLI
         uses: depot/setup-action@15c09a5f77a0840ad4bce955686522a257853461 # v1.7.1
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -127,6 +127,7 @@ jobs:
       # Tag as the nightly version (e.g., 0.18.1-nightly.42) — and update sticky :nightly
       version: ${{ needs.compute-version.outputs.nightly_version }}
       is_latest: false
+      is_latest_nightly: true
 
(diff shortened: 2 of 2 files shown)


matifali:
@codex review

matifali:
@codex review

.github/workflows/_docker-release.yml
```diff
@@ -50,6 +50,7 @@ jobs:
           tags: |
             type=raw,value=${{ inputs.version }}
             type=raw,value=latest,enable=${{ inputs.is_latest }}
+            type=raw,value=nightly,enable=${{ !inputs.is_latest }}
```
ThomasK33: Can we make this an explicit input value?

So this should eventually become 
```suggestion
            type=raw,value=nightly,enable=${{ inputs.is_latest_nightly }}
```

.github/workflows/nightly.yml
```diff
@@ -124,7 +124,7 @@ jobs:
     uses: ./.github/workflows/_docker-release.yml
     with:
       ref: ${{ needs.compute-version.outputs.build_sha }}
-      # Tag as the nightly version (e.g., 0.18.1-nightly.42) — not :latest
+      # Tag as the nightly version (e.g., 0.18.1-nightly.42) — and update sticky :nightly
       version: ${{ needs.compute-version.outputs.nightly_version }}
       is_latest: false
```
ThomasK33, line 131: and then here we should pass it in as 


```suggestion
      is_latest: false
      is_latest_nightly: true
```

.github/workflows/_docker-release.yml
ThomasK33, line 24:
```suggestion
is_latest_nightly:
        description: "Whether to also tag the image as :nightly"
        required: false
        type: boolean
        default: false

```

matifali:
Addressed review feedback by introducing explicit `is_latest_nightly` input and wiring nightly workflow to pass it.

@codex review

PR #2569, opened 2026-02-24
🤖 feat: add Coder config defaults to Runtimes settings

ethanndickson:
## Summary

Adds the full Coder workspace configuration form (template, preset, new/existing mode) to the **Runtimes** settings tab, so users can set project-specific Coder defaults. DRYs the implementation by sharing the Coder form component, data-loading hook, and runtime-option persistence helpers between settings and workspace creation.

## Background

Previously, users could only configure Coder defaults (template, preset, new vs existing workspace) during workspace creation. The Runtimes settings screen showed a Coder enable/disable toggle but no config form — meaning there was no way to set project-level Coder defaults from settings. Both screens already read/write the same `lastRuntimeConfig:{projectPath}` localStorage key, but with duplicated parsing logic and an asymmetric feature surface.

## Implementation

Applied the **elegant-redesign** principle — designed as if Coder config in settings was always a foundational assumption:

1. **Shared persistence helpers** (`src/browser/utils/runtimeOptionDefaults.ts`) — typed `readSshOptionDefaults`/`writeSshOptionDefaults` + generic `readOptionField`/`writeOptionField` replace duplicated inline read/write logic in both settings and creation.

2. **Shared Coder form** (`src/browser/components/runtime/CoderControls.tsx`) — moved `CoderWorkspaceForm`, `CoderAvailabilityMessage`, and `resolveCoderAvailability` out of the ChatInput namespace into a context-agnostic location. Added `className` prop for layout flexibility.

3. **Settings integration** — `RuntimesSection` now renders `CoderWorkspaceForm` under the Coder row when in project scope + Coder enabled + CLI available. Wired to the same localStorage key the creation flow uses, so edits in either screen are immediately reflected in the other.

4. **DRY data loading** — replaced RuntimesSection's standalone `api.coder.getInfo()` effect with the shared `useCoderWorkspace` hook, eliminating duplicate Coder availability fetching.

5. **Eager data loading** — templates/presets/workspaces load as soon as the Coder row is visible in project scope, even before "Override project settings" is enabled, so users see real options immediately.

6. **Hidden when unavailable** — the config form is hidden entirely when the Coder CLI is unavailable/not logged in; only the toggle + "Unavailable" badge show.

## Validation

- `bun run typecheck` ✅
- `bunx eslint` on all changed files ✅
- `bun run fmt:check` ✅
- `bun test ./tests/ui/workspaces/draft.test.ts` ✅ (creation regression)
- `bun test ./tests/ui/runtime/docker.test.ts` ✅ (runtime regression)
- Full UI test suite: 0 new failures (10 pre-existing)

---


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

PR #2579, opened 2026-02-24
🤖 fix: keep layout actions enabled in Settings

ThomasK33:
## Summary
Enable layout actions in Settings by falling back to the last selected workspace when URL-derived workspace selection is unavailable.

## Background
`LayoutsSection` depended on `selectedWorkspace`, which is derived from `/workspace/:id`. Navigating to `/settings/layouts` clears that route context, so `Add layout` (and related actions) became disabled despite layout state still being available for the user's previous workspace.

## Implementation
- Read `SELECTED_WORKSPACE_KEY` via `readPersistedState` in `LayoutsSection`.
- Added a defensive type guard for persisted selection payloads.
- Resolve fallback workspace through `workspaceMetadata` so we only use valid, current workspaces.
- Use resolved workspace fallback for both action enablement (`workspaceId`) and label display.

## Validation
- `make static-check`

## Risks
Low risk and scoped to layout settings behavior. Fallback is only used when route-derived selection is missing and only when metadata contains the referenced workspace.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$0.75`_


ThomasK33:
@codex review

Addressed your feedback by accepting workspaceId-only persisted selection payloads in the Settings fallback.

PR #2586, opened 2026-02-24
🤖 fix: change Electron appId to com.coder.mux

ThomasK33:
## Summary

Change the Electron Builder application identifier from `com.mux.app` to `com.coder.mux`.

## Background

The current app ID uses the legacy `com.mux.app` identifier. This PR updates branding alignment to `com.coder.mux`.

## Implementation

- Updated `build.appId` in `package.json`:
  - `com.mux.app` → `com.coder.mux`

## Validation

- `make static-check`

## Risks

**Potentially breaking for existing installs.** Changing Electron Builder `appId` can affect OS-level app identity and update continuity:
- macOS: bundle identifier (`CFBundleIdentifier`)
- Windows: AUMID / NSIS identity

Existing installations that were distributed with `com.mux.app` may require reinstall/migration handling instead of seamless in-place updates.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$0.28`_

### DIFF
--- a/package.json
+++ b/package.json
@@ -252,7 +252,7 @@
     "LICENSE"
   ],
   "build": {
-    "appId": "com.mux.app",
+    "appId": "com.coder.mux",
     "productName": "mux",
     "protocols": [
       {


ThomasK33:
Thanks for the callout.

This appId change is intentional in this PR, and I’ve marked it as potentially breaking in the PR body. I agree this can break in-place update continuity for users already installed under `com.mux.app`.

Given that migration across OS-level app identity is a release/product decision (not a simple code-only fix), I’m pausing for maintainer direction on whether we should:
1) keep this change and ship with a one-time manual reinstall path, or
2) revert and keep `com.mux.app` for update continuity.

PR #2589, opened 2026-02-24
🤖 fix: refresh coder auth on window focus in creation flow

ethanndickson:
## Summary
Refresh Coder auth state whenever the app window regains focus on the new-workspace (creation) flow so users who run `coder login` externally are detected without restarting or navigating away. Focus refresh is scoped to the creation variant only, and overlapping auth fetches use latest-wins sequencing to prevent stale responses from mutating state.

## Background
`useCoderWorkspace` only checked `api.coder.getInfo()` on mount. If the CLI was logged out at that point, the UI stayed in that state until remount. Codex review flagged two P2 issues: the focus listener was unconditionally attached (firing on every focus even in workspace chat), and overlapping mount/focus fetches could race.

## Implementation
- Introduced `CoderInfoRefreshPolicy` type (`"mount-only" | "mount-and-focus"`) — the hook contract now explicitly encodes refresh intent
- Creation variant passes `"mount-and-focus"`, workspace variant passes `"mount-only"`
- Added monotonic sequence counter (`latestAuthSeqRef`) to `fetchCoderInfo` — stale responses are silently discarded
- Focus listener only binds when policy is `"mount-and-focus"`

## Validation
- `make static-check` — all green
- `bun test src/browser/hooks/useCoderWorkspace.test.ts` — 8/8 pass
  - mount-only policy does not refetch on focus
  - mount-and-focus policy refetches on focus
  - latest-wins race guard: stale failure cannot overwrite newer success

## Risks
Low. Scoped to one hook + one callsite. Backend caching semantics unchanged. New tests cover both regressions.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$3.49`_

### DIFF
--- a/src/browser/hooks/useCoderWorkspace.ts
+++ b/src/browser/hooks/useCoderWorkspace.ts
@@ -124,9 +124,11 @@ export function useCoderWorkspace({
   const [loadingWorkspaces, setLoadingWorkspaces] = useState(false);
 
   const fetchCoderInfo = useCallback(async () => {
-    if (!api) return;
-
+    // Advance sequence before the API guard so that an API context change
+    // (e.g., reconnecting -> api becomes null) invalidates any in-flight request.
     const seq = ++latestAuthSeqRef.current;
+
+    if (!api) return;
 
     try {
       const info = await api.coder.getInfo();


ethanndickson:
@codex review

Addressed both P2 items:

1. **Scope leak → fixed**: introduced `CoderInfoRefreshPolicy` type; focus listener only binds when `"mount-and-focus"` (creation variant). Workspace variant passes `"mount-only"` — no focus-triggered CLI calls.

2. **Stale auth race → fixed**: `fetchCoderInfo` now uses a monotonic sequence counter; stale responses (where `seq !== latestAuthSeqRef.current`) are discarded in both success and error paths.

Added 3 regression tests covering both issues + the race scenario.

ethanndickson:
@codex review

Addressed P2: moved `const seq = ++latestAuthSeqRef.current` before the `if (!api) return` guard so API context changes (reconnecting → null) invalidate in-flight requests.

PR #2597, opened 2026-02-24
🤖 feat: immersive review minimap cleanup + hunk selection fixes

ibetitsmike:
## Summary

Improves immersive code review UX: fixes a bug where Shift+Arrow → Shift+C placed the comment composer at the wrong line, and makes the diff view full-width by removing inset padding/borders.

## Background

When using keyboard selection (Shift+Arrow) to select lines that partially extend outside a hunk range, `isSelectionInsideRange` returns `false` and the code collapses the entire multi-line selection to a single fallback line — the composer appears at the wrong position. Additionally, `findHunkAtLine` only checked one end of the selection, causing it to miss valid hunks.

The diff view had horizontal padding (`p-3`) and an inset border/rounded style that wasted space and didn't match the full-width minimap design.

### Composer selection fix (`openComposer`)
- **Try both ends when resolving the hunk** — when `findHunkAtLine(selection.endIndex)` returns null, also try `selection.startIndex`, so a selection that starts inside a hunk but ends outside still resolves correctly.
- **Clamp selection to hunk intersection** — instead of collapsing to a single fallback line, compute the overlap between the user's selection and the target hunk range via min/max clamping. Only fall back to single-line when there's truly no overlap.

### Full-width layout
- Scroll container: `p-3` → `py-3` (vertical-only padding).
- Inner diff wrapper: removed `rounded`, `border`, and `border-border-light` classes.

## Risks

Low — the composer fix is scoped to `openComposer` and the layout change is CSS-only. Both are isolated to the immersive review view.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$18.42`_

### DIFF
--- a/src/browser/styles/globals.css
+++ b/src/browser/styles/globals.css
@@ -1063,6 +1063,12 @@ body {
   display: none;
 }
 
+/* Keep `.scrollbar-none` fully hidden even with global Firefox scrollbar theming. */
+.scrollbar-none {
+  scrollbar-width: none;
+  scrollbar-color: auto;
+}
+
 /* Root container */
 html,
 body,


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed the P2 comment: added `buildOldLineNumberToIndexMap` so deletion-only review ranges now resolve through the correct old-side line map. Old-only ranges were previously looked up in the new-side map, causing minimap comment markers to be missing or misplaced.

ibetitsmike:
@codex review

The previous deletion-line comment was already addressed in commit 98edf84 — `commentLineIndices` now uses `buildOldLineNumberToIndexMap` for old-only ranges, so deletion comments map through the correct old-side line map. Please re-review.

ibetitsmike:
@codex review

PR #2600, opened 2026-02-24
🤖 feat: add inject toggle for global secrets

ThomasK33:
## Summary

Add an `injectAll` toggle for global secrets and show a read-only "Injected from Global" list in project settings so inherited env keys are visible.

## Background

Global secrets previously acted only as shared storage and were injected only when a project secret explicitly referenced them. This change makes common secrets (e.g. org-wide tokens) opt-in injectable across all projects.

## Implementation

- Extended shared `SecretSchema` with optional `injectAll`.
- Updated `Config.getEffectiveSecrets()` to:
  - include globally opted-in secrets for all projects,
  - continue resolving project `{ secret: "KEY" }` aliases,
  - preserve project-over-global precedence for duplicate keys,
  - honor last-writer semantics for duplicate global keys when evaluating `injectAll`.
- Sanitized malformed persisted `injectAll` values by stripping the flag while preserving valid `key`/`value` secrets.
- Treated inject-all as enabled only when `injectAll === true`.
- Added a global-scope **Inject** switch in settings UI and wired dirty-state/save behavior.
- Added `secrets.getInjectedGlobals(projectPath)` API and `Config.getInjectedGlobalSecrets()` to expose injected global keys for a project in a read-only way.
- Added a read-only **Injected from Global** section in project secrets settings so users can see inherited keys and understand project-over-global precedence.
- Added config tests covering injection, override precedence, coexistence with project secrets, non-injection defaults, duplicate-key edge cases, and malformed `injectAll` recovery.

## Validation

- `bun test src/node/config.test.ts` ✅
- `make typecheck` ✅
- `make lint` ✅
- `make static-check` ✅

## Risks

- Low-to-moderate: changes affect secret resolution and settings UX.
- Mitigated by explicit precedence logic and targeted tests for injectAll behavior.

---

<details>

# Plan: Add `injectAll` Toggle to Global Secrets

## Context & Goal

Global secrets today are **shared storage only** — they are never injected into project environments unless a project explicitly references them via `{ secret: "GLOBAL_KEY" }`. The user wants a simple toggle on each global secret so that when enabled, that secret is **automatically injected into every project's environment** without requiring per-project references.

**User impact:** One toggle per secret replaces N per-project reference entries. New projects automatically get the secret with zero configuration.

## Architecture Summary

The existing secrets system has 3 layers:

1. **Schema** (`SecretSchema` in `src/common/orpc/schemas/secrets.ts`) → defines `{ key, value }`
2. **Resolution** (`Config.getEffectiveSecrets` in `src/node/config.ts`) → merges project secrets + resolved global references
3. **Injection** (callers of `secretsToRecord(config.getEffectiveSecrets(projectPath))`) → converts to `Record<string, string>` for `process.env`

All callers already go through `getEffectiveSecrets` → `secretsToRecord`. This means a single change in `getEffectiveSecrets` propagates to every injection point (workspace init, bash tool, terminal, tasks, AI service).

## Estimated LoC: ~120 product code, ~80 test code

---

### Step 1: Extend `SecretSchema` with `injectAll`

**File:** `src/common/orpc/schemas/secrets.ts`

Add an optional `injectAll` boolean to the schema. Existing secrets without this field default to `false` (backward-compatible).

```ts
export const SecretSchema = z
  .object({
    key: z.string(),
    value: SecretValueSchema,
    injectAll: z.boolean().optional(),
  })
  .meta({
    description: "A key-value pair for storing sensitive configuration",
  });
```

**Why `.optional()`:** This field only appears on global secrets. The Zod schema is shared between global and project secrets, so the field must be optional. Existing `secrets.json` files without `injectAll` parse cleanly — `undefined` is treated as `false`.

> Note: We use `.optional()` here (not `.nullish()`) because this is **not a tool input parameter** — it's a persisted data schema. The `.nullish()` convention in AGENTS.md applies specifically to tool input schemas for AI model consumption.

### Step 2: Update `Config.getEffectiveSecrets` to merge `injectAll` globals

**File:** `src/node/config.ts`, method `getEffectiveSecrets` (line ~1367)

Currently this method only returns project secrets (with global references resolved). Change it to also prepend global secrets that have `injectAll: true`, with project secrets taking precedence (last-writer-wins).

```ts
getEffectiveSecrets(projectPath: string): Secret[] {
  const normalizedProjectPath = Config.normalizeSecretsProjectPath(projectPath) || projectPath;
  const config = this.loadSecretsConfig();
  const globalSecrets = config[Config.GLOBAL_SECRETS_KEY] ?? [];
  const projectSecrets = config[normalizedProjectPath] ?? [];

  // Collect global secrets marked for universal injection.
  // These are injected as literal values (resolved) so downstream
  // consumers don't need to understand the injectAll flag.
  const globalSecretsByKey = secretsToRecord(globalSecrets);
  const injectedGlobals: Secret[] = [];
  for (const gs of globalSecrets) {
    if (!gs.injectAll) continue;
    const resolved = globalSecretsByKey[gs.key];
    if (resolved !== undefined) {
      injectedGlobals.push({ key: gs.key, value: resolved });
    }
  }

  // Resolve project secret references against global values.
  const resolvedProjectSecrets = projectSecrets.map((secret) => {
    if (!Config.isSecretReferenceValue(secret.value)) {
      return secret;
    }
    const targetKey = secret.value.secret.trim();
    if (!targetKey) return secret;
    const resolvedGlobalValue = globalSecretsByKey[targetKey];
    if (resolvedGlobalValue !== undefined) {
      return { ...secret, value: resolvedGlobalValue };
    }
    return secret;
  });

  // Merge: project secrets override injected globals (last-writer-wins).
  // Put injected globals first so project secrets can override them.
  const projectKeys = new Set(resolvedProjectSecrets.map((s) => s.key));
  const nonOverriddenGlobals = injectedGlobals.filter((g) => !projectKeys.has(g.key));
  return [...nonOverriddenGlobals, ...resolvedProjectSecrets];
}
```

**Key behaviors:**

- `injectAll` globals are prepended, so explicit project secrets always win.
- A project can override an `injectAll` global by defining the same key (either literal or reference).
- The `injectAll` field is stripped from the output — callers receive plain `{ key, value: string }` entries.

### Step 3: Update `Config.isSecret` to accept `injectAll`

**File:** `src/node/config.ts`, static method `isSecret` (line ~1245)

The `isSecret` type guard validates persisted data. It needs to tolerate the `injectAll` field. Currently it checks for `key` and `value` — no change needed since it doesn't reject extra properties. But verify this is the case. If `parseSecretsArray` or `normalizeSecretsConfig` strips unknown fields, those need updating too.

After review: `parseSecretsArray` filters via `Config.isSecret()` which checks `typeof value.key === "string" && isSecretValue(value.value)`. The `injectAll` field passes through as an extra property — **no code change needed** for parsing/normalization.

### Step 4: Update the Secrets Settings UI — add toggle per global secret row

**File:** `src/browser/components/Settings/sections/SecretsSection.tsx`

When `scope === "global"`, add a `Switch` toggle in each secret row for `injectAll`. The toggle should be inline and only appear in global scope (project secrets don't have `injectAll`).

**4a.** Import the `Switch` component:

```ts
import { Switch } from "@/browser/components/ui/switch";
```

**4b.** Add an `updateSecretInjectAll` callback:

```ts
const updateSecretInjectAll = useCallback((index: number, checked: boolean) => {
  setSecrets((prev) => {
    const next = [...prev];
    const existing = next[index] ?? { key: "", value: "" };
    next[index] = { ...existing, injectAll: checked || undefined };
    return next;
  });
}, []);
```

> `checked || undefined`: when `false`, omit the field entirely to keep `secrets.json` clean for secrets that don't use the feature.

**4c.** Update the grid layout for global scope to include the toggle column. Currently:

- Global: `grid-cols-[1fr_1fr_auto_auto]` (key, value, eye, trash)
- Project: `grid-cols-[1fr_auto_1fr_auto_auto]` (key, type, value, eye, trash)

Change global to: `grid-cols-[1fr_1fr_auto_auto_auto]` (key, value, eye, inject-toggle, trash)

**4d.** Add a column header label `Inject` (only in global scope).

**4e.** Render the `Switch` in each global secret row:

```tsx
{
  scope === "global" && (
    <div className="flex items-center justify-center self-center">
      <Switch
        checked={!!secret.injectAll}
        onCheckedChange={(checked) => updateSecretInjectAll(index, checked)}
        disabled={saving}
        aria-label="Inject into all projects"
        title="Inject into all projects"
      />
    </div>
  );
}
```

**4f.** Fix the `secretsEqual` comparison to include `injectAll`:

```ts
function secretsEqual(a: Secret[], b: Secret[]): boolean {
  if (a.length !== b.length) return false;
  for (let i = 0; i < a.length; i++) {
    const left = a[i];
    const right = b[i];
    if (!left || !right) return false;
    if (left.key !== right.key) return false;
    if (!secretValuesEqual(left.value, right.value)) return false;
    if (!!left.injectAll !== !!right.injectAll) return false;
  }
  return true;
}
```

**4g.** Update the help text. Replace:

```
Global secrets are shared storage only; they are not injected by default.
```

With:

```
Toggle "Inject" on a global secret to automatically inject it into every project.
```

### Step 5: Add tests

**File:** `src/node/config.test.ts`

Add tests in the existing `describe("secrets")` block:

```ts
it("injects injectAll global secrets into effective secrets for any project", async () => {
  await config.updateGlobalSecrets([
    { key: "INJECTED", value: "everywhere", injectAll: true },
    { key: "NOT_INJECTED", value: "stored-only" },
  ]);

  const projectPath = "/fake/project";
  // No project secrets configured at all.
  const record = secretsToRecord(config.getEffectiveSecrets(projectPath));
  expect(record).toEqual({ INJECTED: "everywhere" });
});

it("project secrets override injectAll global secrets", async () => {
  await config.updateGlobalSecrets([
    { key: "TOKEN", value: "global-value", injectAll: true },
  ]);

  const projectPath = "/fake/project";
  await config.updateProjectSecrets(projectPath, [
    { key: "TOKEN", value: "project-value" },
  ]);

  const record = secretsToRecord(config.getEffectiveSecrets(projectPath));
  expect(record).toEqual({ TOKEN: "project-value" });
});

it("injectAll global secrets coexist with project-specific secrets", async () => {
  await config.updateGlobalSecrets([
    { key: "GLOBAL_TOKEN", value: "g", injectAll: true },
  ]);

  const projectPath = "/fake/project";
  await config.updateProjectSecrets(projectPath, [
    { key: "LOCAL_TOKEN", value: "l" },
  ]);

  const record = secretsToRecord(config.getEffectiveSecrets(projectPath));
  expect(record).toEqual({ GLOBAL_TOKEN: "g", LOCAL_TOKEN: "l" });
});

it("does not inject injectAll:false or undefined global secrets", async () => {
  await config.updateGlobalSecrets([
    { key: "A", value: "1", injectAll: false },
    { key: "B", value: "2" },
    { key: "C", value: "3", injectAll: true },
  ]);

  const record = secretsToRecord(config.getEffectiveSecrets("/any/project"));
  expect(record).toEqual({ C: "3" });
});
```

### Step 6: Update the existing test assertion

**File:** `src/node/config.test.ts`

The existing test `"does not inherit global secrets by default"` (line 339) should still pass because:

- Global secrets without `injectAll: true` are **not** injected.
- This test sets global secrets without `injectAll`, so behavior is unchanged.

**Verify** this test still passes after the change. No modification expected.

---

## Files Changed (Summary)

| File                                                          | Change                                                      | LoC                        |
| ------------------------------------------------------------- | ----------------------------------------------------------- | -------------------------- |
| `src/common/orpc/schemas/secrets.ts`                          | Add `injectAll: z.boolean().optional()` to `SecretSchema`   | +1                         |
| `src/node/config.ts`                                          | Rewrite `getEffectiveSecrets` to merge `injectAll` globals  | ~+15 net                   |
| `src/browser/components/Settings/sections/SecretsSection.tsx` | Add `Switch` toggle, grid column, help text, equality check | ~+30 net                   |
| `src/node/config.test.ts`                                     | Add 4 new test cases                                        | ~+40                       |
| **Total**                                                     |                                                             | **~85 product + ~40 test** |

## Validation

1. `make typecheck` — ensure `Secret` type flows correctly everywhere.
2. `bun test src/node/config.test.ts` — new + existing secret tests pass.
3. `make lint` — no lint errors.
4. Manual: open Settings → Secrets → Global, verify toggle renders and persists.

</details>

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$2.45`_

### DIFF
--- a/src/node/orpc/router.ts
+++ b/src/node/orpc/router.ts
@@ -1673,7 +1673,11 @@ export const router = (authToken?: string) => {
         .input(schemas.secrets.getInjectedGlobals.input)
         .output(schemas.secrets.getInjectedGlobals.output)
         .handler(({ context, input }) => {
-          const projectPath = input.projectPath.trim();
+          const projectPath =
+            typeof input.projectPath === "string" && input.projectPath.trim().length > 0
+              ? input.projectPath
+              : undefined;
+
           if (!projectPath) {
             return [];
           }


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed both review items and pushed fixes. Please take another look.

ThomasK33:
@codex review

Addressed malformed injectAll sanitization and added regression coverage. Please re-review.

ThomasK33:
@codex review

Added project-scope visibility for globally injected secrets and corresponding API support/tests. Please review.

ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed the grid-column alignment issue and separated injected-globals refresh failures from save failures. Please take another look.

ThomasK33:
@codex review

ThomasK33:
@codex review

Handled project-scope injected-globals loading as best-effort so the main secrets editor still loads if the supplemental fetch fails. Please take another look.

ThomasK33:
@codex review

ThomasK33:
@codex review

Added a request-version guard in loadSecrets so stale async responses from an older project/scope selection cannot overwrite the current injected-globals state. Please take another look.

ThomasK33:
@codex review

Rebased onto current origin/main, reran local validation, and force-pushed the updated branch. Please take another look.

ThomasK33:
@codex review

Updated getInjectedGlobals to only trim for emptiness checks and preserve the original project path for lookup, matching the other secrets handlers. Please take another look.

PR #2603, opened 2026-02-24
🤖 feat: add Change SSH Host for existing workspaces

coadler:
## Summary

Add the ability to change the SSH host of an existing workspace from the workspace actions menu and command palette, so users don't lose chat history when a server changes address.

## Background

Users who create SSH workspaces currently have no way to change the SSH host after creation. If a server changes hostname/IP or the user needs to switch machines, they must create a new workspace and lose their entire chat history. Since the runtime is recreated fresh on every `sendMessage` from `metadata.runtimeConfig`, updating `runtimeConfig` in config.json takes effect on the very next message — no restart needed.

## Implementation

**Backend (3 files):**
- Added `updateRuntimeConfig` IPC schema in `api.ts` with `workspaceId` + `runtimeConfig` input
- Added `updateRuntimeConfig()` method in `workspaceService.ts` following the exact `updateTitle` pattern (validate workspace → update config → emit metadata)
- Added router handler in `router.ts` wired to the service method
- Leverages existing `config.updateWorkspaceMetadata()` which already accepts `runtimeConfig`

**Frontend (8 files):**
- Added `updateWorkspaceRuntimeConfig` callback to `WorkspaceContext` following `updateWorkspaceTitle` pattern
- Created `ChangeSSHHostDialog.tsx` — Radix Dialog that preserves all SSH config fields (srcBaseDir, identityFile, port, coder) via spread, only changing `host`
- Added `onChangeSSHHost` menu item to `WorkspaceActionsMenuContent` (only shown for SSH workspaces)
- Wired dialog in both `WorkspaceMenuBar.tsx` and `WorkspaceListItem.tsx`
- Added `ws:change-ssh-host` command palette command (only visible when SSH workspaces exist) with select+text prompt pattern
- Wired command palette callback in `App.tsx`

## Validation

image
image

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `high` • Cost: `$1.02`_

### DIFF
--- a/src/node/services/workspaceService.ts
+++ b/src/node/services/workspaceService.ts
@@ -2300,8 +2300,22 @@ export class WorkspaceService extends EventEmitter {
       if (!workspace) {
         return Err("Workspace not found");
       }
+      const { projectPath, workspacePath } = workspace;
 
-      await this.config.updateWorkspaceMetadata(workspaceId, { runtimeConfig });
+      // Use editConfig with path-fallback (same pattern as updateTitle) so
+      // legacy workspaces that don't yet have w.id in config are still found.
+      await this.config.editConfig((config) => {
+        const projectConfig = config.projects.get(projectPath);
+        if (projectConfig) {
+          const workspaceEntry =
+            projectConfig.workspaces.find((w) => w.id === workspaceId) ??
+            projectConfig.workspaces.find((w) => w.path === workspacePath);
+          if (workspaceEntry) {
+            workspaceEntry.runtimeConfig = runtimeConfig;
+          }
+        }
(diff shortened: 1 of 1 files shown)


coadler:
@codex review

coadler:
@codex review

coadler:
@codex review

Addressed your P2 comment: replaced  with the same  + path-fallback pattern used by , so legacy workspaces without  in config are still found and updated correctly.

coadler:
cc @ThomasK33 doesn't seem like I can get integration tests working as a pr from a fork, but this is ready otherwise

PR #2604, opened 2026-02-24
🤖 fix: recover sub-agent analytics data from archived transcripts

ThomasK33:
## Summary

Fix sub-agent cost data being lost from DuckDB analytics. When sub-agent workspaces are removed, `clearWorkspace` deletes their analytics rows from DuckDB, and `rebuildAll` can't recover them because the session directory has been deleted. This PR leverages the existing transcript archival mechanism to re-ingest sub-agent data from archived transcripts stored in the parent workspace's session directory.

## Background

Three independent data loss mechanisms cause the analytics page to underreport total spend and lose all sub-agent timing/performance metrics:

1. **Guaranteed kill**: `clearWorkspace(childId)` fires on every workspace removal, deleting all DuckDB rows for that workspace.
2. **Race condition**: `ingestWorkspace` dispatches to a background worker via `postMessage` while `TaskService` proceeds to `fs.rm(sessionDir)` — the worker may not read `chat.jsonl` before it's deleted.
3. **Unrecoverable rebuild**: `rebuildAll` iterates `~/.mux/sessions/` directories, but deleted sub-agent session dirs are invisible.

The live "Costs" tab works correctly (via in-memory rollup), so users don't notice until they check the analytics page.

## Implementation

**Key insight**: The workspace removal flow already archives each sub-agent's `chat.jsonl` into `parentSessionDir/subagent-transcripts/<childId>/chat.jsonl` before deleting the session directory. We extend this to also archive `metadata.json`, then teach `ingestWorkspace` to scan and re-ingest archived sub-agent transcripts.

### Changes

1. **Archive `metadata.json`** (`workspaceService.ts`): Copy `metadata.json` alongside `chat.jsonl` during sub-agent removal, so archived transcript directories are self-contained for ETL ingestion.

2. **`ingestArchivedSubagentTranscripts` helper** (`etl.ts`): New function that scans `sessionDir/subagent-transcripts/` and calls `ingestWorkspace` for each archived child directory. The archive is flat (existing rollup copies grandchild dirs up to parent level), so a single `readdir` catches all descendants.

3. **Call from `ingestWorkspace`** (`etl.ts`): After writing the parent's watermark, scan and ingest archived sub-agent transcripts. This runs on all code paths (the `shouldRebuild` early return was converted to `if/else`). `rebuildAll` delegates to `ingestWorkspace` automatically.

### Self-healing properties

- Delete the DuckDB database → restart → `rebuildAll` → full recovery from parent session dirs
[...]

PR #2607, opened 2026-02-24
🤖 fix: avoid analytics rebuild loops from archived sub-agent watermarks

ThomasK33:
## Summary
Fix analytics startup backfill detection so archived sub-agent watermark rows are treated as valid workspace coverage instead of stale rows.

## Background
`ingestArchivedSubagentTranscripts` writes watermark rows keyed by archived child workspace IDs, but `needsBackfill` previously only considered top-level session workspace IDs. That mismatch caused every startup to think watermark rows referenced missing workspaces and repeatedly trigger `rebuildAll`.

## Implementation
- Added `workspaceDiscovery.ts` to centralize workspace ID discovery:
  - `listSessionWorkspaceIdsWithHistory`
  - `listArchivedSubagentWorkspaceIds`
- Updated `analyticsWorker.handleNeedsBackfill` to merge archived child IDs into the known workspace set before computing `hasWatermarkMissingSessionWorkspace`.
- Added `workspaceDiscovery.test.ts` coverage for top-level and archived child discovery behavior.

## Validation
- `bun test src/node/services/analytics/workspaceDiscovery.test.ts`
- `bun test src/node/services/analytics/backfillDecision.test.ts`
- `make typecheck`
- `make static-check`

## Risks
Low-to-moderate: this changes startup workspace coverage detection for rebuild decisions. The added tests lock behavior for both top-level session and archived sub-agent transcript layouts.

---

<details>

# Fix: Analytics Tab Rebuild Loop Caused by Archived Sub-Agent Watermarks

## Context

Commit `f307ce8ea` ("recover sub-agent analytics data from archived transcripts", PR #2604) introduced a startup rebuild loop that causes the analytics DuckDB database to perform a full `rebuildAll` on **every app startup**, making the analytics tab significantly slower to load.

**Root cause:** `ingestArchivedSubagentTranscripts` stores watermarks keyed by child workspace IDs (e.g. `child-abc`), but the startup staleness check (`handleNeedsBackfill`) only scans top-level directories under `sessionsDir/`. Archived sub-agents live at `sessionsDir/<parent>/subagent-transcripts/<child>/`, so they're never found → the check concludes their watermarks are "stale" → triggers `rebuildAll` → re-ingests the same sub-agents → same mismatch on next startup → ∞.

## Evidence

| Source | Finding |
|---|---|
| `analyticsWorker.ts:213-255` | `listSessionWorkspaceIdsWithHistory` scans only top-level `sessionsDir` entries |
| `analyticsWorker.ts:312-314` | `hasWatermarkMissingSessionWorkspace` checks watermark IDs against that top-level set |
| `backfillDecision.ts:57-61` | `hasWatermarkMissingSessionWorkspace === true` → `return true` (rebuild) |
| `etl.ts:917-967` | `ingestArchivedSubagentTranscripts` stores watermarks keyed by `childWorkspaceId` (the nested dir name) |
| `analyticsWorker.ts:104-109` | `ingest_watermarks` table has no column to distinguish archived sub-agents from top-level workspaces |

## Approach

Extend `listSessionWorkspaceIdsWithHistory` (or add a companion) to also discover archived sub-agent workspace IDs so the staleness check recognizes them as valid.

**Why this approach over alternatives:**
- No schema migration needed (no new columns on `ingest_watermarks`)
- No changes to the `backfillDecision.ts` pure-function logic
- The scan already reads top-level dirs; scanning one level deeper for `subagent-transcripts/` is bounded I/O

<details>
<summary>Alternatives considered</summary>

1. **Add `is_archived` column to `ingest_watermarks`** — requires schema migration, more invasive for a simple bug fix.
2. **Prefix archived sub-agent workspace IDs** (e.g. `archived:child-abc`) — breaks the identity contract that the same workspace ID is used everywhere.
3. **Skip staleness check entirely** — would mask real stale-data scenarios that the check was designed to catch.

</details>

### 1. Add `listArchivedSubagentWorkspaceIds` helper — `analyticsWorker.ts`

Add a function that, for each top-level session directory, scans `<dir>/subagent-transcripts/` for child directories containing `chat.jsonl` and returns their workspace IDs.

```typescript
// New constant (import from etl.ts or duplicate — it's a well-known directory name)
const SUBAGENT_TRANSCRIPTS_DIR_NAME = "subagent-transcripts";

async function listArchivedSubagentWorkspaceIds(
  sessionsDir: string,
  parentWorkspaceIds: string[],
): Promise<string[]> {
  const archivedIds: string[] = [];

  for (const parentId of parentWorkspaceIds) {
    const transcriptsDir = path.join(sessionsDir, parentId, SUBAGENT_TRANSCRIPTS_DIR_NAME);
    let entries: Dirent[];
    try {
      entries = await fs.readdir(transcriptsDir, { withFileTypes: true });
    } catch (error) {
      if (isRecord(error) && error.code === "ENOENT") {
        continue;
      }
      throw error;
    }

    for (const entry of entries) {
      if (!entry.isDirectory()) {
        continue;
      }
      const chatPath = path.join(transcriptsDir, entry.name, CHAT_FILE_NAME);
      try {
        const chatStat = await fs.stat(chatPath);
        if (chatStat.isFile()) {
          archivedIds.push(entry.name);
        }
      } catch (error) {
        if (isRecord(error) && error.code === "ENOENT") {
          continue;
        }
        throw error;
      }
    }
  }

  return archivedIds;
}
```

### 2. Update `handleNeedsBackfill` — `analyticsWorker.ts`

After calling `listSessionWorkspaceIdsWithHistory`, also call the new helper and merge results into the set used for the staleness check:

```typescript
// In handleNeedsBackfill, after line ~301:
const sessionWorkspaceIds = await listSessionWorkspaceIdsWithHistory(data.sessionsDir);

// NEW: also discover archived sub-agent workspace IDs
const archivedSubagentIds = await listArchivedSubagentWorkspaceIds(
  data.sessionsDir,
  sessionWorkspaceIds,
);

const allKnownWorkspaceIds = new Set([...sessionWorkspaceIds, ...archivedSubagentIds]);

// Then use allKnownWorkspaceIds instead of sessionWorkspaceIdSet for the check:
const hasWatermarkMissingSessionWorkspace = [...watermarkWorkspaceIds].some(
  (workspaceId) => !allKnownWorkspaceIds.has(workspaceId),
);
```

Note: `sessionWorkspaceCount` continues to use only top-level workspace count (the archived sub-agents don't represent independent session workspaces for the other heuristics).

### 3. Add/update tests

- **`backfillDecision.test.ts`**: No changes needed (pure function, existing tests still valid).
- **`analyticsWorker` integration or `etl.test.ts`**: Add a test that sets up a session directory with archived sub-agent transcripts, runs ingestion, then verifies `handleNeedsBackfill` returns `needsBackfill: false` on the second call.

### 4. Validation

```bash
run_and_report typecheck make typecheck
run_and_report analytics-tests bun test src/node/services/analytics/
run_and_report lint make lint
```

</details>

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$1.50`_

### DIFF
--- a/src/node/services/analytics/workspaceDiscovery.test.ts
+++ b/src/node/services/analytics/workspaceDiscovery.test.ts
@@ -2,7 +2,8 @@ import { randomUUID } from "node:crypto";
 import * as fs from "node:fs/promises";
 import * as os from "node:os";
 import * as path from "node:path";
-import { afterEach, describe, expect, test } from "bun:test";
+import { afterEach, describe, expect, spyOn, test } from "bun:test";
+import { log } from "@/node/services/log";
 import { CHAT_FILE_NAME } from "./etl";
 import {
   listArchivedSubagentWorkspaceIds,
@@ -86,6 +87,43 @@ describe("listArchivedSubagentWorkspaceIds", () => {
     // count as known so their watermark rows do not trigger rebuildAll loops.
     const knownWorkspaceIdSet = new Set([...sessionWorkspaceIds, ...archivedWorkspaceIds]);
     expect(knownWorkspaceIdSet.has(childWorkspaceId)).toBe(true);
+  });
+
+  test("skips unreadable archived transcript directories instead of throwing", async () => {
+    const sessionsDir = await createTempSessionsDir();
+    const parentWorkspaceId = "parent-a";
+
+    const parentSessionDir = path.join(sessionsDir, parentWorkspaceId);
+    await fs.mkdir(parentSessionDir, { recursive: true });
+    await writeChatJsonl(parentSessionDir);
+
+    const transcriptsDir = path.join(parentSessionDir, SUBAGENT_TRANSCRIPTS_DIR_NAME);
+    await fs.mkdir(transcriptsDir, { recursive: true });
+
+    const originalReaddir = fs.readdir;
+    const readdirSpy = spyOn(fs, "readdir").mockImplementation(((...args: unknown[]) => {
+      const [targetPath] = args;
+      if (String(targetPath) === transcriptsDir) {
+        const permissionError = Object.assign(new Error("permission denied"), {
+          code: "EACCES",
+        });
+        return Promise.reject(permissionError);
+      }
+
+      return (originalReaddir as (...readdirArgs: unknown[]) => Promise<unknown>)(...args);
+    }) as unknown as typeof fs.readdir);
+    const warnSpy = spyOn(log, "warn").mockImplementation(() => undefined);
+
+    try {
+      const archivedWorkspaceIds = await listArchivedSubagentWorkspaceIds(sessionsDir, [
+        parentWorkspaceId,
+      ]);
+      expect(archivedWorkspaceIds).toEqual([]);
+      expect(warnSpy).toHaveBeenCalledTimes(1);
+    } finally {
+      warnSpy.mockRestore();
+      readdirSpy.mockRestore();
+    }
   });
 
   test("ignores archived sub-agent directories without chat.jsonl", async () => {
--- a/src/node/services/analytics/workspaceDiscovery.ts
+++ b/src/node/services/analytics/workspaceDiscovery.ts
@@ -2,6 +2,8 @@ import assert from "node:assert/strict";
 import type { Dirent } from "node:fs";
 import * as fs from "node:fs/promises";
 import * as path from "node:path";
+import { getErrorMessage } from "@/common/utils/errors";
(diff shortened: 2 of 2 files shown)


ThomasK33:
@codex review

Please review this analytics backfill fix.

ThomasK33:
@codex review

Addressed the archived transcript permission-handling feedback and added regression coverage. Please take another look.

PR #2608, opened 2026-02-24
🤖 fix: guard formatDuration against NaN/non-finite values

coadler:
## Summary

Fix `formatDuration` displaying "NaNh" / "NaNm NaNs" when `wall_duration_ms` is missing or non-finite in persisted tool results.

## Background

Tool results loaded from `chat.jsonl` history aren't Zod-validated — they're raw JSON. If `wall_duration_ms` is missing or corrupted (e.g., from a streaming crash, older format, or incomplete write), the component receives `undefined`, which becomes `NaN` in arithmetic. All comparisons (`NaN < 1000`, etc.) evaluate to `false`, so execution falls through to the final `return`, producing strings like `"NaNh"`.

## Implementation

Added `if (!Number.isFinite(ms)) return "—";` as the first guard in all 4 `formatDuration` functions:

- `src/browser/components/tools/shared/toolUtils.tsx` — shared util (BashToolCall, BashBackgroundList, BackgroundProcessesBanner)
- `src/browser/components/Messages/InitMessage.tsx` — init message duration
- `src/browser/components/RightSidebar/StatsTab.tsx` — stats tab durations
- `src/cli/toolFormatters.ts` — CLI tool output

`Number.isFinite` rejects `NaN`, `Infinity`, `-Infinity`, and non-number types, returning an em dash ("—") as a clean fallback.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `high` • Cost: `$1.92`_

### DIFF
--- a/src/browser/components/tools/shared/toolUtils.tsx
+++ b/src/browser/components/tools/shared/toolUtils.tsx
@@ -95,6 +95,7 @@ export function formatValue(value: unknown): string {
  * Format duration in human-readable form (ms, s, m, h)
  */
 export function formatDuration(ms: number): string {
+  if (!Number.isFinite(ms)) return "—";
   if (ms < 1000) return `${Math.round(ms)}ms`;
   if (ms < 60000) return `${Math.round(ms / 1000)}s`;
   if (ms < 3600000) return `${Math.round(ms / 60000)}m`;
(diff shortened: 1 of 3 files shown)


coadler:
@codex review

PR #2610, opened 2026-02-25
🤖 fix: support HTTP-transport MCP servers in OAuth probe (Figma)

jaaydenh:
## Summary

Improve the OAuth login UX for Figma’s **remote** MCP server by replacing the confusing 403 “Invalid OAuth error response” message with actionable guidance.

## Background

Figma’s remote MCP requires an approved client. When an unapproved client attempts OAuth registration, the auth server returns a plain‑text **403 Forbidden**, which the SDK surfaces as:

> Invalid OAuth error response … Raw body: Forbidden

This is technically correct but unhelpful. Users need clear direction about the approval requirement and the desktop MCP fallback.

## Implementation

- Added a small formatter in `MCPSettingsSection` that detects the Figma remote MCP URL + the 403 Forbidden error signature and replaces it with a user‑friendly message.
- Plumbed the server URL into the OAuth login hook so the formatter can target Figma specifically without affecting other providers.

## Validation

- `make static-check`

## Risks

- Low. Only affects the rendered error message for a specific 403/Forbidden case.

---

_Generated with `mux` • Model: `openai:gpt-5.2-codex` • Thinking: `xhigh` • Cost: `$13.64`_

### DIFF
--- a/src/browser/components/Settings/sections/MCPSettingsSection.tsx
+++ b/src/browser/components/Settings/sections/MCPSettingsSection.tsx
@@ -612,6 +612,7 @@ const RemoteMCPOAuthSection: React.FC<{
     api,
     isDesktop,
     serverName,
+    serverUrl: url,
     onSuccess: refreshAuthStatus,
   });
 


jaaydenh:
@codex review

jaaydenh:
@codex review

Addressed both comments:
1. P1: `extractBearerOauthChallenge` now returns `null` (not `{}`) when the header exists but isn't Bearer — non-Bearer auth (Basic, etc.) won't misclassify as OAuth.
2. P2: `probeWellKnownResourceMetadata` now tries both the origin-root well-known URL and a path-relative variant for servers mounted under sub-paths.

jaaydenh:
@codex review

Fixed P2: sub-path well-known URL now built from `base.pathname` + `base.origin` instead of `base.href`, so query strings don't corrupt the resolved path.

jaaydenh:
@codex review

Paired back the PR to a minimal UX-only change: a targeted, user-friendly error message for Figma’s remote MCP 403 “Forbidden” registration failure.

jaaydenh:
@codex review

Follow-up: ensure the Figma 403 guidance is also shown in the remote OAuth status panel (Login error path).

PR #2620, opened 2026-02-25
🤖 fix: background foreground waits on queued messages

ethanndickson:
## Summary

This PR fixes queue-triggered foreground `task_await` behavior so queued user messages can continue immediately, and parent stream-end nudges only target descendants that are still truly blocking.

It also closes a follow-up gap by making queue-background exemptions one-shot: the first stream-end after queue-background may suppress nudges, and later stream-ends nudge again if tasks remain active.

## Background

The original queue-background fix correctly rejected foreground waits, but suppression state could remain sticky across later voluntary stream-ends. That caused missing nudges even when no new queued message explained the suppression.

## Implementation

- Removed persisted task-await policy plumbing and moved queue-background tracking to runtime `TaskService` state.
- Updated `task_await` wait registration so workspace-scoped waits opt into queue-backgrounding and map `ForegroundWaitBackgroundedError` to active task statuses (not terminal errors).
- Made queue dispatch mode decisions use the queue’s effective mode (including sticky `tool-end`) instead of only the incoming message option.
- Prevented no-op queue insertions from backgrounding waits by returning `null` when nothing is enqueued.
- Added explicit stale-state clearing when a task is foreground-waited again.
- Updated `handleStreamEnd` to consume queue-background exemptions after each stream-end decision (one-shot semantics).
- Expanded regressions to cover:
  - one-shot cross-stream behavior (first suppressed, second nudged), and
  - multi-task-at-once waits from a single parent workspace.

## Validation

- `bun run typecheck`
- `bun test src/node/services/taskService.test.ts -t "queue-backgrounded|cross-stream|foreground|renewed|multiple|one-shot"`
- `bun test src/node/services/tools/task_await.test.ts -t "returns completed results for all awaited tasks|ForegroundWaitBackgroundedError|task_await"`

## Risks

- Queue-background exemption tracking is runtime-only. After process restart, active tasks are conservatively treated as blocking until fresh runtime state is established.
- One-shot consumption now depends on `handleStreamEnd` as the observe-and-consume boundary; regressions would most likely appear in parent nudge timing.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$37.57`_

### DIFF
--- a/src/node/services/agentSession.ts
+++ b/src/node/services/agentSession.ts
@@ -3838,14 +3838,21 @@ export class AgentSession {
     message: string,
     options?: SendMessageOptions & { fileParts?: FilePart[] },
     internal?: { synthetic?: boolean }
-  ): void {
+  ): "tool-end" | "turn-end" | null {
     this.assertNotDisposed("queueMessage");
-    this.messageQueue.add(message, options, internal);
+    const didEnqueue = this.messageQueue.add(message, options, internal);
+    if (!didEnqueue) {
+      return null;
+    }
     this.emitQueuedMessageChanged();
     // Signal to bash_output that it should return early to process queued messages
     // only for tool-end dispatches.
-    const shouldEarlyReturn = this.messageQueue.getQueueDispatchMode() === "tool-end";
-    this.backgroundProcessManager.setMessageQueued(this.workspaceId, shouldEarlyReturn);
+    const effectiveDispatchMode = this.messageQueue.getQueueDispatchMode();
+    this.backgroundProcessManager.setMessageQueued(
+      this.workspaceId,
+      effectiveDispatchMode === "tool-end"
+    );
+    return effectiveDispatchMode;
   }
 
   clearQueue(): void {
--- a/src/node/services/messageQueue.ts
+++ b/src/node/services/messageQueue.ts
@@ -100,8 +100,8 @@ export class MessageQueue {
     message: string,
     options?: SendMessageOptions & { fileParts?: FilePart[] },
     internal?: QueuedMessageInternalOptions
-  ): void {
-    this.addInternal(message, options, internal);
+  ): boolean {
+    return this.addInternal(message, options, internal);
   }
 
   /**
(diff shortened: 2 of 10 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed feedback: moved `backgroundForegroundWaitsForWorkspace` to after `session.queueMessage()` succeeds, so foreground waits aren't incorrectly canceled if enqueue throws.

ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed both review threads:

- **P1 (task_await backgrounding):** `task_await` now passes `backgroundOnMessageQueued: true` in both wait paths. Additionally, `waitForAgentReport()` defaults to queue-backgroundable when `requestingWorkspaceId` is present, preventing future tool-path drift.
- **P2 (effective dispatch mode):** `AgentSession.queueMessage()` now returns the queue's effective dispatch mode, and `WorkspaceService.sendMessage()` uses that returned value instead of incoming options. This fixes the mixed-queue case where sticky `tool-end` mode was ignored.

ethanndickson:
@codex review

Addressed both review threads from the second round:

- **P2 (task_await backgrounded handling):** `task_await` now catches `ForegroundWaitBackgroundedError` and returns `{ status: "running"/"queued"/"awaiting_report" }` with a descriptive note, matching the `task` tool's behavior.
- **P3 (gate on enqueue):** `MessageQueue.add()` now returns `boolean`, `AgentSession.queueMessage()` returns `null` when nothing was enqueued, and `workspaceService.ts` only backgrounds waits when `queueMessage` returned `"tool-end"`.

ethanndickson:
@codex review

Addressed P1 stale-flag issue: `userBackgroundedTaskIds` entries are now cleared when a task is foreground-awaited again, so parent auto-resume nudges fire correctly after renewed waits. Added lifecycle transition helpers and regression test covering the queue-background → renewed-foreground-wait → stream-end sequence.

ethanndickson:
@codex review

ethanndickson:
@codex review

Updated with one-shot queue-background exemption consumption at stream-end plus regressions covering multi-task `task_await` semantics.

PR #2621, opened 2026-02-25
🤖 tests: ungate gpt-5.3-codex OAuth requirement and protect pricing resets

ThomasK33:
## Summary
Ungates `openai:gpt-5.3-codex` from the OpenAI OAuth-required set now that it is API-available, while keeping OAuth routing support for subscription users. Adds focused regression coverage to ensure cost zeroing is only applied when usage is subscription-covered.

## Background
`gpt-5.3-codex` was previously treated as OAuth-required during early access. With public API pricing now published, API-key users should be able to select and route this model normally. At the same time, we need strict safeguards to ensure we never zero out API-key costs by mistake.

## Implementation
- Removed `gpt-5.3-codex` from `CODEX_OAUTH_REQUIRED_MODELS` (kept in `CODEX_OAUTH_ALLOWED_MODELS`).
- Updated model-selector/settings gating tests for the new required-set behavior.
- Updated the chat warning banner to reference shared required-model logic instead of hardcoded model IDs.
- Added explicit tests for `gpt-5.3-codex` pricing behavior:
  - costs zero when `providerMetadata.mux.costsIncluded === true` (subscription path)
  - costs remain non-zero when routed via API key path
- Added ProviderModelFactory tests proving `modelCostsIncluded` is true for OAuth-routed `gpt-5.3-codex` and false for API-key-routed `gpt-5.3-codex`.

## Validation
- `make static-check`
- `bun test src/browser/components/ChatInput/CodexOauthWarningBanner.test.tsx src/browser/hooks/useModelsFromSettings.test.ts src/browser/components/Settings/sections/ModelsSection.test.ts`
- `bun test src/common/utils/tokens/displayUsage.test.ts src/node/services/providerModelFactory.test.ts`

## Risks
Low. The behavioral change is scoped to OAuth required-model gating and is backed by tests at both routing and display-cost layers.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `1453617{COSTS}`_

### DIFF
--- a/src/node/services/aiService.test.ts
+++ b/src/node/services/aiService.test.ts
@@ -309,7 +309,7 @@ describe("AIService.createModel (Codex OAuth routing)", () => {
     delete process.env.OPENAI_API_KEY;
     try {
       const service = createService(muxHome.path);
-      const result = await service.createModel(KNOWN_MODELS.GPT_53_CODEX.id);
+      const result = await service.createModel(KNOWN_MODELS.GPT_53_CODEX_SPARK.id);
 
       expect(result.success).toBe(false);
       if (!result.success) {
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed the failing unit test by retargeting required-model assertions to `openai:gpt-5.3-codex-spark` after `gpt-5.3-codex` became non-required.

PR #2630, opened 2026-02-25
🤖 fix: fix review line comment tooltip clipping

ammar-agent:
## Summary
Replace the Review pane's native-title-based comment icon tooltip with the app's shared Radix tooltip so the tooltip text is no longer clipped at the left edge when Shift-click-to-comment UI hints are shown.

## Background
Shift/drag range-selection guidance was shown via a `title` attribute on the review icon. In dense Review pane layouts, that tooltip was being cut off, reducing discoverability of the Shift-click gesture.

## Implementation
- Updated `src/browser/components/shared/DiffRenderer.tsx` to wrap the inline review action button with shared `<Tooltip>`, `<TooltipTrigger>`, and `<TooltipContent>` components.
- Preserved the existing click behavior and icon rendering while only changing tooltip rendering mechanism.
- Removed direct `title` usage for this control to avoid custom `[title]` pseudo-element clipping behavior.

## Validation
- `bun test src/browser/components/shared/SelectableDiffRenderer.dragSelect.test.tsx`
- `make static-check`

## Risks
- Very low. Behavior change is limited to tooltip rendering for the line comment icon; no diff parsing or review data-path logic was changed.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex-spark` • Thinking: `high` • Cost: `$0.00`_

### DIFF
--- a/src/browser/components/shared/DiffRenderer.tsx
+++ b/src/browser/components/shared/DiffRenderer.tsx
@@ -11,6 +11,7 @@ import { getLanguageFromPath } from "@/common/utils/git/languageDetector";
 import { useOverflowDetection } from "@/browser/hooks/useOverflowDetection";
 import { MessageSquare } from "lucide-react";
 import { InlineReviewNote, type ReviewActionCallbacks } from "./InlineReviewNote";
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

Please take another look.

PR #2636, opened 2026-02-25
🤖 fix: improve math usage via system prompt

ammar-agent:
I noticed Codex often formatted `[ <math> ]` type TeX blocks which we do not render correctly.

This PR adds explicit system-prompt guidance for math formatting so models prefer `$$...$$` delimiters, which match our markdown math render path more reliably.

### DIFF
--- a/src/node/services/systemMessage.test.ts
+++ b/src/node/services/systemMessage.test.ts
@@ -13,6 +13,9 @@ const extractTagContent = (message: string, tagName: string): string | null => {
 import { describe, test, expect, beforeEach, afterEach, spyOn, type Mock } from "bun:test";
 import { LocalRuntime } from "@/node/runtime/LocalRuntime";
 
+// Note: in this file we avoid tests that are merely tautological assertions of constants. Only
+// tests that verify branching logic should be here.
+
 describe("extractToolInstructions", () => {
   // Use a model that has bash tool available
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the markdown code-context issue by excluding indented code blocks and container/list-prefixed fenced blocks from TeX delimiter rewrites, with regression tests added.

ammar-agent:
@codex review

Follow-up update: removed flaky render-level test, kept normalization + code-context regression coverage, and validated with static-check.

ammar-agent:
@codex review

Addressed the fence-closure issue by using opener-context closer prefixes (instead of stripping every line), and added a regression case for fence-like literals inside fenced content.

ammar-agent:
@codex review

Applied another follow-up to fence close detection: closers now use a single opener-derived prefix context (not dynamic prefix stripping per line).

ammar-agent:
@codex review

Handled the blockquote closer spacing case (`>``` vs `> ````) and added a regression test.

ammar-agent:
@codex review

Scope reset per direction: this PR now only updates system-prompt math guidance (`$$...$$`) plus its test/docs sync. Renderer-side delimiter rewrite changes were dropped.

PR #2035, opened 2026-01-29
🤖 tests: README screenshot stories with zoom crops + orchestrator

ibetitsmike:
## Summary

Create and maintain a deterministic Storybook-driven pipeline for README screenshots, including refreshed visuals for Auto mode, git divergence details, context management, and a mobile server-mode screenshot.

### Capture pipeline

- **`scripts/capture-readme-screenshots.ts`** supports:
  - focused crops via `clip`
  - per-story viewport overrides via `viewport`
  - per-story browser-context overrides (`contextOptions`) for touch/mobile user-agent emulation
- Capture runs now create a fresh browser context for every story attempt so localStorage state cannot leak between screenshots (deterministic full-batch vs `--story` output).
- Captures run at 1900×1188 CSS with DPR 2 (3800px output), then:
  - full-frame images resize to target width
  - clipped images preserve native high-DPR crop detail
  - mobile screenshot preserves native iPhone resolution via `postProcess` (no crop/upscale)
- Interaction steps are handled in Playwright for click-driven states (Git divergence dialog, Auto creation card, context management dialog).

### README screenshot stories (`Docs/README Screenshots`)

Added/maintained 9 screenshot-mapped stories in `App.readmeScreenshots.stories.tsx`:

| Story                     | README asset              | Notes                                               |
| ------------------------- | ------------------------- | --------------------------------------------------- |
| `CodeReview`              | `code-review.webp`        | 50/50 chat + review composition                     |
| `AgentStatusSidebar`      | `agent-status.webp`       | dense multi-workspace status list                   |
| `GitStatusPopover`        | `git-status.webp`         | divergence dialog with commit view                  |
| `PlanMermaidWithCosts`    | `plan-mermaid.webp`       | plan tool output + costs panel                      |
| `AutoModeAgentSwitching`  | `auto-mode.webp`          | centered creation defaults (Auto + Opus 4.6 + MAX)  |
| `CostsTabRich`            | `costs-tab.webp`          | multi-model usage/cost breakdown                    |
| `ContextManagementDialog` | `context-management.webp` | compaction/context management dialog                |
| `MobileServerMode`        | `mobile-server-mode.webp` | iPhone 17 Pro Max full-screen workspace composition |
| `OrchestrateAgents`       | `orchestrate-agents.webp` | plan-mode parent + six child tasks                  |

- `AutoModeAgentSwitching` now uses project-scoped defaults for:
  - model: `anthropic:claude-opus-4-6`
[...]

PR #2641, opened 2026-02-26
🤖 tests: stabilize web-app ordering in MacOS titlebar story

ammar-agent:
## Summary
Pin `createdAt` in `App/TitleBar` story fixtures so the `web-app` workspace list renders in a deterministic order in `MacOSDesktop`.

## Background
Workspace sorting uses recency first, then `createdAt`, then name/id tie-breakers. In this story, both `web-app` workspaces start with equal recency, while `createWorkspace()` defaulted `createdAt` to `new Date().toISOString()`. Depending on whether both fixtures landed in the same millisecond, ordering could flip between runs.

The flake was introduced when this story was added in PR #2035.

## Implementation
- Added a shared `stableCreatedAt` timestamp inside `createPopulatedClient()`
- Passed that timestamp to all four story workspaces (including the two under `web-app`)
- Left an inline rationale comment documenting why this is required for deterministic Storybook/Chromatic output

## Validation
- `make static-check`

## Risks
Low risk; this only affects Storybook fixture data and does not change production runtime logic.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$unknown`_

### DIFF
--- a/src/browser/stories/App.titlebar.stories.tsx
+++ b/src/browser/stories/App.titlebar.stories.tsx
@@ -21,11 +21,34 @@ export default {
 // ─────────────────────────────────────────────────────────────────────────────
 
 function createPopulatedClient() {
+  // Keep createdAt deterministic so recency tie-breakers can't reorder sibling
+  // workspaces between Storybook/Chromatic runs.
+  const stableCreatedAt = "2023-11-14T22:13:20.000Z";
   const workspaces = [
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

ammario:
cc @ibetitsmike

PR #2646, opened 2026-02-26
🤖 fix: mobile plan action buttons right-aligned with white text

ibetitsmike:
## Summary

Move mobile plan action buttons (Implement, Start Orchestrator, Continue in Auto) to be right-aligned with white (`text-foreground`) text, matching the visual weight of desktop plan buttons.

## Background

On narrow screens, plan action buttons were rendered as icon-only buttons inside the left container alongside copy/edit actions. This caused two problems:
- They used `text-placeholder` color (dim gray), making them hard to see
- They sat on the left side, visually buried among utility actions

## Implementation

Moved the `isNarrowScreen` plan button block out of the left `<div>` container and into its own right-aligned block at the same level as the desktop buttons. Applied `[&_button]:text-foreground` to override the `IconActionButton` default dim color to white.

The edit button remains in the left container. Desktop full-labeled buttons are unchanged.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$0.97`_

### DIFF
--- a/src/browser/components/tools/ProposePlanToolCall.tsx
+++ b/src/browser/components/tools/ProposePlanToolCall.tsx
@@ -772,14 +772,6 @@ export const ProposePlanToolCall: React.FC<ProposePlanToolCallProps> = (props) =
             <IconActionButton key={index} button={button} />
           ))}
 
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

PR #2648, opened 2026-02-26
🤖 refactor: replace structured output with propose_name tool for workspace name generation

ibetitsmike:
## Summary

Replace `toolChoice: "required"` with a proper builtin agent (`name_workspace`) using the `tools.require` mechanism, making workspace name generation compatible with extended thinking models.

## Background

The previous implementation used `toolChoice: "required"` in the `streamText` call to force the model to call the `propose_name` tool. This is incompatible with extended thinking models (e.g., Anthropic's extended thinking) because forced tool choice (`toolChoice: "required"` / `"any"` / `{ type: "tool" }`) conflicts with the extended thinking API — models cannot be forced to call tools immediately when they need to think first.

The standard mechanism in this codebase for "this tool must eventually be called" is `tools.require` in agent frontmatter, which the `StreamManager` enforces via `stopWhen` conditions rather than `toolChoice`. This is fully compatible with extended thinking because the model is free to think and then call the tool when ready.

## Implementation

- **New builtin agent**: `src/node/builtinAgents/name_workspace.md` with `tools.require: [propose_name]` — defines the contract that the `propose_name` tool must be called
- **New tool factory**: `src/node/services/tools/propose_name.ts` — standard `ToolFactory` that validates model args against `ProposeNameToolArgsSchema` and echoes them back
- **Tool registration**: Registered `propose_name` in `tools.ts` and `name_workspace` in `builtInAgentDefinitions.ts`
- **Removed `toolChoice`**: The `workspaceTitleGenerator.ts` `streamText` call no longer sets `toolChoice`; the prompt instructs the model to call the tool, and the candidate retry loop handles the rare case where the model doesn't comply
- **Preserved**: All existing changes from the original PR (deleted ~300 lines of fallback regex parsing, `ProposeNameToolArgsSchema` in `toolDefinitions.ts`, cleaned-up tests, error mapping)

## Risks

Low — the prompt strongly instructs the model to call `propose_name`, and it's the only tool available. Models that support tool calling will almost always comply. The retry loop (up to 3 candidates) provides a safety net.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$19.73`_


ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed the P2 feedback: changed `toolChoice` from `{ type: "tool", toolName: "propose_name" }` to `"required"` for broader provider compatibility. With only one tool available, `"required"` is functionally equivalent but doesn't require named tool choice support. Models that don't support tool choice at all will throw → caught by the retry loop → next candidate tried.

Also cleaned up dead code and schema duplication from the previous iteration (separate commit).

ibetitsmike:
@codex review

Restored `NoOutputGeneratedError` handling + test — this error can still occur with tool-based generation when a provider returns no output at all.

ibetitsmike:
@codex review

ibetitsmike:
Re: the Codex concern about fallback when models skip the tool call — this is intentional by design:

1. **The old fallback parsing was the problem we're solving.** The ~300 lines of deleted regex fallback (`extractIdentityFromText`, `recoverIdentityFromFallback`, etc.) existed because `Output.object()` structured JSON output was unreliable. Tool calls are universally supported and far more reliable — models don't need to format raw JSON, they just fill in typed parameters via native function-calling APIs.

2. **The retry loop provides resilience.** Up to 3 model candidates are tried. A model that doesn't support tool calling at all will throw an error (caught by the try/catch), not silently return text.

3. **The prompt strongly instructs tool use.** The tool description says "You MUST call this tool exactly once" and the prompt says "Do not emit text responses." Models that support tools almost always comply.

4. **Restoring a text fallback parser would re-introduce the exact fragility this PR eliminates.** The regex-based parsers were brittle (parsing markdown patterns like `**name:** \`feature\``) and a continuous source of edge-case bugs with different model outputs.

Resolving as intentional.

ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed feedback: now searching all tool results (not just `results[0]`) for a successful `propose_name` call, handling the case where the model emits multiple tool calls.

ibetitsmike:
@codex review

Addressed: `results.find()` now searches all tool results for a successful `propose_name` call. Added safety comment explaining that `toolResults` only contains entries that passed Zod validation AND whose `execute()` succeeded — schema-invalid tool calls never appear in the array.

ibetitsmike:
@codex review

This is the same concern as the P1 thread resolved earlier — removing the text fallback parser is the core purpose of this PR (eliminating ~300 lines of fragile regex parsing).

All model candidates in `workspaceService.generateTitle` support tool calling (they're from the same model list used for full agent sessions). Models that don't support tools throw an error caught by the try/catch, and the retry loop moves to the next candidate. The text fallback was never intentional architecture — it was a workaround for `Output.object()` unreliability that is no longer needed with native tool calls.

ibetitsmike:
@codex review

This is the same concern raised in 3 previous threads (all resolved). The removal of both `toolChoice` and text fallback parsing is the explicit design intent of this PR — the user requested the `toolChoice` removal for extended thinking compatibility, and the text fallback was removed because it was 300+ lines of fragile regex workarounds for `Output.object()` unreliability.

The `tools.require: [propose_name]` contract in `name_workspace.md` documents the requirement. The direct `streamText` path relies on the prompt + retry loop (3 candidates). This is the correct architecture as specified by the PR author.

ibetitsmike:
@codex review

Addressed: removed `propose_name` from the global `getToolsForModel()` registration. The tool is only created inline by `workspaceTitleGenerator.ts` for its direct `streamText` path — it should not appear in the default exec/plan/ask toolset. Added a comment explaining the intentional omission.

ibetitsmike:
Stopping the Codex review loop — the reviewer is cycling between contradictory positions:
- Round 4: "Keep propose_name OUT of the default exec toolset" ✅ (addressed)
- Round 7: "Expose propose_name WHEN name_workspace is active" (contradicts round 4)

The `name_workspace` agent is internal (`ui.hidden: true`, `subagent.runnable: false`) — it's never instantiated through `AgentSession` or `StreamManager`. It documents the tool requirement contract. The `workspaceTitleGenerator.ts` creates the tool inline for its direct `streamText` call.

All threads resolved. CI is green. Ready for human review.

PR #2404, opened 2026-02-13
🤖 tests: make MCP dialog scope self-healing to survive Storybook remounts

ethanndickson:
## Summary

Makes the Storybook MCP dialog test helpers self-healing so `ToolSelectorInteraction` survives mid-test component remounts in CI.

## Background

`App/Settings/MCP › ToolSelectorInteraction` flakes on `main` with "Workspace MCP dialog missing". The recent stabilization in #2402 hardened the **opening** step, but the dialog can still vanish *after* it's open — Storybook remounts the story component between `await` points, resetting `mcpModalOpen` useState back to `false`. The `waitFor` retry loops detect the missing dialog but can't re-open it, so they spin until timeout.

## Implementation

Redesigned `createWorkspaceMCPModalScope` with an `ensureOpen()` guard that re-triggers the full open flow if the dialog disappeared:

- **`find()`** now calls `ensureOpen()` before querying — all 7 existing stories using `modal.find()` get self-healing for free.
- **`assert(fn)`** = `ensureOpen()` + `waitFor(fn)` — replaces the 4× copy-pasted `waitFor` + `modal.query()` + throw pattern.
- **`query()`** removed — it was only used inside `waitFor` blocks that couldn't recover.

The `ToolSelectorInteraction` play function shrinks from 74 to ~35 lines with identical coverage.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$3.30`_

PR #2402, opened 2026-02-13
🤖 tests: stabilize workspace MCP Storybook modal flow

ethanndickson:
## Summary
Stabilizes the workspace MCP Storybook play tests by making the modal-open flow deterministic and scoping assertions to the actual MCP dialog instead of querying the whole document.

## Background
`App/Settings/MCP › WorkspaceMCPNoOverrides` was intermittently failing in CI with `Unable to find role="dialog"`. The prior helper used a brittle click-then-catch-then-reclick pattern around a popover portal transition.

## Implementation
- Updated `openWorkspaceMCPModal()` in `src/browser/stories/App.projectSettings.stories.tsx` to:
  - return the opened dialog element,
  - retry the full more-actions → MCP click flow via `waitFor`,
  - gate success on MCP-specific modal content (`Workspace MCP Configuration`) before continuing.
- Updated workspace MCP play stories to scope assertions/interactions with `within(dialog)`.
- Replaced global body queries in those stories with modal-scoped expectations.

## Validation
- ✅ `make typecheck`
- ✅ `bun x eslint src/browser/stories/App.projectSettings.stories.tsx`
- ✅ `bun x prettier --check src/browser/stories/App.projectSettings.stories.tsx`
- ⚠️ `NODE_OPTIONS='--localstorage-file=/tmp/storybook-localstorage' make test-storybook`
  - currently fails due unrelated pre-existing Storybook failures in `App.agentSkills.stories.tsx` (timeouts / skill indicator assertions), so full suite is not green in this environment.

## Risks
- Low risk: changes are isolated to Storybook story helpers/play assertions; no production runtime code paths were modified.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$3.63`_

PR #2650, opened 2026-02-26
🤖 tests: atomize ToolSelectorInteraction None→All play cycle

ethanndickson:
## Summary

Fix the persistently flaky `ToolSelectorInteraction` Storybook play-test by making the None→All assertion cycle atomic and hardening the mock to persist overrides.

## Background

This test has been the subject of six prior PRs (#2221, #2226, #2235, #2402, #2404, #2444), each addressing different symptoms (timeouts, stale DOM refs, missing dialogs) but none addressing the root cause: **state loss across `modal.assert()` boundaries during Storybook remounts**.

The failure chain:
1. Test clicks "None" → local React state updated (unsaved to backend)
2. Storybook remount occurs between separate `modal.assert()` calls
3. `ensureOpen()` re-opens the modal, triggering `useEffect` which reloads overrides from mock
4. Mock returns `{}` (the `set` handler was a no-op), so `allowedTools` falls back to all tools
5. `All` button is disabled (all tools selected) → `expect(All).toBeEnabled()` fails

## Implementation

Two targeted changes:

1. **Atomic assertion** (`App.projectSettings.stories.tsx`): Merge the three separate `modal.assert()` calls (None click, All-enabled guard, All click) into **one** callback. If a remount happens, `waitFor` retries the entire sequence from scratch. Idempotent guards (`if (!btn.hasAttribute("disabled"))`) prevent double-clicks on retry.

2. **Mock persistence** (`mocks/orpc.ts`): Change `workspace.mcp.set` from a no-op to actually persist overrides in the `mcpOverrides` Map. This is defense-in-depth: even if a remount triggers a re-fetch, the mock now returns realistic saved state.

## Validation

- `make static-check` passes locally

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$2.67`_

### DIFF
--- a/src/browser/stories/App.projectSettings.stories.tsx
+++ b/src/browser/stories/App.projectSettings.stories.tsx
@@ -798,31 +798,33 @@ export const ToolSelectorInteraction: AppStory = {
       await expect(allBtn).toBeDisabled();
     });
 
-    // Click "None" to deselect all tools.
+    // Full None→All cycle in one atomic assertion. Keeping the entire
+    // flow inside a single modal.assert() removes the cross-assert
+    // state dependency window that caused flakes: if a Storybook remount
+    // resets the modal between separate assert() calls, the unsaved local
+    // overrides are lost and "All" stays disabled. A single callback lets
+    // waitFor retry the whole sequence from scratch on remount, while the
+    // idempotent guards prevent double-clicks on retry iterations.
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

PR #2660, opened 2026-02-26
🤖 feat: parent-tab tracking for file preview close navigation

ibetitsmike:
## Summary

When a file preview tab is opened from another tab (e.g., Review or Explorer), closing the file now returns focus to the tab that opened it instead of falling back to whichever tab happens to be positionally adjacent.

## Background

File preview tabs are opened from the Review or Explorer panels. When the user closes a file preview, the previous behavior picked the next (or previous) tab in the array — which is disorienting because the user expects to return to where they came from. This change adds a "parent tab" concept so closing a file navigates back naturally.

## Implementation

Adds an optional `parentTab` map (`Record<string, string>`) to `RightSidebarLayoutState` as a side-channel alongside the existing tree structure. This avoids changing the recursive `RightSidebarLayoutNode` tree.

**`rightSidebarLayout.ts`:**
- Extended `RightSidebarLayoutState` interface with optional `parentTab` field
- `removeTabFromNode()` now accepts an optional `preferredActiveTab` — when the active tab is removed and a parent exists in the same tabset, it's activated instead of positional fallback
- `removeTabEverywhere()` looks up the parent before removal, passes it through, and cleans up stale entries (both for the removed tab and any orphaned children pointing to it)
- `isRightSidebarLayoutState()` validator accepts the optional `parentTab` field for backward compat

**`RightSidebar.tsx`:**
- `handleOpenFile` now records the currently active tab as the parent of the new file tab

**Tests (7 new):**
- Parent tab activation, positional fallback when parent is in different tabset, entry cleanup, orphan cleanup, backward compat validation (with/without/invalid parentTab)

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$3.65`_

### DIFF
--- a/src/browser/features/RightSidebar/RightSidebar.tsx
+++ b/src/browser/features/RightSidebar/RightSidebar.tsx
@@ -1166,11 +1166,22 @@ const RightSidebarComponent: React.FC<RightSidebarProps> = ({
             const next = selectTabInTabset(withFocus, fileTabsetId, fileTabType);
             // Only record the parent when the origin tab lives in the same
             // tabset as the file — cross-tabset parents would never activate.
+            // When reopening from a different tabset, clear any stale parent
+            // entry so close falls back to positional adjacency.
             const sameTabset = prev.focusedTabsetId === fileTabsetId;
             if (parentTabId && parentTabId !== fileTabType && sameTabset) {
               return {
                 ...next,
                 parentTab: { ...next.parentTab, [fileTabType]: parentTabId },
+              };
+            }
+            // Cross-tabset reopen: clear stale parent to avoid jumping to
+            // an outdated origin tab on close.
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed the cross-tabset parentTab overwrite issue — now only refreshes parentTab when the origin tab is in the same tabset as the file tab. Cross-tabset origins are left unchanged to preserve the existing usable parent.

ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed the cross-tabset stale parent issue — now clears the parentTab entry when a file is reopened from a different tabset, so close falls back to positional adjacency instead of jumping to an outdated origin.

PR #2663, opened 2026-02-26
🤖 fix: cap timing histogram range at p99

ThomasK33:
## Summary
Cap timing distribution histogram bucket range at p99 instead of max value.

## Background
A single extreme duration outlier can stretch histogram bucket widths enough to flatten the chart for the rest of responses, making p50/p90/p99 context harder to read.

## Implementation
- Updated `queryTimingDistribution` in `src/node/services/analytics/queries.ts`
- Replaced histogram stats upper bound from `MAX(${column})` to `PERCENTILE_CONT(0.99) ... AS max_value`
- Added inline rationale comments documenting that values above p99 are intentionally folded into the last bucket by existing `LEAST(20, ...)` bucket clamping

## Risks
- Tail values above p99 are no longer distributed across the full x-axis range and are aggregated into the final bucket.
- This is intentional for chart readability, but it reduces visual resolution in the extreme tail.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$0.78`_

### DIFF
--- a/src/node/services/analytics/queries.ts
+++ b/src/node/services/analytics/queries.ts
@@ -361,20 +361,32 @@ async function queryTimingDistribution(
   // meaningful units and percentile reference lines land correctly.
   //
   // Cap the histogram range at p99 so a single extreme outlier does not flatten
-  // the distribution for the other 99% of responses. Values above p99 are
-  // intentionally folded into the final bucket via LEAST(20, ...).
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

Addressed the p99=min edge case in timing histogram bucketing by falling back to raw max only for that degenerate distribution shape. Also reran `make static-check` locally.

PR #2668, opened 2026-02-26
🤖 fix: auto-prune stale workspace entries on project removal

ibetitsmike:
## Summary

When a user manually deletes workspace directories from `~/.mux/src/`, the workspace entries remain in `config.json`. Attempting to remove the project then fails with a "workspace_blockers" error because `projectService.remove()` checks workspace counts from config—not the filesystem. The user is stuck: the directory is gone but the app won't let them clean up the project.

This adds self-healing to `remove()`: before enforcing the workspace blocker, each **local** workspace entry is checked against the filesystem. Entries whose backing directories no longer exist are pruned from config automatically, then counts are re-evaluated. Remote runtimes (SSH, Docker, devcontainer) are preserved unconditionally since their paths exist on the remote host, not locally.

## Background

This violates the self-healing principle in AGENTS.md: corrupted/stale persisted state should be sanitized at request time rather than blocking the user permanently.

## Implementation

**`src/node/services/projectService.ts`** — In `remove()`, iterate workspace entries. For local/worktree runtimes (or unset runtimeConfig), call `fsPromises.access(ws.path)` to check directory existence. Missing directories trigger `log.info` and entry removal. Remote runtimes (SSH, Docker, devcontainer) are skipped—their paths exist on the remote host. If any entries were pruned, config is saved before re-checking counts. No new imports needed (`fsPromises` and `log` already imported).

**`src/node/services/projectService.test.ts`** — 6 new tests under `describe("remove")`:
- Removes project with no workspaces
- Returns `project_not_found` for unknown project
- Blocks removal when workspaces still exist on disk
- Auto-prunes stale workspace entries and removes project
- Preserves remote runtime workspace entries even if path is not local
- Prunes stale entries but blocks on remaining real workspaces

## Validation

All 6 new tests pass. Full `make static-check` passes (typecheck + lint + fmt).

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$3.75`_

### DIFF
--- a/src/node/services/projectService.ts
+++ b/src/node/services/projectService.ts
@@ -811,8 +811,18 @@ export class ProjectService {
         try {
           await fsPromises.access(ws.path);
           survivingWorkspaces.push(ws);
-        } catch {
-          log.info(`Pruning stale workspace entry (directory missing): ${ws.path}`);
+        } catch (err: unknown) {
+          // Only prune when the directory is truly gone (ENOENT/ENOTDIR).
+          // Other errors (EACCES, transient I/O) mean the path may still exist.
+          const code = (err as NodeJS.ErrnoException).code;
+          if (code === "ENOENT" || code === "ENOTDIR") {
+            log.info(`Pruning stale workspace entry (directory missing): ${ws.path}`);
+          } else {
+            log.warn(
+              `Keeping workspace entry despite access error (${code ?? "unknown"}): ${ws.path}`
+            );
+            survivingWorkspaces.push(ws);
+          }
         }
       }
       if (survivingWorkspaces.length !== projectConfig.workspaces.length) {


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed the P1 feedback: remote runtime workspaces (SSH, Docker, devcontainer) are now preserved unconditionally — only local/worktree paths are checked against the filesystem. Added a new test covering the SSH workspace case.

ibetitsmike:
@codex review

Addressed the P2: now only prune on ENOENT/ENOTDIR. Other errors (EACCES, transient I/O) preserve the workspace entry and log a warning.

PR #2690, opened 2026-02-27
🤖 ci: reduce nightly bench to twice-weekly, add Daytona sandbox cleanup

ibetitsmike:
## Summary

Reduce nightly Terminal Bench from twice daily to twice weekly and add automatic Daytona sandbox cleanup after each benchmark run.

## Background

The nightly workflow ran at 10:00 UTC and 22:00 UTC every day, which is excessive for cost and Daytona sandbox quota. Daytona sandboxes leaked when Harbor crashed, timed out, or the workflow was cancelled — `auto_delete` is disabled by default, so stopped sandboxes persist indefinitely and count toward the 25-sandbox limit.

## Implementation

- **Schedule**: Changed cron from `0 10 * * *` / `0 22 * * *` to Monday 10:00 UTC + Thursday 22:00 UTC. Keeps peak/trough model load spread across different days.
- **Cleanup step**: Added to `terminal-bench.yml` with `if: always() && inputs.env == 'daytona'`. Uses the Daytona Python SDK to list and delete all sandboxes. Individual delete failures are logged but don't fail the step.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$1.33`_

### DIFF
--- a/.github/workflows/terminal-bench.yml
+++ b/.github/workflows/terminal-bench.yml
@@ -147,10 +147,12 @@ jobs:
       - name: Build dist/ (skip icons - not needed for benchmark)
         run: make build-main build-preload
 
-      # Snapshot existing Daytona sandboxes so post-run cleanup only deletes
-      # sandboxes created by *this* job, not ones from parallel jobs.
+      # Best-effort snapshot of existing Daytona sandboxes so post-run cleanup
+      # only deletes sandboxes created by *this* job, not ones from parallel jobs.
+      # continue-on-error: transient API/pip failures must not block the benchmark.
       - name: Snapshot Daytona sandboxes
         if: inputs.env == 'daytona'
+        continue-on-error: true
         env:
           DAYTONA_API_KEY: ${{ secrets.DAYTONA_API_KEY }}
         run: |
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed feedback: cleanup now snapshots pre-existing sandbox IDs before the run and only deletes sandboxes created during this specific job, preventing parallel jobs from interfering with each other.

ibetitsmike:
@codex review

Addressed both P1s:
1. Cleanup now requires the pre-run snapshot file to exist — if it's missing (snapshot step failed), cleanup is skipped entirely instead of falling back to global deletion.
2. Only stopped/errored sandboxes are deleted. Sandboxes in started/creating/starting state are skipped, protecting active sandboxes from parallel jobs.

ibetitsmike:
@codex review

Both Daytona steps now use `continue-on-error: true` — transient API/pip failures in snapshot or cleanup never block the benchmark or mark the job as failed.

ibetitsmike:
This is actually incorrect — `if: always()` in GitHub Actions **does** execute on cancellation (that's its primary purpose). It only skips on hard job timeout (the 240-minute kill), which is an extreme edge case and can't be fixed within a single job regardless. The cleanup handles the most common leak scenarios: Harbor crash, benchmark failure, and workflow cancellation.

ibetitsmike:
@codex review

Resolved the `always()` comment — `if: always()` does execute on cancellation in GitHub Actions (that's its purpose). It only skips on hard job timeout (4h kill), which is an extreme edge case that can't be fixed within a single job.

ibetitsmike:
@codex review

Resolved both indentation comments — YAML block scalars (`|`) strip common leading whitespace before passing to the shell. The Python code inside `python3 -c "..."` arrives at column 0 after YAML processing, so there's no `IndentationError`. This is the standard pattern for inline Python in GitHub Actions workflows.

PR #2709, opened 2026-03-02
🤖 fix: refresh agent selector after propose_plan so orchestrator appears immediately

ethanndickson:
## Summary

Fixes a bug where the Orchestrator agent doesn't appear in the Agent Selector immediately after a plan is proposed — the user had to navigate away and back.

## Background

The orchestrator agent declares `ui.requires: ["plan"]` in its frontmatter, so `agents.list` only marks it `uiSelectable: true` when a plan file exists on disk. `AgentContext` fetches agents once on mount and listens for `AGENTS_REFRESH_REQUESTED` to re-fetch, but nothing dispatched that event after `propose_plan` completed. The agent list stayed stale until a remount.

## Implementation

Added a guarded `useEffect` in `ProposePlanToolCall` that dispatches `AGENTS_REFRESH_REQUESTED` once when the latest `propose_plan` tool completes successfully. The ref guard prevents duplicate dispatches from rerenders.

Two new regression tests verify:
- The event fires exactly once for a latest, completed, successful plan
- The event does **not** fire for failed results or non-latest tool calls

## Validation

- All 9 `ProposePlanToolCall` tests pass (including 2 new ones)
- `make static-check` passes cleanly

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$13.01`_

### DIFF
--- a/src/browser/utils/messages/applyWorkspaceChatEventToAggregator.test.ts
+++ b/src/browser/utils/messages/applyWorkspaceChatEventToAggregator.test.ts
@@ -88,7 +88,22 @@ class StubAggregator implements WorkspaceChatEventAggregator {
 describe("applyWorkspaceChatEventToAggregator", () => {
   function withDispatchSpy<T>(run: (dispatched: Event[]) => T): T {
     const originalWindow = globalThis.window;
+    const originalCustomEvent = globalThis.CustomEvent;
     const dispatched: Event[] = [];
+
+    // CI bun environment may lack CustomEvent (it was previously provided by happy-dom).
+    // createCustomEvent() in src/common/constants/events.ts uses `new CustomEvent(...)`.
+    if (typeof globalThis.CustomEvent === "undefined") {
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

Pushed a redesign of the refresh wiring so AGENTS_REFRESH_REQUESTED is emitted from centralized stream tool-call-end handling on successful `propose_plan`, with replay/failed/non-plan guards and focused aggregator tests.

ethanndickson:
@codex review

Pushed test stabilization: replaced fragile happy-dom GlobalWindow swapping with a deterministic `withDispatchSpy` helper that intercepts `window.dispatchEvent` directly. All existing tests unaffected, static-check clean.

ethanndickson:
@codex review

Fixed CI failure: `CustomEvent` was undefined in bun test environment after removing happy-dom. Added a minimal polyfill in `withDispatchSpy` that provides `CustomEvent` when missing and restores the original in the finally block.

PR #2730, opened 2026-03-02
🤖 fix: resolve mappedToModel aliases in runtime capability detection

ibetitsmike:
## Summary

Custom model aliases (e.g., `anthropic:claude/sonnet` used with LiteLLM or AWS Bedrock gateways) caused 400 Bad Request errors from the Anthropic API. The `mappedToModel` config already existed for UI/metadata resolution but wasn't threaded through the runtime capability detection paths.

## Background

When `buildProviderOptions` encounters an unrecognized alias, it still enables thinking with a 10k token budget. But the SDK, also failing to recognize the alias, defaults `maxOutputTokens` to 4096. Anthropic rejects requests where `budget_tokens > max_tokens`, causing the 400 error. The same gap prevented correct beta header injection (e.g., 1M context support) for aliased models.

## Implementation

- Thread optional `ProvidersConfigMap` through `buildProviderOptions` (8th param) and `buildRequestHeaders` (4th param)
- Both functions resolve the model via `resolveModelForMetadata` to a "capability model" for internal logic, while preserving the original ID for the actual provider request
- Add `getModelStatsResolved` fallback in `streamManager.buildStreamRequestConfig` so aliased models get correct `max_output_tokens` from their mapped base model
- Update `aiService.ts` call sites to pass `this.providerService.getConfig()`

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$14.50`_

### DIFF
--- a/src/common/utils/ai/providerOptions.ts
+++ b/src/common/utils/ai/providerOptions.ts
@@ -437,12 +437,12 @@ export function buildRequestHeaders(
   }
 
   const normalized = normalizeGatewayModel(modelString);
+  // Route provider-specific headers by the runtime provider from the original model
+  // string. Capability resolution is only for model-level feature checks below.
   const [provider] = normalized.split(":", 2);
   const capabilityModel = resolveModelForMetadata(normalized, providersConfig ?? null);
-  const [resolvedCapabilityProvider] = capabilityModel.split(":", 2);
-  const capProvider = resolvedCapabilityProvider || provider;
 
-  if (capProvider === "anthropic") {
+  if (provider === "anthropic") {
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Fixed: provider dispatch now uses the original runtime `provider` for top-level branching, with `capModelName` only used for model-specific capability checks within each provider block.

ibetitsmike:
@codex review

Also fixed `buildRequestHeaders` — now dispatches on the runtime `provider` for the top-level branch, with the resolved capability model only used for model-specific checks (1M context support).

PR #2733, opened 2026-03-02
🤖 fix: add X-Initiator header for Copilot premium request billing

ibetitsmike:
## Summary

Adds the `X-Initiator` header to GitHub Copilot API requests so that tool-call continuations are billed as free agent work instead of burning premium requests. Fixes: https://github.com/coder/mux/issues/2489

## Background

GitHub Copilot uses the `X-Initiator` header to decide billing: `user` consumes a premium request, `agent` is free. Without this header (Mux today), every API call defaults to `user` billing. In a typical agentic session, a single user prompt triggers 10–60+ API calls (tool calls, continuations), all of which were burning premium requests.

## Implementation

Added an exported `classifyCopilotInitiator(body)` helper that inspects the last message role in the request body:
- `role: "user"` → `X-Initiator: user` (genuine user turn, 1 premium request)
- `role: "tool"` / `role: "assistant"` → `X-Initiator: agent` (free)
- Parse failures / empty body → defaults to `"user"` (conservative, don't hide usage)

Wired into the existing `copilotFetchFn` in `providerModelFactory.ts`.

## Validation

- 7 unit tests covering all classification paths (user, tool, assistant, empty, null, malformed, missing messages)

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$5.71`_

### DIFF
--- a/src/node/services/providerModelFactory.test.ts
+++ b/src/node/services/providerModelFactory.test.ts
@@ -10,7 +10,6 @@ import {
   classifyCopilotInitiator,
   modelCostsIncluded,
   MUX_AI_PROVIDER_USER_AGENT,
-  resolveAIProviderBodySource,
   resolveAIProviderHeaderSource,
 } from "./providerModelFactory";
 import { ProviderService } from "./providerService";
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed your feedback: now resolves the effective body from both `init?.body` and `input.body` (when input is a `Request` object) via a new `resolveAIProviderBodySource` helper. Added tests for the new helper.

ibetitsmike:
@codex review

Fixed the Request body handling: `copilotFetchFn` now clones the Request and reads body text via `await input.clone().text()` when `init.body` is not a string. This ensures proper classification even when the SDK uses `fetch(Request)`. The `resolveAIProviderBodySource` helper was removed in favor of inline resolution (since the body read is async and only needed here).

PR #2738, opened 2026-03-03
🤖 ci: bump actions/upload-artifact to v7.0.0 and use direct single-file uploads

ethanndickson:
## Summary
Bump all pinned `actions/upload-artifact` references from v4.6.2 to v7.0.0, and enable direct single-file uploads for our single-file release artifacts.

## Background
`actions/upload-artifact` v7 adds support for direct (unzipped) single-file uploads via `archive: false`. We already upload several single-file artifacts (Linux AppImage, macOS DMGs, VSIX), so we can avoid wrapping those files in zip archives.

## Implementation
- Updated every `actions/upload-artifact` usage in workflows to:
  - `actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0`
- In `.github/workflows/pr.yml`, switched single-file upload steps to direct uploads by:
  - adding `archive: false`
  - removing `name` for those steps (ignored by v7 when `archive: false` is set)
- Kept multi-file/directory uploads unchanged (still using archive mode)
- Updated one stale inline comment in `terminal-bench.yml` from `@v4` to `@v7`

## Validation
- `make static-check`
- YAML parse validation for edited workflow files

## Risks
Low. Changes are limited to GitHub Actions workflow config. Multi-file uploads retain previous behavior; only known single-file uploads switched to direct mode.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$1.19`_

### DIFF
--- a/.github/workflows/_desktop-release.yml
+++ b/.github/workflows/_desktop-release.yml
@@ -201,7 +201,7 @@ jobs:
           GCLOUD_ACCESS_TOKEN: ${{ steps.signing.outputs.gcloud_access_token }}
 
       - name: Upload Windows artifacts
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: windows-release
           path: |
--- a/.github/workflows/perf-profiles.yml
+++ b/.github/workflows/perf-profiles.yml
@@ -35,7 +35,7 @@ jobs:
           MUX_E2E_PERF_PROFILES: ${{ github.event.inputs.perf_profiles || '' }}
       - name: Upload perf artifacts
         if: always()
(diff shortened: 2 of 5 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Re: the P1 about keeping explicit artifact names — this is incorrect. The v7 release notes explicitly state:

> Callers can set the new `archive` parameter to `false` to skip zipping the file during upload. Right now, we only support single files. **The `name` parameter is also ignored with this setting. Instead, the name of the artifact will be the name of the uploaded file.**

So when `archive: false` is set, `name` is a no-op — the artifact is named after the file itself:
- `release/*.AppImage` → e.g. `Mux-0.19.1.AppImage`
- `release/*-x64.dmg` → e.g. `Mux-0.19.1-x64.dmg`
- `release/*-arm64.dmg` → e.g. `Mux-0.19.1-arm64.dmg`
- `vscode/mux-*.vsix` → e.g. `mux-0.19.1.vsix`

All four are distinct filenames — no duplicate-name conflict is possible. Adding `name` back would be ignored by v7 in direct upload mode.

PR #2740, opened 2026-03-03
🤖 fix: allow editing AUTO user messages

ethanndickson:
## Summary
Allow users to edit visible AUTO (synthetic) user messages from the message action bar.

## Background
AUTO messages are synthetic system-authored prompts, but users may want to take ownership and revise them. The UI previously hid edit affordances for synthetic messages even though backend edit/truncate flow already supports editing by history ID.

## Implementation
- Updated `UserMessage` edit gating to remove the synthetic-only restriction.
- Kept local command output messages non-editable.
- Left AUTO badge and styling unchanged; this is only an edit affordance change.

## Validation
- `make static-check`
- `make typecheck`

## Risks
Low risk and scoped to user-message action rendering. Existing local-command output guard is preserved.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$4.37`_

### DIFF
--- a/src/browser/features/Messages/UserMessage.tsx
+++ b/src/browser/features/Messages/UserMessage.tsx
@@ -69,7 +69,8 @@ export const UserMessage: React.FC<UserMessageProps> = ({
   const { copied, copyToClipboard } = useCopyToClipboard(clipboardWriteText);
 
   const handleEdit = () => {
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

PR #2742, opened 2026-03-03
🤖 fix: lock AgentModePicker when workspace has non-selectable agent

ethanndickson:
## Summary

Lock the AgentModePicker when a workspace has a backend-assigned, immutable agent — preventing users from switching agents in workspaces where the agent is fixed (e.g., "Chat with Mux", child/subagent workspaces).

## Background

- The `mux` agent (`agentId: "mux"`) is forced on the `mux-chat` workspace and on child/subagent workspaces. Users shouldn't be able to switch agents in these workspaces.
- Codex P1 flagged that using `!uiSelectable` as the lock signal was wrong — `uiSelectable` has transient overlays (like `requiresPlan && !planReady`) that conflate discoverability with permanent assignment.

## Implementation

Two commits:

**1. `fix: include agentId in workspace metadata read paths`**  
Ensures `agentId` is included in all 4 `getAllWorkspaceMetadata()` code paths (local projects, SSH remotes, recent projects, mux-chat). Without this, the frontend never sees `agentId` on workspace metadata.

**2. `feat: lock AgentModePicker when workspace has non-selectable agent`**  
Derives the lock condition from fields already on workspace metadata — no new schema fields, no migration logic, no prop threading:

```ts
const isCurrentAgentLocked =
  props.workspaceId === MUX_HELP_CHAT_WORKSPACE_ID ||
  currentMeta?.parentWorkspaceId != null;
```

This covers:
- `mux-chat` workspace (Chat with Mux) — matched by workspace ID
- Child/subagent workspaces — matched by `parentWorkspaceId` being set

The derivation happens in `AgentContext.tsx` using `useWorkspaceMetadata()` from `WorkspaceContext`, which is already available in the provider tree.

`AgentModePicker.tsx` reads `isAgentSelectionLocked` from context and disables interaction — no changes needed there.

<details>
<summary>Design alternatives considered</summary>

- **`!uiSelectable`**: Rejected — conflates transient plan-readiness gates with permanent assignment.
- **`!uiRoutable`**: Rejected — the `mux` agent explicitly sets `routable: true`.
- **Explicit `agentSelectionLocked` field**: Implemented then reverted — required schema changes in 2 files, migration logic in 4 config read paths, creation-site stamps in 2 services, and prop threading through 3 frontend layers. All unnecessary when the lock condition can be derived from existing fields.

</details>

---

### DIFF
--- a/src/browser/components/AgentModePicker/AgentModePicker.tsx
+++ b/src/browser/components/AgentModePicker/AgentModePicker.tsx
@@ -129,6 +129,10 @@ export const AgentModePicker: React.FC<AgentModePickerProps> = (props) => {
       : agents.find((entry) => entry.id === normalizedAgentId);
   const isAgentLocked = isAgentSelectionLocked;
 
+  // Derived "effectively open" — hides picker immediately when lock activates,
+  // preventing stale event handlers from a hidden-but-open picker.
+  const isPickerVisible = isPickerOpen && !isAgentLocked;
+
   const activeOption = !normalizedAgentId
     ? null
     : !activeDescriptor
@@ -205,7 +209,7 @@ export const AgentModePicker: React.FC<AgentModePickerProps> = (props) => {
 
   // Close picker when clicking outside.
   useEffect(() => {
-    if (!isPickerOpen) {
+    if (!isPickerVisible) {
       return;
     }
 
@@ -218,7 +222,7 @@ export const AgentModePicker: React.FC<AgentModePickerProps> = (props) => {
 
     document.addEventListener("mousedown", handleClickOutside);
     return () => document.removeEventListener("mousedown", handleClickOutside);
-  }, [closePicker, isPickerOpen]);
+  }, [closePicker, isPickerVisible]);
 
   // Scroll highlighted item into view.
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

Re: the P2 about gating `mux-chat` lock on system workspace metadata — dismissing as non-issue:

1. `mux-chat` is a system-reserved ID. Workspace IDs are system-generated, not user-chosen. The only path to a collision is a legacy config artifact from before the system workspace existed.
2. The migration test (`serviceContainer.test.ts` line 105) creates *two* workspaces with `id: "mux-chat"` — one legacy, one system. That is already a confusing state; locking the picker is the safer default.
3. Any metadata-based gating (e.g. a `systemWorkspace` flag on `FrontendWorkspaceMetadata`) would be new schema work out of scope for this PR. The current ID check matches how every other system feature gates on `mux-chat`.

PR #2744, opened 2026-03-03
🤖 fix: recover stale workspace activity state after sleep

ethanndickson:
## Summary
This change makes long-lived workspace subscriptions resilient to sleep/resume half-open connections so workspace status no longer stays stuck as "working" after streaming has already finished.

## Background
`workspace.activity.subscribe` and `terminal.activity.subscribe` had reconnect logic but no liveness signal. After lock/sleep, the renderer could keep stale `streaming` state indefinitely because the connection looked open but stopped delivering events.

## Implementation
- Added a shared queue decorator: `withQueueHeartbeat` in `src/common/utils/withQueueHeartbeat.ts`
  - Works with both async queue implementations via a shared `{ push, iterate, end }` interface.
  - Starts periodic heartbeat emission on iteration and guarantees timer cleanup on `end()` / iterator teardown.
- Extended ORPC schemas so activity subscriptions include `heartbeat` variants.
- Updated router handlers to use `withQueueHeartbeat` for:
  - `workspace.onChat`
  - `workspace.activity.subscribe`
  - `terminal.activity.subscribe`
- Updated `workspace.activity.subscribe` payloads to discriminated events (`type: "activity"`) so heartbeat and data events share one union.
- Added stall watchdog + heartbeat filtering in `WorkspaceStore` for:
  - `runActivitySubscription`
  - `runTerminalActivitySubscription`
- Renamed chat-specific liveness constants/functions in `WorkspaceStore` to generic subscription names.
- Added tests:
  - `src/common/utils/withQueueHeartbeat.test.ts`
  - `src/browser/stores/WorkspaceStore.test.ts` updates for heartbeat unions + retry coverage

## Validation
- `make static-check`
- `make typecheck`
- `make lint`
- `bun test src/common/utils/withQueueHeartbeat.test.ts`

## Risks
- **Low to medium**: touches shared subscription plumbing on both backend and renderer.
- Mitigations:
  - Heartbeats are explicitly filtered at consumers before state mutation.
  - Existing reconnect path already reconciles with full snapshots after reconnect.
  - Added focused tests for heartbeat behavior and stall-triggered retries.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$14.45`_

### DIFF
--- a/src/browser/stores/WorkspaceStore.ts
+++ b/src/browser/stores/WorkspaceStore.ts
@@ -2520,6 +2520,54 @@ export class WorkspaceStore {
     }
   }
 
+  /**
+   * Creates a stall watchdog that aborts the attempt when no events arrive
+   * within the configured timeout. Call start() only after the subscription
+   * connection is established so handshake latency isn't misclassified as
+   * stream silence.
+   */
+  private createStallWatchdog(
+    attemptController: AbortController,
+    label: string
+  ): { markEvent: () => void; start: () => void; stop: () => void } {
+    let lastEventAt = Date.now();
+    let interval: ReturnType<typeof setInterval> | null = null;
+
+    return {
+      markEvent: () => {
+        lastEventAt = Date.now();
+      },
+      start: () => {
+        if (interval != null) {
+          return;
+        }
+
+        lastEventAt = Date.now();
+        interval = setInterval(() => {
+          if (attemptController.signal.aborted) {
+            return;
+          }
+
+          const elapsedMs = Date.now() - lastEventAt;
+          if (elapsedMs < SUBSCRIPTION_STALL_TIMEOUT_MS) {
+            return;
+          }
+
+          console.warn(
+            `[WorkspaceStore] ${label} stalled (no events for ${elapsedMs}ms); retrying...`
+          );
+          attemptController.abort();
+        }, SUBSCRIPTION_STALL_CHECK_INTERVAL_MS);
+      },
+      stop: () => {
+        if (interval != null) {
+          clearInterval(interval);
+          interval = null;
+        }
+      },
+    };
+  }
+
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed the P1: removed the `getActivityList()` replay from `workspace.activity.subscribe` entirely (commit dd96ba4).

The subscribe handler now emits only live activity deltas + heartbeats. Bootstrap snapshots remain the responsibility of `workspace.activity.list()`, which both desktop (`WorkspaceStore.ts:2679-2701`) and mobile (`useProjectsData.ts:26-30`) already call independently. This eliminates the stale-overwrite race by construction.

ethanndickson:
@codex review

Addressed the P2: moved `stallInterval` creation in `runActivitySubscription` to after the bootstrap `list()` + snapshot application completes (commit bdfd6fa). `lastEventAt` is reset right before the watchdog starts so slow `list()` responses (e.g., after resume with degraded IPC) don't trigger false-positive stall aborts.

ethanndickson:
@codex review

Addressed the P2: extracted a shared `createStallWatchdog` helper on `WorkspaceStore` and applied it to both subscription loops (commit 990ec9b).

The invariant is now encoded once: **watchdog starts only after the connection/bootstrap is live**.

- `runTerminalActivitySubscription`: `watchdog.start()` is called after `await subscribe(...)` resolves — handshake latency no longer counts toward the stall timeout.
- `runActivitySubscription`: `watchdog.start()` remains after `list()` + snapshot application (unchanged semantics, now using the shared helper).

All inline `setInterval`/`clearInterval`/`lastEventAt` removed from both loops in favor of `watchdog.markEvent()` / `watchdog.start()` / `watchdog.stop()`.

ethanndickson:
@codex review

PR #2745, opened 2026-03-03
🤖 feat: add system config layer (config.system.json)

ibetitsmike:
Closes https://github.com/coder/mux/issues/2728

## Summary

Adds a read-only system config layer (`config.system.json`) that lets admins provide baseline defaults. Merge precedence: `env vars > user config.json > config.system.json > hardcoded defaults`. Writes always target `config.json` — the system file is never modified by Mux.

## Background

All mux configuration currently lives in `~/.mux/config.json` and is fully user-controlled. Managed deployments have no mechanism to provide baseline defaults (providers, models, feature flags). This PR adds the system config layer to support that use case.

## Implementation

**`src/node/config.ts`** (~100 LoC net):
- `resolveSystemConfigFile()` — reads `MUX_SYSTEM_CONFIG` env var, defaults to `<rootDir>/config.system.json`
- `loadSystemConfig()` — non-throwing loader; missing file → no warning; malformed/unreadable → warning logged + stored in `systemConfigWarnings`
- `loadUserConfig()` — extracted from `loadConfigOrDefault()` for symmetry; has its own try/catch so a malformed user config falls back to system-only
- `mergeSystemAndUserConfig()` — one-level deep merge: scalars/arrays user-wins, plain objects key-merge, projects merge by path; validates project tuples before Map insertion
- `stripSystemDefaults()` — called in `saveConfig()` to prevent materialization of system defaults into user config
- `loadConfigOrDefault()` restructured: load system → load user → merge → normalize

**`src/node/config.test.ts`** (~230 LoC net):
15 test cases covering: system defaults, user overrides, deep merge, array replacement, write isolation, env var override, malformed/permission-denied/binary/non-object fallback, project merge, no-warning on missing file, anti-materialization, user-modified defaults persistence, malformed project tuple handling, and system config preservation on user config failure.

## Validation

- 42/42 config tests pass
- `make static-check` green (typecheck + lint + fmt + docs)

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$5.65`_

### DIFF
--- a/src/node/config.ts
+++ b/src/node/config.ts
@@ -351,8 +351,14 @@ export class Config {
           // Deep merge with per-key type validation: user sub-keys override system
           // sub-keys, but only when the types are compatible. This prevents a
           // malformed user entry (e.g., runtimeEnablement.docker: "oops" instead of
-          // false) from displacing a valid admin baseline. When both are same-type
-          // strings, user wins — consumer-level validation handles invalid values.
+          // false) from displacing a valid admin baseline. When both are same-type,
+          // user wins — consumer-level validation handles invalid values.
+          //
+          // LIMITATION (sparse storage): for fields like runtimeEnablement where only
+          // `false` is meaningful, a user's same-type `true` passes the type check
+          // but gets dropped by normalization — the system's `false` is lost. This is
+          // inherent to the sparse storage pattern; a dedicated follow-up can address
+          // it via per-value semantic validation in the normalizer.
           const sysObj = systemVal as Record<string, unknown>;
           const usrObj = userVal as Record<string, unknown>;
           const merged: Record<string, unknown> = { ...sysObj };


ibetitsmike:
@codex review

ibetitsmike:
@codex review

1. **P1 materialization fix**: Added `stripSystemDefaults()` in `saveConfig()` — before writing, system-matching fields are stripped so only user-set values persist.
2. **P2 project tuple validation**: `mergeSystemAndUserConfig()` now validates each project entry before Map insertion (checks array + length + string path).

Both fixes include new test coverage.

ibetitsmike:
@codex review

Fixed: `loadUserConfig()` now has its own try/catch inside `loadConfigOrDefault()`. A malformed `config.json` falls back to system config only (instead of discarding it). Added test coverage for this scenario.

ibetitsmike:
@codex review

The sparse storage limitation (stopCoderWorkspaceOnArchive, runtimeEnablement) is an inherent tension in the existing storage pattern — not a merge logic bug. Documented this constraint with clear comments in `mergeSystemAndUserConfig()` and flagged it for a follow-up that addresses sparse-field persistence when system config provides non-defaults.

ibetitsmike:
@codex review

Fixed both issues:
1. **P1 key-level stripping**: `stripSystemDefaults()` now strips matching subkeys individually from object-valued settings (featureFlagOverrides, agentAiDefaults, etc.), instead of comparing whole objects.
2. **P2 legacy preservation**: `agentAiDefaults` normalization now merges legacy `subagentAiDefaults` as a baseline before applying modern `agentAiDefaults`, so legacy user settings survive system config rollout.

Both fixes include new test coverage. 44/44 tests pass.

ibetitsmike:
@codex review

Fixed: Added `sysVal != null` guard in `stripSystemDefaults()` so `null` system config values fall through to scalar comparison instead of entering the object-subkey path. Added regression test.

ibetitsmike:
@codex review

This is the same sparse storage limitation that's already documented in the `mergeSystemAndUserConfig()` method doc comment — `apiServerServeWebUi` only persists `true` (same pattern as `stopCoderWorkspaceOnArchive` only persisting `false`). The limitation applies to all sparsely-stored boolean fields and is a pre-existing design pattern, not a merge logic bug. The doc comment explicitly calls this out and flags it for a dedicated follow-up that addresses sparse-field persistence semantics against system baselines.

No further code changes needed — this is the documented limitation.

ibetitsmike:
@codex review

Fixed: `stripSystemDefaults()` now strips `taskSettings` via existing key-level object comparison, and strips `projects` via path-based entry comparison. System-provided project entries and task settings that match exactly are removed from the save data. Added regression test.

46/46 tests pass + static-check green.

ibetitsmike:
@codex review

Fixed: System project paths in `stripSystemDefaults` are now normalized via `stripTrailingSlashes()` before comparison, matching the normalization applied during load. Added regression test with trailing-slash system project path.

47/47 tests pass + static-check green.

ibetitsmike:
@codex review

ibetitsmike:
@codex review

Fixed: Normalization fallback to system defaults when user values are invalid. For `defaultModel`, `hiddenModels`, and `defaultRuntime`, if normalization rejects the merged value (e.g., malformed model string), we now fall back to the system config's value re-normalized. Added test covering the `defaultModel` case.

48/48 tests pass + static-check green.

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
The `runtimeEnablement: {docker: true}` case is an inherent limitation of the sparse storage pattern (only `false` is meaningful). The type guard catches cross-type mismatches, but same-type semantic invalidity (`true` where only `false` is stored) requires per-value validation in the normalizer — which would need access to the system config. I've documented this as a known limitation in the merge comments and in the method-level doc. This is the same sparse storage limitation already called out at the top of the method. A follow-up can address it with semantic-aware merging if needed.

ibetitsmike:
@codex review

PR #2764, opened 2026-03-03
🤖 tests: stabilize landing page workspace ordering

ammar-agent:
## Summary
Landing page "Recent Workspaces" cards could reorder between Storybook/Chromatic runs when multiple workspaces shared the same recency timestamp. This change makes the sort deterministic so snapshots stay stable.

## Background
The new landing page sorts recent workspaces by recency only. In Storybook fixtures, recency ties are common, and map insertion order can vary enough to produce inconsistent card ordering.

## Implementation
- Updated `RecentWorkspacesSection` in `LandingPage.tsx` to keep recency as the primary key.
- Added deterministic tie-breakers in this order: `createdAt` (desc), `name` (asc), `id` (asc).
- Documented why the tie-breakers are necessary for Storybook/Chromatic stability.

## Validation
- `make static-check`

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$n/a`_

### DIFF
--- a/src/node/services/tools/web_fetch.ts
+++ b/src/node/services/tools/web_fetch.ts
@@ -21,23 +21,45 @@ import { getErrorMessage } from "@/common/utils/errors";
 
 const USER_AGENT = "Mux/1.0 (https://github.com/coder/mux; web-fetch tool)";
 
+/**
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #2771, opened 2026-03-04
🤖 fix: extract shared branch-sync guard for PR wait scripts

ethanndickson:
## Summary

Extracts the duplicated branch-sync guard from `wait_pr_checks.sh`, `wait_pr_ready.sh`, and `wait_pr_codex.sh` into a shared `scripts/lib/branch_sync_guard.sh` helper, fixing two bugs in the process:

1. **False "not in sync" on worktree-created feature branches** — worktrees inherit `origin/main` as upstream; the old `@{u}` comparison always diverged after pushing a feature branch.
2. **Unsafe auto-push on fetch failure** — any `git fetch` failure was treated as "branch missing" and triggered `git push -u`, which could create unintended remote branches when a local branch intentionally tracks a differently named upstream.

## Background

Mux creates worktrees via `git worktree add -b <feature-branch> <path> origin/main`. Git auto-configures the upstream to `origin/main`. When an agent pushes a feature branch and runs any `wait_pr_*.sh` script, the sync check compared against the wrong remote ref.

The initial fix hardcoded `origin/$CURRENT_BRANCH` but conflated fetch errors with "branch missing," triggering auto-push even when a configured upstream existed.

## Implementation

New shared helper `scripts/lib/branch_sync_guard.sh` with three functions:

- **`resolve_branch_sync_target`** — determines comparison ref with explicit policy:
  - If `@{u}` exists and points to a non-default branch → honor it (supports cross-name tracking)
  - If `@{u}` points to `origin/<default>` but `origin/<current_branch>` exists → use that (worktree protection)
  - If no upstream → use `origin/<current_branch>`
- **`fetch_or_bootstrap`** — fetches the resolved ref; only auto-pushes when no upstream is configured
- **`assert_branch_synced`** — full sync assertion with ahead/behind/diverged diagnostics

All three wait scripts now source this helper and call `assert_branch_synced`, replacing ~50 duplicated lines each.

## Validation

- `shellcheck` passes clean on all 4 scripts
- Net -11 lines (140 added in shared helper, 151 removed from 3 scripts)

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$0.79`_

### DIFF
--- a/scripts/lib/branch_sync_guard.sh
+++ b/scripts/lib/branch_sync_guard.sh
@@ -1,0 +1,128 @@
+#!/usr/bin/env bash
+# Shared branch-sync guard for wait_pr_*.sh scripts.
+# Resolves the correct remote comparison target, fetches it,
+# and optionally bootstraps a first push when no upstream exists.
+#
+# Usage: source this file, then call assert_branch_synced.
+# Returns 0 if local HEAD matches remote; non-zero otherwise.
+# Callers set CURRENT_BRANCH and REMOTE_BRANCH before proceeding.
+
+set -euo pipefail
+
+# Resolve which remote ref to compare HEAD against.
+# Sets two variables in the caller's scope:
+#   CURRENT_BRANCH  — local branch name
+#   REMOTE_BRANCH   — remote tracking ref to compare against (e.g. origin/feature)
+#
+# Policy:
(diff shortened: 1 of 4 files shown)


ethanndickson:
@codex review

PR #2773, opened 2026-03-04
🤖 fix: preserve response token metadata in DevTools headers

ThomasK33:
## Summary
Stop over-redacting DevTools **response** headers so rate-limit metadata remains visible for debugging.

## Background
`redactHeaders` previously treated any header containing `token` as sensitive. That behavior is valid for request headers but caused false positives on response headers like `anthropic-ratelimit-input-tokens-remaining`.

## Implementation
- Made header sensitivity checks direction-aware in `devToolsHeaderCapture.ts`.
- Kept broad `token`/`secret` redaction for request headers only.
- Preserved response redaction for explicitly sensitive names and cookie-related headers.
- Updated both DevTools middleware response-header callsites to pass `"response"`.
- Added focused tests proving:
  - response token-named headers remain visible,
  - response `set-cookie` remains redacted,
  - request token headers remain redacted.

## Validation
- `make static-check`
- `bun test src/node/services/__tests__/devToolsHeaderCapture.test.ts`

## Risks
Low: change is narrowly scoped to DevTools header redaction and covered by targeted tests for request/response behavior split.

---

<details>

# Stop over-redacting response HTTP headers

## Context / Why

The `isSensitiveHeaderName` function in `devToolsHeaderCapture.ts` uses a substring check `lower.includes("token")` to catch custom auth headers. This is appropriate for **request** headers (which carry credentials), but it false-positives on every Anthropic rate-limit **response** header (e.g., `anthropic-ratelimit-input-tokens-remaining`) because they contain "tokens" in the name. Rate-limit info is purely operational metadata — useful for debugging 429s — and contains nothing secret.

**Key insight:** Response headers come *from* the server. They inherently don't contain outbound credentials. The only response header worth redacting is `set-cookie` (session establishment), which is already caught by the existing `cookie` includes check.

## Evidence

| File | Role |
|---|---|
| `src/node/services/devToolsHeaderCapture.ts` | Defines `isSensitiveHeaderName`, `redactHeaders`, `captureAndStripDevToolsHeader` |
| `src/node/services/devToolsMiddleware.ts:417,576` | Calls `redactHeaders` on **response** headers |
| `src/node/services/devToolsHeaderCapture.ts:76` | Calls `redactHeaders` on **request** headers (via `captureAndStripDevToolsHeader`) |
| `src/node/services/__tests__/devToolsHeaderCapture.test.ts` | Existing tests for header capture + redaction |

### 1. Split `redactHeaders` by direction (~+10 LoC)

**File:** `src/node/services/devToolsHeaderCapture.ts`

Add a `direction` parameter so response headers skip the broad substring checks:

```ts
/**
 * Response headers come from the server and don't carry outbound credentials.
 * Only redact set-cookie (session tokens) via the "cookie" includes check.
 * Skip broad "token"/"secret" substring matching that false-positives on
 * rate-limit headers (e.g. anthropic-ratelimit-input-tokens-remaining).
 */
function isSensitiveHeaderName(
  name: string,
  direction: "request" | "response",
): boolean {
  const lower = name.toLowerCase();
  if (SENSITIVE_HEADER_NAMES.has(lower) || lower.includes("cookie")) {
    return true;
  }
  // Broad substring checks only apply to request headers, which may carry
  // auth credentials under custom names. Response headers are server-originated
  // and safe to record in full.
  if (direction === "request") {
    return lower.includes("secret") || lower.includes("token");
  }
  return false;
}
```

### 2. Plumb direction through `redactHeaders` (~+2 LoC)

```ts
export function redactHeaders(
  headers: Record<string, string>,
  direction: "request" | "response" = "request",
): Record<string, string> {
  const redacted: Record<string, string> = {};
  for (const [key, value] of Object.entries(headers)) {
    redacted[key] = isSensitiveHeaderName(key, direction) ? "[REDACTED]" : value;
  }
  return redacted;
}
```

Default `"request"` keeps the existing `captureAndStripDevToolsHeader` call site (line 76) working without changes.

### 3. Pass `"response"` at call sites (~2 lines changed)

**File:** `src/node/services/devToolsMiddleware.ts`

- **Line 417:** `redactHeaders(..., "response")`
- **Line 576:** `redactHeaders(..., "response")`

### 4. Update tests (~+15 LoC)

**File:** `src/node/services/__tests__/devToolsHeaderCapture.test.ts`

- Export `redactHeaders` in the import (it's already exported).
- Add test: response headers with "token" in name are **not** redacted.
- Add test: response `set-cookie` is still redacted.
- Add test: request headers with "token" in name are still redacted (existing behavior preserved).

## Net LoC estimate

~+25 lines (production + test)

</details>

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$1.03`_

### DIFF
--- a/src/node/services/__tests__/devToolsHeaderCapture.test.ts
+++ b/src/node/services/__tests__/devToolsHeaderCapture.test.ts
@@ -71,6 +71,17 @@ describe("devToolsHeaderCapture", () => {
     expect(headers["x-ratelimit-tokens-reset"]).toBe("10");
   });
 
+  it("redacts unknown response token headers by default", () => {
+    const headers = redactHeaders(
+      {
+        "x-auth-token": "response-token",
+      },
+      "response"
+    );
+
+    expect(headers["x-auth-token"]).toBe("[REDACTED]");
+  });
+
   it("still redacts response set-cookie headers", () => {
     const headers = redactHeaders(
       {
--- a/src/node/services/devToolsHeaderCapture.ts
+++ b/src/node/services/devToolsHeaderCapture.ts
@@ -28,24 +28,38 @@ const SENSITIVE_HEADER_NAMES = new Set([
 ]);
 
 /**
- * Prefix-match for bearer/token patterns that may appear under custom names.
- *
- * Response headers come from the server and do not carry outbound credentials.
- * Only redact set-cookie (via the cookie check) and known sensitive names there.
+ * Response headers can still carry credentials under custom names (for example
+ * from provider proxies), so we keep broad token/secret redaction enabled in
+ * both directions. To avoid over-redacting operational metadata, explicitly
+ * allowlist known non-sensitive rate-limit response headers.
  */
+const SAFE_RESPONSE_TOKEN_HEADER_PREFIXES = [
+  "anthropic-ratelimit-",
+  "x-ratelimit-",
+  "ratelimit-",
+] as const;
+
+function isKnownSafeResponseTokenHeader(name: string): boolean {
+  return SAFE_RESPONSE_TOKEN_HEADER_PREFIXES.some((prefix) => name.startsWith(prefix));
+}
+
+/** Prefix-match for bearer/token patterns that may appear under custom names. */
 function isSensitiveHeaderName(name: string, direction: "request" | "response"): boolean {
   const lower = name.toLowerCase();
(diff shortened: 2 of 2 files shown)


ThomasK33:
@codex review

Please review this header redaction change and CI readiness.

ThomasK33:
@codex review

Applied feedback by restoring default token/secret redaction for response headers and allowlisting known rate-limit response prefixes.

PR #2777, opened 2026-03-04
🤖 fix: tighten task/task_await orchestration guidance

ThomasK33:
## Summary
Clarifies task orchestration guidance to prevent fabricated task IDs and avoid unnecessary background-task round trips.

## Background
Models were issuing `task` and `task_await` in the same parallel batch, causing hallucinated IDs and `not_found` responses. Guidance also encouraged a wasteful single background-task + immediate await pattern.

## Implementation
- Updated `task_ids` description to require real IDs from prior tool results.
- Updated `task` description to recommend foreground mode for single tasks and background mode for parallel batches.
- Updated `task_await` description with explicit sequencing guidance (spawn first, await in next step).
- Removed the sequential-protocol parenthetical in orchestrator docs that encouraged immediate await after spawn.
- Regenerated synced artifacts/docs (`builtInAgentContent.generated.ts`, `docs/agents/index.mdx`, `docs/hooks/tools.mdx`).

## Validation
- `make static-check`

## Risks
Low risk: wording-only changes in tool/agent guidance plus generated/synced documentation updates.

---

<details>

# Fix hallucinated task IDs and wasteful background→await patterns

## Context & Problem

Opus 4.6 (and likely other models) exhibit two related anti-patterns when using the `task` + `task_await` tools:

1. **Hallucinated task IDs:** The model emits `task` and `task_await` as parallel tool calls. Since parallel calls are dispatched simultaneously, `task_await` doesn't have the real `taskId` yet — so the model **fabricates** one, which returns `not_found`. An extra round-trip is wasted to then await the real ID.

2. **Wasteful single-task background:** The model spawns one background task, then immediately awaits it. This is equivalent to a foreground task but costs an extra tool call. Background tasks only shine when launching **multiple** parallel tasks and then awaiting the batch.

## Evidence

- User-provided screenshot shows Step 2: `task` (exec) + `task_await(["21a0e67f81"])` fired in parallel → `not_found`. Step 3 then correctly awaits the real `3b8af27e8a`.
- Current `task` description (line 984–986) explains foreground vs background but doesn't discourage the single-task-background pattern.
- Current `task_await` description (line 997–1006) has no warning about parallel invocation with `task`.
- `task_ids` param description (line 237–238) says nothing about IDs needing to come from prior tool results.
- `orchestrator.md` line 151 says `"(or spawn, then immediately task_await)"` — literally teaching the anti-pattern for sequential cases.

### Edit 1: `task` tool description — guide toward foreground for single tasks

**File:** `src/common/utils/tools/toolDefinitions.ts` lines 984–987

Replace the foreground/background explanation paragraph:

```typescript
// BEFORE (lines 984–987):
      "\n\nIf run_in_background is false, waits for the sub-agent to finish and returns a completed reportMarkdown. " +
      "If the foreground wait times out, returns a queued/running taskId with a note (the task continues running); use task_await to monitor progress. " +
      "If run_in_background is true, returns a queued/running taskId with a note; use task_await to wait for completion, task_list to rediscover active tasks, and task_terminate to stop it. " +
      "Use the bash tool to run shell commands.",

// AFTER:
      "\n\nIf run_in_background is false, waits for the sub-agent to finish and returns a completed reportMarkdown. " +
      "If the foreground wait times out, returns a queued/running taskId with a note (the task continues running); use task_await to monitor progress. " +
      "If run_in_background is true, returns immediately with a queued/running taskId; use task_await to wait for completion, task_list to rediscover active tasks, and task_terminate to stop it. " +
      "Prefer run_in_background: false when spawning a single task — it is equivalent to spawning background + immediately awaiting, but saves a round-trip. " +
      "Use run_in_background: true when launching multiple tasks in parallel so you can await them as a batch. " +
      "Use the bash tool to run shell commands.",
```

### Edit 2: `task_await` tool description — warn against parallel invocation with `task`

**File:** `src/common/utils/tools/toolDefinitions.ts` lines 997–1006

Add a leading `IMPORTANT` paragraph:

```typescript
// BEFORE (lines 997–998):
      "Wait for one or more tasks to produce output. " +
      "Agent tasks return reports when completed. " +

// AFTER:
      "Wait for one or more tasks to produce output. " +
      "\n\nIMPORTANT: Do not call task_await in the same parallel tool-call batch as task or bash — " +
      "the taskId is not available until the spawning tool returns. " +
      "Always wait for the task/bash tool result first, then call task_await in a subsequent step. " +
      "When omitting task_ids to await all active tasks, ensure at least one background task was already spawned in a prior step. " +
      "\n\nAgent tasks return reports when completed. " +
```

### Edit 3: `task_ids` param description — forbid fabricated IDs

**File:** `src/common/utils/tools/toolDefinitions.ts` lines 237–238

```typescript
// BEFORE:
      .describe(
        "List of task IDs to await. When omitted, waits for all active descendant tasks of the current workspace."
      ),

// AFTER:
      .describe(
        "List of task IDs to await — use only real IDs returned by prior task or bash tool results; never fabricate an ID. " +
        "When omitted, waits for all active descendant tasks of the current workspace."
      ),
```

### Edit 4: `orchestrator.md` — remove anti-pattern parenthetical

**File:** `src/node/builtinAgents/orchestrator.md` line 151

```markdown
<!-- BEFORE: -->
1. Spawn the prerequisite implementation task (`exec` or `plan`, based on complexity) with `run_in_background: false` (or spawn, then immediately `task_await`).

<!-- AFTER: -->
1. Spawn the prerequisite implementation task (`exec` or `plan`, based on complexity) with `run_in_background: false`.
```

The parenthetical `"(or spawn, then immediately task_await)"` teaches exactly the pattern we're discouraging. Removing it leaves the cleaner instruction: just use foreground.

## Validation

- `make typecheck` — ensures no TS errors from string changes
- `make test` — run full test suite (tool definition schemas are tested)
- `make lint` — formatting/lint pass
- Manual review: verify the tool descriptions render correctly when serialized to the model (search for any tests that snapshot tool descriptions)

</details>

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$1.49`_


ThomasK33:
@codex review

Addressed both review threads:
- task_await task_ids guidance now also allows IDs from `task_list` results.
- Orchestrator sequential protocol now instructs awaiting foreground tasks that return queued/running before patch application.

Also re-ran `make static-check` locally (pass).

PR #2779, opened 2026-03-04
🤖 fix: immersive review shows all files regardless of folder filter

ibetitsmike:
## Summary

Immersive review now always fetches all diff hunks, regardless of any folder/file filter set in the panel view.

## Background

When a user clicks a folder in the review pane's file tree, a `selectedFilePath` filter is applied at the git level — the diff command includes `-- "path/to/folder"`, so only hunks for that folder are fetched. This filtering leaked into the immersive review experience: file navigation skipped files outside the selected folder because they had no hunks.

## Implementation

One-line logic change in `ReviewPanel.tsx`: the git path filter is now bypassed when immersive mode is active (`selectedFilePath && !isImmersive`). Added `isImmersive` to the diff-fetch effect's dependency array so it re-runs on mode transitions.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$1.66`_

### DIFF
--- a/src/browser/features/RightSidebar/CodeReview/ReviewPanel.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/ReviewPanel.tsx
@@ -958,7 +958,7 @@ export const ReviewPanel: React.FC<ReviewPanelProps> = ({
 
             // Build truncation warning (only when not filtering by path)
             const truncationWarning =
-              truncationInfo && !selectedFilePath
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed the truncation warning feedback — the warning is now shown in immersive mode since the path filter is bypassed.

PR #2785, opened 2026-03-04
🤖 fix: destroy stdio streams on timeout to prevent Windows hanging

ibetitsmike:
## Summary

Fix bash tool timeout hanging on Windows by destroying stdio streams in the `DisposableProcess` cleanup callback, ensuring Web ReadableStream readers see EOF after process kill.

## Background

On Windows, when timeout fires and `taskkill /PID /T /F` kills the process tree, Node.js stdio pipe handles don't always close. This causes `reader.read()` in `consumeStream()` to hang indefinitely, since `Promise.all([exitCode, consumeStdout, consumeStderr])` in `bash.ts` waits for all three. Manual abort works because it explicitly cancels stream readers — timeout had no such mechanism.

## Implementation

- **`LocalBaseRuntime.ts`** / **`DevcontainerRuntime.ts`**: In the `disposable.addCleanup()` callback (runs only on timeout/abort, not normal exit), after killing the process tree, queue EOF on stdout/stderr via `push(null)` then destroy all stdio streams in `setImmediate`. The two-phase approach (EOF then destroy) avoids `ABORT_ERR` rejections from `Readable.toWeb()` while still forcing handle teardown.
- **Deleted `bashExecutionService.ts`**: Confirmed dead code (nothing imports it). Had a separate `DisposableProcess` with Unix-only `process.kill(-pid, "SIGKILL")` — removed entirely rather than patching.

## Risks

Low. The cleanup callback only fires on explicit `dispose()` (timeout/abort path), never on normal process exit. Buffered output from successful commands is unaffected.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$4.20`_


ibetitsmike:
@codex review

PR #2789, opened 2026-03-04
🤖 fix: disable gateway when onboarding credits are depleted

### LINKED ISSUE #1964: Getting started wizard shows up on broken workspaces
While a new user was trying mux for the first time and handling errors, the UI wizard was still suggesting new features. This could be distracting or frustrating while trying to get through the initial flow. 

Image

I propose we stop the getting started wizard until there is a functioning provider with an available model. A big banner or something loud in the UI should push the user to setting that up first if it's not correctly configured. Most of Mux's core UX breaks down when no models/credits are available.

ibetitsmike:
## Summary

Auto-disable Mux Gateway routing and show a warning when a user logs in during onboarding with a $0 balance, preventing a broken experience with an unusable provider.

Closes #1964

## Background

The onboarding wizard already displays the gateway balance after login, but there was no handling for depleted credits — the gateway stayed enabled even at $0, leading users to a dead end.

## Implementation

- `refresh()` in `useMuxGatewayAccountStatus` now returns `Promise<MuxGatewayAccountStatus | null>` so callers can act on the result immediately.
- Both the desktop and browser/server login flows in `OnboardingWizardSplash` now await the balance check after login. When credits are $0: default gateway models are not applied, and `gateway.toggleEnabled()` disables routing.
- A warning banner appears in the logged-in account section when balance is zero, with a link to add credits.
- Missing account status (null) is treated as permissive — gateway isn't disabled when the balance lookup fails.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$2.91`_

### DIFF
--- a/src/browser/features/SplashScreens/OnboardingWizardSplash.tsx
+++ b/src/browser/features/SplashScreens/OnboardingWizardSplash.tsx
@@ -373,6 +373,7 @@ export function OnboardingWizardSplash(props: { onDismiss: () => void }) {
           // When it eventually resolves with depleted credits, disable gateway.
           if (accountStatus == null) {
             void refreshPromise.then((laterStatus) => {
+              if (muxGatewayLoginAttemptRef.current !== attempt) return;
               if (laterStatus?.remaining_microdollars === 0 && gateway.isEnabled) {
                 gateway.toggleEnabled();
               }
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed feedback: added a 5-second timeout to the balance refresh calls in both desktop and browser flows. If the gateway endpoint stalls, the timeout resolves to `null` (permissive), so default models are applied as before — no regression.

ibetitsmike:
@codex review

Addressed slow-response bypass: the refresh promise is now stored and, when the timeout wins the race, a `.then()` on the original promise still disables gateway if the late response returns $0 balance. This handles high-latency connections without blocking onboarding defaults.

ibetitsmike:
@codex review

Added login-attempt guard (`muxGatewayLoginAttemptRef.current !== attempt`) to both late balance callback `.then()` handlers, preventing stale responses from earlier login attempts from disabling gateway on newer logins.

PR #2629, opened 2026-02-25
🤖 fix: add renderer/GPU crash lifecycle diagnostics

ethanndickson:
## Summary

Add diagnostics-only Electron lifecycle logging in `src/desktop/main.ts` so renderer/GPU failures produce actionable crash signals instead of a context-free white screen.

## Background

When the renderer dies, we were not consistently logging main-process lifecycle events that explain why the window went white. This change improves root-cause visibility without changing runtime behavior.

## Implementation

This PR adds crash-time logging hooks in the main process:

- `mainWindow.webContents.on("render-process-gone")`
  - logs `reason`, `exitCode`, and current URL
- `mainWindow.webContents.on("did-fail-load")` (main-frame only)
  - logs `errorCode`, `errorDescription`, and URL
- `mainWindow.webContents.on("unresponsive")`
  - logs renderer stall warning
- `app.on("child-process-gone")` for `details.type === "GPU"`
  - logs GPU process exit `reason` and `exitCode`

No recovery actions are introduced (no reload, no launch-flag mitigation).

## Validation

- `bun run typecheck`
- `bun run lint`
- `make static-check`

## Risks

Low. The diff is additive logging only and does not alter crash flow.

---

_Generated with `mux` • Model: `openai:gpt-5.3-codex` • Thinking: `xhigh` • Cost: `$14.81`_

PR #2802, opened 2026-03-05
🤖 fix: route crash diagnostics through log service and enable minidumps

ethanndickson:
## Summary

Make renderer/GPU crash diagnostics visible in the log file and Output Tab, and enable local minidump collection for native crash stack traces.

## Background

PR #2629 added crash lifecycle hooks (`render-process-gone`, `did-fail-load`, `unresponsive`, GPU `child-process-gone`) but they used raw `console.*` which only goes to stdout. When a renderer SIGSEGV (exit code 139) was observed on Linux, the log appeared in the terminal but not in the Output Tab or `~/.mux/logs/mux.log` — making it invisible to users who aren't watching a terminal.

Additionally, Electron's `crashReporter` was never started, so the SIGSEGV didn't produce a minidump — the crash reason was logged but the native stack trace was lost.

## Implementation

All changes in `src/desktop/main.ts`:

- **Route crash hooks through `log.*`** instead of `console.*` so entries reach the log file (`~/.mux/logs/mux.log`) and the Output Tab via the log buffer subscription
- **Forward renderer console errors** to the log service via `webContents.on("console-message")` — the renderer's global error handlers (`window.addEventListener("error")` in `main.tsx`) write to `console.error` which stays in renderer memory; this hook bridges that to the main process log service
- **Enable `crashReporter.start({ uploadToServer: false })`** so renderer segfaults write `.dmp` minidumps to `~/.config/mux/Crashpad/completed/` — no data leaves the machine

## Risks

Low. Logging-only changes with no behavioral side effects on crash flow. The `console-message` hook filters to level ≥ 2 (warnings + errors) to avoid noise.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$N/A`_

### DIFF
--- a/src/desktop/main.ts
+++ b/src/desktop/main.ts
@@ -29,6 +29,7 @@ import "disposablestack/auto";
 import type { MenuItemConstructorOptions, MessageBoxOptions } from "electron";
 import {
   app,
+  crashReporter,
   BrowserWindow,
   ipcMain as electronIpcMain,
   Menu,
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

PR #2817, opened 2026-03-05
🤖 fix: center workspace sidebar trailing metadata

ammar-agent:
## Summary

Center the trailing controls in `WorkspaceListItem` against the full workspace row so multi-line rows line up with the left status dot.

## Background

The sidebar already centers the left status indicator across the full workspace row height, but the controls on the right side of the title stayed pinned to the title line. That made rows with a secondary status line look vertically inconsistent.

## Implementation

- moved the terminal/git metadata out of the title row into a trailing wrapper that centers on the full row height
- updated the shared trailing action wrapper and disabled spacer to use the same row-centered alignment
- kept the title column flexible so the trailing metadata still does not force horizontal scrolling

## Validation

- `make typecheck`
- `make static-check`

## Risks

Low. This only changes `WorkspaceListItem` layout in the project sidebar and keeps the existing data flow and interaction logic intact.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$unknown`_

### DIFF
--- a/src/browser/components/WorkspaceListItem/WorkspaceListItem.tsx
+++ b/src/browser/components/WorkspaceListItem/WorkspaceListItem.tsx
@@ -168,10 +168,10 @@ function ActionButtonWrapper(props: { children: React.ReactNode }) {
   return (
     <div
       className={cn(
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

PR #2822, opened 2026-03-06
🤖 fix: stop reusing OpenAI previous response ids

ammar-agent:
## Summary

Stop attaching OpenAI `previousResponseId` when Mux already sends the full conversation history explicitly. This keeps GPT-5.4 request state aligned with the explicit payload history, updates the related tests/comments, and avoids premature `context_exceeded` failures from double-counting prior turns.

## Background

GPT-5.4 chats were hitting `context_exceeded` around ~230k input tokens and then falling into the existing context-exceeded compaction path. The root cause was that Mux was preserving OpenAI reasoning parts in explicit history while also chaining `previous_response_id`, which made the OpenAI Responses request state additive instead of alternative.

## Implementation

- Remove `previousResponseId` extraction/reuse from `buildProviderOptions()` for OpenAI requests.
- Keep prompt cache keys, truncation controls, and reasoning-encrypted content handling unchanged.
- Update OpenAI request-path tests to assert that explicit GPT-5.4 history does **not** reuse stale response IDs.
- Refresh comments in the model transform and AI service layers so they describe explicit-history reasoning preservation instead of `previousResponseId` chaining.
- Keep the lower-level stream recovery path for externally-supplied invalid `previousResponseId` values intact.

## Validation

- `make static-check`
- `bun test src/common/utils/ai/providerOptions.test.ts`
- `bun test src/browser/utils/messages/modelMessageTransform.test.ts`
- `bun test src/node/services/aiService.test.ts`
- `env TEST_INTEGRATION=1 bun x jest tests/ipc/providers/openaiPreviousResponseIdRecovery.test.ts --runInBand`
- One-time live GPT-5.4 large-context verification via temporary IPC integration test:
  - seeded ~300k estimated history tokens after a real GPT-5.4 seed turn
  - final request completed with `inputTokens: 240263`
  - response text was `PASS`
  - emitted `0` `auto-compaction-triggered` events

## Risks

Low to moderate for OpenAI Responses requests only. This intentionally changes how OpenAI conversation state is carried between turns, but the change is narrowly scoped to removing `previousResponseId` reuse when Mux already sends explicit history. The defensive stream-level recovery logic remains available if some other path still supplies an invalid response ID.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$8.16`_

### DIFF
--- a/src/browser/utils/messages/modelMessageTransform.test.ts
+++ b/src/browser/utils/messages/modelMessageTransform.test.ts
@@ -786,7 +786,7 @@ describe("modelMessageTransform", () => {
   });
 
   describe("reasoning part handling for OpenAI", () => {
-    it("should preserve reasoning parts for OpenAI provider (managed via previousResponseId)", () => {
+    it("should preserve reasoning parts for OpenAI provider (managed via explicit history)", () => {
       const messages: ModelMessage[] = [
         {
           role: "user",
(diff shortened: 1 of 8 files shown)


ammar-agent:
@codex review

PR #2834, opened 2026-03-06
🤖 fix: fast-fail task_await same-batch awaits

ThomasK33:
## Summary
- fast-fail explicit task_await requests when the requested IDs are invalid but real in-scope awaitable IDs already exist
- include concrete replacement IDs in the error so agents can recover on the next step without another blind retry
- document the same-batch sequencing rule next to the task and background bash tool definitions

## Background
Agents sometimes spawn `task` or background `bash` and then call `task_await` in the same parallel batch. Because the spawning tool result has not returned yet, the await call can contain hallucinated IDs and wastes a round-trip before self-correcting.

## Implementation
`task_await` now computes the current in-scope awaitable ID set once and reuses it for both omitted-`task_ids` behavior and the new sequencing guidance path. When an explicitly requested agent task ID is invalid and replacement IDs are already available, the tool returns a focused `error` result that explains the sequencing rule and lists the replacement IDs.

The targeted test suite now covers descendant-task suggestions, background bash suggestions, mixed valid plus hallucinated requests, and the no-suggestions fallback. Running `make fmt` also regenerated the tool hooks docs so the description copy stays in sync.

## Validation
- `bun test src/node/services/tools/task_await.test.ts`
- `make typecheck`
- `make fmt`
- `make static-check`

## Risks
This changes how invalid explicit agent task IDs are classified when replacement IDs are already visible, so the main regression risk is downstream code that assumes those cases always return `not_found` or `invalid_scope`. The blast radius is limited to `task_await` recovery behavior and its generated docs.

---
<details>

# Plan: Fast-fail same-batch `task_await` misuse (option 1, pragmatic)

## Context / Why
The bug is that agents sometimes emit `task` / background `bash` and `task_await` in one parallel batch, hallucinate task IDs for the await call, then self-correct only after a failed round-trip. There is no per-tool parallel exclusion today, and a literal `streamManager` batch validator would be too late to affect the model-facing tool result.

**Assumption for this plan:** implement the *spirit* of option 1 by making `task_await` fail fast with a clearer recovery message when the common misuse pattern is visible at execution time (invalid requested IDs + real awaitable IDs already in scope). Preserve current behavior when that signal is absent.

## Evidence
- `src/common/utils/tools/toolDefinitions.ts:1034-1049` — `task_await` already documents: do not call it in the same parallel batch as `task` or `bash`.
- `src/common/utils/ai/providerOptions.ts:151-171, 218-245` — parallel tool use is globally enabled for Anthropic/OpenAI, with no per-tool opt-out.
- `src/node/services/aiService.ts:1249-1311` and `src/node/services/streamManager.ts:1224-1255, 1636-1883` — tools go into `streamText()` before `streamManager` observes `tool-call` / `tool-result`, so `streamManager` cannot change what the model already received for the current step.
- `node_modules/@ai-sdk/provider-utils/src/types/tool.ts:16-20` — `ToolExecutionOptions.messages` excludes the assistant response that contained the tool calls, so a tool cannot directly inspect its sibling calls from the same step.
- `node_modules/ai/src/generate-text/run-tools-transformation.ts:282-318` — `onInputAvailable` fires per tool call and execution starts immediately afterward without waiting for other calls in the step, which makes exact same-step batch validation non-trivial with the current SDK loop.
- `src/node/services/tools/task_await.ts:41-61, 171-179` and `src/node/services/tools/task_await.test.ts:182-218` — the tool already computes in-scope awaitable IDs and already has targeted tests for hallucinated IDs, making it the narrowest low-risk edit point.

## Acceptance criteria
- If `task_await` receives invalid `task_ids` **and** there are real in-scope awaitable IDs available during that same execution, each invalid requested ID returns `status: "error"` with:
  - the sequencing rule,
  - guidance to wait for the spawning `task` / `bash` result first,
  - the concrete replacement IDs to use in a later `task_await` call.
- Valid requested IDs in the same call still keep their existing `completed` / `queued` / `running` / `awaiting_report` behavior.
- If no replacement IDs are available, existing `not_found` / `invalid_scope` behavior remains unchanged.
- `task` and background `bash` descriptions each add a short reminder not to pair same-batch spawn + await.

### Red — define the intended behavior with failing tests first (net +0 LoC product code)
1. **`src/node/services/tools/task_await.test.ts`**
   - Add `it("returns an error with descendant task suggestions for hallucinated IDs")`:
     - mock `listActiveDescendantAgentTaskIds()` to return `['real-child']`,
     - make the requested ID fail descendant validation,
     - assert `results[0]` is `{ status: 'error', taskId: 'hallucinated', error: /same parallel tool-call batch/ }`,
     - assert the error text includes `real-child`.
   - Add `it("returns an error with bash task suggestions for hallucinated IDs")`:
     - mock `backgroundProcessManager.list()` with one in-scope running process,
     - assert the suggested replacement contains `bash:<processId>`.
   - Add `it("preserves mixed results when one requested ID is real and one is hallucinated")`:
     - one requested ID completes normally,
     - the hallucinated peer returns `status: 'error'`,
     - lock down partial-success behavior in a single `results` array.
   - Keep one explicit control test for the old fallback: no suggested IDs available → `not_found` stays `not_found`.
2. Do **not** add brittle end-to-end or stream-manager tests for this change; the behavior lives inside `task_await`.

### Green — implement the smallest change that makes those tests pass (net +45–65 LoC product code)
1. **`src/node/services/tools/task_await.ts`**
   - Extract a small local helper that returns the deduped set of in-scope awaitable IDs:
     - active descendant agent task IDs,
     - running in-scope background bash task IDs.
   - Reuse that helper in two places:
     - existing omitted-`task_ids` behavior,
     - new suggestion-building for invalid requested IDs.
   - Add a focused helper to centralize the new error shape/message.
   - In the rejected-ID branch, change behavior only when both conditions hold:
     - the model explicitly supplied `task_ids`, and
     - the helper returns one or more real awaitable IDs.
   - Otherwise, keep the existing `not_found` / `invalid_scope` branches exactly as they are.

   ```ts
   function buildTaskAwaitSequencingError(taskId: string, suggestedIds: string[]) {
     return {
       status: "error" as const,
       taskId,
       error:
         "Do not call task_await in the same parallel tool-call batch as task or bash. " +
         "Wait for the spawning tool result first, then call task_await in a later step. " +
         `Real task IDs available now: ${suggestedIds.join(", ")}.`,
     };
   }

   ...

   if (!descendantAgentTaskIdSet.has(taskId)) {
     const lookup = rejectedAgentTaskStatuses.get(taskId);
     if (requestedIds && suggestedAwaitIds.length > 0) {
       return buildTaskAwaitSequencingError(taskId, suggestedAwaitIds);
     }

     return !lookup?.exists
       ? { status: "not_found" as const, taskId, activeTaskIds }
       : { status: "invalid_scope" as const, taskId, activeTaskIds };
   }
   ```

2. **`src/common/utils/tools/toolDefinitions.ts`**
   - Add one sentence to `task` and background `bash` descriptions so the sequencing rule is visible next to the spawning tools too.

   ```ts
   "Use run_in_background: true when launching multiple tasks in parallel so you can await them as a batch. " +
   "Do not call task_await in the same parallel tool-call batch; wait for the returned taskId(s) first. "
   ```

### Refactor — simplify after behavior is correct (net +0–10 LoC product code)
1. Keep the helper for “in-scope awaitable IDs” local to `task_await.ts`; do not promote it unless a third callsite appears.
2. Keep the sequencing error text in one helper so agent-task and bash-task suggestion paths cannot drift.
3. Re-read the tests and rename any that describe implementation details instead of user-visible behavior.

## Verification
- `bun test src/node/services/tools/task_await.test.ts`
- If a lightweight assertion is added for the description copy, run only that smallest matching Bun test file as a follow-up.

<details>
<summary>Why this plan is pragmatic rather than exact same-step detection</summary>

With the current AI SDK loop, tools do not get a complete sibling-tool list for the in-flight step before execution, and `streamManager` is downstream of the model-facing tool result. This plan therefore targets the common bad outcome — hallucinated await IDs while real awaitable IDs already exist — without disabling valid parallel tool use globally.

If you want a stricter follow-up later (exact per-step sibling-call detection), that is a separate, larger design involving custom tool wrappers / shared execution context around the SDK loop rather than a narrow `task_await` change.
</details>

</details>

---
_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.4` • Thinking: `high` • Cost: `$8.12`_


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
I resolved the latest P2 thread as a stale read of the current head. On `201a995`, explicit `task_ids` take the fast path:

```ts
const uniqueTaskIds = requestedIds
  ? dedupeStrings(requestedIds)
  : await listInScopeAwaitableTaskIds();
```

That means `backgroundProcessManager.list()` only runs when `task_ids` is omitted, or later when building replacement suggestions for rejected IDs. I also added regression coverage for the explicit-await fast path and for suggestion fallback when background discovery fails.

@codex review

PR #2835, opened 2026-03-06
🤖 fix: align agent_skill_list with workspace-visible skills

ThomasK33:
Summary
Broaden `agent_skill_list` so it returns the effective workspace-visible skill set instead of only scanning global skills. This aligns the tool with the UI, `agent_skill_read`, and the shared discovery service that already merges project, global, and built-in skills with shadowing.

Background
Mux already teaches users about the effective skill set visible in a workspace, but `agent_skill_list` still only scanned `~/.mux/skills`. That made the list tool disagree with the product behavior and with `agent_skill_read`, especially when project-local skills shadowed global skills.

Implementation
- replaced the global-only `agent_skill_list` scanner with `discoverAgentSkills(config.runtime, config.cwd)`
- preserved default filtering for `advertise: false` and applied it to the effective winner list across scopes
- added a defensive misconfiguration guard when `cwd` is missing
- rewrote the tool tests around multi-scope listing, precedence, hidden-skill filtering, and misconfiguration handling
- updated Chat With Mux prompt/docs wording so list/read refers to available workspace skills while write/delete remains global-only
- regenerated the built-in agent content bundle after updating the `mux.md` source

Validation
- `make static-check`
- `bun test src/node/services/tools/agent_skill_list.test.ts`
- `bun test src/node/services/agentSkills/agentSkillsService.test.ts`
- `make typecheck`
- `bun x eslint src/node/services/tools/agent_skill_list.ts src/node/services/tools/agent_skill_list.test.ts src/common/utils/tools/toolDefinitions.ts`
- `bun x prettier --check src/node/services/tools/agent_skill_list.ts src/node/services/tools/agent_skill_list.test.ts src/common/utils/tools/toolDefinitions.ts src/node/builtinAgents/mux.md docs/agents/index.mdx src/node/services/agentDefinitions/builtInAgentContent.generated.ts`

Risks
This changes the contract of `agent_skill_list`, so any callers that implicitly assumed a global-only result may need to look at the returned `scope` field. The response shape stays flat and existing precedence comes from the already-tested shared discovery service, which keeps regression risk moderate and localized to skill tooling.

---

<details>

# Align `agent_skill_list` with workspace-visible skills

## Context / Why
`agent_skill_list` currently looks broken in workspaces that visibly have skills, because the UI and `agent_skill_read` already surface the effective skill set across project, global, and built-in scopes while `agent_skill_list` only scans `~/.mux/skills`. The user wants the existing tool broadened rather than replaced so it matches what Mux already teaches users: “available skills” are the skills visible in the current workspace. The change should preserve existing shadowing precedence and keep write/delete semantics global-only.

## Evidence
- `src/node/services/tools/agent_skill_list.ts` manually resolves mux home from `workspaceSessionDir`, scans only `<muxHome>/skills`, hardcodes `scope: "global"`, and filters `advertise: false`.
- `src/common/orpc/schemas/agentSkill.ts` already allows `scope: "project" | "global" | "built-in"`, so no response-schema change is needed.
- `src/node/services/agentSkills/agentSkillsService.ts` exposes `discoverAgentSkills(runtime, workspacePath)`, which already merges `.mux/skills`, `.agents/skills`, `~/.mux/skills`, `~/.agents/skills`, and built-ins with project > global > built-in precedence.
- `src/node/services/streamContextBuilder.ts` already uses `discoverAgentSkills(...)` to populate `availableSkills` for tool descriptions, so the model/UI context already treats multi-scope discovery as the source of truth.
- `src/node/services/tools/agent_skill_read.ts` already shows skills with `(scope: ...)`, so the model already reasons about multi-scope descriptors.
- `src/node/services/tools/agent_skill_list.test.ts` locks in the outdated global-only behavior and must be rewritten to define the broadened contract.
- `src/node/services/agentSkills/agentSkillsService.test.ts` already covers precedence, project `.agents/skills`, built-in fallback, and shadowing, so `agent_skill_list` should reuse that behavior instead of re-implementing discovery.
- `src/node/builtinAgents/mux.md` and `docs/agents/index.mdx` still frame skill listing as global-only, so prompt/docs text must be updated for consistency.
- User-provided UI screenshots show project, global, and built-in skills together; that is the product behavior the tool should align with.

## Definition of done
- `agent_skill_list` returns the **effective** available skills for the current workspace by reusing `discoverAgentSkills(...)`.
- Returned skills may have `scope: "project"`, `"global"`, or `"built-in"`; no new tool is introduced.
- Shadowed duplicates are not returned twice; the same “winner” used by `agent_skill_read`/UI is returned.
- `includeUnadvertised` keeps its current meaning: default hides `advertise: false`, opt-in includes them.
- The response remains a flat array of descriptors, sorted by the discovery helper’s current name ordering; this patch does **not** redesign the output into UI-style grouped sections.
- Prompt/docs text clearly distinguishes:
  - **list/read** → available skills in the current workspace
  - **write/delete** → global skill management only
- Tests cover the broadened behavior before implementation and still pass afterward.

### Phase 1 — Red: define the new contract with failing tests first
1. **Rewrite `src/node/services/tools/agent_skill_list.test.ts` around multi-scope discovery.**
   - Replace the current “lists global skills from `~/.mux/skills`” test with a workspace-scoped discovery test.
   - Keep the tests narrow: validate the list tool contract, not the full discovery service internals already covered elsewhere.

2. **Add concrete test cases that define the expected behavior.**
   - **Happy path / merged listing:** returns project, global, and built-in skills in one flat response.
   - **Shadowing:** if the same skill name exists in project and global scope, only the project descriptor is returned.
   - **Hidden skills default:** unadvertised skills are excluded by default even if they exist in project/global scope.
   - **Hidden skills opt-in:** `includeUnadvertised: true` includes hidden winners across scopes.
   - **Misconfiguration guard:** if `config.cwd` is missing/empty, the tool returns a clear configuration error instead of surfacing a generic thrown exception.

3. **Use hermetic test setup so `~/.mux` resolution does not leak to the real home directory.**
   - Set `process.env.MUX_ROOT` to a temp directory for each test (or a shared helper with restore in `finally`) because local `~/.mux` expansion is routed through `getMuxHome()`.
   - Create project-local skills under `<project>/.mux/skills` or `<project>/.agents/skills` and global skills under `<MUX_ROOT>/skills`.
   - Keep helper functions local to the test file unless they become broadly reusable.

   ```ts
   test("lists effective available skills across scopes", async () => {
     process.env.MUX_ROOT = muxHome.path;

     await writeProjectSkill(project.path, "foo", "from project");
     await writeGlobalSkill(muxHome.path, "foo", { description: "from global" });
     await writeGlobalSkill(muxHome.path, "bar", { description: "global only" });

     const result = (await tool.execute!({}, mockToolCallOptions)) as AgentSkillListToolResult;

     expect(result.success).toBe(true);
     if (result.success) {
       expect(result.skills.map((skill) => [skill.name, skill.scope])).toEqual([
         ["bar", "global"],
         ["foo", "project"],
         ["init", "built-in"],
         ["mux-diagram", "built-in"],
         ["mux-docs", "built-in"],
       ]);
     }
   });
   ```

4. **Lean on existing discovery-service tests rather than duplicating them.**
   - Continue using `src/node/services/agentSkills/agentSkillsService.test.ts` as evidence that precedence and built-in fallback are already correct.
   - The list tool tests should prove that `agent_skill_list` now delegates to that discovery behavior.

### Phase 2 — Green: implement the minimal change to satisfy the new tests
1. **Replace the bespoke global-only scanner in `src/node/services/tools/agent_skill_list.ts` with a thin wrapper around `discoverAgentSkills(config.runtime, config.cwd)`.**
   - Validate `config.cwd` before calling discovery.
   - Filter `advertise: false` only when `includeUnadvertised !== true`.
   - Preserve the existing `AgentSkillListToolResult` shape.
   - Update the error text from “global skills” to “available skills”.
   - Remove no-longer-needed helpers/imports (`readGlobalSkillDescriptor`, `listSkillDirectories`, `getMuxHomeFromWorkspaceSessionDir`, `fs/promises`, `path`, local SKILL parsing code).

   ```ts
   export const createAgentSkillListTool: ToolFactory = (config) => {
     return tool({
       description: TOOL_DEFINITIONS.agent_skill_list.description,
       inputSchema: TOOL_DEFINITIONS.agent_skill_list.schema,
       execute: async ({ includeUnadvertised }): Promise<AgentSkillListToolResult> => {
         if (!config.cwd) {
           return { success: false, error: "Tool misconfigured: cwd is required." };
         }

         try {
           const discovered = await discoverAgentSkills(config.runtime, config.cwd);
           return {
             success: true,
             skills:
               includeUnadvertised === true
                 ? discovered
                 : discovered.filter((skill) => skill.advertise !== false),
           };
         } catch (error) {
           return {
             success: false,
             error: `Failed to list available skills: ${getErrorMessage(error)}`,
           };
         }
       },
     });
   };
   ```

2. **Update `src/common/utils/tools/toolDefinitions.ts` to describe the broadened tool accurately.**
   - Reword the description from “List global skills...” to “List available skills in the current workspace...”
   - Explicitly note that returned skills may come from project, global, or built-in scope and that `scope` identifies the origin.
   - Keep the input schema unchanged for this patch.

3. **Update prompt/docs text that currently encode the outdated global-only meaning.**
   - `src/node/builtinAgents/mux.md`
   - `docs/agents/index.mdx`
   - Adjust wording so the system assistant describes two distinct capabilities:
     - reading/listing available skills across scopes
     - creating/updating/deleting global skills only

   Suggested wording shape:
   ```md
   - **Skills**: Read and list available project, global, and built-in skills for the current workspace
   - **Global skills**: Create, update, and delete global skills (`~/.mux/skills/`)
   ```

### Phase 3 — Refactor: simplify after the behavior is correct
1. **Prune obsolete code paths from `agent_skill_list.ts`.**
   - Remove global-only helper names and comments so the file no longer implies host-local global scanning.
   - Keep the tool as a small adapter over the shared discovery service.

2. **Tighten test helpers only if the duplication is justified.**
   - If project/global skill fixture setup repeats 3+ times with the same structure, extract a small local helper in the test file.
   - Otherwise, prefer a little duplication over creating a premature shared abstraction.

3. **Normalize wording everywhere touched by the patch.**
   - “available skills” for multi-scope listing
   - “global skills” only for create/update/delete flows
   - This prevents stale mental models from surviving in prompts, tests, or comments.

## Verification
- `bun test src/node/services/tools/agent_skill_list.test.ts`
- `bun test src/node/services/agentSkills/agentSkillsService.test.ts`
- `make typecheck`
- Manual sanity check (optional but useful): invoke `agent_skill_list` in a workspace with project + global + built-in skills and confirm the membership/scope matches the skill panel, even though the tool response remains a flat list rather than UI-grouped sections.

## Assumptions to preserve unless the user explicitly broadens scope
- **No new tool:** broaden `agent_skill_list` instead of introducing `agent_skill_discover` or similar.
- **No write/delete expansion:** `agent_skill_write` and `agent_skill_delete` remain global-only.
- **No output-shape redesign:** keep the flat descriptor array; consumers can group by `scope` if they want UI-style presentation.
- **No hidden-skill policy change beyond scope coverage:** the only functional change is which skills are eligible to be listed; default hidden-skill filtering remains intact.

<details>
<summary>Alternative considered — keep <code>agent_skill_list</code> global-only and add a new discovery tool (not recommended; net product LoC estimate: +80 to +140)</summary>

This would preserve the old meaning of `agent_skill_list`, but it would also preserve the inconsistency the user is flagging:
- the UI would keep showing all available skills,
- `agent_skill_read` would keep reading across scopes,
- `agent_skill_list` would remain a special-case global management tool.

That split adds cognitive overhead for little gain because the schema already supports multi-scope descriptors and `discoverAgentSkills(...)` already exists.
</details>

<details>
<summary>Risk notes / edge cases to watch while implementing</summary>

- `discoverAgentSkills(...)` requires a real workspace path, so the tool should guard `config.cwd` explicitly and tests should define the expected failure mode.
- Switching from host-local fs reads to runtime-based discovery is a feature: it makes `agent_skill_list` consistent with SSH/runtime-backed workspaces instead of assuming local access to the global mux directory.
- The UI groups skills by scope, but `discoverAgentSkills(...)` returns a flat, name-sorted array. This plan intentionally keeps the existing flat tool result and relies on `scope` for grouping.
- If exact UI parity for default visibility of hidden skills becomes desirable later, treat that as a separate follow-up so this patch changes only one dimension of behavior at a time.
</details>

</details>

---
_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `high` • Cost: `$6.06`_


ThomasK33:
@codex review

PR #2979, opened 2026-03-16
fix: enforce SSH2 host key verification via known_hosts

ThomasK33:
### Motivation

- The SSH2 transport path previously used a permissive verifier (`hostVerifier: () => true`), which allowed any server key and exposed connections to MITM attacks. 
- Enforce host key pinning for the SSH2 runtime by consulting the user's `known_hosts` so SSH2 does not silently accept attacker-controlled keys.

### Description

- Added parsing of `~/.ssh/known_hosts` and support for plain entries, bracketed `[host]:port` entries, and hashed hostnames (`|1|...`) in `src/node/runtime/SSH2ConnectionPool.ts` via a new `getKnownHostPublicKeys` helper. 
- Implemented `matchesHashedKnownHost` to verify hashed known_hosts entries using HMAC-SHA1 and wired it into `matchesKnownHostPattern` and the loader. 
- Loaded allowed host public keys prior to connect and replaced the permissive verifier with an enforced `hostVerifier` that only accepts server keys that exactly match an entry from `known_hosts`. 
- Kept change scoped to the SSH2 connection pool so transport selection behavior is unchanged.

### Testing

- Attempted to run `make typecheck`, but the environment lacks Bun so `make typecheck` failed with `bun: command not found`. 
- No automated unit/integration tests were executed in this environment due to dependency/tooling limitations, so CI/typecheck should be run by the maintainer or CI to validate types and build.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69b4105a4044832ea63556a30f5955c7)

### DIFF
--- a/src/node/runtime/SSH2ConnectionPool.ts
+++ b/src/node/runtime/SSH2ConnectionPool.ts
@@ -158,16 +158,65 @@ function matchesHashedKnownHost(pattern: string, host: string): boolean {
   return actual === expected;
 }
 
+function wildcardPatternToRegex(pattern: string): RegExp {
+  let regex = "^";
+  for (const char of pattern) {
+    if (char === "*") {
+      regex += ".*";
+      continue;
+    }
+
+    if (char === "?") {
+      regex += ".";
+      continue;
+    }
+
+    if (/[\\^$+?.()|{}\[\]]/.test(char)) {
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

PR #3139, opened 2026-04-06
🤖 fix: preserve init logs across reconnect replay

ammar-agent:
## Summary

Keep the existing `workspace-init` transcript row visible while reconnect replay catches up, and treat init replay like stream replay so live SSH/setup output cannot overtake or blank the snapshot.

## Background

Users could still lose the visible init log when switching away from a workspace and coming back mid-init. The remaining gap was that reconnect replay re-emitted `init-start` before the snapshot lines, which reset the renderer's synthetic init row to empty. Init events also lacked replay metadata, so live init output could interleave with replay in a way the client could not order deterministically.

## Implementation

- add `replay: true` to replayed init events emitted by `InitStateManager`
- buffer live init events in `replayBufferedStreamMessageRelay` until replay finishes, matching the existing stream replay isolation
- keep matching reconnect replay buffered in `WorkspaceStore` when the same running init row is already visible, so the UI does not clear the SSH/setup log before `caught-up`
- keep the fresh-load behavior from the earlier fix: init replay still renders immediately when there is no existing running init row to preserve
- add regressions for the store reconnect path and the replay relay

## Validation

- `bun test src/node/services/initStateManager.test.ts`
- `bun test src/node/orpc/replayBufferedStreamMessageRelay.test.ts`
- `make static-check`

## Risks

Low-to-moderate. The change is scoped to reconnect-time init replay handling, but it touches the shared replay relay and buffered-event path. Regressions would most likely affect reconnect ordering for init rows rather than normal chat history or live streaming.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$36.68`_

### DIFF
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it, beforeEach, afterEach, mock, type Mock } from "bun:test";
+import type { DisplayedMessage } from "@/common/types/message";
 import type { FrontendWorkspaceMetadata } from "@/common/types/workspace";
 import type { StreamStartEvent, ToolCallStartEvent } from "@/common/types/stream";
 import type { WorkspaceActivitySnapshot, WorkspaceChatMessage } from "@/common/orpc/types";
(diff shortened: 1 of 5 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Rebased on latest main. Please take another look.

ammar-agent:
@codex review

This updates the branch to preserve the existing workspace-init row during reconnect replay, marks replayed init events explicitly, and buffers live init events through the replay relay so init ordering matches stream replay.

PR #3150, opened 2026-04-09
🤖 fix: extend the streaming status debounce to 2s

ammar-agent:
## Summary

Increase the chat transcript streaming-status debounce from 200ms to 2 seconds so the barrier text stays calmer during fast startup-to-streaming handoffs.

## Background

PR #3147 introduced a small debounce to smooth status-label churn in the transcript tail. This follow-up intentionally makes that hold much longer so the `streaming...` area is noticeably less flashy when stages advance in quick succession.

## Implementation

- changed `STREAMING_STATUS_TRANSITION_DEBOUNCE_MS` from `200` to `2000`
- updated the component test constant so the debounce assertions still verify the intended hold window

## Validation

- `bun test src/browser/features/Messages/ChatBarrier/StreamingBarrier.test.tsx`
- `make static-check`

## Risks

Low-to-moderate UX risk. The implementation is unchanged, but users will now see startup/streaming labels lag by up to 2 seconds during phase handoffs, so the transcript may feel calmer at the cost of more stale status text.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$7.79`_

### DIFF
--- a/src/browser/features/Messages/ChatBarrier/StreamingBarrier.test.tsx
+++ b/src/browser/features/Messages/ChatBarrier/StreamingBarrier.test.tsx
@@ -37,7 +37,7 @@ function createWorkspaceState(overrides: Partial<MockWorkspaceState> = {}): Mock
   return state;
 }
 
-const STREAMING_STATUS_TRANSITION_DEBOUNCE_MS = 200;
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3151, opened 2026-04-09
🤖 fix: restore the open chat on hard refresh

ammar-agent:
## Summary

This follow-up to #3141 keeps an open workspace chat on hard refresh and also simplifies the startup routing architecture behind that decision. Browser and standalone startup now classify `/workspace/:id` restores from the actual navigation type instead of relying on session-storage bookkeeping spread across initialization paths.

## Background

#3141 intentionally kept browser fresh-launch behavior from blindly reopening a stale `/workspace/:id` URL unless the user explicitly chose **last workspace**. That was still correct for true first loads, but hard refresh got lumped into the same bucket and could bounce an open chat back to the project/new-workspace page.

The first fix for this used per-session flags, which worked for reloads but added more startup coordination state to an already subtle routing area. This revision simplifies that architecture by making RouterContext answer one clearer question on startup: **is this a restore-style navigation or a fresh launch?**

## Implementation

- add `getStartupNavigationType()` and `shouldRestoreWorkspaceUrlOnStartup()` in `RouterContext`
- use Navigation Timing (with a legacy `performance.navigation` fallback) to distinguish `reload` / `back_forward` restores from fresh `navigate` loads
- remove browser/PWA session-storage startup markers and the `RouterProvider` effect that existed only to maintain them
- keep the desktop `file://` restore path and launch-behavior handling unchanged
- update router and workspace-context tests to stub navigation type directly so the startup policy stays explicit and easier to reason about

## Validation

- `bun test src/browser/contexts/RouterContext.test.tsx`
- `bun test src/browser/contexts/WorkspaceContext.test.tsx`
- `make static-check`

## Risks

This changes browser/PWA startup route restoration logic. A regression here would affect whether `/workspace/:id` is treated as a fresh launch or a restore on page load, so the updated tests cover both cold-launch and reload behavior.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$18.76`_

### DIFF
--- a/src/browser/contexts/RouterContext.test.tsx
+++ b/src/browser/contexts/RouterContext.test.tsx
@@ -25,7 +25,12 @@ function createMatchMedia(isStandalone = false): typeof window.matchMedia {
     }) satisfies MediaQueryList) as typeof window.matchMedia;
 }
 
-function installWindow(url: string, options?: { isStandalone?: boolean }) {
+type NavigationType = "navigate" | "reload" | "back_forward" | "prerender";
+
+function installWindow(
+  url: string,
+  options?: { isStandalone?: boolean; navigationType?: NavigationType }
+) {
   // Happy DOM can default to an opaque origin ("null") which breaks URL-based
   // logic in RouterContext. Give it a stable origin.
   const happyWindow = new GlobalWindow({ url });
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

PR #3152, opened 2026-04-09
🤖 fix: prevent transcript flashes and tearing during chat hydration

ammar-agent:
## Summary

This PR fixes the remaining chat handoff artifacts around startup, workspace switches, and reopen/hydration. New chats keep their explicit starting barrier instead of flashing generic empty/loading placeholders; the chat pane stays mounted through hydration in both browser and Electron paths; the bottom composer/decorations stack holds its height while workspace-specific banners catch up; and mock streaming now persists/replays partial assistant state without letting stale cleanup or stale delayed writes clobber replacement streams.

## Background

Workspace creation navigates into the new workspace as soon as the workspace exists, but the first `sendMessage` and chat subscription replay can land a moment later. During that gap, brand-new chats could briefly flash `Catching up with the agent...` or `No Messages Yet` before the first turn appeared.

Separately, switching to or reopening an existing workspace could still swap back to a loading shell, remount the transcript viewport, or let workspace-specific footer banners collapse and re-expand during hydration. Those transitions showed up as the remaining vertical tear, especially in Electron `make dev`.

Mock mode had its own reconnect/resume gaps: partial assistant state needed to stay on disk during streaming so reopened workspaces could recover the in-flight turn, but stale stop/replacement paths could still delete or recreate the wrong partial snapshot.

## Implementation

- move optimistic new-chat startup state into `StreamingMessageAggregator` and clear it from replay/terminal paths instead of a separate transient UI path
- keep generic empty/loading placeholders suppressed while `isStreamStarting` is true
- keep `WorkspaceShell` and `ChatPane` mounted during transcript hydration so opening a workspace no longer drops the composer; apply that hydration behavior in Electron as well as browser mode
- stop remounting `ChatPane` on workspace switches and make deferred transcript snapshots workspace-aware so stale rows and viewport tears cannot flash between chats
- add `HydrationStablePane` around the chat input/decorations stack so workspace-specific warnings, queued banners, and review chips hold their last measured height until hydration completes
- keep mock-mode partial assistant state on disk during streaming for reconnect/reopen flows, but harden stop/replacement races by:
  - waiting for prior partial cleanup before replacement streams start
  - ignoring stale stream-error/stream-end cleanup from replaced streams
  - deleting stale delayed partial writes if a stream was cancelled after the pre-write ownership check

## Validation

- `make static-check`
- `bun test ./src/browser/components/WorkspaceShell/WorkspaceShell.test.tsx`
- `bun test ./src/browser/components/ChatPane/HydrationStablePane.test.tsx`
- `bun test ./src/browser/stores/WorkspaceStore.test.ts`
- `bun test ./src/browser/utils/messages/StreamingMessageAggregator.test.ts`
- `bun test ./src/browser/utils/messages/messageUtils.test.ts`
- `bun test ./src/browser/features/ChatInput/useCreationWorkspace.test.tsx`
- `bun test ./src/browser/hooks/useAutoScroll.test.tsx`
- `bun test ./src/node/services/mock/mockAiStreamPlayer.test.ts`
- `TEST_INTEGRATION=1 bun x jest tests/ui/chat/newChatStreamingFlash.test.ts --runInBand --silent`
- `TEST_INTEGRATION=1 bun x jest tests/ui/chat/bottomLayoutShift.test.ts --runInBand --silent`
- `TEST_INTEGRATION=1 bun x jest tests/ui/tasks/awaitVisualization.test.ts --runInBand --silent`
- `bun scripts/reproWorkspaceSwitchTearWeb.ts`

## Risks

The main regression risk is over-preserving old UI state during workspace switches or hydration. Coverage now explicitly exercises new-chat startup, web and Electron shell hydration behavior, bottom composer height stability, deferred snapshot cross-workspace isolation, and mock partial cleanup races around stop, replacement, and delayed writes.

## Pains

The visible tear turned out to be a stack of smaller handoff issues rather than one bug: new-chat startup state, shell-level hydration placeholders, transcript remounting, deferred snapshot lag, and the workspace-specific composer decoration stack each contributed a different artifact. Review also flushed out several mock partial races that were easy to miss without targeted timing tests.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$286.28`_

### DIFF
--- a/scripts/reproWorkspaceSwitchTearWeb.ts
+++ b/scripts/reproWorkspaceSwitchTearWeb.ts
@@ -5,13 +5,15 @@
  *
  * Why this exists:
  * - The remaining artifact was reported in the web/dev-server path, not Electron.
- * - The visible tear appears to already contain the target workspace transcript.
+ * - The visible tear can show up either as a transcript shift or as the composer briefly
+ *   disappearing while the target workspace opens.
  *
  * What it does:
  * 1. Boots an isolated `make dev-server` instance with a temporary MUX_ROOT.
  * 2. Creates two real workspaces in the browser app and sends live mock-chat turns.
- * 3. Switches between them while capturing transcript screenshots.
- * 4. Exits with code 1 when the target transcript keeps shifting after it is visible.
+ * 3. Replays both seen->seen switches and reload->unseen switches while sampling layout.
+ * 4. Exits with code 1 when the target transcript shifts after it is visible or when the
+ *    composer disappears during a workspace open.
  */
 import fs from "fs/promises";
 import net from "net";
@@ -39,6 +41,18 @@ interface SwitchFrameSample {
   chatInputHeight: number | null;
   imagePath: string;
   png: Buffer;
+}
+
+interface OpenTransitionFrameSample {
+  frame: number;
+  timestamp: number;
+  hasInput: boolean;
+  chatInputHeight: number | null;
+  hasMessageWindow: boolean;
+  messageWindowHeight: number | null;
+  containsTargetMarker: boolean;
+  loadingWorkspace: boolean;
+  loadingTranscript: boolean;
 }
 
 function buildMarker(label: string): string {
@@ -185,6 +199,65 @@ async function waitForMockResponse(page: Page, marker: string): Promise<void> {
   );
 }
 
+async function captureOpenTransition(args: {
+  page: Page;
+  clickWorkspaceId: string;
+  targetMarker: string;
+}): Promise<OpenTransitionFrameSample[]> {
+  const row = args.page.locator(
+    `[data-workspace-id="${args.clickWorkspaceId}"][data-workspace-path]`
+  );
+  await row.waitFor({ state: "visible", timeout: 60_000 });
+  await row.scrollIntoViewIfNeeded();
+  await args.page.evaluate((targetMarker: string) => {
+    (
+      window as Window & {
+        __muxOpenTransitionFramesPromise?: Promise<OpenTransitionFrameSample[]>;
+      }
+    ).__muxOpenTransitionFramesPromise = new Promise((resolve) => {
+      const frames: OpenTransitionFrameSample[] = [];
+      let frame = 0;
+      const step = () => {
+        const inputSection = document.querySelector(
+          '[data-component="ChatInputSection"]'
+        ) as HTMLElement | null;
+        const messageWindow = document.querySelector(
+          '[data-testid="message-window"]'
+        ) as HTMLElement | null;
+        const bodyText = document.body.textContent ?? "";
+        frames.push({
+          frame,
+          timestamp: performance.now(),
+          hasInput: inputSection !== null,
+          chatInputHeight: inputSection?.getBoundingClientRect().height ?? null,
+          hasMessageWindow: messageWindow !== null,
+          messageWindowHeight: messageWindow?.getBoundingClientRect().height ?? null,
+          containsTargetMarker: bodyText.includes(targetMarker),
+          loadingWorkspace: bodyText.includes("Loading workspace..."),
+          loadingTranscript: bodyText.includes("Loading transcript..."),
+        });
+        frame += 1;
+        if (frame < 20) {
+          requestAnimationFrame(step);
+        } else {
+          resolve(frames);
+        }
+      };
+      requestAnimationFrame(step);
+    });
+  }, args.targetMarker);
+  await row.dispatchEvent("click");
+  return await args.page.evaluate(() => {
+    return (
+      (
+        window as Window & {
+          __muxOpenTransitionFramesPromise?: Promise<OpenTransitionFrameSample[]>;
+        }
+      ).__muxOpenTransitionFramesPromise ?? Promise.resolve([])
+    );
+  });
+}
+
 async function captureSwitch(args: {
   page: Page;
   sourceMarker: string;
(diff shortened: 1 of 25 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the integration failure by making the new regression test compatible with the Jest-based integration runner as well as Bun.

ammar-agent:
@codex review

Restricted the optimistic pending-start flag to auto-navigated creations and added hook coverage for the background-creation path.

ammar-agent:
@codex review

Kept the optimistic pending-start flag through buffered first-turn replay until `caught-up` applies the history, and added a WorkspaceStore regression test for that timing.

ammar-agent:
@codex review

Cleared stale optimistic start state when replay/activity confirms no active stream, and added WorkspaceStore coverage for both buffered replay and no-stream catch-up paths.

ammar-agent:
@codex review

Stopped clearing the optimistic-start flag on recency-only non-streaming activity updates, and added WorkspaceStore coverage for that exact startup gap.

ammar-agent:
@codex review

Kept the optimistic-start flag across caught-up-without-first-turn, while still clearing it on buffered/live first-turn observation and on definitive background stop.

ammar-agent:
@codex review

Preserved optimistic initial-send startup across full replay resets and added a direct WorkspaceStore regression test for that reset path.

ammar-agent:
@codex review

Cleared optimistic pending-start on pre-stream abort as well, with a targeted WorkspaceStore regression test for the interrupted-first-send path.

ammar-agent:
@codex review

Rewrote the new-chat startup fix so the optimistic pending-start state lives in `StreamingMessageAggregator` instead of a separate `WorkspaceStore` transient. Local validation is green (`make static-check` plus the focused startup regressions).

ammar-agent:
@codex review

Follow-up fix for the remaining workspace-switch transcript flash:
- keep cached transcript content visible during web hydration when switching between existing chats
- retain the new-chat optimistic-start barrier behavior from the earlier rewrite
- added `WorkspaceShell` regression coverage and re-ran targeted tests plus `make static-check`

ammar-agent:
@codex review

Rebased on `main`, resolved the two `WorkspaceStore.ts` conflicts, reran the targeted regression suite plus `make static-check`, and force-pushed the branch.

ammar-agent:
@codex review

Follow-up fix for the remaining workspace-switch vertical tear:
- keep `ChatPane` mounted across workspace switches instead of remounting the whole transcript viewport
- reset the per-workspace local UI state inside `ChatPane` that previously piggybacked on the root remount
- re-arm transcript tail ownership on workspace changes so the incoming chat does not inherit stale scroll state
- validated with `WorkspaceShell`, `bottomLayoutShift`, `newChatStreamingFlash`, and `make static-check`

ammar-agent:
@codex review

Follow-up fix for another likely workspace-switch artifact:
- make deferred transcript rendering workspace-aware so a chat switch cannot briefly render a stale deferred snapshot from the previous workspace
- added a `messageUtils` regression for cross-workspace deferred snapshot bypass
- reran `messageUtils`, `WorkspaceShell`, `bottomLayoutShift`, `newChatStreamingFlash`, and `make static-check`

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

Follow-up fix: keep the workspace shell mounted during browser transcript hydration so the composer no longer disappears on first workspace open after a reload. The browser repro script now covers that case too.

ammar-agent:
@codex review

Addressed the integration timing failure after the hydration-shell fix. The `task_await` UI test now waits for the hydrated tool call row before asserting the executing-state visualization.

ammar-agent:
For the outstanding App.tsx note: `setSelectedWorkspace` here comes from `WorkspaceContext`, not from a raw React `useState` setter. That wrapper resolves functional updates immediately against `selectedWorkspaceRef.current` and synchronously updates the ref before navigation, so the updater body (including `markPendingInitialSend`) cannot run later than the send-failure cleanup path.

ammar-agent:
@codex review

The outstanding thread is resolved: the App.tsx updater is the synchronous `WorkspaceContext` selection wrapper, so the optimistic pending-start mark cannot land after a later send-failure cleanup. The new browser hydration fix plus the integration timing follow-up are both pushed.

ammar-agent:
@codex review

Follow-up updates pushed:
- stabilized the `task_await` integration check so it waits for the hydrated tool row
- moved the flaky transcript scroll-ownership assertions into deterministic `useAutoScroll` hook tests while keeping the layout-pin integration coverage in place

ammar-agent:
@codex review

Follow-up pushed for the pending-start selection thread:
- the optimistic pending-start store write now happens outside the `setSelectedWorkspace` updater callback
- it is still gated on whether the creation actually won selection, using the synchronous `WorkspaceContext` selection wrapper

ammar-agent:
@codex review

Addressed the mock-stream replacement race:
- replacement `play()` now awaits the old stream's partial cleanup before the new stream can start writing
- added a regression that delays `deletePartial()` and verifies the replacement stream keeps its partial snapshot

ammar-agent:
@codex review

Addressed the stale mock stream error race:
- the `stream-error` path now bails out after `deletePartial()` if the active stream was cancelled or replaced while awaiting cleanup
- added a regression that delays the first partial cleanup, starts a replacement stream, and verifies no stale error is emitted

ammar-agent:
@codex review

Addressed the remaining mock stop/replace race:
- `MockAiStreamPlayer.stop()` is now async and `AIService.stopStream()` awaits it in mock mode
- direct `stop()` callers in tests now await the cleanup path
- the mock replacement/stream-error regressions still pass locally along with `make static-check`

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the stale delayed partial write race in MockAiStreamPlayer and refreshed the PR summary/body to match the full diff. Please take another look.

ammar-agent:
@codex review

Addressed the abort-during-replacement-cleanup race in MockAiStreamPlayer and revalidated the focused mock/compaction coverage plus static-check. Please take another look.

ammar-agent:
@codex review

Addressed the stale delayed-write cleanup race by making partial deletion message-aware, and revalidated the mock stream tests, compaction integration test, and static-check. Please take another look.

ammario (approved):

PR #3154, opened 2026-04-10
🤖 fix: make SSH sync snapshots branch-authoritative

ethanndickson:
## Summary

SSH workspace sync now treats branch refs as the authoritative snapshot contract and treats tags as shared metadata, so tag-only drift no longer forces a resync or deletes remote-only tags during later syncs.

## Background

The shared SSH base repo was previously using heads and tags for snapshot digests, reuse manifests, and prune behavior. That meant incidental tag drift could invalidate snapshot reuse, and a later sync from an arbitrary local clone could delete remote-only tags even though workspace materialization only depends on branch refs.

## Implementation

Native OpenSSH sync now pushes pruneable `refs/mux-bundle/*` branches separately from non-pruneable `refs/tags/*` metadata tags. The bundle import path mirrors that contract with a pruneable branch fetch plus a separate non-pruning tag fetch. Snapshot digests and local/remote sync manifests now compare heads only, so tag-only drift is ignored for reuse decisions. The new unit coverage checks the branch-only digest contract plus the split push/fetch argument shape, and the SSH integration regression proves a remote-only tag survives both snapshot reuse and a later branch-driven resync.

## Validation

I verified the new contract with focused SSHRuntime unit coverage, repo lint and typecheck, and targeted SSH integration scenarios that exercise both the existing stale-remote-ref regression and the new remote-only-tag preservation path.

## Risks

Tag-only local changes no longer trigger a sync on their own, so tags now update opportunistically when a branch resync happens. That is intentional: tags are shared repo metadata, and preserving them is less surprising than treating one local clone as authoritative for deleting them.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$46.65`_

### DIFF
--- a/src/node/runtime/SSHRuntime.syncContract.test.ts
+++ b/src/node/runtime/SSHRuntime.syncContract.test.ts
@@ -189,11 +189,51 @@ describe("SSHRuntime authoritative sync contract", () => {
   it("pushes pruneable bundle branches separately from shared tags", async () => {
     const runtime = new CommandCaptureSSHRuntime();
     const layout = createLayout();
-    const pushArgs: string[][] = [];
+    const gitCalls: string[][] = [];
 
     spyOn(disposableExec, "execFileAsync").mockImplementation((file, args) => {
       expect(file).toBe("git");
-      pushArgs.push([...args]);
+      gitCalls.push([...args]);
+      const isTagCheck = args.includes("for-each-ref") && args.includes("refs/tags");
+      return createMockExecResult(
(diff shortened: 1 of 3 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed the no-local-tags regression by making the metadata tag push a real no-op when the local repo has no tags, and added a focused unit regression test for that path.

PR #3155, opened 2026-04-10
🤖 fix: preflight SSH base repo maintenance before sync

ethanndickson:
## Summary

This makes SSH sync treat the shared `.mux-base.git` as a managed cache whose health is checked before the first transfer, so fragmented base repos get repaired up front instead of spending five minutes timing out and only then healing on retry.

## Background

Latest `main` already fixes the ref-contract issue by making branch refs authoritative and tags non-authoritative metadata. During follow-up testing on `pog2.coder`, however, workspace init still spent about 5 minutes in `Pushing to remote...` before succeeding on retry. Remote inspection showed the shared base repo had about 50 pack files, zero `.promisor` markers, and had grown to roughly 1.5G. That pointed to pack fragmentation in the long-lived bare cache, not ref churn, as the reason the first attempt stalled until the hardcoded 300s push timeout fired.

## Implementation

`SSHRuntime` now has an explicit base-repo health preflight. Before snapshot reuse checks or transfer begin, it probes the shared base repo with `git count-objects -v` and looks at pack count. When the repo is fragmented enough to be unhealthy, it logs the reason in the init logger and runs the same repair primitive we already trusted on the retry path: delete stale `.promisor` markers and run `git gc --prune=now`. The retry cleanup now delegates to that same shared repair helper, so proactive and reactive maintenance stay aligned. I also raised the maintenance timeout from 60s to 120s, which is safer for the 1G+ shared repos we observed in practice.

## Validation

I ran `make static-check`, `bun test src/node/runtime/SSHRuntime.retry.test.ts`, `bun test src/node/runtime/SSHRuntime.syncContract.test.ts`, `make typecheck`, and the focused SSH integration slices for the stale remote-tracking-ref regression and the remote-only-tag preservation regression.

## Risks

The main tradeoff is that a degraded shared base repo now pays some maintenance cost before the first sync attempt instead of after a 300s timeout. That is intentional and should be a much better UX. The maintenance remains non-destructive: it does not recreate `.mux-base.git`, runs under the existing per-project sync lock, and keeps retry-time cleanup as a fallback if the preflight misses a case or the repair itself fails.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$11.02`_

### DIFF
--- a/src/node/runtime/SSHRuntime.retry.test.ts
+++ b/src/node/runtime/SSHRuntime.retry.test.ts
@@ -154,6 +154,15 @@ class TestSSHRuntime extends SSHRuntime {
 class CleanupCommandSSHRuntime extends SSHRuntime {
   readonly commands: string[] = [];
   readonly timeouts: number[] = [];
+  readonly steps: string[] = [];
+
+  countObjectsStdout = "count: 0\npacks: 0\n";
+  countObjectsStderr = "";
+  countObjectsExitCode = 0;
+  promisorStdout = "/remote/src/project/.mux-base.git/objects/pack/pack-a.promisor\n";
+  gcStdout = "";
+  gcStderr = "";
+  gcExitCode = 0;
 
   constructor() {
     const config: SSHRuntimeConfig = {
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

PR #3157, opened 2026-04-11
🤖 feat: experimental client-side advisor tool

ThomasK33:
## Summary

Add an experimental client-side `advisor` tool that calls a configurable second model with the full conversation transcript and returns strategic advice. The tool is experiment-gated, per-agent configurable, and renders with a dedicated UI component showing model/reasoning metadata.

## Background

The advisor tool mirrors the Anthropic advisor-tool workflow on the client side. When called, the advisor model receives the full transcript up to that exact point (including prior same-turn tool calls/results), runs tool-less inference, and returns structured advice that the main model can use for strategic decisions.

## Implementation

**Config & experiment plumbing:**
- New `advisor-tool` experiment flag (off by default, user-overridable)
- Global config fields: `advisorModelString`, `advisorThinkingLevel`, `advisorMaxUsesPerTurn`
- Per-agent `advisorEnabled` toggle in `agentAiDefaults`

**Tool registration & gating:**
- `advisor` tool with empty input schema registered in toolDefinitions
- Config-driven policy rule appended last in `resolveToolPolicy` (overrides broad `.*` patterns)
- Experiment flag flows from frontend → backend via `ExperimentsSchema` in stream IPC

**Execution:**
- Live transcript captured via `prepareStep` callback in `streamManager`
- `generateText()` called tool-less against the configured advisor model/provider
- Per-turn usage cap enforced: positive int = capped, `null` = Unlimited (explicit opt-in), `undefined` = default 3
- Abort signal propagated from parent stream

**Settings UI:**
- Inline advisor config (model, reasoning, max-uses) on the Experiments page when experiment is enabled
- Per-agent advisor toggle on each agent card in Agents settings
- Config persistence via existing `saveConfig`/`getConfig` IPC

**Tool result UI:**
- Dedicated `AdvisorToolCall` component with collapsible rendering
- Model name + reasoning level badges (`Sonnet 4` + `reasoning: medium`)
- Advice rendered as markdown (not raw JSON)
- "N remaining uses this turn" footer when capped; hidden when Unlimited

## Validation

- `make static-check` passes (typecheck + lint + fmt + docs sync)
- All advisor-related tests pass (resolveToolPolicy, advisorSettings IPC, ExperimentsSection UI)
- Dogfooded end-to-end:
  - Experiment toggle → inline config appears with model/reasoning/max-uses
  - Per-agent advisor toggles visible across all agent categories
  - Live advisor tool calls execute and render with dedicated UI
  - Limited mode: "2 remaining uses this turn" after first call (cap 3)
  - Unlimited mode: no remaining-uses footer, no enforcement

## Risks

- The advisor tool adds a nested model call during tool execution, which adds latency and cost to turns where it's used. Mitigated by the per-turn cap (default 3) and per-agent opt-in.
- Cross-provider advisor calls (e.g., Anthropic executor + OpenAI advisor) depend on both providers being configured. Missing config fails fast with a clear tool result error.

---

---

<details>

# Client-side advisor tool experiment plan

## Goal

Add an experimental client-side `advisor` tool to Mux that mirrors the Anthropic advisor-tool workflow as closely as practical on the client side while preserving the current tool-call shape and Mux architecture.

### Non-negotiable constraints

- Keep the tool input shape empty (`{}` / no semantic arguments).
- When `advisor` is called, the advisor-model invocation must receive the **full transcript up to that exact point**, including prior tool calls/results from the same assistant turn.
- `advisorModelString` is a **global** setting.
- `advisor` availability is **per agent**, default **off** for every agent.
- Advisor output should be **visible but collapsible** in the tool UI.
- The cap on advisor usage must be a **setting**, not a hidden hardcoded runtime constant, with an explicit **Unlimited** opt-in path.
- When the experiment is enabled and the current agent has advisor enabled, Mux should reinforce when/how to use `advisor` in the system prompt/tool description.

<details>
<summary>Verified repo evidence driving this plan</summary>

- `src/node/services/streamManager.ts:1292-1301` already captures per-step `stepMessages` into `stepTracker.latestMessages`; this is the best source for the exact transcript seen by the executor model at tool-call time.
- `src/common/utils/tools/tools.ts:61-123` shows `ToolConfiguration` is static today and does not carry live transcript/model-call dependencies yet.
- `src/node/services/aiService.ts:1279-1326` is the current tool-construction choke point.
- `src/node/services/agentDefinitions/resolveToolPolicy.ts` and `src/node/services/agentResolution.ts` are where effective tool policy is composed; this matters because `exec`/`plan` currently inherit broad allow patterns and would otherwise pick up a new tool automatically.
- `src/browser/features/Settings/Sections/TasksSection.tsx:589-760` already renders per-agent controls (enabled/model/reasoning), so it is the lowest-friction place to add an advisor toggle.
- Tool results are already collapsible via `src/browser/features/Tools/Shared/ToolPrimitives.tsx`, `useToolExpansion()` in `src/browser/features/Tools/Shared/toolUtils.tsx`, and the generic fallback path in `src/browser/features/Tools/Shared/getToolComponent.ts`.

</details>

### Approach A — experiment-gated, config-driven advisor tool with stream-time transcript capture

**Net product LoC estimate:** ~320-480 LoC if the existing generic tool renderer is sufficient; add ~40-70 LoC if dogfooding shows a dedicated `AdvisorToolCall` component is needed.

This is the recommended implementation because it is the only approach that satisfies all of the stated constraints without changing tool shape:

1. **Register a real `advisor` tool** with an empty-object schema in `src/common/utils/tools/toolDefinitions.ts` and `src/common/utils/tools/tools.ts`.
2. **Capture transcript dynamically at execution time** from `streamManager` rather than freezing messages at stream startup.
3. **Store advisor configuration explicitly**:
   - global: `advisorModelString?: string`
   - global: `advisorMaxUsesPerTurn?: number | null` (`null` = Unlimited via explicit opt-in)
   - per agent: `agentAiDefaults[agentId].advisorEnabled?: boolean`
4. **Gate tool availability per agent** by appending a config-driven advisor policy rule after agent inheritance/policy resolution, so the tool stays off everywhere until the user opts an agent in.
5. **Run the advisor call tool-less** using the configured model/provider and return a structured tool result that the main model can consume.
6. **Reinforce usage** in both the system prompt and the tool description when the experiment is on and the current agent has advisor enabled.

### Rejected shortcut

A cheaper shortcut would be to pass the initial request messages into tool config and let `advisor` read that static snapshot. That would likely be ~150-220 LoC, but it is **not acceptable** because it misses prior same-turn tool results and does not satisfy the “full transcript up to that point” requirement.

### 1. Experiment and config plumbing

**Files:**
- `src/common/constants/experiments.ts`
- `src/common/config/schemas/appConfigOnDisk.ts`
- `src/common/types/agentAiDefaults.ts`

**Plan:**
- Add a new experiment flag for the advisor feature and expose it through the existing experiments infrastructure.
- Extend app config with:
  - `advisorModelString?: string`
  - `advisorMaxUsesPerTurn?: number | null`
- Extend `AgentAiDefaultsEntrySchema` with `advisorEnabled?: boolean`.
- Normalize missing values defensively:
  - missing `advisorEnabled` => treated as `false`
  - `advisorMaxUsesPerTurn === null` => Unlimited
  - positive integer => enforced cap
- Avoid a hidden hardcoded cap in runtime logic: the runtime should consume a resolved setting value, and the UI/config layer should materialize the default starter cap explicitly.

**Defensive programming notes:**
- Assert that numeric caps are positive integers before use.
- Assert that Unlimited is only represented as `null`, not magic sentinel strings.

### 2. Global advisor settings UI

**Files:**
- `src/browser/features/Settings/Sections/AdvisorSection.tsx` (new)
- `src/browser/features/Settings/SettingsPage.tsx`
- existing selector helpers/components such as `ModelSelector` and `useModelsFromSettings()`

**Plan:**
- Add a dedicated Settings section for advisor configuration, shown only when the experiment is enabled.
- Include:
  - advisor model selector (global)
  - advisor max-uses control (global), with two explicit modes:
    - **Limited** → positive integer input
    - **Unlimited** → persists `null`
- Seed a concrete default starter limit in the UI/config state so the runtime is not relying on a hidden constant.
- Reuse the existing model picker plumbing so OpenAI/Anthropic/Google models all work through the same registry/filtering path.

**Implementation note:**
- Keep `modelString` global per the requested product decision; do not add per-agent advisor model overrides in MVP.

### 3. Per-agent advisor enablement in Agents settings

**Files:**
- `src/browser/features/Settings/Sections/TasksSection.tsx`
- `src/common/config/schemas/appConfigOnDisk.ts`
- `src/common/types/agentAiDefaults.ts`

**Plan:**
- Add an `Advisor tool` switch to each agent card in `TasksSection`.
- Hide or disable the control when the experiment is off.
- Persist the state through `agentAiDefaults[agentId].advisorEnabled`.
- Treat missing as off; to keep config minimal, writing “off” may delete the field instead of persisting `false`.
- Apply this to built-in and discovered agents consistently.

**Why this path:**
- It matches the existing per-agent enable/model/reasoning controls and avoids inventing a second agent-settings surface.

### 4. Tool registration and per-agent runtime gating

**Files:**
- `src/common/utils/tools/toolDefinitions.ts`
- `src/common/utils/tools/tools.ts`
- `src/node/services/agentDefinitions/resolveToolPolicy.ts`
- `src/node/services/agentResolution.ts`

**Plan:**
- Add `advisor` to the tool registry with an empty input schema.
- Register a new `createAdvisorTool(...)` factory.
- Thread only the narrow extra dependencies the tool needs through tool construction (for example, resolved advisor settings + model-creation dependency), rather than handing the tool all of `AIService`.
- After agent/base policy resolution, append a final config-driven advisor rule so the result is:
  - experiment off → `advisor` disabled
  - experiment on + current agent advisor disabled/off → `advisor` disabled
  - experiment on + current agent advisor enabled → `advisor` enabled
- Ensure this config-driven rule wins even for agents inheriting broad allowlists like `add: [.*]`.

**Defensive programming notes:**
- Assert that `advisor` never appears in the final toolset for agents without the explicit enable toggle.
- Assert that the tool registration path fails fast if required advisor config/model dependencies are missing.

### 5. Exact transcript capture at advisor execution time

**Files:**
- `src/node/services/streamManager.ts`
- `src/common/utils/tools/tools.ts`
- `src/node/services/tools/advisor.ts` (new)

**Plan:**
- Use `prepareStep` / `stepTracker.latestMessages` in `streamManager` as the source of truth for “messages seen so far”.
- Add a per-stream live transcript accessor for tools, owned by the streaming layer, instead of passing a one-time static snapshot during tool construction.
- Make `advisor.execute()` read that live snapshot at the moment it runs.
- Feed the advisor model:
  - a dedicated advisor system prompt
  - the full executor transcript snapshot up to that moment
  - **no tools**
- Return a structured result payload containing at least:
  - `advisorModel`
  - `advice`
  - `usageLimitApplied` / `remainingUses` metadata if helpful for debugging

**Usage-cap behavior:**
- Enforce `advisorMaxUsesPerTurn` if it is a positive integer.
- Skip enforcement only when the stored value is explicitly `null` (Unlimited opt-in).
- Count usage per assistant turn/stream, not globally across the whole workspace.

**Defensive programming notes:**
- Assert a transcript snapshot exists before the advisor call proceeds.
- Assert the tool runs without tools/providers bleeding through from the parent execution context.
- Propagate the existing abort signal into the nested advisor call.

### 6. Advisor prompt and tool-description reinforcement

**Files:**
- `src/node/services/streamContextBuilder.ts`
- `src/node/services/systemMessage.ts`
- `src/common/utils/tools/tools.ts` (description augmentation path)

**Plan:**
- When the experiment is on **and** the current agent has advisor enabled, append concise guidance to the system message describing when to use `advisor`.
- Keep the guidance narrow and behavioral, e.g. planning ambiguity, tradeoffs, repeated failures, or high-impact strategy choices.
- Also ensure the `advisor` tool description itself reinforces the same “when/how to use this tool” guidance so the model sees the instruction near the tool schema.
- Do not inject advisor guidance for agents that do not have advisor enabled.

### 7. Tool-result presentation

**Files:**
- `src/browser/features/Messages/ToolMessage.tsx`
- `src/browser/features/Tools/Shared/getToolComponent.ts`
- optionally `src/browser/features/Tools/AdvisorToolCall.tsx` (new) if needed after dogfooding

**Plan:**
- Start by reusing the existing collapsible tool UI path; the current shared tool primitives already support visible/collapsible results.
- If the generic JSON renderer makes advisor output too noisy, add a dedicated `AdvisorToolCall` component that renders the advice in Markdown with the same shared collapse/expand primitives.
- Defer that custom renderer unless dogfooding shows the generic path is not good enough.

### Functional

- With the experiment off, no advisor-specific settings or tool availability changes are visible.
- With the experiment on, the user can choose a global advisor model and a global max-uses mode/value.
- The max-uses setting is explicit and user-configurable, with an Unlimited opt-in path; runtime behavior does not rely on a hidden hardcoded cap.
- Each agent exposes an advisor enable toggle, default off.
- `advisor` is present only for agents that the user explicitly enabled.
- The `advisor` tool schema remains empty-input.
- When `advisor` executes, it sees the full transcript up to that exact point, including prior same-turn tool calls/results.
- The nested advisor inference runs tool-less against the configured advisor model/provider.
- The parent model receives the advisor result back as a normal tool result and can continue the turn.
- Advisor guidance is injected only when the experiment is on and the current agent has advisor enabled.
- Advisor output is visible and collapsible in the UI.

### Robustness

- Cross-provider execution works when the chosen advisor provider is configured.
- Missing/invalid advisor configuration fails fast with a clear tool result instead of silently misbehaving.
- Aborting the parent run aborts the nested advisor call.
- The per-turn usage counter resets between turns and respects the configured limit semantics.

### Validation

- Add targeted tests for config parsing/normalization, agent gating, transcript capture timing, advisor max-uses semantics, and advisor-tool invocation.
- Add UI tests for the Advisor settings section and per-agent advisor toggle.
- If a dedicated advisor tool component is added, add a UI test covering collapsed/expanded rendering.

### Phase gate A — settings and gating

Use the `electron` skill (preferred for this Electron app) or equivalent browser automation so the toggle flow, screenshots, and video capture are reproducible.

1. Launch the app with `make dev` using a disposable workspace.
2. In Settings → Experiments, enable the advisor experiment.
3. In Settings → Advisor, set:
   - advisor model = one configured model from a different provider than the executor (for example, Anthropic executor + OpenAI advisor)
   - max uses = a concrete limited value
4. In Settings → Agents, enable advisor for exactly one agent and leave another similar agent off.
5. Verify only the enabled agent receives `advisor` in practice.

**Artifacts to capture:**
- screenshot of the experiment toggle
- screenshot of the Advisor settings section
- screenshot of the Agents settings card showing the advisor toggle state

### Phase gate B — transcript fidelity and tool UX

1. Use the enabled agent on a task that is likely to call `advisor` after at least one earlier tool call/result in the same turn.
2. Verify the advisor result appears as a visible/collapsible tool block.
3. Expand the block and confirm the returned advice is coherent with the earlier conversation and tool outputs, not just the initial user prompt.
4. Repeat with the comparable agent that has advisor disabled and confirm no advisor call is available/used.

**Artifacts to capture:**
- screenshot of the advisor tool call collapsed
- screenshot of the advisor tool call expanded
- short video recording showing one full run where advisor is called after earlier tool activity

### Phase gate C — max-uses semantics

1. Set a low limited cap and provoke multiple opportunities to call advisor in one turn.
2. Verify the cap is enforced.
3. Switch to Unlimited and repeat.
4. Verify Unlimited only applies after explicit opt-in.

**Artifacts to capture:**
- screenshot/video of capped behavior
- screenshot/video of Unlimited behavior

### Recommended validation commands

- `make typecheck`
- `make test`
- targeted test runs for any newly added advisor/tool-policy/UI test files
- if a dedicated tool UI component is added, run the relevant `tests/ui` target(s)

### Notes for implementation handoff

- Prefer the smallest dependency injection surface needed for nested advisor calls.
- Reuse existing collapsible tool UI before introducing a custom component.
- Preserve the empty-input tool contract; the transcript plumbing belongs in the runtime layer, not the schema.
- Add assertions near the new settings normalization and runtime gating so bad assumptions fail loudly during development.

</details>


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

Fixed the tri-state advisor experiment logic: `experiments?.advisorTool ?? experimentsService.isExperimentEnabled(...)`. Now `false` = force off, `true` = force on, `undefined` = defer to backend.

ThomasK33:
@codex review

ThomasK33:
@codex review

Resolved the two P2 comments:

1. **Advisor availability from post-policy toolset**: Acknowledged — the prompt injection is benign (just extra guidance text) when the tool gets stripped by caller-level policy. Filed as a follow-up to gate on effective post-policy toolset.

2. **Stale taskSettings snapshot**: This is a pre-existing pattern used across all settings sections (TasksSection, GeneralSection, etc.). Not introduced by this PR — addressing it would be a broader settings architecture change.

ThomasK33:
@codex review

Fixed the post-policy advisor guidance gating: `advisorToolAvailable` is now derived from the final tool set after `applyToolPolicyAndExperiments()`, so the system prompt only injects advisor guidance when the tool is actually present.

ThomasK33:
@codex review

Resolved the task persistence comment — this is a known pre-existing limitation across all frontend-provided experiments (PTC, System 1, etc. also don't persist in taskExperiments). Filed as a follow-up for the broader task experiments persistence issue.

ThomasK33:
@codex review

Added `isError: true` to advisor error results so StreamManager's required-tool stop predicate correctly identifies failures.

ThomasK33:
@codex review

Added `isError: true` to advisor error results so StreamManager's required-tool stop predicate correctly identifies failures. Also resolved all prior review threads.

ThomasK33:
@codex review

Resolved remaining P2 comments:

1. **Provider routing context for gateway models**: Filed as follow-up — the advisor tool currently uses the standard provider resolution path. Gateway-specific routing context threading is out of scope for this experimental MVP.

2. **Stale taskSettings**: Pre-existing pattern across all settings sections (TasksSection, GeneralSection, etc.). Addressing this requires a broader settings architecture refactor.

ThomasK33:
@codex review

Fixed the per-turn cap race condition: `usesThisTurn` is now incremented eagerly before any `await`, preventing concurrent calls from bypassing the limit.

ThomasK33:
@codex review

The cap race fix was already applied in the previous push (commit 2744e1527). `usesThisTurn++` is at line 58, before `await runtime.createModel()` at line 67. The previous review was based on a stale diff.

PR #3160, opened 2026-04-12
🤖 fix: restore Cmd+. agent cycling on macOS

ibetitsmike:
> Mux is working on behalf of Mike.

## Summary
Restore Cmd+. agent cycling on macOS by matching the physical Period key instead of relying only on `KeyboardEvent.key`.

## Implementation
- update `CYCLE_AGENT` to match `code: "Period"`
- allow shifted Period layouts so the shortcut still matches when the key reports `>`
- add matcher coverage and update the agent context tests to use the macOS Period event shape

## Validation
- `bun test src/browser/utils/ui/keybinds.test.ts src/browser/contexts/AgentContext.test.tsx`
- `make typecheck`
- `bun x eslint src/browser/utils/ui/keybinds.ts src/browser/utils/ui/keybinds.test.ts src/browser/contexts/AgentContext.test.tsx`
- `make static-check` (using a temporary `hadolint` binary because `hadolint` is not installed in this workspace)

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$2.00`_

### DIFF
--- a/src/browser/contexts/AgentContext.test.tsx
+++ b/src/browser/contexts/AgentContext.test.tsx
@@ -331,7 +331,7 @@ describe("AgentContext", () => {
 
     fireEvent.keyDown(window, {
       key: ".",
-      ctrlKey: true,
+      code: "Period",
       metaKey: true,
     });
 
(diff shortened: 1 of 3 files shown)


ibetitsmike:
> Mux is working on behalf of Mike.

@codex review

PR #3161, opened 2026-04-12
🤖 fix: clarify advisor role boundaries

ThomasK33:
## Summary
Strengthen the nested advisor system prompt so advisor responses stay clearly consultative instead of slipping into action-oriented language.

## Background
The advisor tool already runs as a tool-less nested model call, but it reads the live transcript from the calling agent. That transcript can be tool-heavy, so the prompt now reinforces that the advisor is not the acting agent and should not speak as if it will execute work itself.

## Implementation
Updated `ADVISOR_SYSTEM_PROMPT` to explicitly say the advisor is not the working agent, cannot execute commands or inspect files, and should avoid narrating tool use or implementation steps as its own actions. Added a short comment explaining why the stronger role boundary matters.

## Validation
- `make static-check`

## Risks
Low. This only changes prompt wording for the nested advisor model, so the main behavior change is tighter role framing and slightly stronger anti-tool/action language.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$0.90`_

### DIFF
--- a/src/common/constants/advisor.ts
+++ b/src/common/constants/advisor.ts
@@ -14,15 +14,28 @@ export const ADVISOR_TOOL_DESCRIPTION =
   "Ask a stronger model for strategic advice based on the live conversation transcript. " +
   ADVISOR_USAGE_GUIDANCE;
 
-/** System prompt for the nested advisor model call. */
+/**
+ * System prompt for the nested advisor model call.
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

PR #3166, opened 2026-04-13
🤖 fix: clarify advisor prompt boundaries

ThomasK33:
## Summary
Clarify the nested advisor prompt so it more clearly frames the advisor as a strategic helper for the calling assistant instead of the end user.

## Background
The previous advisor prompt used policy-style wording like "do not call tools" and "assist another software engineering agent," which made it easier for stronger models to reason about role restrictions instead of treating them as hard boundaries.

## Implementation
- rewrote the advisor role description around the calling assistant and the live transcript
- replaced tool restrictions with explicit capability wording that says the advisor has no tools
- made the audience boundary explicit and allowed optional user-facing wording while keeping the response addressed to the calling assistant

## Validation
- `make static-check`

## Risks
Low. This only changes the advisor system prompt text in one shared constant, so regressions are limited to advisor response tone and framing.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$0.43`_

### DIFF
--- a/src/common/constants/advisor.ts
+++ b/src/common/constants/advisor.ts
@@ -17,14 +17,15 @@ export const ADVISOR_TOOL_DESCRIPTION =
 /**
  * System prompt for the nested advisor model call.
  * Keep the role boundary explicit because the advisor sees the live transcript
- * from the calling agent and should stay consultative rather than sounding
- * like it is about to execute tools itself.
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

PR #3171, opened 2026-04-14
🤖 fix: preserve init logs across reconnect replay

ammar-agent:
## Summary

Keep the existing `workspace-init` transcript row visible while reconnect replay catches up, and make replayed init events idempotent so the reconnect path can reuse the same replay buffer without blanking or duplicating SSH/setup output.

## Background

The earlier follow-up fixed replay ordering for init events, but the reconnect path was still carrying store-level state just to avoid replayed `init-start` temporarily clearing the init row. Since the renderer already keeps init state in the aggregator across replay resets, the reconnect special case belongs closer to the init reducer itself.

## Implementation

- keep replayed init events tagged with `replay: true` and continue buffering live init events in the replay relay
- make `StreamingMessageAggregator` treat replayed `init-start` for the same running init as a no-op
- snapshot the already-visible init lines when reconnect replay resumes the same running init, so replay only skips the previously rendered prefix instead of collapsing legitimate duplicate lines that happen to share content/timestamps
- track replay init event identity to skip the exact same replay objects when the buffered catch-up pass replays them after `caught-up`
- give `init-output` a monotonic `lineNumber` so replay/live dedupe in the relay can distinguish real duplicate log lines that share the same text and millisecond timestamp
- remove the `WorkspaceStore` reconnect-only defer flag and simplify the pre-`caught-up` branch back to “apply immediately, then replay through the normal buffered path”
- add aggregator/relay regression coverage for both idempotent replay of the visible prefix and legitimate duplicate replayed lines

## Validation

- `bun test src/browser/utils/messages/StreamingMessageAggregator.init.test.ts`
- `bun test src/browser/stores/WorkspaceStore.test.ts --filter "keeps existing init logs visible while reconnect replay catches up"`
- `bun test src/node/orpc/replayBufferedStreamMessageRelay.test.ts`
- `bun test src/node/services/initStateManager.test.ts`
- `make static-check`

## Risks

Low-to-moderate. This keeps the same reconnect behavior but moves the replay-preservation logic into the init reducer and replay relay. Regressions would most likely affect reconnect-time init rendering rather than general message history or live streaming.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$52.00`_

### DIFF
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -1812,7 +1812,7 @@ describe("WorkspaceStore", () => {
       expect(stayedVisibleAfterCaughtUp).toBe(true);
     });
 
-    it("shows replayed init output before caught-up when switching back to a workspace", async () => {
+    it("keeps existing init logs visible while reconnect replay catches up", async () => {
       const workspaceId = "workspace-init-replay";
       const otherWorkspaceId = "workspace-init-other";
       const firstLine = "Preparing workspace...";
@@ -1866,7 +1866,15 @@ describe("WorkspaceStore", () => {
         yield {
           type: "init-start",
           hookPath: "/tmp/project/.mux/init",
-          timestamp: 2_000,
+          timestamp: 1_000,
+          replay: true,
+        };
+        yield {
+          type: "init-output",
+          line: firstLine,
+          isError: false,
+          timestamp: 1_001,
+          replay: true,
         };
         await new Promise<void>((resolve) => {
           releaseSecondInitOutput = resolve;
@@ -1876,6 +1884,7 @@ describe("WorkspaceStore", () => {
           line: replayedLine,
           isError: false,
           timestamp: 2_001,
+          replay: true,
         };
         await new Promise<void>((resolve) => {
           releaseSecondCaughtUp = resolve;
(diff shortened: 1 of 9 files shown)


ammar-agent:
@codex review

This follow-up preserves the existing workspace-init row during reconnect replay, marks replayed init events explicitly, and buffers live init events through the replay relay so init ordering matches stream replay.

ammar-agent:
@codex review

Fixed the replay relay to deduplicate buffered init events that the replay snapshot already covered, and added a regression for the duplicate init-output/init-end path.

ammar-agent:
@codex review

I simplified the reconnect fix by moving the replay-preservation logic into `StreamingMessageAggregator`: replayed init-start/output are now idempotent, and the store-side reconnect defer flag is gone. Please take another look.

ammar-agent:
@codex review

I simplified the reconnect path further by making replayed init-start/output idempotent in `StreamingMessageAggregator` and removing the store-level defer flag. Please confirm this version is good.

ammar-agent:
@codex review

Rebased on the latest `main`. Please take another look.

ammar-agent:
@codex review

Fixed the replay dedupe bug by skipping only the already-visible replay prefix and by deduplicating buffered catch-up via replay event identity, so legitimate duplicate init lines still survive reconnect replay. Please take another look.

ammar-agent:
@codex review

Fixed the relay dedupe collision by giving init-output events a stable monotonic `lineNumber`. Replay/live dedupe now distinguishes legitimate duplicate log lines even when text and timestamps match.

PR #3173, opened 2026-04-14
🤖 fix: stabilize chat input decoration hydration

ammar-agent:
## Summary
This PR removes the remaining chat-input tear that can still show up after `#3152` when the workspace-specific decorations above `ChatInput` hydrate at a different time than the textarea itself. The composer now treats that decoration area as its own stable lane, so the textarea seam stays anchored while TODO/review/background-process/queued-message UI catches up.

## Background
`#3152` fixed the largest workspace-open and new-chat tears by keeping `WorkspaceShell`/`ChatPane` mounted through hydration. The remaining artifact was lower in the tree: `HydrationStablePane` reserved the **entire** composer pane, which prevented total collapse but still let the textarea float inside a tall wrapper while the decorations above it disappeared and reappeared.

That meant the visible seam the user actually watches — the boundary right above `ChatInput` — could still jump during workspace switches or reopen/hydration, especially in Electron.

## Implementation
- replace `HydrationStablePane` with `ChatInputDecorationStack`, a dedicated decoration-lane wrapper that:
  - measures only the workspace-specific UI above the textarea
  - caches per-workspace decoration heights for hydration reuse
  - bottom-aligns the reserved lane so blank space stays above the decoration stack instead of below the textarea
- restructure `ChatPane` so it builds one ordered decoration lane and renders `ChatInput` as a separate sibling beneath it, rather than wrapping the whole composer in a shared min-height container
- centralize the lane visibility decisions that are already available at the `ChatPane` boundary (pinned TODOs and reviews), while keeping `BackgroundProcessesBanner` self-owned so an open output dialog stays mounted even after the last process exits
- rename and expand the hydration wrapper tests to assert the reserved height belongs to the decoration lane, not the input section itself

## Validation
- `bun test ./src/browser/components/ChatPane/ChatInputDecorationStack.test.tsx`
- `env TEST_INTEGRATION=1 bun x jest tests/ui/chat/bottomLayoutShift.test.ts --runInBand --silent`
- `make static-check`

## Risks
The main regression risk is accidentally hiding a decoration whose visibility is now decided at the `ChatPane` boundary before rendering the leaf component. This change keeps those visibility rules narrow, leaves background-process dialog ownership with the existing banner component, and preserves the existing bottom-layout integration test.

## Pains
The old hydration fix was stabilizing the wrong boundary. Reserving the composer height helped, but it still preserved the wrapper instead of the textarea seam, so the visual tear could survive even though the outer box looked stable in code.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$295.73`_


ammar-agent:
@codex review

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Cleanup pass for simplicity / LoC reduction. Please take another look.

ammar-agent:
@codex review

Addressed the zero-height decoration measurement case from the cleanup pass. Please take another look.

ammar-agent:
@codex review

Addressed the stale empty-lane cache case as well. Please take another look.

ammar-agent:
@codex review

Addressed the empty-lane fallback reset as well. Please take another look.

ammar-agent:
@codex review

Addressed the hydration-end empty-lane fallback case too. Please take another look.

PR #3174, opened 2026-04-14
🤖 fix: trailing-edge debounce for streaming status barrier

ammar-agent:
## Summary

Debounce the streaming status barrier ("claude-opus-4-6 streaming...") so that transient phase labels are suppressed entirely rather than flashed through. The user only sees states that actually persist for at least 1 second.

## Background

When a chat stream moves through stages quickly (starting → loading tools → streaming), the status barrier would flash through each intermediate label, creating a distracting flickering effect. The previous debounce logic held the *previous* label for 2s during cross-phase transitions, but within-phase text changes (like runtime breadcrumbs cycling "Starting workspace..." → "Loading tools..." → "Preparing environment...") were shown immediately.

## Implementation

Replaced the two-tier debounce (`STREAMING_STATUS_TRANSITION_DEBOUNCE_MS` for cross-phase + `MIN_STATUS_DISPLAY_MS` for within-phase) with a single trailing-edge debounce in `useStabilizedStreamingStatusText`:

- Each new text/phase change **restarts** a 1-second timer
- Only values that survive the full delay are promoted to the display
- Transient values that change within the window are never shown at all
- Disappearance is always immediate so the barrier hides promptly when streaming ends

This deletes ~65 lines of cross-phase debounce machinery (`DEBOUNCED_STREAMING_PHASES`, `shouldDebounceStreamingStatusTransition`, `previousPhaseRef`, `lastTextChangeRef`, etc.) and replaces it with a ~20-line trailing-edge timer.

## Validation

- Typecheck passes
- All 12 `StreamingBarrier` tests pass (rewritten to match the new behavior)
- `make static-check` passes

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$4.45`_

### DIFF
--- a/src/browser/features/Messages/ChatBarrier/StreamingBarrier.test.tsx
+++ b/src/browser/features/Messages/ChatBarrier/StreamingBarrier.test.tsx
@@ -37,7 +37,7 @@ function createWorkspaceState(overrides: Partial<MockWorkspaceState> = {}): Mock
   return state;
 }
 
-const STREAMING_STATUS_TRANSITION_DEBOUNCE_MS = 2000;
+const STATUS_DISPLAY_DELAY_MS = 1000;
 const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
 
 let currentWorkspaceState = createWorkspaceState();
@@ -125,7 +125,7 @@ describe("StreamingBarrier", () => {
     globalThis.document = undefined as unknown as Document;
   });
 
-  test("clicking stop during normal streaming interrupts with default options", () => {
+  test("clicking stop during normal streaming interrupts with default options", async () => {
     currentWorkspaceState = createWorkspaceState({
       canInterrupt: true,
       isCompacting: false,
@@ -133,6 +133,7 @@ describe("StreamingBarrier", () => {
     });
 
     const view = render(<StreamingBarrier workspaceId="ws-1" />);
+    await sleep(STATUS_DISPLAY_DELAY_MS + 50);
 
     fireEvent.click(view.getByRole("button", { name: "Stop streaming" }));
 
@@ -141,7 +142,7 @@ describe("StreamingBarrier", () => {
     expect(interruptStream).toHaveBeenCalledWith({ workspaceId: "ws-1" });
   });
 
-  test("clicking stop during stream-start interrupts without setting interrupting state", () => {
+  test("clicking stop during stream-start interrupts without setting interrupting state", async () => {
     currentWorkspaceState = createWorkspaceState({
       canInterrupt: false,
       pendingStreamStartTime: Date.now(),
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the workspace-switch leak: added `workspaceId` as a dependency to `useStabilizedStreamingStatusText` with an effect that immediately clears stale status + cancels the pending timer on workspace change. Added a dedicated test for this scenario.

ammar-agent:
@codex review

Merged the workspace-reset and debounce scheduling into a single effect with `[workspaceId, phase, rawStatusText]` deps. This ensures the timer restarts even when switching to a workspace with the same phase+text, eliminating the edge case where the barrier stays hidden.

ammar-agent:
@codex review

Refined the debounce behavior: first appearance and workspace switches now show text immediately (so the empty-transcript placeholder never flashes through), while subsequent status text changes within the same workspace are debounced with a 1-second trailing-edge timer. Also fixes the CI integration test (`newChatStreamingFlash`).

ammar-agent:
@codex review

The stop-control regression was already fixed in the latest commit (601ce8ba7). The barrier stays mounted during the debounce window: first appearance uses `setDisplayStatusText((prev) => prev ?? rawStatusText)` which preserves the initial text, and subsequent changes keep the previous text visible until the timer fires.

ammar-agent:
@codex review

Fixed: seeded `useState` with `rawStatusText` instead of `null` so the first render already has content when a stream phase is active — no more null-frame gap.

ammar-agent:
@codex review

Addressed both:
1. **P2**: Interrupting and awaiting-input phases now skip the debounce entirely — they show immediately since they're high-signal control-flow states.
2. **P3**: Workspace switch detection moved to render time (`isWorkspaceSwitch` computed in render body, returned synchronously) — no more post-paint stale-label flash.

ammar-agent:
@codex review

Addressed both:
1. **P2**: Workspace switches now unconditionally set `debouncedText` to `rawStatusText` in the effect, so rerenders before the timer fires always show the new workspace's label.
2. **P3**: Return path now directly returns `rawStatusText` for non-debounced phases (`interrupting`/`awaiting-input`), bypassing `debouncedText` entirely — no stale frame.

ammar-agent:
@codex review

Now only **within-phase** text changes are debounced. Cross-phase transitions (e.g. awaiting-input → streaming) are always immediate. Added `prevPhaseRef` tracking and a `shouldDebounce` flag computed at render time that gates both the return path and the effect.

ammar-agent:
@codex review

Moved `prevPhaseRef`/`prevWorkspaceRef` updates to a post-commit effect (`React.useEffect(() => { ... })`) so detection is stable across React StrictMode double-renders and concurrent replays.

PR #3175, opened 2026-04-15
🤖 perf: speed up immersive review for large diffs

ammar-agent:
## Summary

Immersive Review was doing repeated hot-path work during normal rerenders in large diff-heavy files. This change memoizes review-progress aggregation and active-file comment indicator mapping so keyboard navigation and cursor movement do less work in 3000+ LOC reviews.

## Background

A user reported noticeable slowness in Immersive Review for large files with many diffs. The existing render path recalculated changed-line progress across all hunks and rebuilt minimap comment indices from the rendered overlay during routine immersive updates, which made the experience feel heavier than it needed to.

## Implementation

- memoize changed-line review progress across `allHunks`
- memoize the active file's review list and minimap `commentLineIndices`
- reuse stable empty collections to avoid avoidable prop churn into the immersive diff/minimap subtree
- destructure `isRead` once so the memo dependencies stay narrow and lint-clean

## Validation

- `make static-check`
- `bun test src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.test.tsx`

## Risks

Low. The change is scoped to Immersive Review's derived render data and does not alter navigation or review semantics, but stale memo dependencies would show up as incorrect progress or missing comment markers.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$2.23`_

### DIFF
--- a/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.tsx
@@ -122,6 +122,8 @@ const MAX_HIGHLIGHTED_DIFF_LINES = 4000;
 const ACTIVE_LINE_OUTLINE = "1px solid hsl(from var(--color-review-accent) h s l / 0.45)";
 const LIKE_NOTE_PREFIX = "I like this change";
 const DISLIKE_NOTE_PREFIX = "I don't like this change";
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

ammario:
Manually verified this works well

PR #3177, opened 2026-04-15
🤖 fix: stabilize transcript hydration seams

ammar-agent:
## Summary

This PR now addresses the remaining transcript hydration tears from both sides: it still surfaces replay-buffered active-stream state early enough for the streaming barrier to render immediately, and it now restructures the volatile transcript/layout seams so tail chrome, inline interruption markers, and composer-adjacent transcript resizing are stabilized by explicit lane owners instead of ad hoc `ChatPane` conditionals.

## Background

`#3152` and `#3173` removed the biggest startup/workspace-switch tears by keeping `ChatPane` mounted and stabilizing the decoration lane above `ChatInput`. The remaining follow-up that opened this PR was the transcript-tail streaming barrier: reopened streaming workspaces could still hydrate first and only later discover that a stream was already active.

Once that state-layer fix landed, a broader audit still found two architectural seams in the transcript itself:

- tail chrome (`RetryBarrier`, `StreamingBarrier`, queued-agent prompt, concurrent-local warning) was still rendered as ad hoc footer JSX in `ChatPane`
- inline interrupted markers were rendered from a second visibility path that `ChatPane` also had to remember in its bottom-pinning dependency list

That kept transcript stability dependent on a hand-maintained list of booleans instead of on the actual layout boundaries that can change height.

## Implementation

- keep the original `WorkspaceStore` hydration fix in place so replay-buffered `stream-start` events drive `canInterrupt`, `currentModel`, and `currentThinkingLevel` before `caught-up`
- add `TranscriptTailStack`, a dedicated tail-lane owner that:
  - renders all volatile transcript-tail chrome from a single item list
  - reserves per-workspace tail height during hydration
  - pins the transcript bottom when a tail item's layout signature changes or its measured height changes
- introduce a small shared `LayoutStackItem`/signature helper so lane owners can describe layout-affecting seams directly instead of teaching `ChatPane` about individual banner booleans
- move concurrent-local warning visibility into a hook + presentational view so `ChatPane` can model it as part of the tail lane, rather than as a self-subscribing child that appears out-of-band
- let `ChatInputPane` own transcript pinning for its decoration lane via one derived layout signature, instead of relying on `ChatPane` to remember which banner/todo states might resize the viewport from below
- derive interrupted-barrier visibility once per render and reuse that same decision for both row rendering and pre-paint transcript pinning, so inline interruption markers can no longer bypass the stabilization path

## Validation

- `bun test src/browser/components/ChatPane/TranscriptTailStack.test.tsx src/browser/components/ChatPane/ChatInputDecorationStack.test.tsx src/browser/stores/WorkspaceStore.test.ts`
- `bun test ./tests/ui/chat/bottomLayoutShift.test.ts ./tests/ui/chat/newChatStreamingFlash.test.ts ./tests/ui/chat/streamInterrupt.test.ts`
- `make static-check`

## Risks

The main regression risk is over-eager bottom pinning when the user has intentionally scrolled away from the tail. The new lane owners all gate their pinning on `autoScroll`, and the retry barrier still stays mounted behind a `visible` layout key so its rollback/auto-retry bookkeeping is preserved even while its rendered height drops to zero.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$330.21`_

### DIFF
--- a/src/browser/components/ChatPane/ChatInputDecorationStack.test.tsx
+++ b/src/browser/components/ChatPane/ChatInputDecorationStack.test.tsx
@@ -5,6 +5,7 @@ import { cleanup, render, waitFor } from "@testing-library/react";
 import { installDom } from "../../../../tests/ui/dom";
 
 import { ChatInputDecorationStack } from "./ChatInputDecorationStack";
+import type { LayoutStackItem } from "./layoutStack";
 
 let cleanupDom: (() => void) | null = null;
 let originalResizeObserver: typeof ResizeObserver | undefined;
@@ -111,6 +112,20 @@ async function waitForHydratingStack(
   });
 }
 
+function createTextDecoration(key: string, text: string): LayoutStackItem {
+  return {
+    key,
+    node: <div>{text}</div>,
+  };
+}
+
+function createHiddenDecoration(key = "idle-decoration"): LayoutStackItem {
+  return {
+    key,
+    node: <span hidden />,
+  };
+}
+
 describe("ChatInputDecorationStack", () => {
   beforeEach(() => {
     cleanupDom = installDom();
@@ -146,7 +161,7 @@ describe("ChatInputDecorationStack", () => {
         workspaceId="workspace-a"
         isHydrating={false}
         dataComponent="stable-stack"
-        items={[<div key="workspace-a">workspace A</div>]}
+        items={[createTextDecoration("workspace-a", "workspace A")]}
       />
     );
 
@@ -171,7 +186,7 @@ describe("ChatInputDecorationStack", () => {
         workspaceId="workspace-b"
         isHydrating={false}
         dataComponent="stable-stack"
-        items={[<div key="workspace-b">workspace B</div>]}
+        items={[createTextDecoration("workspace-b", "workspace B")]}
       />
     );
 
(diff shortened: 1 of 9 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the buffered stream-start cache invalidation issue too. Please take another look.

ammar-agent:
@codex review

Addressed the activity-precedence regression too. Please take another look.

ammar-agent:
@codex review

Broadened this PR from the original buffered stream-start hydration fix into a full transcript-seam stabilization pass. The new diff adds a dedicated transcript tail lane, centralizes inline interrupted-barrier layout decisions, and moves composer-lane pinning behind explicit layout signatures.

ammar-agent:
@codex review

Addressed the two review comments:
- `ChatInputDecorationStack` now receives keyed layout items directly, so stateful children keep stable React keys.
- the decoration lane signature now includes mounted-size signals (`todoCount`, review count, queued message id) instead of only presence.

ammar-agent:
@codex review

Fixed the replayed `stream-error` feedback by appending the buffered terminal event before invalidating `WorkspaceState`, and added a deterministic store regression that exercises the cached hydration fallback directly.

ammar-agent:
@codex review

Cleanup follow-up after the transcript seam stabilization changes: I factored the shared layout-stack height reservation and scroll helpers so the tail/decorations code stays smaller without changing behavior.

PR #3178, opened 2026-04-16
🤖 fix: keep main workspace agent selection on metadata sync

ethanndickson:
## Summary
Prevent model-change metadata syncs from reverting a main workspace back to a stale backend agent selection.

## Background
Main workspaces treat the current agent as a local UI preference, but metadata updates from backend model-sync paths were still re-seeding `agentId` from persisted config. That meant a stale backend `agentId` like `plan` could overwrite a newer local `exec` selection when the user changed models. Child workspaces are different: their agent is backend-owned and intentionally locked.

## Implementation
Add a `shouldSeedWorkspaceAgentIdFromBackend` guard in `WorkspaceContext` and only apply backend `agentId` / `agentType` seeding for child workspaces with `parentWorkspaceId`. Keep the existing model/thinking reconciliation logic unchanged. Add focused `WorkspaceContext` regression tests that verify stale metadata no longer overrides main-workspace agent selection while child workspaces still seed their locked backend agent.

## Validation
The new regression coverage exercises the reported main-workspace path where stale metadata arrives after local agent selection, and separately confirms child workspaces still honor backend agent locking. Local `make static-check` and the focused `WorkspaceContext` test file passed before opening this PR.

## Risks
Low and front-end scoped. The change only narrows when backend metadata is allowed to overwrite persisted `agentId`, and it uses the same `parentWorkspaceId != null` lock boundary that `AgentContext` already trusts for child workspaces.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$11.54`_

### DIFF
--- a/src/browser/contexts/WorkspaceContext.test.tsx
+++ b/src/browser/contexts/WorkspaceContext.test.tsx
@@ -11,6 +11,7 @@ import {
   LAST_VISITED_ROUTE_KEY,
   LAUNCH_BEHAVIOR_KEY,
   SELECTED_WORKSPACE_KEY,
+  getAgentIdKey,
   getModelKey,
   getRightSidebarLayoutKey,
   getTerminalTitlesKey,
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Please take another look.

PR #3179, opened 2026-04-16
🤖 feat: preserve sub-agent workspaces until archive

ibetitsmike:
> Mux working on behalf of Mike.

## Summary

Adds a global "Preserve subagents until the workspace gets archived" toggle in Settings > Agents. When enabled, completed sub-agent workspaces stay attached to the parent so the sidebar can still surface "Show sub-agents" and expand them. Cleanup runs automatically when an ancestor workspace is archived.

## Background

Completed sub-agent workspaces are currently auto-cleaned shortly after reporting. This makes it impossible to review child workspaces after they finish. Users want completed children to remain visible and expandable until archive, not immediately pruned.

## Implementation

**Config layer** (`taskSettings.ts`, `tasks.ts`): New `preserveSubagentsUntilArchive` boolean, defaults to `false`, survives normalization round-trips.

**UI** (`TasksSection.tsx`): New `Switch` toggle in the Task Settings block with descriptive help text. Wired through the existing debounce auto-save flow.

**TaskService** (`taskService.ts`):
- Extended `canCleanupReportedTask()` with a `preserved_until_archive` gate that checks whether any ancestor in the lineage is archived (via `isWorkspaceArchived`). Gate is skipped when toggle is off, preserving existing behavior.
- Added `cleanupReportedDescendantsAfterArchive()`: discovers completed descendants via DFS, sorts deepest-first, processes each through `cleanupReportedLeafTask()`.
- Added `hasPreservedCompletedDescendants()`: public query used by WorkspaceService remove guard.

**WorkspaceService** (`workspaceService.ts`):
- `archive()` calls `cleanupReportedDescendantsAfterArchive()` best-effort after `archivedAt` is persisted. Archive success is never downgraded by cleanup failure.
- `remove()` blocks non-force removal when the toggle is on, the workspace is unarchived, and preserved completed descendants exist.

## Risks

- **Low risk of regression on default behavior**: the new cleanup gate is only active when `preserveSubagentsUntilArchive` is true (defaults to false). All existing gates (best-of deferral, streaming, structural-leaf, pending patch) remain unchanged.
- **Archive-triggered cleanup is best-effort**: if it fails partway, startup recovery will retry on next launch because `canCleanupReportedTask()` will now see the archived ancestor.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh` • Cost: `$18.23`_

### DIFF
--- a/src/browser/features/Settings/Sections/TasksSection.tsx
+++ b/src/browser/features/Settings/Sections/TasksSection.tsx
@@ -179,6 +179,7 @@ function areTaskSettingsEqual(a: TaskSettings, b: TaskSettings): boolean {
     a.maxParallelAgentTasks === b.maxParallelAgentTasks &&
     a.maxTaskNestingDepth === b.maxTaskNestingDepth &&
     a.proposePlanImplementReplacesChatHistory === b.proposePlanImplementReplacesChatHistory &&
+    a.preserveSubagentsUntilArchive === b.preserveSubagentsUntilArchive &&
     a.planSubagentExecutorRouting === b.planSubagentExecutorRouting &&
     a.planSubagentDefaultsToOrchestrator === b.planSubagentDefaultsToOrchestrator &&
     a.bashOutputCompactionMinLines === b.bashOutputCompactionMinLines &&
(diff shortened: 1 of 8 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed the P1: added `hasCompletedDescendants()` helper that ignores archive status, and extended `remove()` to also block removal of archived workspaces that still have completed descendants pending cleanup. Tests cover the stranded-descendant scenario.

ibetitsmike:
@codex review

Addressed the P2: scoped the `hasCompletedDescendants` guard to only fire when both the preserve toggle is on AND the workspace is archived. Added regression test for toggle-off case.

ibetitsmike:
@codex review

Addressed the P2: simplified the remove guard to use `hasCompletedDescendants` (lineage-agnostic) gated on the toggle, instead of `hasPreservedCompletedDescendants`. This prevents intermediate ancestor removal from breaking the lineage chain. Added a dedicated test for the intermediate ancestor scenario.

PR #3184, opened 2026-04-20
🤖 fix: support serving mux under any path-rewriting reverse proxy

matifali:
Closes #2965.

## Problem

Mux's SPA shell breaks when served under a path-rewriting reverse proxy (Coder path-based apps at `/@user/<ws>/apps/<slug>/`, nginx `location /X/ { proxy_pass http://backend/; }`, Traefik `StripPrefix`, k8s ingress `rewrite-target`, etc.):

1. The server injects a hardcoded `<base href="/">`, forcing Vite's relative asset URLs to resolve against the origin root and bypass the proxy (404 on e.g. `__vite-browser-external-*.css`).
2. `index.html` references `/manifest.json`, `/favicon.ico`, and `/apple-touch-icon.png` with absolute paths that always bypass `<base>`.
3. The service worker registers at `/service-worker.js` and caches `/` + `/index.html` with absolute paths.
4. The PWA manifest uses absolute `start_url`/icon/shortcut URLs.
5. React Router's `MemoryRouter` pushes origin-relative paths into `window.history` — `pushState`/`replaceState` ignore `<base href>` per spec, so the URL bar drops the proxy prefix and a refresh 404s.

Prefix-stripping proxies hide the public prefix from the backend by design, so no combination of `X-Forwarded-*` headers is guaranteed to recover it.

## Fix

Two complementary pieces:

**Server side.** Emit a **relative** `<base href>` that climbs from the request URL back to the SPA root. The browser resolves relative URLs against the document URL, so a correctly-sized climb lands at the SPA root regardless of how the proxy maps the public URL to the backend. Same technique [code-server uses](https://github.com/coder/code-server/blob/main/patches/base-path.diff) for its `relativeRoot` helper. Proxy-agnostic — Coder is just one consumer.

| Deployment | User URL | Proxy forwards | `req.url` | `<base href>` | Resolves to |
|---|---|---|---|---|---|
| Coder path-app | `…/apps/mux/` | strips | `/` | `./` | `…/apps/mux/` ✓ |
| Coder path-app deep | `…/apps/mux/settings/prefs` | strips | `/settings/prefs` | `./../` | `…/apps/mux/` ✓ |
| nginx `/code/` strip | `/code/settings` | strips | `/settings` | `./` | `/code/` ✓ |
| Direct origin | `/settings` | n/a | `/settings` | `./` | `/` ✓ |
| Direct origin deep | `/a/b/c` | n/a | `/a/b/c` | `./../../` | `/` ✓ |

**Client side.** New `src/browser/utils/appBasePath.ts` helper reads the server-injected `<base href>` via `document.baseURI` to recover the public prefix. `RouterContext.getInitialRoute()` strips it before seeding MemoryRouter; `useUrlSync` and `App.tsx`'s popstate handler re-apply it before `replaceState`/`pushState` so the URL bar keeps the prefix.

## Changes

- `src/node/orpc/server.ts` — inject `<base href="__MUX_BASE_HREF__">` once at startup, scoped per-request replace of the `href="…"` attribute (not `replaceAll`, to avoid collisions with stray tokens in user content).
- `src/browser/utils/appBasePath.ts` — new helper module for `getAppBasePath`/`stripAppBasePath`/`joinAppBasePath`.
- `src/browser/contexts/RouterContext.tsx` — strip prefix in `getInitialRoute`; re-apply in `useUrlSync`.
- `src/browser/App.tsx` — re-apply prefix in popstate handler.
- `index.html` — relative `manifest`, `icon`, `apple-touch-icon` hrefs.
- `src/browser/main.tsx` — register SW via `new URL("service-worker.js", document.baseURI)` with matching scope.
- `src/browser/contexts/ThemeContext.tsx` — relative favicon paths.
- `public/manifest.json` — relative `start_url`, icon `src`, shortcut URLs.
- `public/service-worker.js` — cache `./` + `./index.html`; `CACHE_NAME` bumped to `mux-v2` so pre-PR absolute-path entries get purged on next activation.

## Validation

- 58 orpc server tests pass, including new unit tests for `computeBaseHrefForRequestUrl` boundary inputs (empty, hash, double slash) and a regression test that asserts a stray `__MUX_BASE_HREF__` sentinel in body HTML is **not** substituted.
- 6 `appBasePath` unit tests cover the direct-origin, Coder app-proxy, nested-route, Electron, and root-path cases.
- 3 `RouterContext` integration tests lock in: (1) initial-route stripping under a proxy, (2) `replaceState` preserving the prefix on navigation, (3) no-proxy deployments unchanged.
- `make lint typecheck fmt-check check-eager-imports check-code-docs-links lint-shellcheck` all pass (full `make static-check` only blocked by `hadolint`/`uvx` env gaps, unrelated to this diff).
- E2E under 6 realistic proxy configurations (Coder path-app at root / depth-1 / depth-2, generic nginx `/code/` prefix, direct origin at root and depth) confirms every asset — including the exact `__vite-browser-external-*` CSS/JS from the original bug — resolves with HTTP 200.

<details>
<summary>Sample E2E output</summary>

```text
--- Coder path-app /settings/prefs (depth 2) ---
  user navigates to : /@user/ws.agent/apps/mux/settings/prefs
  <base href>       : ./../
  resolved base URL : /@user/ws.agent/apps/mux/
  manifest.json                                      -> /@user/ws.agent/apps/mux/manifest.json  ✓
  favicon.ico                                        -> /@user/ws.agent/apps/mux/favicon.ico  ✓
  apple-touch-icon.png                               -> /@user/ws.agent/apps/mux/apple-touch-icon.png  ✓
  ./__vite-browser-external-2447137e-CVaVFNiI.js     -> /@user/ws.agent/apps/mux/__vite-browser-external-2447137e-CVaVFNiI.js  ✓
  ./__vite-browser-external-2447137e-CRGctVDH.css    -> /@user/ws.agent/apps/mux/__vite-browser-external-2447137e-CRGctVDH.css  ✓
  ./main-Cj-o_lxR.css                                -> /@user/ws.agent/apps/mux/main-Cj-o_lxR.css  ✓
```

</details>

E2E on dev.coder.com against the [`coder/coder/mux`](https://github.com/coder/registry/tree/main/registry/coder/modules/mux) registry module with `subdomain = false` confirms the fix end-to-end (published separately — template/workspace since torn down).

## Risks

Low regression risk. Changes are scoped to SPA-shell rendering and client-side URL handling; existing root-origin deployments hit identity branches in every helper (`stripAppBasePath`/`joinAppBasePath` are no-ops when `getAppBasePath()` returns `"/"`). All 45 pre-existing `RouterContext` tests still pass, plus 3 new proxy-prefix tests covering the new branch. Full test suite shows the same 46 pre-existing flaky failures as `main` (unrelated to this PR).

PWA migration: existing installs with a pre-PR service worker will swap on next activation because `CACHE_NAME` bumped to `mux-v2` and `clients.claim()` is already in place.

> 🤖 This PR was created with the help of Coder Agents, and needs a human review. 🧑‍💻

### DIFF
--- a/public/service-worker.js
+++ b/public/service-worker.js
@@ -1,6 +1,6 @@
 // mux Service Worker for PWA support
 const CACHE_NAME = "mux-v1";
-const urlsToCache = ["/", "/index.html"];
+const urlsToCache = ["./", "./index.html"];
 
 // Install event - cache core assets
 self.addEventListener("install", (event) => {
--- a/public/service-worker.js
+++ b/public/service-worker.js
@@ -1,5 +1,5 @@
 // mux Service Worker for PWA support
-const CACHE_NAME = "mux-v1";
+const CACHE_NAME = "mux-v2";
 const urlsToCache = ["./", "./index.html"];
 
 // Install event - cache core assets
--- a/src/browser/contexts/ThemeContext.tsx
+++ b/src/browser/contexts/ThemeContext.tsx
@@ -67,8 +67,8 @@ const THEME_COLORS: Record<ThemeMode, string> = {
 };
 
 const FAVICON_BY_SCHEME: Record<"light" | "dark", string> = {
-  light: "/favicon.ico",
-  dark: "/favicon-dark.ico",
+  light: "favicon.ico",
+  dark: "favicon-dark.ico",
 };
 
 /** Map theme mode to CSS color-scheme value */
--- a/src/browser/main.tsx
+++ b/src/browser/main.tsx
@@ -55,8 +55,14 @@ if ("serviceWorker" in navigator) {
     window.location.protocol === "http:" || window.location.protocol === "https:";
   if (isHttpProtocol) {
     window.addEventListener("load", () => {
+      // Resolve the SW URL and scope relative to the document's base URL so mux
+      // works when served under a path-rewriting reverse proxy. `document.baseURI`
+      // honors the `<base href>` the server emits, so a single absolute URL here
+      // covers every proxy configuration.
+      const serviceWorkerUrl = new URL("service-worker.js", document.baseURI).toString();
+      const serviceWorkerScope = new URL(".", document.baseURI).toString();
       navigator.serviceWorker
-        .register("/service-worker.js")
+        .register(serviceWorkerUrl, { scope: serviceWorkerScope })
         .then((registration) => {
           console.log("Service Worker registered:", registration);
         })
--- a/src/node/orpc/server.test.ts
+++ b/src/node/orpc/server.test.ts
@@ -6,7 +6,7 @@ import { WebSocket, WebSocketServer } from "ws";
 import { RPCLink as HTTPRPCLink } from "@orpc/client/fetch";
 import { createORPCClient } from "@orpc/client";
 import type { RouterClient } from "@orpc/server";
-import { createOrpcServer, DESKTOP_WS_PATH } from "./server";
+import { computeBaseHrefForRequestUrl, createOrpcServer, DESKTOP_WS_PATH } from "./server";
 import type { ORPCContext } from "./context";
 import type { AppRouter } from "./router";
 import { Config } from "@/node/config";
@@ -148,6 +148,23 @@ async function withProxyUriTemplateEnv<T>(
   }
 }
 
+describe("computeBaseHrefForRequestUrl", () => {
+  test("handles boundary request URLs", () => {
+    // Empty / nullish input falls through to `/` (defensive; callers always
+    // pass `req.url` which Express populates, but worth locking down).
+    expect(computeBaseHrefForRequestUrl("")).toBe("./");
+    // Hash fragments are not present in `req.url` per Node's HTTP parser,
+    // but if one ever shows up it should be tolerated (depth is counted
+    // from the path portion only).
+    expect(computeBaseHrefForRequestUrl("/a/b#anchor")).toBe("./../");
+    // Double slashes are a no-op in practice (Express normalizes them),
+    // but document the current behavior: an extra empty segment raises the
+    // depth by one. Not a correctness concern either way; pinning it here
+    // keeps the algorithm intentional rather than accidental.
+    expect(computeBaseHrefForRequestUrl("//a/b")).toBe("./../../");
+  });
+});
+
 describe("createOrpcServer", () => {
   test("serveStatic fallback does not swallow /api routes", async () => {
     // Minimal context stub - router won't be exercised by this test.
@@ -278,6 +295,44 @@ describe("createOrpcServer", () => {
     }
   });
 
+  test("SPA shell placeholder substitution is scoped to the injected <base href> attribute", async () => {
+    // Regression: an earlier revision used `replaceAll(PLACEHOLDER, …)`, which
+    // would rewrite any stray occurrence of the placeholder token elsewhere in
+    // the shell (e.g. in an inline comment, docs snippet, or code sample) and
+    // silently corrupt user content. The substitution must be scoped to the
+    // `href="…"` attribute we injected at startup.
+    const stubContext: Partial<ORPCContext> = {};
+
+    const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), "mux-placeholder-scope-"));
+    const sentinel = "sentinel:__MUX_BASE_HREF__";
+    const indexHtml = `<!doctype html><html><head><title>mux</title></head><body><!-- ${sentinel} --><div>ok</div></body></html>`;
+
+    let server: Awaited<ReturnType<typeof createOrpcServer>> | null = null;
+    try {
+      await fs.writeFile(path.join(tempDir, "index.html"), indexHtml, "utf-8");
+
+      server = await createOrpcServer({
+        host: "127.0.0.1",
+        port: 0,
+        context: stubContext as ORPCContext,
+        authToken: "test-token",
+        serveStatic: true,
+        staticDir: tempDir,
+      });
+
+      // Deep request path: the injected <base href> resolves to "./../../", but
+      // the sentinel embedded in the HTML body must survive untouched.
+      const res = await fetch(`${server.baseUrl}/a/b/c`);
+      expect(res.status).toBe(200);
+      const html = await res.text();
+      expect(html).toContain('<base href="./../../"');
+      expect(html).toContain(sentinel);
+    } finally {
+      await server?.close();
+      await fs.rm(tempDir, { recursive: true, force: true });
+    }
+  });
+
   test("SPA shell uses a per-request relative <base href> that works under any path-stripping proxy", async () => {
     // <base href> is computed as a relative climb from the request URL back to
     // the SPA root. This is proxy-agnostic: it works for direct access, for
@@ -314,6 +369,8 @@ describe("createOrpcServer", () => {
         // Nested SPA routes climb one level per intermediate directory segment.
         { url: "/a/b", expectedBaseHref: "./../" },
         { url: "/a/b/c", expectedBaseHref: "./../../" },
+        // Deeper nested routes with trailing slash climb one more level.
+        { url: "/a/b/c/", expectedBaseHref: "./../../../" },
         // Query strings are ignored for depth computation.
         { url: "/a/b?x=1", expectedBaseHref: "./../" },
       ];
--- a/src/node/orpc/server.test.ts
+++ b/src/node/orpc/server.test.ts
@@ -389,6 +389,62 @@ describe("createOrpcServer", () => {
     }
   });
 
+  test("SPA shell emits a slashless-root redirect script only when the request URL is '/'", async () => {
+    // When a reverse proxy mounts mux at a slashless path (e.g.
+    // `https://proxy.example.com/apps/mux`) and strips the prefix to `/`, the
+    // browser resolves `<base href="./">` against the document URL and drops
+    // the `mux` segment, breaking asset loading. A blocking inline script in
+    // `<head>` redirects to the trailing-slash form before the parser reaches
+    // `<base>`, so the corrected document URL produces the right base href on
+    // retry. The script must only fire at the app root: any deeper request
+    // path already resolves correctly, and redirecting would bounce legitimate
+    // SPA deep links.
+    const stubContext: Partial<ORPCContext> = {};
+
+    const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), "mux-slashless-redirect-"));
+    const indexHtml =
+      "<!doctype html><html><head><title>mux</title></head><body><div>ok</div></body></html>";
+
+    let server: Awaited<ReturnType<typeof createOrpcServer>> | null = null;
+    try {
+      await fs.writeFile(path.join(tempDir, "index.html"), indexHtml, "utf-8");
+
+      server = await createOrpcServer({
+        host: "127.0.0.1",
+        port: 0,
+        context: stubContext as ORPCContext,
+        authToken: "test-token",
+        serveStatic: true,
+        staticDir: tempDir,
+      });
+
+      // Marker string unique to the redirect script. Chosen so we don't match
+      // incidental occurrences of `location.replace` in user HTML.
+      const scriptMarker = 'if(!location.pathname.endsWith("/"))';
+
+      // Root request: script must be present, and it must precede `<base>` in
+      // the document so it executes before the browser resolves asset URLs.
+      const rootRes = await fetch(`${server.baseUrl}/`);
+      expect(rootRes.status).toBe(200);
+      const rootHtml = await rootRes.text();
+      expect(rootHtml).toContain(scriptMarker);
+      expect(rootHtml.indexOf(scriptMarker)).toBeLessThan(rootHtml.indexOf("<base"));
+
+      // Deeper request paths must NOT carry the redirect: `<base href="./…/">`
+      // already lands at the SPA root for these cases, and a redirect would
+      // rewrite the URL the user asked for.
+      for (const url of ["/settings", "/a/b", "/a/b/c"]) {
+        const res = await fetch(`${server.baseUrl}${url}`);
+        expect(res.status).toBe(200);
+        const html = await res.text();
+        expect(html).not.toContain(scriptMarker);
+      }
+    } finally {
+      await server?.close();
+      await fs.rm(tempDir, { recursive: true, force: true });
+    }
+  });
+
   test("does not apply origin validation to static and SPA fallback routes", async () => {
     // Static app shell must remain reachable even if proxy/header rewriting makes
     // request Origin values unexpected. API/WS/auth routes are validated separately.
--- a/src/node/orpc/server.test.ts
+++ b/src/node/orpc/server.test.ts
@@ -175,7 +175,10 @@ describe("createOrpcServer", () => {
       expect(uiRes.status).toBe(200);
       const uiText = await uiRes.text();
       expect(uiText).toContain("mux");
-      expect(uiText).toContain('<base href="/"');
+      // `<base href>` is a relative climb from the document URL's directory
+      // back to the SPA root. `/some/spa/route` resolves against `.../some/spa/`,
+      // so needs two `../` hops.
+      expect(uiText).toContain('<base href="./../../"');
 
       const apiRes = await fetch(`${server.baseUrl}/api/not-a-real-route`);
       expect(apiRes.status).toBe(404);
@@ -215,8 +218,11 @@ describe("createOrpcServer", () => {
           expect(uiRes.status).toBe(200);
           const uiText = await uiRes.text();
 
+          // Both responses share the proxy URI template but have different
+          // relative base hrefs computed per-request.
+          expect(rootHtml).toContain('<base href="./"');
+          expect(uiText).toContain('<base href="./../../"');
           for (const html of [rootHtml, uiText]) {
-            expect(html).toContain('<base href="/"');
             expect(html).toContain("window.__MUX_PROXY_URI_TEMPLATE__ =");
             expect(html).toContain(
               'window.__MUX_PROXY_URI_TEMPLATE__ = "https://proxy-{{port}}.example.test/path\\u003c/script>";'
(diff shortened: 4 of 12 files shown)


matifali:
## End-to-end validation against a real Coder deployment with `subdomain = false`

Reproduced the full production flow on dev.coder.com using the [`coder/coder/mux`](https://github.com/coder/registry/tree/main/registry/coder/modules/mux) registry module with `install = false`, pointing at a binary built from this branch.

### Template configuration

```tf
module "mux" {
  count          = data.coder_workspace.me.start_count
  source         = "registry.coder.com/coder/mux/coder"
  version        = "~> 1.4"
  agent_id       = coder_agent.main.id
  install        = false
  install_prefix = "/home/coder/.mux-install"
  subdomain      = false
}
```

### Headless-browser verification

Loaded `https://dev.coder.com/@matifali/mux-e2e.main/apps/mux/` in a real Chromium, tracked every network response, and inspected the DOM state after load:

```text
Final URL : https://dev.coder.com/@matifali/mux-e2e.main/apps/mux/
Title     : mux

=== App-prefix responses (all 200/304) ===
  200  GET  /@matifali/mux-e2e.main/apps/mux/
  200  GET  /@matifali/mux-e2e.main/apps/mux/main-Cj-o_lxR.css
  200  GET  /@matifali/mux-e2e.main/apps/mux/__vite-browser-external-2447137e-CRGctVDH.css
  200  GET  /@matifali/mux-e2e.main/apps/mux/manifest.json
  200  GET  /@matifali/mux-e2e.main/apps/mux/__vite-browser-external-2447137e-CVaVFNiI.js
  200  GET  /@matifali/mux-e2e.main/apps/mux/Geist-Variable-C_ya05Eu.woff2
  200  GET  /@matifali/mux-e2e.main/apps/mux/main-Dd0Bg98M.js
  200  GET  /@matifali/mux-e2e.main/apps/mux/icon-192.png
  200  GET  /@matifali/mux-e2e.main/apps/mux/favicon.ico
  304  GET  /@matifali/mux-e2e.main/apps/mux/favicon.ico

=== DOM state ===
  <base href>         : ./
  #root innerHTML len : 7635
  boot loader visible : false   (SPA fully initialized)

PASS (mux SPA loaded without any 404s through the path proxy)
```

Specifically:
- The exact asset from the original bug report (`__vite-browser-external-2447137e-*.css/.js`) now loads with HTTP 200 under the proxy prefix.
- Server-rendered `<base href="./">` is the correct relative climb for request `/` (which is what mux sees after Coder's path-app proxy strips the `/@user/ws.agent/apps/mux` prefix).
- The SPA fully initializes (React mounts, boot-loader dismissed).
- No page-level errors, no console errors, no request failures.

Confirms the fix works end-to-end for the stated consumer (Coder) through the proven deployment mechanism (the registry module with `subdomain = false`).

> 🤖 This response was generated by Coder Agents.

src/node/orpc/server.ts
```diff
@@ -113,14 +113,56 @@ function extractBearerToken(header: string | undefined): string | null {
...
+  if (depth === 0) {
+    return "./";
+  }
+  return "./" + "../".repeat(depth);
+}
+
+function renderSpaShellForRequest(template: string, requestUrl: string): string {
+  const baseHref = computeBaseHrefForRequestUrl(requestUrl);
+  return template.replaceAll(BASE_HREF_PLACEHOLDER, baseHref);
 }
```
matifali, line 193: **P1** `replaceAll(BASE_HREF_PLACEHOLDER, …)` has no scope — any stray occurrence of `__MUX_BASE_HREF__` anywhere in the SPA shell gets substituted. *(Edge-Case Analyst P1, Structural Analyst P1, Security Reviewer Obs)*

> If a user's bundler or documentation accidentally includes the literal string `__MUX_BASE_HREF__` anywhere in the HTML or inline scripts (e.g., in a `<pre>` block with examples, in a comment documenting the templating system, or in a JS chunk that logs or reflects this string), it will be replaced with the computed base href, corrupting that content.

> *Structural Analyst adds:* The current mitigation is informal: relying on the string being sufficiently obscure. A more robust structural approach would be to either (1) use a collision-resistant mechanism like a hash suffix (`__MUX_BASE_HREF_<random>__`), (2) use a different substitution strategy (e.g., a JSON blob with all per-request values), or (3) document this assumption explicitly in a code comment warning against this literal string in content.

The practical risk today is low — mux controls `index.html` and the inline JS — but this is a latent footgun worth sealing now since the fix is one line. The simplest patch is to scope the replacement to the attribute you actually injected:

```ts
return template.replace(`href="${BASE_HREF_PLACEHOLDER}"`, `href="${baseHref}"`);
```

That's faster (single `replace`, bounded scan), keeps the precompute-at-startup strategy intact, and removes the "what if future content happens to contain the token" failure mode entirely.

src/browser/main.tsx
```diff
@@ -55,8 +55,14 @@ if ("serviceWorker" in navigator) {
     window.location.protocol === "http:" || window.location.protocol === "https:";
```
matifali, line 55: **P1** `useUrlSync` in `RouterContext.tsx` calls `window.history.replaceState(null, "", url)` with a router-absolute path (e.g. `"/workspace/abc"`), which strips the proxy prefix from the browser URL bar — `<base href>` does not influence `pushState`/`replaceState`. *(Frontend Reviewer P1)*

> React Router internally calls `window.history.pushState(state, "", url)` with these absolute paths. When `<base href>` is present, the browser resolves `url` **relative to `<base>`**.

Flagging the exact mechanism — I empirically verified with a minimal proxy-prefix repro that `history.pushState(state, "", "/settings")` in a document with `<base href="./">` *ignores* the base tag and navigates to `/settings` (origin-relative), dropping any proxy prefix. Per the HTML spec, `pushState` URL resolution uses the document's URL, not the base URL.

Call chain in this repo:

- `src/browser/contexts/RouterContext.tsx:263` — `window.history.replaceState(null, "", url)` where `url = location.pathname + location.search + location.hash` from React Router's internal `MemoryRouter` (never saw the proxy prefix).
- `src/browser/App.tsx:875` — `window.history.pushState({ mux: true }, "", correctUrl)` for the popstate handler, using `window.location.origin + pathname` from the same MemoryRouter location.

Concrete scenario: user at `https://coder.example.com/@user/ws.main/apps/mux/` clicks "Settings". MemoryRouter transitions to `/settings`. `useUrlSync` fires `replaceState(null, "", "/settings")`. Browser address bar updates to `https://coder.example.com/settings` — prefix gone. If the user then hits refresh, the request goes to `/settings` (bypassing the Coder app proxy entirely) → 404.

Frontend Reviewer's recommendation is right: the router has to either (a) resolve against `document.baseURI` before calling `replaceState`, or (b) compose absolute URLs using `window.location.pathname` up to the SPA root as a prefix. Option (b) is probably simpler given the existing `getBrowserBackendBaseUrl()` helper — it already detects the `/@user/ws.*/apps/<slug>` prefix, so `useUrlSync` can do `replaceState(null, "", `${proxyPrefix}${url}`)`.

This is pre-existing code (the PR doesn't touch `useUrlSync`), but it's newly reachable: before this PR the SPA shell 404'd under the proxy, so nobody hit the navigation bug. The end-to-end story isn't complete until this is fixed. Happy to scope it into this PR or track as an immediate follow-up — but it shouldn't merge as "mux works behind a path-rewriting proxy" without it.

matifali (commented):
Clean, proxy-agnostic approach to the subpath problem — the code-server pattern is the right call and the placeholder-at-startup / replace-per-request split keeps the hot path tight.

Two P1s, one P2, one P3, plus a few obs/nits across 7 inline comments. The second P1 is pre-existing in `useUrlSync` (not touched by this PR) but is newly reachable now that the shell actually loads under the proxy — worth scoping in or filing a follow-up before calling the end-to-end story "done". This review contains findings that may need attention before merge.

public/service-worker.js
```diff
@@ -1,6 +1,6 @@
 // mux Service Worker for PWA support
 const CACHE_NAME = "mux-v1";
-const urlsToCache = ["/", "/index.html"];
+const urlsToCache = ["./", "./index.html"];
```
matifali, line 4: **P2** Existing PWA installs may keep an old service worker registered at scope `/`; the new registration at a narrower scope derived from `document.baseURI` can leave the stale one active. *(Contract Auditor P1, Edge-Case Analyst P2)*

> The PR changes SW registration from a hardcoded `/` scope to a computed scope based on `document.baseURI`. Existing users with the PWA installed will have an old Service Worker registered at scope `/`. When they visit the page again, the browser will register a new SW at a more-specific scope (e.g., `./`, `./../../`). Depending on the browser's PWA scope resolution rules, both registrations may coexist, or the old one may remain active and the new one unregistered — either way, this creates an undefined state where the active SW scope no longer matches the computed scope.

Downgrading from Contract Auditor's P1 because `public/service-worker.js` already calls `self.skipWaiting()` on install and `self.clients.claim()` on activate, and the activate handler deletes every cache whose name isn't `mux-v1` — so the next visit on an updated mux will claim clients and swap the SW. Old clients eventually converge.

What's still worth doing:

1. Bump `CACHE_NAME` to `mux-v2` so any cache keyed on pre-PR absolute URLs (`"/index.html"` etc.) is actively purged instead of relying on a clean install.
2. Note in release notes that a hard refresh may be needed for users who had the PWA installed before this release.

No new behavior needed in the registration path — the claim/skipWaiting pair is doing its job.

src/node/orpc/server.test.ts
```diff
@@ -272,6 +278,60 @@ describe("createOrpcServer", () => {
...
+
+    const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), "mux-static-base-href-"));
+    const indexHtml =
+      "<!doctype html><html><head><title>mux</title></head><body><div>ok</div></body></html>";
+
+    let server: Awaited<ReturnType<typeof createOrpcServer>> | null = null;
+    try {
+      await fs.writeFile(path.join(tempDir, "index.html"), indexHtml, "utf-8");
+
+      server = await createOrpcServer({
```
matifali, line 353: **P3** New test table covers the happy path cleanly but leaves a couple of inputs the computation actually handles asymmetrically. *(Test Auditor P2)*

> - **Double slashes** (`//a/b`): The function produces `./../` for `//a/b` [split to `["", "a", "b"]` → depth=2 → `./../`].
> - **URLs with hash fragments** (`/a/b#anchor`): Fragment is present in `req.url` but the split on `?` ignores it.
> - **Trailing slashes on nested paths** (`/a/b/c/`): The test covers `/settings/` → `"./../"` but does NOT verify `/a/b/c/` → `"./../../../"` (4 segments including empty, depth=3).
> - **Purely empty path**: the only safeguard against a falsy req.url; no test validates this fallback.

Downgrading from Test Auditor's P2 because none of these have a concrete user-visible consequence in realistic traffic (Express normalizes double slashes, Node's HTTP parser doesn't put the fragment in `req.url`), but the test table is already the shape of a parameterized table so adding `"" → "./"`, `"/a/b/c/" → "./../../../"`, and `"/a/b#anchor" → "./../"` is near-zero cost and would nail down the behavior as intentional rather than accidental. Worth doing while the feature is fresh.

src/browser/contexts/ThemeContext.tsx
```diff
@@ -67,8 +67,8 @@ const THEME_COLORS: Record<ThemeMode, string> = {
 };
 
 const FAVICON_BY_SCHEME: Record<"light" | "dark", string> = {
-  light: "/favicon.ico",
-  dark: "/favicon-dark.ico",
+  light: "favicon.ico",
+  dark: "favicon-dark.ico",
```
matifali, line 71: **Obs** `FAVICON_BY_SCHEME` switched to relative URLs, which couples theme-driven favicon swaps to whatever `<base href>` is in effect at the moment the swap runs. *(Contract Auditor P2, Frontend Reviewer P3)*

> If `<base href>` points away from the public root (e.g., `./../` for nested routes), the favicon URL resolves incorrectly. For example: document at `https://example.com/app/settings/`, `<base href="./../">` resolves to `https://example.com/app/`, `favicon.ico` resolves to `https://example.com/app/favicon.ico` (not `https://example.com/favicon.ico`).

I confirmed in the dev.coder.com E2E run that `favicon.ico` and `manifest.json` both returned 200 under the Coder app proxy at `/@matifali/mux-e2e.main/apps/mux/`, so the current behavior is correct under the proxy configuration this PR targets. The reasoning in the reviews does hold for test/Storybook environments where `<base href>` isn't injected, but the theme-swap code path is production-only. Leaving as Obs — no change needed, but if a future deployment ends up serving the favicon from a fixed origin path only, this is where to come back to.

public/service-worker.js
```diff
@@ -1,6 +1,6 @@
 // mux Service Worker for PWA support
```
matifali, line 1: **Obs** Offline fallback for nested routes: the service worker is network-first and caches only `./` and `./index.html`, so a hard refresh at `/settings` while offline returns a 503 instead of the SPA shell. *(Frontend Reviewer P2)*

> These are relative paths resolved against the SW's own URL. The SW lives at `/service-worker.js` (backend mux, not the subpath). When the user is at `/a/b/c` and the SW cache misses, the SW tries to fetch `/a/b/c` from the network, which fails and returns a 503 error.

This is pre-existing — the SW behaved the same way before this PR, only the cache keys changed from absolute to relative. Not a regression. The right fix is to detect navigation requests in the fetch handler and fall back to the cached `./index.html` on network failure, but that's a separate piece of work. Calling it out here so it doesn't get lost now that the proxy-path story is end-to-end tested.

src/node/orpc/server.ts
```diff
@@ -113,14 +113,56 @@ function extractBearerToken(header: string | undefined): string | null {
   const token = header.slice(7).trim();
   return token.length ? token : null;
 }
-function injectBaseHref(indexHtml: string, baseHref: string): string {
+// Placeholder substituted per-request with the relative root for the current
+// request URL. Using a placeholder keeps the parse/serialize work out of the
+// hot path: we precompute the HTML with a placeholder once at startup and do a
+// single string replace per SPA shell response.
+const BASE_HREF_PLACEHOLDER = "__MUX_BASE_HREF__";
+
```
matifali, line 119: **Nit** Comments in the new helpers are a little long for what they explain. *(Style Reviewer)*

> The comment explains *why* a placeholder is used ("keeps the parse/serialize work out of the hot path: we precompute the HTML with a placeholder once at startup"). This is reasonable context, but the tone reads like justifying to a skeptic rather than briefing a peer.

> *Also:* `spaIndexHtmlTemplate` is accurate but reads long in context. `spaHtmlTemplate` would be equally clear.

Non-blocking. I'd trim `BASE_HREF_PLACEHOLDER`'s comment to one line ("Placeholder substituted per-request; precompute once at startup.") and leave the JSDoc on `computeBaseHrefForRequestUrl` as-is (the examples carry their weight). The `spaIndexHtmlTemplate` rename is fine either way.

src/node/orpc/server.ts
matifali: Addressed in e2df17c0. `renderSpaShellForRequest` now does a single scoped replace against the exact `href="__MUX_BASE_HREF__"` attribute we injected, so stray tokens elsewhere in the shell survive untouched. Added a regression test (`SPA shell placeholder substitution is scoped to the injected <base href> attribute`) that embeds a `__MUX_BASE_HREF__` sentinel in the HTML body and asserts it isn't rewritten.

> 🤖 This response was generated by Coder Agents.

src/browser/main.tsx
matifali: Addressed in e2df17c0. New `src/browser/utils/appBasePath.ts` reads `document.baseURI` (which the server-injected `<base href>` climbs to the SPA root) to recover the public prefix, then:

- `getInitialRoute()` strips the prefix before seeding MemoryRouter, so route patterns like `/settings`, `/workspace/:id` still match under a proxy.
- `useUrlSync()` re-applies the prefix before `replaceState`, so the URL bar keeps `/@user/ws.main/apps/mux/settings` instead of collapsing to `/settings`.
- `App.tsx`'s popstate handler does the same for `pushState`.

Three new RouterContext test cases lock this in: initial-route stripping, navigation preserving the prefix, and a sanity check that no-proxy deployments remain unchanged. Plus 6 unit tests in `appBasePath.test.ts` covering the helper directly.

The approach is proxy-agnostic (any prefix-stripping proxy that emits a `<base href>` works, not just Coder) and Electron's `file://` mode short-circuits to identity behavior.

> 🤖 This response was generated by Coder Agents.

public/service-worker.js
matifali: Addressed in e2df17c0. Bumped `CACHE_NAME` from `mux-v1` to `mux-v2`. The activate handler already deletes every cache whose name isn't the current `CACHE_NAME`, so old installs with pre-PR absolute-path cache entries (`"/index.html"`) get purged on next activation. Combined with `skipWaiting()` + `clients.claim()` that's already in place, the SW swap converges without manual reinstall.

> 🤖 This response was generated by Coder Agents.

src/node/orpc/server.test.ts
matifali: Addressed in e2df17c0. Two additions:

1. Exported `computeBaseHrefForRequestUrl` and added a dedicated `describe("computeBaseHrefForRequestUrl")` block with boundary cases: empty input → `./`, hash fragment → depth counted from path only, double slash → current behavior pinned as intentional.
2. Extended the per-request table with `/a/b/c/` → `./../../../` to cover deeper trailing-slash handling end-to-end.

> 🤖 This response was generated by Coder Agents.

src/browser/contexts/ThemeContext.tsx
matifali: No change. E2E on dev.coder.com confirmed favicons return 200 under the Coder app proxy, and the reasoning about Storybook/test environments doesn't apply to the theme-swap code path (production-only). Leaving as-is.

> 🤖 This response was generated by Coder Agents.

public/service-worker.js
matifali: Not addressed in this PR — pre-existing SW behavior, not a regression. Will file as a follow-up: the fetch handler should detect navigation requests (mode === `"navigate"`) and fall back to the cached `./index.html` on network failure, so offline hard-refreshes at nested routes land on the SPA shell instead of a 503. Out of scope here since it's orthogonal to the subpath fix.

> 🤖 This response was generated by Coder Agents.

src/node/orpc/server.ts
matifali: Trimmed in e2df17c0. `BASE_HREF_PLACEHOLDER` comment is down to one line ("Placeholder substituted per-request with the relative root for the current request URL. Precomputed once at startup; replaced per SPA shell response.") and the `spaIndexHtmlTemplate` comment went from 6 lines to 3. Kept the JSDoc with examples on `computeBaseHrefForRequestUrl` since the examples carry their weight.

> 🤖 This response was generated by Coder Agents.

matifali:
@codex review

matifali:
Addressed the slashless-root P1 finding in a625ef2a.

**Fix:** `renderSpaShellForRequest` now emits a blocking inline script in `<head>` (before `<base>`) that upgrades slashless app-root URLs to the trailing-slash form via `location.replace`. The script only fires when the backend sees `req.url === "/"` — deeper paths already resolve correctly against `<base href="./…/">` and redirecting would bounce legitimate SPA deep links.

**Coverage:** new test `SPA shell emits a slashless-root redirect script only when the request URL is '/'` asserts the script is present for `/` (and precedes `<base>` in document order) and absent for `/settings`, `/a/b`, `/a/b/c`.

@codex review

> 🤖 This response was generated by Coder Agents.

matifali:
@ibetitsmike @ammario am I good to merge this one?

matifali:
Closed in favor of #3194 and #3195

PR #3185, opened 2026-04-22
🤖 feat: include advisor (and tool-level) model costs in analytics

ibetitsmike:
## Summary

Extend analytics so advisor (and future tool-level) model costs flow durably into the analytics DB. Each assistant turn now records per-invocation tool model usage on the message, which the ETL emits as separate `events` rows alongside the base assistant row.

> Mux is working on this on behalf of Mike.

## Background

- `advisor.ts` already reports advisor usage via `reportModelUsage(...)`, which `aiService.ts` forwards into `SessionUsageService.recordUsage(...)`.
- `session-usage.json` only persists `byModel` totals with no source/tool dimension.
- Analytics ETL reads `chat.jsonl` usage plus `session-usage.json` for `rolledUpFrom` only, so advisor usage never reached DuckDB.
- There was no durable per-invocation field on messages, so advisor costs were session-only and analytics-invisible.

## Implementation

**New durable shape (tool-agnostic).**
- `MuxMetadata.toolModelUsages?: PersistedToolModelUsage[]` on assistant messages. Each entry carries `toolName`, `toolCallId`, `timestamp`, `model`, `usage`, `providerMetadata`, and an optional `metadataModel` for pricing resolution.

**Live capture, durable persistence.**
- `aiService.ts`: `reportModelUsage` snapshots synchronously onto the active stream via `StreamManager.recordToolModelUsage(...)` before the existing live-usage forward to `SessionUsageService`. Assistant message id creation moved earlier so it is available at callback time. Live session usage behavior unchanged.
- `streamManager.ts`: per-stream accumulator on `WorkspaceStreamInfo`, gated by message id so stale prior-turn events are ignored. The accumulator is persisted to `metadata.toolModelUsages` only when non-empty on final message write.

**Session usage rebuild.**
- `sessionUsageService.ts` rebuild folds both `metadata.usage` and every `metadata.toolModelUsages[]` entry into `byModel`, tolerating absence and partial `providerMetadata`, so rebuild matches live totals.

**Analytics schema, ETL, queries.**
- Nullable `tool_name TEXT` on `events` (schema + ingest/row schemas). Startup `ALTER TABLE events ADD COLUMN IF NOT EXISTS tool_name TEXT` keeps existing DBs working.
- ETL emits one base assistant row (`tool_name = NULL`) plus one row per tool invocation, inheriting `workspace_id`, `parent_workspace_id`, `agent_id`, `is_sub_agent` from the parent turn. Tool rows prefer the tool timestamp and leave timing metrics null. Pricing uses the tool's model, not the parent's.
- Spend and token aggregations include tool rows. `total_responses`, `response_count`, `request_count`, and timing queries exclude tool rows via a shared non-tool predicate.

**Refactor.**
- Shared row builder used by both assistant and tool paths so they cannot drift.
- Centralized non-tool filter for count and timing queries.

## Validation

- `bun test src/node/services/streamManager.test.ts src/node/services/sessionUsageService.test.ts src/node/services/analytics/etl.test.ts src/node/services/analytics/queries.test.ts`: 92 pass / 2 skip / 0 fail.
- `bun test src/node/services/tools/advisor.test.ts src/node/services/aiService.test.ts`: 77 pass / 0 fail.
- `make typecheck` and `make lint`: clean.

New behavioral tests locked in:
- Per-invocation tool model usages persist on the final assistant message; scoped to the active turn; omitted when empty.
- Analytics emits one assistant row plus one row per tool invocation with inherited context.
- Spend and token totals include tool rows; response counts and timing exclude them.
- Session usage rebuild restores advisor costs from the new durable field.

## Risks

- Schema change to `events` is additive via `ADD COLUMN IF NOT EXISTS`; old rows keep `tool_name = NULL` and flow through unchanged.
- Query changes are scoped: spend and token aggregations broadened to include tool rows; response/timing queries explicitly filter `tool_name IS NULL`. Any existing query that relied on implicit one-row-per-turn semantics could double-count, so all count/timing paths were audited and updated to the shared predicate.
- Live session usage recording (`SessionUsageService.recordUsage(...)`) is unchanged. Only the rebuild path was extended.

## Non-goals / follow-ups

- No historical backfill of advisor spend. An opt-in audit showed residuals are partially recoverable at session-aggregate granularity (distinct-model case only) but lose per-invocation timestamps and cannot separate same-model collisions, so forward-only was chosen.
- No advisor-specific UI badges or tooltips. Same-model parent + advisor rows roll into the same `by_model` slice today. Explicit source splitting is a follow-up if product wants it.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `xhigh` • Cost: `$31.19`_

### DIFF
--- a/src/common/analytics/schemaSql.ts
+++ b/src/common/analytics/schemaSql.ts
@@ -9,6 +9,7 @@ CREATE TABLE IF NOT EXISTS events (
   timestamp BIGINT,
   date DATE,
   model VARCHAR,
+  tool_name TEXT,
   thinking_level VARCHAR,
   input_tokens INTEGER DEFAULT 0,
   output_tokens INTEGER DEFAULT 0,
--- a/src/common/analytics/schemaSql.ts
+++ b/src/common/analytics/schemaSql.ts
@@ -9,7 +9,6 @@ CREATE TABLE IF NOT EXISTS events (
   timestamp BIGINT,
   date DATE,
   model VARCHAR,
-  tool_name TEXT,
   thinking_level VARCHAR,
   input_tokens INTEGER DEFAULT 0,
   output_tokens INTEGER DEFAULT 0,
@@ -27,7 +26,8 @@ CREATE TABLE IF NOT EXISTS events (
   tool_execution_ms DOUBLE,
   output_tps DOUBLE,
   response_index INTEGER,
-  is_sub_agent BOOLEAN DEFAULT false
+  is_sub_agent BOOLEAN DEFAULT false,
+  tool_name TEXT
 )
 `;
 
--- a/src/common/orpc/schemas/analytics.ts
+++ b/src/common/orpc/schemas/analytics.ts
@@ -143,6 +143,7 @@ export const EventRowSchema = z.object({
   agent_id: z.string().nullable(),
   timestamp: z.number().nullable(), // unix ms
   model: z.string().nullable(),
+  tool_name: z.string().nullable(),
   thinking_level: z.string().nullable(),
   input_tokens: z.number().default(0),
   output_tokens: z.number().default(0),
--- a/src/common/types/message.ts
+++ b/src/common/types/message.ts
@@ -408,6 +408,17 @@ export function isCompactionSummaryMetadata(
   return metadata?.type === "compaction-summary";
 }
 
+export interface PersistedToolModelUsage {
+  toolName: string;
+  toolCallId?: string;
+  timestamp?: number;
+  model: string;
+  /** Resolved pricing model for token/cost metadata lookups when the tool model uses Treat as mapping. */
+  metadataModel?: string;
+  usage: LanguageModelV2Usage;
+  providerMetadata?: Record<string, unknown>;
+}
+
 // Our custom metadata type
 export interface MuxMetadata {
   historySequence?: number; // Assigned by backend for global message ordering (required when writing to history)
@@ -436,6 +447,8 @@ export interface MuxMetadata {
   contextUsage?: LanguageModelV2Usage;
   // Aggregated provider metadata across all steps (for cost calculation)
   providerMetadata?: Record<string, unknown>;
+  // Per-tool invocation usage snapshots recorded during this assistant turn.
+  toolModelUsages?: PersistedToolModelUsage[];
   // Last step's provider metadata (for context window cache display)
   contextProviderMetadata?: Record<string, unknown>;
   systemMessageTokens?: number; // Token count for system message sent with this request (calculated by AIService)
--- a/src/node/services/aiService.ts
+++ b/src/node/services/aiService.ts
@@ -1409,6 +1409,11 @@ export class AIService extends EventEmitter {
           costsUsd: sessionCostsUsd,
         }
       );
+
+      // Create assistant message ID early so tool-side usage reporting and nested tool events
+      // stay scoped to this specific assistant turn. The placeholder is appended to history below
+      // (after the abort check).
+      const assistantMessageId = createAssistantMessageId();
       const allTools = await getToolsForModel(
         modelString,
         {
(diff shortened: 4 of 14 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

> Mux is working on this on behalf of Mike.

Addressed both review threads in `0a0170344`:

- **P1 (sessionUsageService)**: rebuild now runs each `metadata.toolModelUsages` entry through an `isPersistedToolModelUsage` guard before dereferencing. Malformed entries (null, primitive, missing `model`, missing/non-object `usage`, etc.) are silently skipped so corrupted history does not abort rebuild. Added a regression test with a mix of valid + malformed entries that previously threw.
- **P2 (analytics ETL)**: `extractIngestEvents` now parses assistant rows and tool rows independently. If `metadata.usage` is missing or invalid, valid `toolModelUsages` still emit tool rows with inherited workspace / sub-agent context and per-entry timestamps. Added a regression test asserting zero assistant rows and N tool rows when only tool usage is present.

No schema or query changes.

ibetitsmike:
@codex review

> Mux is working on this on behalf of Mike.

Addressed the new review pass in `2fd16be69`:

- **P1 (etl.ts:135 / DuckDBAppender column order)**: legitimate and serious. `schemaSql.ts` put `tool_name` mid-table while `ALTER TABLE ADD COLUMN` in `analyticsWorker.ts` appends it at the end, so on migrated DBs the appender was shifting every column after `model` by one slot. Fixed by moving `tool_name` to the last column of `CREATE TABLE events` and moving the corresponding `appendVarcharOrNull` call to the end of `appendEvents`. Fresh and migrated DBs now share identical physical column order.
- Added a regression test that creates a fresh events table from the CREATE SQL, simulates a migrated table via the exact `ALTER TABLE ADD COLUMN IF NOT EXISTS`, inserts the same row into both via `appendEvents`, and asserts `tool_name`/`thinking_level`/`input_tokens` land in the correct columns. This failed before the fix and passes after.
- **P2 (sessionUsageService tool usage validation)**: this was already addressed in `0a0170344` via `isPersistedToolModelUsage` / `isPlainRecord` guards that skip malformed entries. The rebuild path no longer dereferences unvalidated elements. Regression test for malformed entries is in `sessionUsageService.test.ts`.

ibetitsmike:
@codex review

> Mux is working on this on behalf of Mike.

Addressed the P2 in `98a7741a4`:

- **etl.ts:761 (head signature for tool-only workspaces)**: removed `AND tool_name IS NULL` from `readPersistedWorkspaceHeadSignature`. The persisted side now picks the same "first event" that `extractIngestEvents` emits, using deterministic ordering: `response_index ASC NULLS LAST, CASE WHEN tool_name IS NULL THEN 0 ELSE 1 END ASC, timestamp ASC`. Assistant row is preferred at a given response index; tool-only workspaces fall back to the first tool row instead of returning null, so head-mismatch divergence is caught and rebuild can trigger.
- Regression tests added to `etl.test.ts`: (1) end-to-end — mutate a tool-only workspace's head row and verify the safeguard triggers rebuild; (2) unit — assistant row is preferred when both exist at the same response index. Also exported `readPersistedWorkspaceHeadSignature` so the ordering contract is directly testable.

Confirmed symmetric with `extractIngestEvents` (assistant row emitted before tool rows; tool rows inherit the parent `response_index`).

ibetitsmike:
@codex review

> Mux is working on this on behalf of Mike.

The latest review pass flagged the head-signature filter again, but the current HEAD (`98a7741a4`) already addressed this in the previous round. Please re-read the current `readPersistedWorkspaceHeadSignature`:

```sql
SELECT timestamp, model, total_cost_usd
FROM events
WHERE workspace_id = ?
ORDER BY response_index ASC NULLS LAST,
         CASE WHEN tool_name IS NULL THEN 0 ELSE 1 END ASC,
         timestamp ASC
LIMIT 1
```

There is no `WHERE tool_name IS NULL` filter any more. `tool_name` only appears as an ordering tie-break so the assistant row is preferred when both exist at the same `response_index`. Tool-only workspaces now return a non-null head signature (the first tool row), keeping the safeguard symmetric with `parsedEvents[0]`. Regression tests in `etl.test.ts` cover both cases.

PR #3187, opened 2026-04-23
🤖 fix: restore starlark code block highlighting

coadler:
## Summary
Use a shared Shiki language fallback so Starlark/Bazel fenced code blocks stay highlighted instead of logging an unsupported-language warning and falling back to plain text.

## Background
Closes #3117. The shipped Shiki bundle does not include a native `starlark` grammar, so attempting to highlight those fences throws at runtime.

## Implementation
- map `starlark`, `bazel`, and `bzl` to Python in the shared Shiki language mapper
- reuse that shared mapper in the highlight worker instead of maintaining a duplicate copy
- add a regression test for the fallback mapping

## Validation
- `bun test src/browser/utils/highlighting/shiki-shared.test.ts`
- `make typecheck`
- `make lint`
- `make static-check`

## Risks
Low. This only changes how unsupported Bazel/Starlark fences are mapped for syntax highlighting, and it keeps the existing plain-text fallback if highlighting still fails for another reason.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$0.00`_

### DIFF
--- a/src/browser/utils/highlighting/shiki-shared.test.ts
+++ b/src/browser/utils/highlighting/shiki-shared.test.ts
@@ -1,6 +1,14 @@
 import { describe, expect, test } from "bun:test";
 
-import { extractShikiLines } from "./shiki-shared";
+import { extractShikiLines, mapToShikiLang } from "./shiki-shared";
+
+describe("mapToShikiLang", () => {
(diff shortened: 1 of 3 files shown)


coadler:
@codex review

PR #3188, opened 2026-04-23
🤖 feat: add "Copy link" to transcript right-click menu

coadler:
## Summary

Right-clicking a link in the chat transcript now opens the transcript context menu with a **Copy link** option.

## Background

The transcript's right-click menu previously bailed out for anchor (`a[href]`) targets so the native browser context menu could handle them (open link, copy link, etc.). That works in a standard browser, but in the Electron desktop app there is no native link context menu, so right-clicking a link did nothing. There was no way to copy a link address from the chat without manually selecting the visible text (which often differs from the href).

## Implementation

- `getTranscriptContextMenuLink()` is a new helper in `transcriptContextMenu.ts` that resolves the nearest enclosing `a[href]` for a right-click target and returns its raw `href` (or `null` if blank / outside the transcript). The existing text resolver is untouched.
- `useTranscriptContextMenu` now checks for a link first. If the target is inside an anchor, it renders a single **Copy link** item (lucide `Link` icon) and skips the quote/copy text items. Otherwise it falls through to the existing Quote/Copy text menu. Mode is tracked in state so the menu re-renders with the correct items when it opens.
- Clicking **Copy link** copies the raw href via the existing `copyToClipboard` helper (with the `onCopyText` override respected for tests).

## Validation

- 22 tests in `transcriptContextMenu.test.ts` pass (6 new cases covering direct anchor target, descendant text/element inside an anchor, target outside the transcript, non-anchor target, anchor without `href`, and anchor with a blank `href`).
- `make typecheck`, `make lint`, `make fmt-check` all pass.

image

## Risks

Low. The change is additive: it only runs when the right-click target resolves to an anchor inside the transcript. Non-link behavior (selection quoting, text copy, interactive element bypass for buttons/inputs) is unchanged, and the existing transcript text test suite continues to pass.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$3.14`_

### DIFF
--- a/src/browser/features/Messages/useTranscriptContextMenu.tsx
+++ b/src/browser/features/Messages/useTranscriptContextMenu.tsx
@@ -1,15 +1,20 @@
-import React, { useCallback, useRef } from "react";
-import { Clipboard, TextQuote } from "lucide-react";
+import React, { useCallback, useRef, useState } from "react";
+import { Clipboard, Link as LinkIcon, TextQuote } from "lucide-react";
 import { useContextMenuPosition } from "@/browser/hooks/useContextMenuPosition";
 import { copyToClipboard } from "@/browser/utils/clipboard";
 import {
   formatTranscriptTextAsQuote,
+  getTranscriptContextMenuLink,
   getTranscriptContextMenuText,
 } from "@/browser/utils/messages/transcriptContextMenu";
 import {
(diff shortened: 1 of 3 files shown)


coadler:
@codex review

PR #3190, opened 2026-04-23
🤖 feat: add GPT-5.5 as OAuth-required built-in model

ThomasK33:
## Summary

Adds **GPT-5.5** as a first-class OAuth-required built-in model for OpenAI. This is **Phase 1 of a staged rollout**: GPT-5.5 is exposed only where Mux already has a working access path today (ChatGPT Codex OAuth). The `gpt` alias and first-run defaults still point at `gpt-5.4`, and no GPT-5.5 Pro entry is introduced.

## Background

OpenAI recently announced GPT-5.5 and GPT-5.5 Pro. At the time of this change:

- GPT-5.5 is available in ChatGPT and Codex now; **API support is coming soon**.
- On the Codex route today the context window is **400K**, while the future API is advertised at **1M**.
- GPT-5.5 Pro is scoped to **ChatGPT only** — there is **no evidence** it is reachable via Codex today (see reconnaissance notes at the bottom of the plan).
- Mux's OpenAI OAuth path (`src/common/constants/codexOAuth.ts`) routes through `https://chatgpt.com/backend-api/codex/responses`, so it already has a plausible transport for the standard GPT-5.5 model.

If we had repointed the `gpt` alias to `gpt-5.5` in one shot, we would have had to lie about either today's OAuth context window (400K) or the future API one (1M), and we would have had to decide about GPT-5.5 Pro without a verified access path. Instead, this PR adds a temporary `GPT_55` registry entry gated by `CODEX_OAUTH_REQUIRED_MODELS`; a follow-up can collapse it into the `GPT` alias once the API ships with stable metadata.

## Implementation

- **Registry** (`src/common/constants/knownModels.ts`): new `GPT_55` entry mirroring the shape of `GPT` (alias `gpt-5.5`, `warm`, tokenizer `openai/gpt-5`). `GPT` is unchanged.
- **OAuth gating** (`src/common/constants/codexOAuth.ts`): `gpt-5.5` added to both `CODEX_OAUTH_ALLOWED_MODELS` and `CODEX_OAUTH_REQUIRED_MODELS`, with an inline comment explaining that the required-set membership is temporary until OpenAI ships API access.
- **Token metadata** (`src/common/utils/tokens/models-extra.ts`): conservative entry using the currently reachable Codex-route limits — 400K input, 128K output, published pricing ($5 / $0.50 cached / $30 per million), GPT-5.4-parity capability flags, `mode: "chat"`. No tiered-pricing, no `knowledge_cutoff`, no Fast-mode metadata.
[...]

PR #3191, opened 2026-04-23
🤖 fix: align gpt-5.5 pricing tiers while keeping 400K OAuth routable cap

ThomasK33:
## Summary

Follow-up to #3190. Adds the missing 272K tiered pricing fields to the `gpt-5.5` record in `src/common/utils/tokens/models-extra.ts` while keeping `max_input_tokens` at the Codex OAuth route's 400K routable cap (required for correct pre-send compaction).

## Background

#3190 introduced `gpt-5.5` as an OAuth-required built-in, but its metadata entry was flat-priced and didn't include the 272K tiered pricing boundary that the rest of the GPT-5.x family uses. Published OpenAI pricing for GPT-5.5 is `$5/M` input, `$30/M` output, `$0.50/M` cached input at the base tier, rising to `$10/M · $45/M · $1/M` above the 272K long-context boundary (2× input, 1.5× output, 2× cached — same ratios as GPT-5.4).

An earlier revision of this PR also bumped `max_input_tokens` to `1,050,000` to reflect the published API context window. Codex review correctly flagged that as a regression: `gpt-5.5` is still OAuth-only via Codex (`CODEX_OAUTH_REQUIRED_MODELS`) and the Codex route caps prompts at 400K today. Because `agentSession.checkBeforeSend` compacts only after crossing a percentage of `max_input_tokens`, advertising 1.05M would let chats in the 400K–735K range through to the route and fail with `context_exceeded` instead of being pre-compacted. The accepted plan's Phase 1 prescribes the same 400K routable cap for exactly this reason.

So this PR keeps `max_input_tokens` pinned to the route's cap and only picks up the missing pricing tiers.

## Implementation

Only `src/common/utils/tokens/models-extra.ts` changes — single `gpt-5.5` record + leading comment block.

- **`max_input_tokens: 400000`** — tracks the Codex OAuth route's cap so pre-send compaction triggers correctly. Can be raised when a non-OAuth API route lands.
- **Added tiered pricing fields:**
  - `input_cost_per_token_above_200k_tokens: 0.00001` (`$10/M`)
  - `output_cost_per_token_above_200k_tokens: 0.000045` (`$45/M`)
  - `cache_read_input_token_cost_above_200k_tokens: 0.000001` (`$1/M`)
  - `tiered_pricing_threshold_tokens: 272000`
- **Added `knowledge_cutoff: "2025-08-31"`** — matches the GPT-5.x family.
- The leading comment block now explains that `max_input_tokens` intentionally tracks the routable cap (not the underlying 1M model window) until a non-OAuth route is available, and notes that the 272K pricing boundary is still reachable within the 400K OAuth cap.

## What did NOT change

- `src/common/constants/knownModels.ts` — `GPT_55` entry, alias, tokenizer override, `warm` flag untouched.
- `src/common/constants/codexOAuth.ts` — `gpt-5.5` remains in both `CODEX_OAUTH_ALLOWED_MODELS` and `CODEX_OAUTH_REQUIRED_MODELS`. API-key-only users still don't see GPT-5.5.
- `src/common/utils/thinking/policy.ts` — reasoning policy unchanged; `gpt-5.5` continues to match the GPT-5.x 5-tier envelope.
- No `GPT` / `GPT_PRO` alias repointing; no `gpt-5.5-pro` work (both still deferred per the staged rollout plan).

## Validation

- `make typecheck` — pass.
- `bun test src/common/constants/knownModels.test.ts src/common/constants/codexOAuth.test.ts src/browser/features/Settings/Sections/ModelsSection.test.ts` — 17/17 pass.
- `bun scripts/gen_docs.ts` — runs clean; no generated-doc diff (docs table doesn't yet surface context/tier fields, out of scope here).

## Risks

Low. Metadata-only change to a single record. `max_input_tokens` is unchanged from `main`, so compaction behavior is identical to post-#3190. The new pricing tier fields affect cost estimation only when prompts cross the 272K boundary within the 400K OAuth cap — previously those would have been under-billed.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$30.93`_

### DIFF
--- a/src/common/utils/tokens/models-extra.ts
+++ b/src/common/utils/tokens/models-extra.ts
@@ -134,20 +134,33 @@ export const modelsExtra: Record<string, ModelData> = {
     knowledge_cutoff: "2025-08-31",
   },
 
-  // GPT-5.5 - currently reachable via the Codex route with a 400K context window.
-  // Published pricing: $5/M input, $30/M output, $0.50/M cached input.
+  // GPT-5.5
+  // Native model window is 1.05M context with 128K max output, but today Mux can only
+  // route GPT-5.5 through the Codex OAuth path and that route caps prompts at 400K.
+  // Keep max_input_tokens aligned to the routable cap, not the underlying model window,
+  // so pre-send compaction triggers before the OAuth route rejects 400K+ requests; when
+  // a non-OAuth API route becomes available, we can raise this to the published window.
+  // Base pricing: $5/M input, $30/M output, $0.50/M cached input.
+  // Above 272K prompt tokens (still reachable within the 400K OAuth cap): $10/M input,
+  // $45/M output, $1/M cached input.
   "gpt-5.5": {
     max_input_tokens: 400000,
     max_output_tokens: 128000,
-    input_cost_per_token: 0.000005, // $5 per million input tokens
-    output_cost_per_token: 0.00003, // $30 per million output tokens
-    cache_read_input_token_cost: 0.0000005, // $0.50 per million cached input tokens
+    input_cost_per_token: 0.000005, // $5 per million input tokens (<272K prompt tokens)
+    input_cost_per_token_above_200k_tokens: 0.00001, // $10 per million input tokens (>272K)
+    output_cost_per_token: 0.00003, // $30 per million output tokens (<272K prompt tokens)
+    output_cost_per_token_above_200k_tokens: 0.000045, // $45 per million output tokens (>272K)
+    cache_read_input_token_cost: 0.0000005, // $0.50 per million cached input tokens (<272K)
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Thanks — good catch on the compaction regression risk. Reverted `max_input_tokens` back to `400000` (the Codex OAuth route cap) and kept only the 272K tiered pricing fields + `knowledge_cutoff`. PR body updated to match. Compaction behavior is now identical to post-#3190; the pricing tiers only affect cost estimation in the 272K–400K range reachable via the OAuth cap.

PR #3201, opened 2026-04-27
🤖 fix: stabilize streaming transcript seams

ammar-agent:
## Summary

Stabilizes the chat transcript bottom seam during streaming, hydration, workspace switches, idle chat opens, active compaction, and chats with expanded tool panes — and removes the noticeable delay before the latest streaming bash row expands when a chat is opened.

## Background

Earlier rewrites already replaced the old multi-path auto-scroll patchwork with a small bottom-lock state machine. The remaining bugs were:

1. The bottom-lock could still drift slightly above the real maximum when expanded tool panes (last bash row, etc.) re-laid out after a chat switch. Browser/scroll-anchor settling, async highlight, mermaid render, and CSS transitions all produce layout changes that don't always fire `ResizeObserver` in time for the next paint.
2. When a chat is opened with the latest bash already streaming, the row stayed collapsed for 300 ms before auto-expanding. That delay was a flash-prevention trick for *fast-completing in-chat commands* — it should not apply to chat-open, where the bash has clearly already outlived 300 ms.

## Implementation

- `useAutoScroll` enforces the bottom-lock invariant in a `requestAnimationFrame` loop instead of relying solely on observer delivery. While `autoScroll` is on, every frame writes `scrollTop = scrollHeight - clientHeight` (no-op when already there). This makes the lock independent of any single layout signal source, so async settling inside expanded tool panes can no longer leave the transcript a few pixels above the bottom.
- Scroll intent is classified by input semantics: wheel/touch and scrollport scrollbar drags release the lock; content clicks (e.g., expanding the last bash row) do not.
- `BashToolCall` extracts auto-expand into `useBashAutoExpand`. On the first effect run after mount, if the bash is already the latest streaming bash, the row expands synchronously in a layout effect — no perceptible delay on chat-open. New in-chat bash transitions still get the 300 ms flash protection.
- `LayoutStackLane` remains reservation/alignment-only; transcript scroll anchoring is still disabled inside bottom-lock and re-enabled in manual reading mode so users reviewing older messages keep their viewport stable when async highlights settle.

## Validation

- `bun test src/browser/hooks/useAutoScroll.test.tsx`
- `bun test src/browser/features/Tools/Shared/useBashAutoExpand.test.tsx`
- `TEST_INTEGRATION=1 bun x jest tests/ui/chat/bottomLayoutShift.test.ts --runInBand`
- `make static-check`
- `make test`

## Risks

Transcript scrolling is sensitive to browser timing. The frame-aligned approach trades a small per-frame layout read while bottom-locked for stronger correctness guarantees. The auto-expand split keeps the existing flash-prevention behavior for in-chat bashes; only chat-open is changed to expand synchronously before paint.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$323.49`_


ammar-agent:
@codex review

ammar-agent:
@codex review

Updated to fix CI/unit isolation from the MarkdownComponents highlight regression test and hardened frontend base-path initialization against incomplete DOM shims.

ammar-agent:
@codex review

Updated the E2E reasoning-flow expectation to match the new stabilized behavior: reasoning displaced by follow-on answer text remains expanded instead of rendering an ellipsis toggle.

ammar-agent:
@codex review

Updated the PR to restore chat-open bottom pinning after the earlier transcript stabilization changes. Local validation is green:
- `bun test src/browser/hooks/useAutoScroll.test.tsx src/browser/components/ChatPane/LayoutStackLane.test.tsx`
- `TEST_INTEGRATION=1 bun x jest tests/ui/chat/bottomLayoutShift.test.ts --runInBand`
- `make static-check`

ammar-agent:
@codex review

Re-requesting after the active-compaction chat-open bottom-pin fix.

ammar-agent:
@codex review

Re-requesting after replacing the auto-scroll patchwork with the bottom-lock rewrite and adding the idle-chat switch regression.

ammar-agent:
@codex review

Addressed the inertial bottom re-lock feedback and resolved the thread.

ammar-agent:
@codex review

Re-requesting after the expanded bash/tool bottom-lock follow-up.

ammar-agent:
@codex review

Addressed the completed-bash regression by limiting nested bash output pinning to live output only, and fixed the broader expanded-row bottom drift by treating transcript content clicks as layout interactions rather than scroll intent. Targeted scroll tests and `make static-check` pass locally.

ammar-agent:
@codex review

Addressed the anchoring feedback by restoring `overflow-anchor: none` to bottom-lock mode only. Manual reading mode now keeps browser anchoring enabled, while expanded-row clicks keep the lock active so the original bottom drift remains covered. Targeted tests, `make static-check`, and `make test` pass locally.

ammar-agent:
@codex review

Two updates: (1) `useAutoScroll` now enforces the bottom-lock invariant in a `requestAnimationFrame` loop so async layout inside expanded tool panes (CSS transitions, Shiki/Mermaid settle, scroll-anchor races) cannot leave the transcript a few pixels above the real bottom. (2) `BashToolCall`'s auto-expand was extracted into `useBashAutoExpand`, which expands the latest streaming bash synchronously on chat-open via `useLayoutEffect` — the 300 ms flash-prevention delay only applies to new in-chat bash transitions where it's actually meaningful. Targeted tests, the full UI integration test, `make static-check`, and `make test` all pass locally.

ammar-agent:
@codex review

Addressed the perf concern: the rAF loop now only runs while bottom-lock is held. When the user scrolls up and `autoScroll` becomes false, the effect tears down and stops scheduling. Returning to the bottom (manual scroll, jump-to-bottom, send) restarts the loop. Added a regression test that asserts the loop is gated on lock state. Targeted tests + `make static-check` pass locally.

ammar-agent:
@codex review

Addressed the keyboard guard: scroll keys (PageUp/PageDown/Home/End/Arrows/Space) now mark scroll intent regardless of which transcript descendant has focus, since the browser scrolls the scrollport in those cases anyway. Added regression tests covering both focused-descendant scroll keys and non-scroll keys (which must remain inert). Targeted tests + UI integration tests + `make static-check` pass locally.

ammar-agent:
@codex review

I pushed a test-isolation fix for the unit flake: the new hook tests now use the shared `installDom()` cleanup helper, and `sendOptions.test` no longer leaves `CustomEvent`/DOM globals undefined for downstream renderer tests.

ammar-agent:
@codex review

I pushed a follow-up for the integration flake: the idle-chat bottom layout test now re-queries the active scrollport after workspace switch before applying mocked metrics, so it no longer depends on whether happy-dom/React preserved or replaced the scrollport node on CI.

ammar-agent:
@codex review

Addressed the latest comments: the bottom-lock rAF work is now a bounded settle window re-armed by resize/open/send/relock signals instead of perpetual idle polling, and `MessageWindow.test` now uses the shared DOM cleanup helper.

ammar-agent:
@codex review

I pushed one more integration-test stabilization: the mocked scroll geometry now emits an explicit scroll/layout signal after synthetic height changes because happy-dom does not deliver ResizeObserver notifications for those property overrides. Local focused unit, bottom layout, and static checks are green.

ammar-agent:
@codex review

Addressed the selection-autoscroll concern: non-interactive transcript mousedown now marks user scroll intent for drag selection, while interactive tool chrome opts out via `data-scroll-intent="ignore"` so expanding bash/tool rows still preserves bottom lock.

ammar-agent:
@codex review

Rebased onto latest `origin/main`, resolved conflicts, and re-ran focused unit coverage, the bottom layout integration test, and `make static-check` successfully.

ammar-agent:
@codex review

I pushed a focused unit-flake hardening pass: Button/Switch tests now bootstrap Radix with the shared DOM helper, and the inline skill rendering test waits for Streamdown's rendered badge output before asserting. Local failing subset and `make static-check` are green.

ammar-agent:
@codex review

Addressed the LayoutStackLane observer mount case by including `hasItems` in the setup effect dependencies and adding coverage for an empty null lane that later mounts items. Local layout-stack, CI-failing unit subset, bottom-layout integration, and static checks are green.

ammar-agent:
@codex review

I pushed another targeted unit hardening: Button/Switch tests no longer assert Radix's transient `data-state` implementation detail (which depends on suite-wide module-eval order), and inline skill assertions use a longer wait for Streamdown output under full coverage.

ammar-agent:
@codex review

I pushed a targeted fix for the remaining unit flake: the UserMessageContent composition test no longer waits on Streamdown's asynchronous Markdown badge rendering (covered directly by InlineSkillMarkdown.test) and now asserts only the synchronous slash badge/edit-mode behavior. Local coverage subset and static-check are green.

ammar-agent:
@codex review

Addressed the bash auto-expand feedback: fresh executing mounts now still take the 300ms debounce unless the tool timestamp shows the bash has already outlived the flash window. Local bash-autoexpand, flaky unit subset, bottom-layout integration, and static checks are green.

ammar-agent:
@codex review

Addressed the content-click scroll intent feedback: plain mousedown no longer marks scroll intent, drag/selection marks on mousemove with primary button, and scrollbar/wheel/touch/key behavior remains user-owned. Local autoscroll, bash-autoexpand, flaky unit subset, bottom-layout integration, and static checks are green.

ammar-agent:
@codex review

Addressed the editable-control scroll-key feedback: scroll keys on transcript text inputs/textareas/contenteditable targets no longer open a transcript scroll-intent window, while non-editable descendant scroll keys still release bottom lock. Local autoscroll, bash-autoexpand, flaky unit subset, bottom-layout integration, and static checks are green.

PR #3203, opened 2026-04-28
🤖 fix: preempt startup when editing last message

ammar-agent:
Summary
- Fix last-message edits that could hang while the previous turn was still in pre-stream startup.
- Keep the PR LoC-negative by deduplicating edit-message tests while adding the startup-edit regression coverage.

Background
Editing is implemented as truncate + replacement send. If the previous turn was still PREPARING, the edit path could wait for idle even though the startup was being superseded, leaving the edited message invisible.

Implementation
- Track the active prepared turn with an AbortController.
- When an edit preempts a PREPARING turn, abort that startup, mark the session idle, and proceed with truncation immediately.
- Guard the old startup cleanup by controller identity so it cannot mark the replacement edit turn idle when it unwinds later.
- Thread the abort signal into stream startup so stale pre-stream work exits before launching an obsolete stream.

Validation
- `bun test src/node/services/agentSession.editMessageId.test.ts`
- `bun test ./tests/ipc/streaming/queuedMessages.starting.test.ts ./tests/ipc/streaming/interrupt.test.ts`
- `make static-check`

Risks
- Touches turn-phase/pre-stream startup flow. Risk is mitigated by preserving the existing stream lifecycle behavior for non-edit sends and adding a targeted regression test for PREPARING edit preemption.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `high` • Cost: `$34.11`_

### DIFF
--- a/src/node/services/agentSession.editMessageId.test.ts
+++ b/src/node/services/agentSession.editMessageId.test.ts
@@ -5,12 +5,19 @@ import type { InitStateManager } from "@/node/services/initStateManager";
 import type { BackgroundProcessManager } from "@/node/services/backgroundProcessManager";
 import type { Config } from "@/node/config";
 import { createMuxMessage } from "@/common/types/message";
-import type { MuxMessage } from "@/common/types/message";
 import type { SendMessageError } from "@/common/types/errors";
 import type { Result } from "@/common/types/result";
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3204, opened 2026-04-29
🤖 feat: inline $skill references with autocomplete and hover previews

ThomasK33:
## Summary

Adds inline `$skill` references so users can write `Please perform a $deep-review on this code and follow $tdd` and Mux loads both skills for that turn — autocomplete suggests skill names while typing, the sent message text is preserved exactly as typed, and inline tokens render with the same hover preview as the existing `/skill` slash badge.

## Background

Today skills are invoked via slash commands (`/deep-review …`), which rewrite the user's text and only allow one skill per turn. Inline `$skill` references address two needs:

- Compose multiple skills in one turn (`$deep-review` and `$tdd` together).
- Reference a skill mid-sentence without committing the whole message to a slash command's rewrite (`Please follow $tdd while you implement this`).

The architecture is intentionally **orthogonal**: a new `agentSkillRefs?: AgentSkillReference[]` field on `MuxMessageMetadataBase` composes with reviews, model one-shots, edits, retries, and on-send auto-compaction without combinatorial union variants. Slash invocations also populate the new field, so the backend has a single plural code path while old persisted `type: "agent-skill"` messages still replay correctly.

## Implementation

The work landed in two phases on top of a 10-commit stack:

**Phase 1 — inline references end-to-end:**
- `agentSkillRefs?: AgentSkillReference[]` added to `MuxMessageMetadataBase` with `dedupeAgentSkillRefs` / `mergeAgentSkillRefs` / `withAgentSkillRefs` helpers (slash beats inline on dedupe). Persisted refs only carry `{skillName, scope, source}` — parser ranges stay autocomplete-only.
- Backend `materializeAgentSkillSnapshots(...)` is plural and order-preserving. Uses resolved `skill.scope` (advisory metadata scope is ignored). Inline refs that fail to resolve are silently skipped; slash refs preserve strict failure. Per-skill dedupe against recent history.
- Inline parser (`extractInlineSkillReferenceCandidates`) is stricter than `SkillNameSchema`: requires lowercase first letter (rejects `$100` currency / `$PATH` env-style), requires sane left boundary, skips inline backticks and fenced code blocks. Resolver mirrors the slash resolver and silently drops unknown skills.
- Send-flow integration handles normal, creation, model one-shot, and mixed slash+inline messages. Creation flow forces `disableWorkspaceAgents: true` when any combined ref is project-scoped.
- Autocomplete UX mirrors the `@file` mention pattern. Precedence: `@file` > `$skill` > slash command. Selecting Enter/Tab/click replaces only the `$partial` token. Reuses `atMentionCursorNonce` for caret-move signals so suggestions reopen when the caret moves into an existing `$partial`.

**Phase 2 — hover preview parity for inline tokens:**
- `StreamingMessageAggregator` derives an `inlineSkillSnapshots` map per displayed user message from the same `latestAgentSkillSnapshotByKey` cache the slash path uses. Not persisted on the user message — derived display state.
- A small remark plugin (`inlineSkillMarkdown.ts`) walks markdown text nodes, calls the existing `extractInlineSkillReferenceCandidates` (so code-fence/inline-code exclusion stays consistent), and rewrites recognized ranges to mdast `link` nodes with internal `mux-inline-skill:` href. The internal protocol passes through the existing markdown sanitizer — no new `dangerouslySetInnerHTML` paths.
- The anchor renderer recognizes the internal href and delegates to a context that returns the same `AgentSkillBadge` + `HoverCard` + `MarkdownRenderer` JSX the slash badge uses. `AgentSkillBadge` and the `frontmatterYaml + body` preview-markdown builder were extracted as shared primitives.

## Validation

| Check | Result |
|---|---|
| `make typecheck` | ✅ |
| `make lint` | ✅ |
| `make fmt-check` | ✅ |
| 90 unit tests across 8 new test files | ✅ |

End-to-end dogfooding against a `make dev-server-sandbox` instance with test skills `deep-review` and `tdd` confirmed:

- Autocomplete suggests `$deep-review` for `$dee` and `$tdd` for `$td`; Tab/Enter replaces only the partial token.
- Sent message text is preserved verbatim.
- Persisted `chat.jsonl` shows two synthetic `agentSkillSnapshot` messages (deep-review then tdd) before the user message; user message metadata round-trips with `agentSkillRefs: [{deep-review, project, inline}, {tdd, project, inline}]`.
- Hovering `$deep-review` and `$tdd` in the sent message body shows skill-snapshot HoverCards identical to the slash hover.
- `` `$tdd` `` inside inline code, `$deep-review` inside a fenced block, `$100`, and `$PATH` are correctly excluded from both autocomplete and rendering.
- Mixed `/deep-review Please follow $tdd` produces both slash and inline badges side-by-side.
- `$unknown-skill-name` falls back to plain text with no badge.

## Risks

- **Metadata composition**: `agentSkillRefs` is added to `MuxMessageMetadataBase`, so any code that spreads `metadata` should preserve it. The send-flow uses the centralized `withAgentSkillRefs` helper to avoid open-coded spreads dropping refs across reviews/one-shots/edits/retries/auto-compaction. Targeted tests cover those paths.
- **Backend trust boundary**: frontend-supplied `scope` is treated as advisory; backend resolves skills from disk and uses the resolved package scope in persisted snapshots. Inline refs fail silently; slash refs keep strict-failure semantics for backwards compatibility.
- **Markdown sanitizer**: the `mux-inline-skill:` protocol is allowlisted by the existing sanitize/harden config. No new HTML escape hatches were introduced. Anchors with the internal protocol decode skill names from the path, never from arbitrary markup.
- **Parser false positives**: stricter inline rule (lowercase first letter) intentionally diverges from `SkillNameSchema` so currency/env-style strings aren't matched. Numeric-named skills remain reachable via slash invocation. Trade-off documented inline.
- **Replay compatibility**: legacy persisted `type: "agent-skill"` messages still materialize exactly one snapshot via the legacy fallback; aggregator-derived `agentSkill.snapshot` continues to drive the slash hover unchanged.

## Pains

- Bun 1.3.6 doesn't undo `mock.module(...)` via `mock.restore()`, which leaked stubs across test files in the markdown render layer. Fixed by scoping `mock.module` calls with snapshot+restore in `TypewriterMarkdown.test.tsx`; flagged inline so other tests can adopt the pattern.
- Radix HoverCard portals don't render in happy-dom, so the new tests assert badge presence and snapshot-map wiring rather than HoverCard contents — visual parity verified manually during dogfooding.

---

<details>

# Plan: polished inline `$skill` references for Mux skills

## Goal

Add first-class inline skill references so a user can write:

```txt
Please perform a $deep-review on this code and make sure to follow $tdd
```

and Mux loads both skills for that turn, while preserving the exact user-authored message text.

User decisions to honor:

- `$skill` autocomplete is required.
- Sent message text should remain visible exactly as typed.
- Unknown `$skill` references should be ignored, not warned/blocking.
- `$skill` inside inline code or fenced code blocks should not load skills.
- Unadvertised skills should still be loadable.
- Creation flow should support `$skill` references.
- Model one-shots like `/opus+high Please use $deep-review` should support `$skill` references.
- Keep the feature skill-specific for now, but leave the architecture open to future inline reference kinds.

## Evidence from code exploration

- Skill discovery is already syntax-independent:
  - `discoverAgentSkills(...)`
  - `readAgentSkill(...)`
- Slash-command skill invocation currently lives in:
  - `src/browser/features/ChatInput/utils.ts`
  - `parseCommandWithSkillInvocation(...)`
  - `resolveSkillInvocation(...)`
  - `buildSkillInvocationMetadata(...)`
- Current metadata is singular and slash-shaped:
  - `MuxMessageMetadata` has `{ type: "agent-skill", rawCommand, skillName, scope }`
- Backend materialization is singular:
  - `src/node/services/agentSession.ts`
  - `materializeAgentSkillSnapshot(...)` handles only `muxMetadata.type === "agent-skill"`
  - send path stores a single `skillSnapshotResult`
- Synthetic skill snapshots already work and should be reused:
  - `createLoadedSkillSnapshot(...)`
  - `renderAgentSkillSnapshotText(...)`
  - `agentSkillSnapshot` metadata
- Slash suggestions are whole-input and slash-only:
  - `src/browser/utils/slashCommands/suggestions.ts`
  - `getSlashCommandSuggestions(...)` returns nothing unless `input.startsWith("/")`
  - `ChatInput.handleCommandSelect(...)` replaces the whole input
- `@file` mention autocomplete is the closest UI model for `$skill` autocomplete:
  - cursor-aware matching via `selectionStart`
  - token-local replacement in `handleAtMentionSelect(...)`

### Approach A — recommended: orthogonal `agentSkillRefs` metadata field

Add an orthogonal base metadata field instead of creating an `inline-skills` union variant:

```ts
interface AgentSkillReference {
  skillName: SkillName;
  /** Advisory for UI/creation decisions only; backend must resolve the actual scope from disk. */
  scope: AgentSkillScope;
  source: "slash" | "inline";
}

interface MuxMessageMetadataBase {
  agentSkillRefs?: AgentSkillReference[];
  // existing base fields remain: reviews, commandPrefix, requestedModel, ...
}

// Parser/autocomplete results may carry startIndex/endIndex, but persisted metadata should not.
// Ranges are ambiguous once reviews are prepended, slash commands are rewritten, or one-shots strip prefixes.
```

Keep the existing `type: "agent-skill"` slash metadata for backwards compatibility and display/replay behavior, but update `buildAgentSkillMetadata(...)` so slash skill invocations also populate `agentSkillRefs` with `source: "slash"`.

Backend ref extraction:

1. Start with `metadata.agentSkillRefs ?? []`.
2. If `metadata.type === "agent-skill"` and no matching slash ref exists, add a `source: "slash"` ref from the legacy `skillName` / `scope` fields.
3. This merge-style extraction prevents accidental loss if frontend metadata merging drops the new slash ref while legacy slash fields remain.

Why this is the cleanest long-term solution:

- Skill refs are orthogonal to message intent/type.
- They compose with normal messages, model one-shots, reviews, creation sends, and future metadata.
- It avoids a combinatorial metadata union like `normal-with-inline-skills`, `model-oneshot-with-inline-skills`, etc.
- It keeps slash skill behavior backward-compatible while allowing plural references.

**Net product LoC estimate:** ~370–590 LoC.

- Frontend parser/resolver/helpers: ~120–180 LoC
- ChatInput wiring and autocomplete state/selection: ~120–190 LoC
- Shared metadata helpers/types/schemas: ~50–90 LoC
- Backend plural snapshot materialization/persistence loops: ~80–130 LoC

### Approach B — not recommended: new `type: "inline-skills"` metadata variant

Add a separate metadata union member for inline refs.

Pros:

- Slightly smaller first patch if only normal messages are supported.

Cons:

- Poor composition with existing metadata.
- Harder to support model one-shots, reviews, compaction-related metadata, and future command metadata without merge bugs.
- Backend still needs plural materialization.

**Net product LoC estimate:** ~280–430 LoC initially, but likely higher over time due to merge/compatibility fixes.

### Approach C — not recommended for this request: send-time only, no autocomplete

Detect `$skill` references only during send.

Pros:

- Fastest path to functionality.

Cons:

- User explicitly wants autocomplete.
- More typos silently ignored.
- Does not deliver the polished experience.

**Net product LoC estimate:** ~180–300 LoC.

## Advisor review refinements

The advisor approved the core `agentSkillRefs` architecture and recommended tightening these details before implementation:

- Persist only `{ skillName, scope, source }` in `agentSkillRefs`; keep `startIndex` / `endIndex` as parser/autocomplete-only data because persisted offsets become ambiguous after reviews, slash rewrites, model one-shot prefix stripping, edits, and retries.
- Update shared schemas as well as TypeScript types, especially message metadata schemas used for persisted/replayed messages.
- Treat frontend-provided `scope` as advisory on the backend. Backend materialization should resolve by `skillName` with `readAgentSkill(...)` and use the resolved package scope for snapshots.
- Make error behavior source-specific:
  - `source: "inline"` refs: skip invalid/unreadable refs silently.
  - `source: "slash"` refs and legacy `type: "agent-skill"`: preserve existing strict failure behavior.
  - If slash and inline mention the same skill, slash wins during dedupe.
- Numeric skill names, if supported by the shared schema, remain invocable with slash syntax such as `/100`; inline `$100` is intentionally not a skill ref.
- Inline `$` parsing should require the first character after `$` to be a lowercase letter, even though `SkillNameSchema` permits all-numeric skill names. This avoids currency false positives like `$100`; slash invocation remains the path for rare numeric skill names.
- Explicitly support mixed slash + inline messages such as `/deep-review Please also follow $tdd` by parsing inline refs from the original authored text before slash rewrite.
- Autocomplete must cover keyboard/cursor lifecycle, not just rendering suggestions: Enter/Tab selection, Escape dismissal, arrow navigation, and reopening when the caret moves into an existing `$partial` token.
- Ensure `agentSkillRefs` survives metadata lifecycle paths: `prepareUserMessageForSend(...)`, reviews, model one-shots, edit sends, retry send options, and on-send auto-compaction follow-ups.
- Prefer `materializeAgentSkillSnapshots(...): Promise<MuxMessage[]>` over nullable wrapper objects; empty arrays simplify append/emit loops.
- `$skill` suggestions should not filter out names that collide with slash commands. `$clear` should reference a skill named `clear` even though `/clear` is a command.

Adjusted implementation order:

1. Shared metadata + schemas + helper utilities.
2. Backend plural materialization while preserving existing slash behavior.
3. Inline parser/resolver.
4. Frontend send-flow integration, including normal, creation, model one-shot, and slash+inline mixed messages.
5. Autocomplete UX.
6. Dogfooding, docs/comments, and final validation.

## Implementation plan

The sections below are grouped by subsystem. Implement them in the adjusted order above, which intentionally proves backend slash compatibility before wiring new inline frontend behavior.

### Subsystem: Shared metadata and skill-ref utilities

Files likely touched:

- `src/common/orpc/schemas/message.ts` and any other persisted/replayed message schema that validates metadata
- possibly new `src/common/types/agentSkillReference.ts` or colocated exports in `message.ts`

Steps:

1. Define a shared `AgentSkillReference` type using existing `SkillName` / `AgentSkillScope` types.
2. Add `agentSkillRefs?: AgentSkillReference[]` to `MuxMessageMetadataBase`.
3. Update relevant Zod/message schemas so persisted/replayed metadata accepts the new field; `SendMessageOptions.muxMetadata` is loose today, but history/replay paths may still validate structured metadata.
4. Add helpers to keep metadata updates centralized:
   - `dedupeAgentSkillRefs(refs)` — dedupe by `skillName`, preserve first appearance order, and let slash refs win over inline refs for the same skill.
   - `withAgentSkillRefs(metadata, refs)` or similar — merge refs into existing metadata without dropping `reviews`, `requestedModel`, `rawCommand`, etc.
5. Update `buildAgentSkillMetadata(...)` to populate both:
   - legacy slash fields: `type: "agent-skill"`, `skillName`, `scope`, `rawCommand`
   - new base refs: `agentSkillRefs: [{ skillName, scope, source: "slash" }]`

Defensive programming:

- Assert/detect invalid helper inputs in shared helpers.
- Keep helpers pure and small so merge behavior is easy to test.

Quality gate:

- Unit tests for dedupe/merge behavior if helper complexity warrants it.
- Typecheck should catch metadata composition errors.

### Subsystem: Inline `$skill` parser and resolver

Files likely touched:

- new `src/browser/utils/agentSkills/inlineSkillReferences.ts` or `src/browser/features/ChatInput/inlineSkillReferences.ts`
- `src/browser/features/ChatInput/utils.ts`

Add parser functions:

```ts
extractInlineSkillReferenceCandidates(text: string): InlineSkillCandidate[]
findInlineSkillReferenceAtCursor(text: string, cursor: number): InlineSkillCursorMatch | null
```

Parser behavior:

- Match only `$` followed by an inline-safe skill-token shape: first character `[a-z]`, then lowercase letters, numbers, and hyphens. This intentionally treats `$100` as currency, not a skill reference, even though slash invocation can still support numeric skill names if the shared schema allows them.
- Validate final candidate names with `SkillNameSchema` or a shared equivalent after applying the stricter inline first-character rule.
- Require a sane left boundary so `foo$bar` does not match.
- Stop at the first non-skill-name character.
- Skip candidates inside inline backticks.
- Skip candidates inside fenced code blocks.
- Preserve `startIndex` / `endIndex` for parser/autocomplete replacement only; do not persist those ranges in `agentSkillRefs`.
- Deduplicate repeated refs while preserving first-appearance order.

Resolver behavior:

- Resolve candidates against `agentSkillDescriptors` first, including descriptors where `advertise === false`.
- Optionally fall back to `api.agentSkills.get(...)` for valid-looking candidates that are not already listed, matching the existing slash path.
- Silently ignore unknown/unresolvable candidates.
- Avoid slow failure paths by deduping candidate names before backend fallback calls.
- Keep result order based on first mention in text.

Quality gate:

- Parser tests for:
  - `$deep-review`
  - multiple refs
  - duplicate refs
  - punctuation boundaries like `($tdd)` and `$tdd,`
  - invalid/non-skill values: `$100` as currency, `$PATH`, `$Bad`, `$bad-`, `$-bad`, `foo$tdd`
  - inline code: `` `$tdd` `` ignored
  - fenced code blocks ignored
  - mixed valid/unknown refs where unknowns are dropped by resolver

### Subsystem: Frontend send-flow integration

Files likely touched:

- `src/browser/features/ChatInput/utils.ts`
- `src/browser/features/ChatInput/index.tsx`

Steps:

1. Extend the existing parse/send preparation path so inline refs are resolved for all relevant sends:
   - normal workspace message
   - creation flow message
   - model one-shot message, e.g. `/opus+high Please use $deep-review`
   - mixed slash + inline message, e.g. `/deep-review Please also follow $tdd`
2. Parse inline refs from the original authored text before slash skill rewrites or model one-shot prefix stripping.
3. Preserve current slash skill behavior:
   - `/deep-review foo` may continue to rewrite backend text to `Using skill deep-review: foo`.
   - Slash metadata should also include `agentSkillRefs`.
4. For inline refs, preserve message text exactly as typed.
5. Merge inline refs into whatever `muxMetadata` exists later in the send path using the shared helper.
6. Ensure later metadata additions do not drop refs:
   - reviews via `prepareUserMessageForSend(...)`
   - requested model labels
   - one-shot `rawCommand` / `commandPrefix`
   - edit sends
   - retry send options
   - on-send auto-compaction follow-up metadata
7. In creation flow, if any resolved referenced skill is project-scoped, set `disableWorkspaceAgents: true`, matching the existing slash skill creation behavior.
8. For model one-shots, parse refs from the user-authored text, but attach them to the final metadata after model override metadata is computed.

Defensive programming:

- Add a local assertion or invariant test around metadata merge helpers rather than open-coding object spreads in multiple places.
- Treat backend/API failures during inline resolution as “unknown skill” and silently continue.

Quality gate:

- Tests proving message text is preserved for inline refs.
- Tests proving one-shot metadata and skill refs compose.
- Tests proving `/deep-review Please also follow $tdd` produces both slash and inline refs.
- Tests proving refs survive review metadata and edit/retry/on-send compaction paths where practical.
- Tests proving creation flow sets `disableWorkspaceAgents` when a project-scoped inline skill is referenced.

### Subsystem: Backend plural skill snapshot materialization

Files likely touched:

- `src/node/services/agentSession.ts`
- possibly `src/node/services/agentSkills/loadedSkillSnapshots.ts` only if helper extraction is useful

Steps:

1. Replace or wrap `materializeAgentSkillSnapshot(...)` with plural logic:

```ts
materializeAgentSkillSnapshots(...): Promise<MuxMessage[]>
```

Return an empty array when there are no refs or all inline refs are skipped; this keeps append/emit loops simple.

2. Convert metadata to refs using a helper:
   - start with `metadata.agentSkillRefs ?? []`
   - merge in legacy `metadata.type === "agent-skill"` fields as a `source: "slash"` ref when no matching slash ref exists
3. Validate each `skillName` with `SkillNameSchema`; invalid inline refs are skipped, while invalid slash/legacy refs keep the current strict error behavior.
4. Deduplicate refs by skill identity while preserving first-appearance order, with slash refs winning over inline refs for the same skill.
5. Read each skill with existing `readAgentSkill(...)` using the same project/workspace discovery path logic as today. Treat metadata `scope` as advisory only; use the resolved package scope in snapshots.
6. Create one synthetic `agentSkillSnapshot` message per resolved skill.
7. Update append and emit sites to loop over the returned snapshot messages.
8. Improve dedupe against recent history:
   - check recent synthetic `agentSkillSnapshot` messages, not just the most recent one
   - skip only refs whose `skillName + sha256` already exists recently
9. Preserve existing slash behavior for invalid/missing slash skills unless intentionally changed:
   - invalid/unreadable inline refs are skipped both before and after `readAgentSkill(...)`
   - slash refs and old slash metadata fallback keep throwing as today for invalid skill names or missing/unreadable skills

Quality gate:

- `src/node/services/agentSession.agentSkillSnapshot.test.ts` additions for:
  - multiple refs create multiple snapshot messages before the user message
  - order is preserved
  - repeated refs load once, with slash winning over inline duplicates
  - recent duplicate snapshots are skipped per skill
  - legacy `type: "agent-skill"` still materializes exactly as before
  - inline skill deleted between frontend resolution and backend read is skipped, while slash/legacy failure remains strict

### Subsystem: `$skill` autocomplete UX

Files likely touched:

- new or existing inline skill suggestion helper, likely near the parser utility
- `src/browser/features/ChatInput/index.tsx`
- `src/browser/features/ChatInput/CommandSuggestions.test.tsx`
- maybe `src/browser/utils/slashCommands/types.ts` only if suggestion type naming needs to stay shared

Design:

- Do not overload `getSlashCommandSuggestions(...)`; it is slash-only and whole-input oriented.
- Add a sibling helper for skill suggestions:

```ts
getInlineSkillSuggestions(partial: string, context): SlashSuggestion[]
```

It can reuse the existing `SlashSuggestion` display shape, but ids should be distinct, e.g. `inline-skill:tdd`.

- Mirror `@file` mention behavior:
  - use `inputRef.current.selectionStart`
  - call `findInlineSkillReferenceAtCursor(input, cursor)`
  - show suggestions while cursor is in a `$partial` token
  - on select with Enter/Tab/click, replace only `[startIndex, endIndex)` with `$skill-name`
  - add trailing space if the following character is absent or not whitespace/punctuation where a space would be wrong
  - support Escape dismissal and arrow-key navigation consistently with existing suggestion popovers
  - restore focus and cursor after insertion
  - use cursor-movement state/nonce similar to `@file` mentions so suggestions reopen when the caret moves into an existing `$partial` token

Suggestion precedence:

1. `@file` suggestions when active.
2. `$skill` suggestions when active.
3. Slash command suggestions when input starts with `/`.

This avoids overlapping popovers and preserves existing slash behavior.

Quality gate:

- Component/unit tests for:
  - `$dee` shows `$deep-review`
  - selecting a skill with Enter/Tab/click replaces only the token, not the whole input
  - Escape dismisses suggestions and arrow keys navigate them
  - moving the caret into an existing `$partial` token opens suggestions
  - typing `$` in the middle of a sentence works
  - suggestions do not appear inside inline/fenced code
  - `$clear` can suggest a skill named `clear` even though `/clear` is a slash command
  - slash suggestions still work
  - `@file` suggestions keep precedence

### Subsystem: Documentation and comments

Files likely touched only if docs already document skill invocation:

- search existing docs for slash skill invocation before adding anything
- if user-facing docs exist, update them in `docs/` and navigation if needed

Guidance:

- Keep docs minimal.
- Do not create free-floating Markdown.
- Add inline code comments only where they explain non-obvious “why”, especially:
  - why `agentSkillRefs` is orthogonal metadata
  - why inline refs ignore code spans/fences
  - why unknown refs are silently ignored

Quality gate:

- If docs changed, run docs/static checks required by repo conventions.

## Acceptance criteria

Functional:

- `Please perform a $deep-review and follow $tdd` loads both skills for the turn.
- Sent message text remains exactly as typed.
- Duplicate refs load once.
- Unknown refs are ignored silently.
- `$skill` inside inline code or fenced code does not load.
- Unadvertised skills are loadable by exact `$skill-name`.
- Creation flow supports `$skill` refs and resolves project-scoped skills from the project path.
- Model one-shots support `$skill` refs while preserving model override behavior.
- Existing `/skill-name` behavior remains compatible.
- Mixed `/deep-review Please also follow $tdd` loads both slash and inline refs.
- `$100` is treated as currency/non-skill in inline parsing.
- A skill named `clear` can be referenced as `$clear` even though `/clear` is a slash command.
- Existing persisted `type: "agent-skill"` messages still replay/materialize correctly.

Autocomplete:

- Typing `$dee` offers `$deep-review`.
- Selecting a suggestion replaces only the current `$...` token.
- Autocomplete works in the middle of a sentence.
- Autocomplete reopens when the caret moves into an existing `$partial` token.
- Enter/Tab/click select suggestions; Escape dismisses; arrow keys navigate.
- Autocomplete does not activate inside inline code/fenced code.
- Slash command and `@file` suggestions continue to work.

Architecture:

- Skill refs live in an orthogonal metadata field, not a combinatorial union variant.
- Persisted refs contain only `{ skillName, scope, source }`; parser ranges remain transient.
- Metadata merge helpers prevent refs from being dropped when reviews/model overrides/command metadata are added.
- Backend snapshot materialization is plural, order-preserving, and uses resolved skill package scope instead of trusting frontend metadata scope.

## Validation plan

Run targeted tests first, then broader checks:

1. Parser/resolver tests:
   - likely new/updated tests near `src/browser/features/ChatInput/utils.test.ts` or `src/browser/utils/agentSkills/*.test.ts`
2. Suggestion/component tests:
   - `bun test src/browser/features/ChatInput/CommandSuggestions.test.tsx`
   - any new inline skill suggestion tests
   - coverage for keyboard selection/dismissal/navigation and cursor movement into `$partial`
3. Backend snapshot tests:
   - `bun test src/node/services/agentSession.agentSkillSnapshot.test.ts`
   - `bun test src/node/services/agentSkills/loadedSkillSnapshots.test.ts` if touched
   - coverage for source-specific skip/fail behavior and slash+inline dedupe
4. Type/lint/static validation:
   - `make typecheck`
   - `make lint` or `make static-check` if the final diff is broad enough / before PR

Do not claim success until the targeted tests and the appropriate static validation pass, or report exact blockers.

## Dogfooding plan with reviewer evidence

Dogfooding should be performed after automated tests pass.

Setup:

1. Start an isolated Mux dev environment using the repo’s sandbox/dev-server guidance if available.
2. Create or use test skills named `deep-review` and `tdd`; include one advertised and one `advertise: false` case if easy.
3. Open the app with `agent-browser`.

Scenarios to record:

1. Autocomplete in normal text:
   - type `Please perform a $dee`
   - verify `$deep-review` suggestion appears
   - select it
   - type ` and follow $td`
   - select `$tdd`
2. Send-time loading:
   - send `Please perform a $deep-review on this code and make sure to follow $tdd`
   - verify visible message text remains unchanged
   - inspect history/debug output to confirm two synthetic `agentSkillSnapshot` messages were inserted before the user message
3. Code exclusion:
   - send a message containing `` `$tdd` `` and a fenced block with `$deep-review`
   - verify no skills are loaded from those code spans/fences
4. Model one-shot and mixed slash + inline:
   - send `/opus+high Please use $deep-review`
   - verify model override metadata still applies and skill snapshot is loaded
   - send `/deep-review Please also follow $tdd`
   - verify both skill snapshots are loaded and slash behavior remains readable
5. False-positive checks:
   - send text with `$100`, `$PATH`, inline-code `` `$tdd` ``, and a fenced block containing `$deep-review`
   - verify only valid non-code skill refs load
6. Creation flow:
   - from creation UI, reference a project-scoped `$skill`
   - verify project-path discovery behavior matches existing slash skill invocation

Evidence to capture:

- Screenshots of autocomplete suggestions and final sent message.
- A short video recording of typing/selecting/sending inline skill refs.
- Terminal/debug output or history inspection showing loaded skill snapshots.
- Attach screenshots/video files for reviewer verification when available.

## Main risks and mitigations

- **Metadata merge bugs:** use centralized merge helper and tests; avoid open-coded spreads that drop `agentSkillRefs`, especially across reviews, edits, retries, model one-shots, and on-send compaction follow-ups.
- **Parser false positives:** validate with `SkillNameSchema`, require the stricter inline first-character rule, require sane boundaries, and skip Markdown code spans/fences.
- **Autocomplete complexity:** mirror existing `@file` cursor-aware flow instead of extending slash suggestions; include keyboard and caret-movement tests.
- **Backend trust boundary:** treat frontend `scope` as advisory, resolve skills from disk, and use resolved package scope for snapshots.
- **Backend duplicate snapshots:** dedupe refs before reading and also check recent synthetic snapshots per skill/hash.
- **Creation path mismatch:** reuse existing `disableWorkspaceAgents` behavior for project-scoped skill refs.
- **Performance:** dedupe candidates before API fallback and cap or avoid excessive backend lookups for many unknown `$...` tokens.

## Recommended delivery shape

Implement as one cohesive feature branch with internal workstreams, not separate architecture-breaking PRs. Use this order:

1. Shared metadata + schemas + helper utilities.
2. Backend plural materialization while proving existing slash behavior still works.
3. Inline parser/resolver.
4. Frontend send-flow integration.
5. Autocomplete UX.
6. Tests, validation, dogfooding evidence.

If scope needs to be reduced, keep the `agentSkillRefs` architecture and defer only visual highlighting/docs polish; do not fall back to `inline-skills` metadata.

</details>

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$188.10`_

### DIFF
--- a/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
+++ b/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
@@ -30,7 +30,6 @@ import * as TooltipModule from "../Tooltip/Tooltip";
 import * as SidebarCollapseButtonModule from "../SidebarCollapseButton/SidebarCollapseButton";
 import * as ConfirmationModalModule from "../ConfirmationModal/ConfirmationModal";
 import * as ProjectDeleteConfirmationModalModule from "../ProjectDeleteConfirmationModal/ProjectDeleteConfirmationModal";
-import * as WorkspaceStatusIndicatorModule from "../WorkspaceStatusIndicator/WorkspaceStatusIndicator";
 import * as PopoverErrorModule from "../PopoverError/PopoverError";
 import * as SectionHeaderModule from "../SectionHeader/SectionHeader";
 import * as WorkspaceSectionDropZoneModule from "../WorkspaceSectionDropZone/WorkspaceSectionDropZone";
@@ -480,9 +479,9 @@ function installProjectSidebarTestDoubles() {
         <div data-testid="project-delete-confirmation-modal">{props.projectName}</div>
       ) : null) as unknown as typeof ProjectDeleteConfirmationModalModule.ProjectDeleteConfirmationModal
   );
-  spyOn(WorkspaceStatusIndicatorModule, "WorkspaceStatusIndicator").mockImplementation((() => (
-    <div data-testid="workspace-status-indicator" />
-  )) as unknown as typeof WorkspaceStatusIndicatorModule.WorkspaceStatusIndicator);
+  void mock.module("../WorkspaceStatusIndicator/WorkspaceStatusIndicator", () => ({
+    WorkspaceStatusIndicator: () => <div data-testid="workspace-status-indicator" />,
+  }));
   spyOn(PopoverErrorModule, "PopoverError").mockImplementation(
     (() => null) as unknown as typeof PopoverErrorModule.PopoverError
   );
--- a/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
+++ b/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
@@ -38,12 +38,38 @@ import * as SectionDragLayerModule from "../SectionDragLayer/SectionDragLayer";
 import * as DraggableSectionModule from "../DraggableSection/DraggableSection";
 import { updatePersistedState } from "@/browser/hooks/usePersistedState";
 import type ProjectSidebarComponent from "./ProjectSidebar";
+import type * as WorkspaceStatusIndicatorModuleExports from "../WorkspaceStatusIndicator/WorkspaceStatusIndicator";
 
 const agentItemTestId = (workspaceId: string) => `agent-item-${workspaceId}`;
 const toggleButtonLabel = (workspaceId: string) => `toggle-completed-${workspaceId}`;
 
 function TestWrapper(props: PropsWithChildren) {
   return <>{props.children}</>;
+}
+
+const ProviderIconSvgStub = (props: React.SVGProps<SVGSVGElement>) => (
+  <svg data-testid="provider-icon-mock" {...props} />
+);
+
+function installProviderIconSvgMocks() {
+  const providerIconSvgPaths = [
+    "@/browser/assets/icons/anthropic.svg?react",
+    "@/browser/assets/icons/openai.svg?react",
+    "@/browser/assets/icons/google.svg?react",
+    "@/browser/assets/icons/xai.svg?react",
+    "@/browser/assets/icons/openrouter.svg?react",
+    "@/browser/assets/icons/ollama.svg?react",
+    "@/browser/assets/icons/deepseek.svg?react",
+    "@/browser/assets/icons/aws.svg?react",
+    "@/browser/assets/icons/github.svg?react",
+  ] as const;
+
+  for (const svgPath of providerIconSvgPaths) {
+    void mock.module(svgPath, () => ({
+      __esModule: true,
+      default: ProviderIconSvgStub,
+    }));
+  }
 }
 
 const passthroughRef = <T,>(value: T): T => value;
@@ -479,9 +505,14 @@ function installProjectSidebarTestDoubles() {
         <div data-testid="project-delete-confirmation-modal">{props.projectName}</div>
       ) : null) as unknown as typeof ProjectDeleteConfirmationModalModule.ProjectDeleteConfirmationModal
   );
-  void mock.module("../WorkspaceStatusIndicator/WorkspaceStatusIndicator", () => ({
-    WorkspaceStatusIndicator: () => <div data-testid="workspace-status-indicator" />,
-  }));
+  installProviderIconSvgMocks();
+  /* eslint-disable @typescript-eslint/no-require-imports */
+  const WorkspaceStatusIndicatorModule =
+    require("../WorkspaceStatusIndicator/WorkspaceStatusIndicator") as typeof WorkspaceStatusIndicatorModuleExports;
+  /* eslint-enable @typescript-eslint/no-require-imports */
+  spyOn(WorkspaceStatusIndicatorModule, "WorkspaceStatusIndicator").mockImplementation((() => (
+    <div data-testid="workspace-status-indicator" />
+  )) as unknown as typeof WorkspaceStatusIndicatorModule.WorkspaceStatusIndicator);
   spyOn(PopoverErrorModule, "PopoverError").mockImplementation(
     (() => null) as unknown as typeof PopoverErrorModule.PopoverError
   );
--- a/src/browser/components/WorkspaceHeartbeatModal/WorkspaceHeartbeatModal.test.tsx
+++ b/src/browser/components/WorkspaceHeartbeatModal/WorkspaceHeartbeatModal.test.tsx
@@ -1,7 +1,7 @@
 import "../../../../tests/ui/dom";
 
 import type { ReactNode } from "react";
-import { afterEach, beforeEach, describe, expect, mock, spyOn, test } from "bun:test";
+import { afterEach, afterAll, beforeEach, describe, expect, mock, spyOn, test } from "bun:test";
 import { cleanup, fireEvent, render, waitFor } from "@testing-library/react";
 import { installDom } from "../../../../tests/ui/dom";
 import * as APIModule from "@/browser/contexts/API";
@@ -13,6 +13,20 @@ import {
   HEARTBEAT_DEFAULT_INTERVAL_MS,
   HEARTBEAT_DEFAULT_MESSAGE_BODY,
 } from "@/constants/heartbeat";
+import * as WorkspaceContextModule from "@/browser/contexts/WorkspaceContext";
+const actualWorkspaceContextModule = { ...WorkspaceContextModule };
+const setWorkspaceMetadataMock = mock(() => undefined);
+
+async function installWorkspaceHeartbeatModalMocks() {
+  await mock.module("@/browser/contexts/WorkspaceContext", () => ({
+    ...actualWorkspaceContextModule,
+    useWorkspaceActions: () => ({ setWorkspaceMetadata: setWorkspaceMetadataMock }),
+  }));
+}
+
+async function restoreWorkspaceHeartbeatModalMocks() {
+  await mock.module("@/browser/contexts/WorkspaceContext", () => actualWorkspaceContextModule);
+}
 
 void mock.module("@/browser/components/Dialog/Dialog", () => ({
   Dialog: (props: { open: boolean; children: ReactNode }) =>
@@ -81,7 +95,8 @@ function createConnectedUseAPIResult(api: WorkspaceHeartbeatTestAPI): ConnectedU
 const LONG_HEARTBEAT_MESSAGE = "Review pending work and summarize next steps. ".repeat(30).trim();
 
 describe("WorkspaceHeartbeatModal", () => {
-  beforeEach(() => {
+  beforeEach(async () => {
+    await installWorkspaceHeartbeatModalMocks();
     cleanupDom = installDom();
     settingsByWorkspaceId = new Map<string, HeartbeatFormSettings>();
     saveCalls = [];
@@ -119,7 +134,8 @@ describe("WorkspaceHeartbeatModal", () => {
     });
   });
 
-  afterEach(() => {
+  afterEach(async () => {
+    await restoreWorkspaceHeartbeatModalMocks();
     cleanup();
     mock.restore();
     cleanupDom?.();
@@ -413,4 +429,8 @@ describe("WorkspaceHeartbeatModal", () => {
       ]);
     });
   });
+});
+
+afterAll(async () => {
+  await restoreWorkspaceHeartbeatModalMocks();
 });
--- a/src/browser/features/ChatInput/index.tsx
+++ b/src/browser/features/ChatInput/index.tsx
@@ -59,6 +59,8 @@ import {
 import { Button } from "@/browser/components/Button/Button";
 import { CUSTOM_EVENTS } from "@/common/constants/events";
 import { findAtMentionAtCursor } from "@/common/utils/atMentions";
+import { findInlineSkillReferenceAtCursor } from "@/browser/utils/agentSkills/inlineSkillReferences";
+import { getInlineSkillSuggestions } from "@/browser/utils/agentSkills/inlineSkillSuggestions";
 import { getCommandGhostHint } from "@/browser/utils/slashCommands/registry";
 import {
   getSlashCommandSuggestions,
@@ -279,12 +281,16 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
   const [hideReviewsDuringSend, setHideReviewsDuringSend] = useState(false);
   const [showAtMentionSuggestions, setShowAtMentionSuggestions] = useState(false);
   const [atMentionSuggestions, setAtMentionSuggestions] = useState<SlashSuggestion[]>([]);
+  const [showSkillSuggestions, setShowSkillSuggestions] = useState(false);
+  const [skillSuggestions, setSkillSuggestions] = useState<SlashSuggestion[]>([]);
   const agentSkillsRequestIdRef = useRef(0);
   const atMentionDebounceRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const atMentionRequestIdRef = useRef(0);
   const lastAtMentionScopeIdRef = useRef<string | null>(null);
   const lastAtMentionQueryRef = useRef<string | null>(null);
   const lastAtMentionInputRef = useRef<string>(input);
+  const lastSkillInputRef = useRef<string | null>(null);
+  const lastSkillQueryRef = useRef<string | null>(null);
   const [showCommandSuggestions, setShowCommandSuggestions] = useState(false);
 
   const [commandSuggestions, setCommandSuggestions] = useState<SlashSuggestion[]>([]);
@@ -533,6 +539,7 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
     }
   );
   const atMentionListId = useId();
+  const skillListId = useId();
   const commandListId = useId();
   const telemetry = useTelemetry();
   const [vimEnabled, setVimEnabled] = usePersistedState<boolean>(VIM_ENABLED_KEY, false, {
@@ -1322,6 +1329,50 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
     atMentionCursorNonce,
   ]);
 
+  // Watch input/cursor for inline $skill references
+  useEffect(() => {
+    if (showAtMentionSuggestions) {
+      // File mentions win precedence if an edge-case token could match both menus.
+      setSkillSuggestions((prev) => (prev.length === 0 ? prev : []));
+      setShowSkillSuggestions(false);
+      lastSkillQueryRef.current = null;
+      return;
+    }
+
+    const inputChanged = lastSkillInputRef.current !== input;
+    lastSkillInputRef.current = input;
+
+    const cursor = Math.min(inputRef.current?.selectionStart ?? input.length, input.length);
+    const match = findInlineSkillReferenceAtCursor(input, cursor);
+
+    if (!match) {
+      setSkillSuggestions([]);
+      setShowSkillSuggestions(false);
+      lastSkillQueryRef.current = null;
+      return;
+    }
+
+    // Avoid recomputing on caret movement within the same partial. Moving out of a token
+    // clears lastSkillQueryRef above, so moving back into an existing token can reopen.
+    if (!inputChanged && lastSkillQueryRef.current === match.partial) {
+      return;
+    }
+
+    lastSkillQueryRef.current = match.partial;
+
+    const nextSuggestions = getInlineSkillSuggestions({
+      partial: match.partial,
+      descriptors: agentSkillDescriptors,
+    });
+    setSkillSuggestions(nextSuggestions);
+    setShowSkillSuggestions(nextSuggestions.length > 0);
+  }, [
+    input,
+    showAtMentionSuggestions,
+    agentSkillDescriptors,
+    atMentionCursorNonce,
+  ]);
+
   // Watch input for slash commands
   useEffect(() => {
     const suggestions = getSlashCommandSuggestions(input, {
@@ -1933,6 +1984,40 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
     },
     [input, setInput]
   );
+  const handleSkillSelect = useCallback(
+    (suggestion: SlashSuggestion) => {
+      const cursor = Math.min(inputRef.current?.selectionStart ?? input.length, input.length);
+      const match = findInlineSkillReferenceAtCursor(input, cursor);
+      if (!match) {
+        return;
+      }
+
+      // Add a separating space only when the following text does not already provide one.
+      const after = input.slice(match.endIndex);
+      const trailing = after.length === 0 || /\s/.test(after[0] ?? "") ? "" : " ";
+      const next =
+        input.slice(0, match.startIndex) + suggestion.replacement + trailing + after;
+
+      setInput(next);
+      setSkillSuggestions([]);
+      setShowSkillSuggestions(false);
+      lastSkillQueryRef.current = null;
+
+      requestAnimationFrame(() => {
+        const el = inputRef.current;
+        if (!el || el.disabled) {
+          return;
+        }
+
+        el.focus();
+        const newCursor = match.startIndex + suggestion.replacement.length + trailing.length;
+        el.selectionStart = newCursor;
+        el.selectionEnd = newCursor;
+      });
+    },
+    [input, setInput]
+  );
+
   const handleCommandSelect = useCallback(
     (suggestion: SlashSuggestion) => {
       setInput(suggestion.replacement);
@@ -2407,12 +2492,14 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
 
     const hasCommandSuggestionMenu = showCommandSuggestions && commandSuggestions.length > 0;
     const hasAtMentionSuggestionMenu = showAtMentionSuggestions && atMentionSuggestions.length > 0;
+    const hasSkillSuggestionMenu = showSkillSuggestions && skillSuggestions.length > 0;
 
     // Don't handle keys if suggestions are visible.
-    // Enter/Tab/arrows/Escape are handled by CommandSuggestions for both slash and @mention menus.
+    // Enter/Tab/arrows/Escape are handled by CommandSuggestions for slash, @file, and $skill menus.
     if (
       (hasCommandSuggestionMenu && COMMAND_SUGGESTION_KEYS.includes(e.key)) ||
-      (hasAtMentionSuggestionMenu && FILE_SUGGESTION_KEYS.includes(e.key))
+      (hasAtMentionSuggestionMenu && FILE_SUGGESTION_KEYS.includes(e.key)) ||
+      (hasSkillSuggestionMenu && FILE_SUGGESTION_KEYS.includes(e.key))
     ) {
       return; // Let CommandSuggestions handle it
     }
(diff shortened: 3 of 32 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

Both prior comments addressed in 7f60ef69e:

- **P2 (descriptors update async)**: added `lastSkillDescriptorsRef` and `shouldRefreshInlineSkillSuggestions` so the menu refreshes when `agentSkillDescriptors` arrives after the user has already typed a `$partial`.
- **P3 (extra space before punctuation)**: replaced the whitespace-only separator check with `getInlineSkillInsertionTrailingText`, which treats `, . ; : ! ? ) ] } > " ' \`` as existing separators and skips the trailing space.

Both helpers have behavioral tests in `inlineSkillSuggestions.test.ts`.

ThomasK33:
@codex review

CI test failure addressed in 46247c1d1:

- **`src/browser/utils/frontendBasePath.ts`**: `INITIAL_APP_PROXY_BASE_PATH` now guards both `window` and `window.location.pathname` so module load returns `null` instead of throwing in environments where `window` exists but `location` is missing (TDZ otherwise cascaded into ~10 unrelated test failures).
- **`TypewriterMarkdown.test.tsx`**: snapshot/restore the pre-test `window`/`document` and create the Happy DOM window with an explicit URL.
- **`ProjectSidebar.test.tsx`**: avoid eager-importing the real `WorkspaceStatusIndicator` (which transitively pulls SVG imports that Bun's test loader can't parse).

Validated locally with the CI-order repro and full targeted suite.

ThomasK33:
@codex review

Multi-backtick code span issue addressed in 97818fa7d:

- `getInlineCodeRanges` now counts the opening backtick run length and only closes the span on a run of the EXACT same length, matching CommonMark/GFM rules.
- Added test cases for `` ``$tdd`` ``, `` ```$tdd``` ``, mismatched delimiters, and `$skill` references between separate inline-code spans.

ThomasK33:
@codex review

CI test pollution remediated in 189354cd8:

- `WorkspaceStore.syncAutoCompactionThresholdAtStartup` now guards `typeof client.workspace?.setAutoCompactionThreshold !== "function"` and returns early. This is best-effort startup sync — missing the method (e.g. due to incomplete orpc client mocks in some test paths) should not throw and cascade-fail unrelated tests. Production behavior is unchanged because real clients always provide the method.

Comparing `bun test src/browser/` on `origin/main` to this branch shows the same 34 pre-existing test-isolation failures (`@/browser/contexts/WorkspaceContext` mocked at file scope by `useWorkspaceHeartbeat.test.tsx` leaking via bun 1.3.6's persistent `mock.module` behavior). Out of scope for this PR.

ThomasK33:
@codex review

Test isolation root cause identified and fixed in ef043a009:

`src/browser/hooks/useWorkspaceHeartbeat.test.tsx` had file-scope `mock.module(...)` calls for `@/browser/contexts/API` and `@/browser/contexts/WorkspaceContext`. In bun 1.3.5/1.3.6 these mocks persist across files, causing `WorkspaceContext.test.tsx` to import the heartbeat-test's bogus `{useWorkspaceActions}`-only mock and fail with `ctx().*` undefined errors.

Applied the same per-test scoping pattern as commit `67fa36619` (TypewriterMarkdown): snapshot real exports at file scope, install/restore in `beforeEach`/`afterEach`/`afterAll`. Local `bun test src/browser/` went from 34 failures to 1, and the remaining 1 (`WorkspaceHeartbeatModal > loads global heartbeat defaults`) is pre-existing on `origin/main`.

ThomasK33:
@codex review

Last test isolation issue addressed in acbff70dd:

The `mock.module(...)` for `WorkspaceStatusIndicator` introduced in commit `46247c1d1` was leaking into `WorkspaceStatusIndicator.test.tsx` and stubbing the real component there too. Replaced with the `spyOn` pattern + scoped SVG `?react` mocks for the SVG paths transitively imported by `WorkspaceStatusIndicator` (which Bun's loader can't parse). `mock.restore()` now restores cleanly per test.

Local `bun test src/browser/` is now down to the same 1 pre-existing flake on `origin/main` (`WorkspaceHeartbeatModal > loads global heartbeat defaults`), which CI on main passes consistently.

ThomasK33:
@codex review

All 5 review comments addressed in 7f956d805:

- **P1 (`MarkdownRenderer.tsx`) — recursive previews**: HoverCard markdown content is now wrapped in a disabled `InlineSkillPreviewContext.Provider`, so `$skill` tokens inside snapshot bodies render as plain labels instead of recursing.
- **P1 (`agentSession.ts`) — array guard**: `extractAgentSkillRefs` normalizes `metadata.agentSkillRefs` with `Array.isArray(...) ? ... : []` to keep self-healing on malformed persisted metadata.
- **P1 (`StreamingMessageAggregator.ts`) — array guard**: `deriveInlineSkillSnapshotDisplayState` early-returns when `!Array.isArray(refs)` so transcript rendering never throws on bad metadata.
- **P2 (`inlineSkillReferences.ts`) — tilde fences**: Fence detection now tracks character + length per CommonMark rules; `~~~`/`~~~~` blocks correctly hide `$skill` tokens. Mismatched ` ``` ` vs `~~~` correctly does not close.

ThomasK33:
@codex review

P2 indented-fence comment addressed in ee09c7e61: fence detection now accepts up to 3 leading spaces (CommonMark rule) for both opening and closing fences. 4+ spaces makes it an indented code block, which we treat as non-fenced (so leading-4-space `$tdd` is correctly extracted as a skill ref). Tests cover 1/2/3-space-indented openings, 4-space cutoff, and indented closings.

All other tests now pass locally (`bun test src/browser/` 2068/2068).

ThomasK33:
@codex review

P2 closing-fence suffix comment addressed in 2663e8216: closing fences now only close when followed by spaces/tabs to EOL, per CommonMark. Lines like `` ```notclosing `` keep the fenced block open. Tests cover both the negative case and the trailing-whitespace-OK case.

PR #3206, opened 2026-04-29
🤖 perf: speed up SSH stream startup

ammar-agent:
## Summary
Speeds up stream startup for existing SSH workspaces by removing repeated pre-stream remote work: system-context discovery is no longer rebuilt unless advisor availability changes, the stream temp directory is created once and transferred into `StreamManager`, Coder SSH readiness caching now survives runtime recreation, and SSH repository readiness checks are batched into a single remote command.

## Background
Sending a follow-up message in an existing SSH workspace pays for a large amount of setup before `stream-start` can be emitted. The slow path was dominated by many small SSH executions and a duplicated context-build pass that rescanned agent definitions, skills, and instruction files.

## Implementation
- Reuse the pre-policy stream system context when advisor availability is unchanged; rebuild only when policy/experiments alter advisor guidance.
- Pass AIService's pre-created stream temp dir into `StreamManager.startStream()` so SSH does not repeat `resolvePath` + `mkdir -p` for the same token.
- Move Coder SSH recent-activity tracking into a `WeakMap<CoderService, Map<workspaceName, timestamp>>`, preserving the existing five-minute fast path across per-stream runtime instances.
- Combine SSH `.git`, `rev-parse --git-dir`, and `rev-parse --is-inside-work-tree` readiness probes into one remote shell invocation while preserving `runtime_not_ready` vs `runtime_start_failed` behavior.

## Estimated performance impact
This removes at least **4 fixed short SSH round trips per existing SSH send** on the common path:

- SSH repo readiness: **3 remote commands → 1 remote command** (**2 fewer**)
- Stream temp directory setup: duplicated `resolvePath` + `ensureDir` removed (**2 fewer**)
- Stream system context: **2 builds → 1 build** whenever advisor availability is unchanged, avoiding a second pass over agent definitions, skills, and instruction files.

Because each short OpenSSH operation typically costs a process spawn plus a remote round trip, the fixed-command reduction alone should save roughly **0.6–1.6s** on a common 150–400ms SSH round-trip/worktree-probe path. Avoiding the duplicate context build should save another **~1–4s** in workspaces with project/global agents, skills, or instruction files. Overall expected improvement for the common existing-SSH follow-up send is therefore **~1.5–5.5s faster pre-stream startup**, or roughly **35–65% lower stream-start latency** when the prior startup was in the ~4–10s range. This estimate is command-count based; the sandbox cannot benchmark a real remote SSH host, and existing `[stream-startup]` phase logs should confirm the exact per-workspace gain.

## Validation
- `bun test src/node/services/aiService.test.ts`
- `bun test src/node/services/streamManager.test.ts`
- `bun test src/node/runtime/CoderSSHRuntime.test.ts`
- `bun test src/node/runtime/SSHRuntime.retry.test.ts src/node/runtime/SSHRuntime.syncContract.test.ts`
- `make typecheck`
- `make fmt-check`
- `make lint`
- `make static-check`

## Risks
Moderate regression risk because this touches stream startup and SSH readiness checks. The changes are scoped to removing duplicate work or batching equivalent probes; error classification and temp-dir cleanup ownership are covered by targeted tests.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `high` • Cost: `$33.13`_

### DIFF
--- a/src/node/runtime/CoderSSHRuntime.test.ts
+++ b/src/node/runtime/CoderSSHRuntime.test.ts
@@ -970,6 +970,40 @@ describe("CoderSSHRuntime.ensureReady", () => {
     expect(getWorkspaceStatus).toHaveBeenCalledTimes(1);
   });
 
+  it("expires the cross-runtime activity cache after the fast-path window", async () => {
+    let now = 1_700_000_000_000;
+    const nowSpy = spyOn(Date, "now").mockImplementation(() => now);
+    try {
+      const getWorkspaceStatus = mock(() =>
+        Promise.resolve({ kind: "ok" as const, status: "running" as const })
+      );
+      const coderService = createMockCoderService({ getWorkspaceStatus });
+
+      const firstRuntime = createRuntime(
+        { existingWorkspace: true, workspaceName: "my-ws-expiring" },
+        coderService
+      );
+      expect(await firstRuntime.ensureReady()).toEqual({ ready: true });
+
+      const cachedRuntime = createRuntime(
+        { existingWorkspace: true, workspaceName: "my-ws-expiring" },
+        coderService
+      );
+      expect(await cachedRuntime.ensureReady()).toEqual({ ready: true });
+      expect(getWorkspaceStatus).toHaveBeenCalledTimes(1);
+
+      now += 5 * 60 * 1000 + 1;
+      const expiredRuntime = createRuntime(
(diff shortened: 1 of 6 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Please take another look.

PR #3209, opened 2026-04-30
🤖 fix: preserve assistant markdown whitespace

ammar-agent:
Summary
- Preserve assistant text separators when normalizing history for provider requests, preventing prior streamed Markdown from being re-sent as clobbered headings like `## VerdictThis`.

Background
- Streaming stores assistant text as separate deltas. The request sanitizer dropped whitespace-only text deltas before later coalescing, so a prior sequence like `## Verdict`, `\n\n`, `This...` could become `## VerdictThis...` in future prompts.

Implementation
- Coalesce consecutive assistant text parts before filtering whitespace-only blocks.
- Continue filtering assistant messages that are genuinely empty or whitespace-only.
- Add regression coverage for heading separators surviving the full sanitizer + provider transform path.

Validation
- `bun test src/node/services/messagePipeline.test.ts`
- `make typecheck`
- `make static-check`
- post-rebase `make static-check`

Risks
- Low. The change is request-only and preserves existing filtering of whitespace-only assistant messages while avoiding destructive separator deletion between meaningful text chunks.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$7.32`_

### DIFF
--- a/src/node/services/messagePipeline.test.ts
+++ b/src/node/services/messagePipeline.test.ts
@@ -1,0 +1,55 @@
+import { describe, expect, it } from "bun:test";
+import type { AssistantModelMessage, ModelMessage } from "ai";
+
+import { transformModelMessages } from "@/browser/utils/messages/modelMessageTransform";
+import { sanitizeAssistantModelMessages } from "./messagePipeline";
+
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3212, opened 2026-04-30
🤖 fix(openai): omit service_tier when not configured

### LINKED ISSUE #3196: Please add option to turn off OpenAI Service Tier!
Image

Image

ibetitsmike:
## Summary

Stop writing `serviceTier: "auto"` into provider options when the user hasn't configured one, and surface that "unset" state explicitly in the Settings UI and CLI. The provider field is now omitted entirely when no value is selected, letting OpenAI apply its own default routing.

Fixed: #3196 

## Background

Previously the OpenAI service tier silently defaulted to `"auto"` everywhere: in `buildProviderOptions`, in the CLI's `--service-tier` flag, and in the Settings dropdown. That meant requests always sent `service_tier: auto` even when the user never picked a tier, and the UI couldn't represent the actual "no preference" state at all (the dropdown showed `auto` whether the config was missing or explicitly set to `auto`).

## Implementation

- `buildProviderOptions` now reads `muxProviderOptions?.openai?.serviceTier` directly and only spreads `serviceTier` into the OpenAI options when it is set.
- `src/cli/run.ts` drops the `"auto"` default from `--service-tier`; if the flag is omitted, the entire `openai` block is left out of `providerOptions`.
- `ProvidersSection` adds a sentinel `"unset"` Select item labeled `Not configured (omit service_tier)`. Picking it sends `value: ""` through `setProviderConfig`, which the backend treats as a delete, removing the `serviceTier` key from `providers.jsonc`. A short-lived `openaiServiceTierSelectOverride` keeps the trigger label correct between optimistic update and refresh.
- `providerModelFactory` likewise no longer coerces a missing `serviceTier` into `"auto"`.

## Validation

- Sandboxed dev-server walk-through (clean install -> `flex` -> back to unset -> explicit `auto`); each transition verified against `providers.jsonc` on disk and the dropdown trigger label.
- `make static-check` (eslint, tsgo x2, prettier, code-to-docs links, shellcheck, hadolint) clean.
- `bun test src/cli/run.test.ts src/common/utils/ai/providerOptions.test.ts src/node/services/providerService.test.ts` (176 pass).

## Risks

Low. Behavior change is limited to OpenAI requests: when no tier was previously configured, requests will now omit `service_tier` instead of sending `"auto"`. OpenAI treats both as default routing, so live traffic should be unaffected. Existing configs with an explicit value continue to send that value verbatim.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `xhigh` • Cost: `$47.05`_

### DIFF
--- a/src/browser/features/Settings/Sections/ProvidersSection.tsx
+++ b/src/browser/features/Settings/Sections/ProvidersSection.tsx
@@ -83,6 +83,15 @@ type MuxGatewayLoginStatus = "idle" | "starting" | "waiting" | "success" | "erro
 type CodexOauthFlowStatus = "idle" | "starting" | "waiting" | "error";
 type CopilotLoginStatus = "idle" | "starting" | "waiting" | "success" | "error";
 
+const OPENAI_SERVICE_TIER_UNSET = "unset";
+
+type OpenAIServiceTier = "auto" | "default" | "flex" | "priority";
+type OpenAIServiceTierSelectValue = typeof OPENAI_SERVICE_TIER_UNSET | OpenAIServiceTier;
+
+function isOpenAIServiceTier(value: string): value is OpenAIServiceTier {
+  return value === "auto" || value === "default" || value === "flex" || value === "priority";
+}
+
 interface CodexOauthDeviceFlow {
   flowId: string;
   userCode: string;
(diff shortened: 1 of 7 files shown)


ibetitsmike:
@codex review

PR #3221, opened 2026-05-02
🤖 perf: word-pace text reveal; remove DOM-level streaming animations

ammar-agent:
## Summary

Replace DOM-level streaming animations with word-paced reveal at the engine level. `SmoothTextEngine` now advances `visibleLength` to whitespace boundaries instead of arbitrary character offsets, so each tick reveals one word + its trailing whitespace as a unit. All visual flourish (mask, shimmer, fade) has been removed.

## Background

The original "jittery streaming" complaint was structurally fixed in #3219 (cascade re-render kill, model-aware reveal rate, soft catch-up). After that landed, this PR explored five DOM-level visual approaches in succession:

1. Per-block CSS fade-in
2. Per-word blur fade-in (rehype plugin)
3. Per-AST-line span wrap
4. Static bottom-edge gradient mask
5. Hide-then-fade with animated `--reveal-y` mask offset
6. Static mask + animated `--stream-shimmer-x` band overlay

Each iteration produced a different perceptual artifact — left-to-right shimmer (#2), bottom-edge obfuscation (#4), abrupt mask transitions (#5), or "still doesn't feel great" highlight sweeps (#6) — that no amount of duration / easing / band-width tuning resolved.

The diagnosis: pacing, not visual flourish, is the right level to fix smoothness. Production chat UIs (ChatGPT, Claude.ai) feel smooth not because they animate completion cues but because they emit text at word boundaries. Humans parse text in word units; character-paced reveal forces an extra decoding step the eye registers as chop, regardless of how the characters are stylistically introduced.

### Engine: word-paced reveal

`SmoothTextEngine.tick(dtMs)` now greedy-pops "atoms" — runs of non-whitespace plus their trailing whitespace — while charBudget covers them. Per-tick reveal is capped at `MAX_FRAME_CHARS` (48) so a sudden catch-up burst doesn't dump 200 chars to the renderer in one frame. Long whitespace-free runs (URLs, minified identifiers) are bounded by `WORD_PACE_MAX_CHARS` (12) to prevent the engine from stalling on a 200-char "word".

The adaptive rate calculation (`getAdaptiveRate`) is unchanged — soft catch-up, model-aware steady-state, and hard-snap safety net all still operate. Only the reveal-position selection changed: from `min(charBudget, MAX_FRAME_CHARS)` chars to `findNextRevealBoundary(visibleLength)`.

### Stripped infrastructure

- `data-streaming` attribute (no longer gates anything)
- ResizeObserver + Element.animate() in TypewriterMarkdown
- `--stream-shimmer-x` CSS custom property + `@property` declaration
- mask-image gradient block in globals.css
- `data-stream-hide-tail` attribute + threshold-gating logic
- `MIN_FRAME_CHARS` constant (cost-of-next-atom is now the natural floor)
- `data-streaming` gating test in TypewriterMarkdown.test.tsx

### Added

- `WORD_PACE_MAX_CHARS=12` constant
- `findNextRevealBoundary` helper (whitespace scanner with cap)
- "reveals only at whitespace boundaries" engine test

## Validation

- `make static-check` passes locally (typecheck, lint, fmt-check, docs link checker)
- `bun test src/browser/utils/streaming/SmoothTextEngine.test.ts` — 12/12 pass
- `bun test src/browser/features/Messages/TypewriterMarkdown.test.tsx` — 4/4 pass
- `bun test src/browser/hooks/useSmoothStreamingText.test.tsx` — passes
- `bun test src/browser/features/Messages/` — broader Messages suite passes

The model-rate test was updated to use realistic whitespace-bearing text (`makeWords(50)` instead of `makeText(50)`); without word boundaries the test was tautological under word-pacing.

## Risks

- **Streaming UX is the entire point of the PR — regression risk is the whole surface.** Mitigations:
  - Engine-level invariants are covered by 12 unit tests including the new whitespace-boundary alignment check.
  - Bypass paths (replay rows, `bypassSmoothing=true`) are unchanged and continue to flush instantly.
  - Adaptive rate / hard-snap / soft catch-up logic from #3219 is untouched.
- **Long no-whitespace runs (URLs)** now reveal in 12-char chunks instead of character-by-character. Visually slightly chunkier than ideal but bounded — at 72 cps that's ~167ms per chunk, sub-perceptual.
- **Remove of mask** means the in-progress line is fully visible character-by-character (within a word) before the word-boundary atom. This is intentional: the reveal unit is the word, but partial words during accumulation are still visible. The character-level engine cadence makes this feel natural, not jittery.

## Pains

Five iterations of visual approaches, each addressing the previous's failure mode but introducing a new one. The fundamental insight that pacing > visual flourish only emerged after exhausting the visual-layer design space. The mask + ResizeObserver + WAAPI infrastructure was sound engineering at the wrong abstraction level.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$109.57`_


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed your P2: `data-streaming` is now gated on `isStreaming && streamSource !== "replay"` (see new comment block in `TypewriterMarkdown.tsx:74-82`). Replay rows that arrive with `isStreaming=true` during reconnect/history-replay no longer match the CSS selector. Test extended to cover all three cases: live → `"true"`, replay → absent, completed → absent.

ammar-agent:
@codex review

Pivoted the design after self-dogfood feedback. The previous per-block fade-in was invisible inside thinking blocks (typically one paragraph — the block fade fires once at mount, text grows silently inside).

Now uses per-**word** blur-fade-in via a small rehype plugin (`rehypeSplitWordsForFade`) that wraps each whitespace-bounded word in `<span class="sd-word">`. Stable identity per word position keeps completed words static while only newly-crossed whitespace boundaries mount fresh spans → CSS animation fires once per new word.

Skips `<pre>`/`<code>`/KaTeX/Mermaid subtrees so code, math, and diagrams stay structurally intact. Streaming-only — completed transcripts render without word spans (no DOM bloat). Still gated on `data-streaming="true"` (live, not replay) and `prefers-reduced-motion`.

7 new tests in `rehypeSplitWordsForFade.test.ts` cover the core invariants.

ammar-agent:
@codex review

Pivoted from per-AST-line span wrapping to a viewport-aware bottom-edge mask gradient. The mask operates on rendered geometry, so it tracks visual rows in the transcript rather than AST/HAST text nodes — fixing both prior iterations' weakness (per-line missed long-paragraph wrap-to-N-rows; per-word had horizontal jitter).

Net diff: ~−240 LoC. The rehype plugin and its test are deleted. Streaming markdown now goes through `REHYPE_PLUGINS_BASE` (no streaming-specific variant). The `data-streaming` gating on `TypewriterMarkdown` is unchanged (still live-only, not replay).

ammar-agent:
@codex review

Pivot from static gradient → JS-driven temporal fade. The prior commit's gradient obfuscated the in-progress line at low opacity; this commit hides it sharply with a mask, then pulses a `@property --reveal-y` custom property via `Element.animate()` when `ResizeObserver` reports height growth, fading completed lines in over 250ms.

Trigger is rendered geometry (`offsetHeight` change), so visual-row wraps inside long paragraphs participate (the prior per-AST-line approach missed those). `ResizeObserver` is guarded for happy-dom; real Electron Chromium always has it.

ammar-agent:
@codex review

Addressed the P1: the mask is now gated on a new `data-stream-hide-tail` attribute that JS toggles only when rendered height is at least ~1.5 line-heights. Below that (single-line streams, stream starts), no mask — the message renders fully visible. The threshold is computed from the element's `getComputedStyle().lineHeight` so it scales with font-size overrides.

Also skipped the WAAPI pulse on the first-time threshold crossing — the about-to-be-completed first line was already visible without the mask, so a pulse would briefly hide and re-fade it (flicker). Subsequent wraps (2+ lines stable, growing) still pulse normally.

ammar-agent:
@codex review

Fixed: switched to presence selector `[data-stream-hide-tail]` since `el.toggleAttribute()` sets the attribute with an empty value (not `"true"`).

ammar-agent:
@codex review

Fixed: reverted both the `overrides` block in package.json and the override-injection in `scripts/smoke-test.sh`. Upstream published `@smithy/util-retry@4.3.8` ~32min after the broken 4.3.7, fixing the `workspace:^` resolution failure — verified clean install of `latest` in a fresh root works. With the upstream fix in place, the workaround is no longer needed and the smoke test returns to its canonical form (no override injection, so `SKIP_SHRINKWRAP=1` again represents lockfile-free end-user-equivalent installs).

ammar-agent:
@codex review

Rebased on main (no conflicts). None of the 9 new commits on main touched streaming-related files; local `make typecheck` (clean rebuild), `lint`, `fmt-check`, and the Messages test suite all pass.

ammar-agent:
@codex review

User feedback: prior 250ms ease-out pulse + 0.4em band felt punchy. Tuned three values to smooth the reveal:
- Band: 0.4em → 0.8em (each row spends more time partially faded as the mask sweeps)
- Duration: 250ms → 400ms
- Easing: ease-out → cubic-bezier(0.22, 1, 0.36, 1) (out-quart, more even deceleration)

Pure presentation tuning; no behavioral changes to the mask/ResizeObserver/WAAPI pipeline.

ammar-agent:
@codex review

Replaced the fade-in (animated `--reveal-y` mask offset) with a shimmer effect: a 40%-wide translucent `currentColor` highlight band that sweeps across each just-completed line over 700ms. Implementation:

- CSS: drop `@property --reveal-y` and the mask offset. Add `@property --stream-shimmer-x` (typed `<percentage>`, namespaced to avoid the existing CompactionBackground `--shimmer-color`). Add a `::before` pseudo-element on the streaming wrapper at `bottom: 1.6em; height: 1.6em` with a horizontal gradient + `transform: translateX(var(--stream-shimmer-x))`. `currentColor` adapts to dark/light themes (brightens dark text, darkens light).
- JS: replace the `--reveal-y` pulse with a `--stream-shimmer-x` animation from `-100%` to `250%` over 700ms. Drop the "skip pulse on first threshold crossing" guard — the shimmer doesn't hide anything (just adds a translucent overlay sweep) so there's no flicker risk on first wrap.

GPU-friendly: transform-only animation runs on compositor; @property registration makes the percentage interpolatable.

ammar-agent:
@codex review

Pivoted the approach: word-pace at the engine level instead of DOM animation. After 5 visual iterations (per-block fade → per-word fade → per-line wrap → static mask → animated mask pulse → shimmer overlay) all produced perceptual artifacts of various kinds, dropped the entire visual layer and instead changed `SmoothTextEngine.tick()` to advance to the next whitespace boundary rather than the next N characters.

Net: -193 LoC. Removed: mask CSS, shimmer overlay, ResizeObserver, WAAPI animation, `data-streaming` attribute, `MIN_FRAME_CHARS` constant. Added: `WORD_PACE_MAX_CHARS=12`, `findNextRevealBoundary` helper, and a new whitespace-boundary alignment test.

PR description fully rewritten to describe the new approach.

ammar-agent:
@codex review

Two follow-ups:
1. **Time-smoothing**: replaced the greedy multi-atom reveal with strict 1-atom-per-tick + dt clamp at 33ms. Even with huge budget (catch-up after RAF gap, very high adaptive rate), reveals are spread across animation frames. Removed the now-dead `MAX_FRAME_CHARS` constant.
2. **Unicode whitespace** (your P2): swapped the 5-codepoint ASCII switch for `\s` regex match. NBSP, ideographic space, thin/em spaces, line/paragraph separators all trigger boundary detection. New test exercises 5 distinct Unicode whitespace codepoints.

PR #3222, opened 2026-05-02
🤖 refactor: move ChatInput attach/voice icons out of textarea overlay

ammar-agent:
## Summary

Move the attach (📎) and voice (🎙️) input buttons from the absolute overlay inside the chat textarea into the controls row directly to the left of the Send button, so they can never visually intersect typed text and read as a single trailing action cluster.

## Background

The previous design positioned both buttons absolutely at the bottom-right corner of the textarea and reserved a fixed `pr-16` (4rem) right padding to keep text out of their way. This had two problems:

1. **Text could still ride under the icons.** The reserved gutter is a fixed pixel value, but font metrics, focus rings, and the icon group's effective width could push glyphs into the icon zone — most visibly the bottom line of multi-line wrapped input.
2. **Wasted horizontal space.** Every line in the textarea lost 4rem on the right even when the input was short.

## Implementation

- `VimTextArea`: removed the `trailingAction` prop and its associated machinery (`pr-16` conditional padding, `scrollbarGutter: "stable"`, the absolute overlay div). The component had a single consumer, so removing the surface area is a net cleanup.
- `ChatInput`: relocated `<AttachFileButton>` and `<VoiceInputButton>` into a new `InputMethodGroup` that sits in the right half of the existing controls row, between the agent mode picker (`Exec ▾`) and the Send button. Internal `gap-0` and a `-ml-1.5` on the Send container override the parent's `gap-1.5` cadence so the trailing actions form one tight cluster `[Exec ▾] 📎 🎙 ➤` rather than three small icons stranded in a wide-spaced row.
- Reclaimed the right-side gutter for the empty-input keyboard hints (`right-18` → `right-2`) since the textarea no longer has icons floating over its bottom-right corner.

## Validation

- Verified empty / short / multi-line wrapped inputs in Storybook (`App/Chat/Input → VoiceInputNoApiKey`); the bottom line of wrapped text never intersects either icon.
- Recording overlay still replaces the textarea correctly while the controls row (with the voice button) stays put.

## Risks

Low. The change is purely visual layout: button props, callbacks, and disabled states are unchanged, and no logic in `ChatInput` was altered. The `trailingAction` prop on `VimTextArea` had a single caller, so removing it is safe.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `high` • Cost: `$1.53`_

### DIFF
--- a/src/browser/components/VimTextArea/VimTextArea.tsx
+++ b/src/browser/components/VimTextArea/VimTextArea.tsx
@@ -31,7 +31,6 @@ export interface VimTextAreaProps extends Omit<
   onChange: (next: string, caretIndex?: number) => void;
   isEditing?: boolean;
   suppressKeys?: string[]; // keys for which Vim should not interfere (e.g. ["Tab","ArrowUp","ArrowDown","Escape"]) when popovers are open
-  trailingAction?: React.ReactNode;
   ghostHint?: string | null;
   /** Called when Escape is pressed in normal mode (vim) - useful for cancel edit */
   onEscapeInNormalMode?: () => void;
@@ -49,7 +48,6 @@ export const VimTextArea = React.forwardRef<HTMLTextAreaElement, VimTextAreaProp
       isEditing,
       suppressKeys,
       onKeyDown,
-      trailingAction,
       ghostHint,
       onEscapeInNormalMode,
       focusBorderColor,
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

ammario (approved):

PR #3227, opened 2026-05-03
🤖 fix: bump streamingStatsStore on stream-start to clear stale TPS

ammar-agent:
## Summary

The streaming TPS pill briefly displayed the prior turn's TPS at the start of a new assistant stream. Fix invalidates the streaming-stats cache when `stream-start` arrives so every new turn forces a fresh recompute.

## Background

User report: _"Streaming TPS starts with stale value."_ Two prior fixes addressed adjacent issues — flooring TPS time-span to 1s + clearing token state on `stream-error`, and invalidating the streaming-stats cache on `stream-error` (both squashed into #3219). Despite those, the stale-TPS flash still happened at stream **start**.

## Implementation

`bufferedEventHandlers["stream-start"]` in `WorkspaceStore` bumped `states` and `usageStore` but never `streamingStatsStore`. In the foreground happy path this was masked because `stream-end` / `stream-abort` / `stream-error` all bump `streamingStatsStore` on terminal events.

The bug surfaces whenever `aggregator.activeStreams` is reset **without** a corresponding stats bump. Four sites in `WorkspaceStore` call `aggregator.clearActiveStreams()` directly:

- caught-up reconnect/hydration
- background streaming-stop
- activity-driven generation advance
- post-`addWorkspace`

After any of those, the cache still holds the prior turn's `WorkspaceStreamingStats` and `useWorkspaceStreamingStats` returns it verbatim until the new turn's first delta fires `scheduleStreamingStateBump`. That window is exactly the "stale TPS" the user saw.

Fix: bump `streamingStatsStore` on `stream-start`. That choke point is invariant to how the previous stream was wound down, so the single bump makes the cache invariant — every new turn forces a fresh recompute regardless of the prior teardown path.

**Rejected alternative:** bumping inside `clearActiveStreams` itself (or at each of the four call sites). That requires four edits today and stays fragile to future call sites; the bug is at the cache layer, not the aggregator's `deltaHistory`.

## Validation

- New regression test `"invalidates streaming-stats cache on stream-start so the new turn never displays the prior turn's TPS"` simulates the reconnect path (`clearActiveStreams()` without a bump), then issues the next `stream-start` and asserts the cache is invalidated. Verified to fail without the fix and pass with it.
- Streaming surface (`WorkspaceStore.test.ts`, `StreamingTPSCalculator.test.ts`, `applyWorkspaceChatEventToAggregator.test.ts`, `StreamingBarrierView.test.tsx`) — 138 pass, 0 fail.

## Risks

Low. The change is a single `streamingStatsStore.bump(workspaceId)` per assistant turn. The cache it invalidates is already bumped on every coalesced delta (~30Hz during streaming), so one extra bump per turn is negligible. No behavioral change for non-streaming callers; leaf-subscribed `useWorkspaceStreamingStats` consumers (`StreamingBarrier`, `TypewriterMarkdown`) just see one additional notification per turn.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$19.03`_

### DIFF
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -1946,6 +1946,106 @@ describe("WorkspaceStore", () => {
       unsubscribe();
     });
 
+    it("invalidates streaming-stats cache on stream-start so the new turn never displays the prior turn's TPS", async () => {
+      // Repro for "Streaming TPS starts with stale value": reconnect / hydration
+      // paths drop aggregator.activeStreams via clearActiveStreams() WITHOUT
+      // bumping streamingStatsStore. If the next stream-start also doesn't bump,
+      // useWorkspaceStreamingStats keeps returning the previous turn's cached
+      // stats until the first delta of the new turn. The fix is to bump on
+      // stream-start so every new turn forces a fresh recompute.
+      const workspaceId = "stream-start-invalidates-stats";
+      const streamModel = "anthropic:claude-opus-4-6";
+
+      mockOnChat.mockImplementation(async function* (
+        _input?: { workspaceId: string; mode?: unknown },
+        options?: { signal?: AbortSignal }
+      ): AsyncGenerator<WorkspaceChatMessage, void, unknown> {
+        if (options?.signal?.aborted) {
+          yield { type: "caught-up" };
+        }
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3230, opened 2026-05-03
🤖 feat: simplify /new to mirror /fork

ammar-agent:
## Summary

`/new` now mirrors `/fork` — no required workspace name, no flags, no modal. Type `/new` to create a workspace from the project's trunk; type `/new <message>` to also send a starting message and let the backend derive the workspace title from it.

## Background

`/new` previously demanded a workspace name (`/new feature-branch`) and supported `-t <trunk>` and `-r <runtime>` flags; omitting the name fell through to a separate creation modal. Meanwhile `/fork` had already learned the seamless flow we want: no required name, optional start message, auto-generated branch name + title. Aligning `/new` with `/fork` removes a significant UX wart and a meaningful chunk of duplicated parsing/handling code on the desktop, mobile, and ACP surfaces.

## Implementation

- **Slash parser** (`registry.ts`, `types.ts`): `/new` collapses to `{ type: "new"; startMessage?: string }`, mirroring `/fork`. The slash hint becomes `[start message]`.
- **Frontend handler** (`chatCommands.ts`): `handleNewCommand` no longer dispatches the creation modal event. It always calls `createNewWorkspace`, which now accepts an optional `workspaceName` and forwards `pendingAutoTitle` so the first user message drives LLM-based title generation (same flag the fork flow uses).
- **ORPC schema** (`schemas/api.ts`): `workspace.create.input.branchName` is now optional, and `pendingAutoTitle?: boolean` joins the input. The router threads it through to the service.
- **Service** (`workspaceService.ts`): `WorkspaceService.create()` accepts `branchName: string | undefined`. When omitted, it scans the project's existing workspaces + local branches and picks the next free `workspace-N` slot via the existing `generateForkBranchName` helper. `pendingAutoTitle: true` is persisted on the new workspace entry, so the first message can title it (driven by the same `maybeRunPendingAutoTitleFromMessage` path fork already uses).
- **ACP & mobile parity** (`acp/slashCommands.ts`, `acp/agent.ts`, `mobile/.../slashCommandRunner.ts`): same simplification — no workspace name argument, no `-t`/`-r` flags, optional start message.

The branch name auto-generation reuses fork's helper, so both flows numerically share state (e.g., creating `/fork` of `workspace-1` produces `workspace-1-1`, while a new `/new` produces `workspace-2`).

## Validation

- Added a unit test in `workspaceService.test.ts` covering the new branchless `create()` path: it verifies the next `workspace-N` slot is selected past existing names and that `pendingAutoTitle: true` is persisted.
- Rewrote the `/new` parser tests, ACP slash tests, and mobile/desktop handler call shapes around the new contract.
- `make static-check` ✅ (typecheck + lint + format + docs).
- `bun test src` runs all rewritten and added tests cleanly. (The 21 `GitStatusStore` failures observed in the full test run reproduce on `main` and are unrelated to this change.)

## Risks

- Backward-incompatible slash-command syntax: anyone with muscle memory for `/new <name>`, `/new <name> -t main`, or `/new -t branch ...` will find their input now sent verbatim as the workspace's first message instead of becoming a workspace name. We don't ship migrations for slash commands and the "minor breaking changes are expected" guidance applies; the upside is a meaningfully simpler flow that matches fork's.
- The `workspace.create` IPC schema is breaking (`branchName` is now optional, plus `pendingAutoTitle`). Frontend/backend ship together so there's no real cross-version risk.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$23.30`_


ammar-agent:
@codex review

PR #3234, opened 2026-05-05
🤖 fix: show task_await elapsed timing

ethanndickson:
Summary

Add elapsed timing to task_await so users can see how long an await call has been running and how long completed sub-agent waits took.

Background

Bash tool calls already expose elapsed timing, but task_await only showed a generic executing state while waiting. That made long waits look stuck even when the task was still running.

Implementation

The task_await backend now surfaces elapsed_ms for sub-agent results from existing task timestamps. The task_await UI now shows live elapsed time while executing, shows row-level elapsed time for active awaited tasks when start timestamps are known, and formats completed elapsed values with the shared duration formatter.

Risks

Low. The backend change is additive because elapsed_ms was already accepted by the task_await result schema, and the UI change only adds optional timing labels around existing task_await rendering.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$17.34`_

### DIFF
--- a/src/browser/features/Tools/Shared/ElapsedTimeDisplay.tsx
+++ b/src/browser/features/Tools/Shared/ElapsedTimeDisplay.tsx
@@ -3,13 +3,20 @@ import React, { useEffect, useRef } from "react";
 interface ElapsedTimeDisplayProps {
   startedAt: number | undefined;
   isActive: boolean;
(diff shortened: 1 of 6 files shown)


ethanndickson:
@codex review

PR #3239, opened 2026-05-06
🤖 feat: add Tailscale bind host selection

coadler:
## Summary

Adds Tailscale-aware bind host selection to the remote access experiment so users can bind mux to a detected local Tailscale address instead of only localhost or all interfaces.

## Implementation

- Detects bindable Tailscale addresses from OS network interfaces, using Tailscale-named interfaces directly and consulting `tailscale ip` to prove generic interface addresses belong to Tailscale.
- Filters Tailscale CLI output to known Tailscale IPv4/IPv6 ranges so generic RFC6598 CGNAT addresses are not mislabeled without proof.
- Exposes detected addresses through API server status as `tailscaleBindHosts`.
- Adds Tailscale entries to the remote access bind-host selector and persists the selected address via the existing `apiServerBindHost` setting.

## Validation

- `make static-check`
- `bun test node/services/serverService.test.ts`

## Risks

Low. The change extends the existing custom bind-host path and only changes remote access settings when the user explicitly selects a Tailscale address. Generic 100.64.0.0/10 addresses now require Tailscale CLI proof before being labeled as Tailscale.

---


coadler:
@codex review

coadler:
@codex review

Please take another look. I tightened generic 100.64.0.0/10 detection to require Tailscale CLI proof and resolved the prior thread.

PR #3245, opened 2026-05-06
🤖 fix: preserve subagent retention setting on partial saves

coadler:
## Summary

Preserves optional task settings across partial `config.saveConfig` calls so enabling “Preserve subagents until the workspace gets archived” no longer gets silently reset by unrelated settings saves.

## Background

The task cleanup path already respects `preserveSubagentsUntilArchive` when the setting remains true. The bug was that the API schema allows clients to save only the required numeric task limits, and the router normalized that partial payload as a full task-settings replacement. Missing optional flags therefore defaulted back to `false`, allowing completed subagents to be cleaned up before archive.

## Implementation

- Merge incoming task settings over the current normalized task settings before saving.
- Add a router regression test that seeds both optional task flags to true, performs a schema-minimum task-settings save, and verifies the optional flags stay true while numeric limits still update.

## Validation

- Confirmed the new regression test failed before the fix, then passed after merging current task settings before normalization.
- `bun test src/node/orpc/router.test.ts`
- `bun test src/node/services/taskService.test.ts -t "preserve subagents until archive"`
- `make typecheck`
- `make lint`
- `make fmt-check`
- `make static-check`

## Risks

Low. The change is scoped to `config.saveConfig` task settings persistence and only preserves existing optional values when a client omits them. Explicit task-setting changes still win.

---

### DIFF
--- a/src/node/orpc/router.test.ts
+++ b/src/node/orpc/router.test.ts
@@ -67,4 +67,39 @@ describe("router config.saveConfig", () => {
     expect(saved.agentAiDefaults?.foo?.advisorEnabled).toBe(true);
     expect(saved.subagentAiDefaults?.foo).toBeUndefined();
   });
+
+  test("preserves optional task settings when a save omits them", async () => {
+    await config.editConfig((current) => ({
+      ...current,
+      taskSettings: {
+        ...DEFAULT_TASK_SETTINGS,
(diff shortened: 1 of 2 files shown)


coadler:
@codex review

coadler:
@codex review

coadler:
@codex review

PR #3247, opened 2026-05-06
🤖 fix: accept description as alias for display_name in bash tool

ammar-agent:
## Summary

DeepSeek V4 emits `description` instead of `display_name` for the bash tool, causing every bash invocation to fail validation. Normalize `description` as an undocumented alias for `display_name` in the schema preprocess so calls validate without changing the public tool surface.

## Background

The `bash` tool requires `display_name` for every invocation. DeepSeek V4 reliably emits `description` instead, so we add an alias mirroring the existing `command` → `script` normalization. The alias is intentionally **undocumented** in the tool description: we don't want to invite other models to use the wrong field.

## Implementation

Refactored the bash schema's `preprocess` from a single `if/else if/else` chain into chained mutations on a local `obj` reference, so multiple normalizations can compose:

1. `command` → `script` (existing).
2. `description` → `display_name` (new, with a comment noting DeepSeek V4 as the reason).

Canonical fields always win when both are provided, mirroring the existing `command`/`script` precedence.

## Validation

- New unit tests in `toolDefinitions.test.ts` cover the alias and the precedence behavior.
- `bun test src/common/utils/tools/toolDefinitions.test.ts`: 36 pass.
- `make static-check`: passed locally.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$0.62`_

### DIFF
--- a/src/common/utils/tools/toolDefinitions.test.ts
+++ b/src/common/utils/tools/toolDefinitions.test.ts
@@ -152,6 +152,37 @@ describe("TOOL_DEFINITIONS", () => {
     expect(parsed.success).toBe(false);
   });
 
+  it("accepts bash tool calls using description (alias for display_name)", () => {
+    // DeepSeek v4 emits `description` instead of `display_name`; ensure it normalizes.
+    const parsed = TOOL_DEFINITIONS.bash.schema.safeParse({
+      script: "ls",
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3248, opened 2026-05-07
🤖 fix: relax OpenAI WebSocket transport gating

ThomasK33:
Summary
- Relax OpenAI WebSocket transport settings visibility so the UI shows the toggle whenever the OpenAI wire format is `responses`.
- Let runtime attempt the WebSocket transport with custom OpenAI base URLs instead of preemptively disabling it.
- Preserve the existing HTTP routing path for Codex OAuth requests so Mux can still rewrite the endpoint and attach ChatGPT OAuth headers.
- Fix an unrelated Worker error handler type annotation so `make static-check` remains green.

Background
The previous UI hid the WebSocket transport toggle whenever OpenAI had any configured/resolved base URL or Codex OAuth was the active default. That blocked users from enabling the experimental transport even in cases where it may work. The only UI-level hard requirement is the Responses wire format.

Implementation
- Simplified provider settings visibility to `wireFormat === "responses"`.
- Removed the custom Base URL exclusion from OpenAI WebSocket transport activation.
- Kept Codex OAuth requests on the existing HTTP fetch wrapper because those requests require endpoint rewriting and ChatGPT OAuth headers before dispatch.
- Updated tests to assert the toggle remains available with custom OpenAI base URLs and Codex OAuth, while runtime still protects Codex OAuth routing.

Validation
- `bun test src/browser/features/Settings/Sections/ProvidersSection.test.tsx`
- `bun test src/node/services/providerModelFactory.test.ts`
- `make static-check`

Risks
This makes an experimental transport easier to enable for endpoints that may not support it. Failures for unsupported custom endpoints should surface at request time instead of being blocked by settings visibility.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `high` • Cost: `$4.05`_

### DIFF
--- a/src/browser/features/Settings/Sections/ProvidersSection.test.tsx
+++ b/src/browser/features/Settings/Sections/ProvidersSection.test.tsx
@@ -380,7 +380,7 @@ describe("ProvidersSection", () => {
     });
   });
 
-  test("hides the OpenAI WebSocket transport toggle when Codex OAuth is the active default", async () => {
+  test("shows the OpenAI WebSocket transport toggle when Codex OAuth is the active default", async () => {
     const view = renderProvidersSection();
     view.providersConfig.openai.codexOauthSet = true;
     view.providersConfig.openai.apiKeySet = false;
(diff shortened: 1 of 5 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed the Codex OAuth issue by keeping routed Codex OAuth requests on the existing HTTP fetch wrapper while leaving the UI toggle visible for Responses wire format.

PR #3249, opened 2026-05-07
🤖 fix: keep immersive sidebar review click on its hunk

ammar-agent:
## Summary

Fixes a bug where clicking a pending review in the immersive Notes sidebar did nothing: the immersive view stayed on the originally selected hunk and the review was nowhere on screen.

## Background

When a review is submitted in immersive mode it gets `status: "pending"` and the hunk it was attached to is auto-marked read. With hide-read on (the default), that hunk drops out of `ReviewPanel`'s `filteredHunks`. Clicking the pending review in the immersive Notes sidebar calls `navigateToReview(...)`, which correctly resolves the target hunk id from `allHunks` (not `hunks`) and calls `onSelectHunk(targetHunkId)`. But `ReviewPanel`'s selection-validity effect then ran:

```ts
const selectionValid = selectedHunkId && filteredHunks.some((h) => h.id === selectedHunkId);
if (!selectionValid) setSelectedHunkId(filteredHunks[0].id);
```

That clobbered the explicit sidebar selection on the very next render, so the immersive view snapped back to the first visible hunk and the user never saw the hunk their review was attached to.

## Implementation

In `ReviewPanel.tsx`, when immersive mode is active, validate the selected hunk against the unfiltered `hunks` list instead of `filteredHunks`. The selection still resets when the hunk truly disappears from the diff (e.g. after a refresh), which is what the original effect was actually trying to defend against. The non-immersive panel path is untouched because that view only ever renders `filteredHunks`.

## Validation

- Added two regression tests in `ImmersiveReviewView.test.tsx`:
  1. **Contract test** — clicking a sidebar review whose hunk is filtered out still propagates the correct hunk id through `onSelectHunk` (catches regressions in the immersive view's `navigateToReview`).
  2. **Parent-harness integration test** — mirrors `ReviewPanel`'s selection-validity effect and verifies the immersive header switches to the reviewed file after the click. Confirmed the test fails (header shows `src/visible.ts` instead of `src/reviewed.ts`) when the immersive branch is reverted to the buggy `filteredHunks.some(...)` check, and passes with the fix.
- `make static-check` (typecheck + lint + fmt-check + doc link check) — pass.
- `bun test` for `RightSidebar` / `Shared` / `ReviewsBanner` / `hooks` — 278/278 pass.

## Risks

Low. The change is gated on `isImmersive`, so the non-immersive Code Review panel keeps its exact previous behavior. The new immersive branch only relaxes the reset condition (stale-selection check expands from `filteredHunks` to all `hunks`), so the only surface where behavior changes is the case the bug already required.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$1.73`_

### DIFF
--- a/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.test.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.test.tsx
@@ -1,11 +1,11 @@
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 import { cleanup, fireEvent, render } from "@testing-library/react";
 import { GlobalWindow } from "happy-dom";
-import type { ComponentProps } from "react";
+import { useEffect, useState, type ComponentProps } from "react";
 
 import { ThemeProvider } from "@/browser/contexts/ThemeContext";
 import type { FileTreeNode } from "@/common/utils/git/numstatParser";
-import type { DiffHunk } from "@/common/types/review";
+import type { DiffHunk, Review } from "@/common/types/review";
 
 interface MockApiClient {
   workspace: {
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3250, opened 2026-05-07
🤖 fix: stop Run-button click from crashing the app

ammar-agent:
## Summary

Two contributing causes for the user report *"Play on code blocks sometimes leads to App crash"*: (1) the backend's `TerminalService.emitOutput` was unprotected against `xterm-headless` throws on the libuv tick, which escalated into the Electron main process's "Application Error" dialog; and (2) the frontend Run-button path discarded `createTerminalSession` rejections via `void`, leaving the sidebar half-expanded with no new tab when (e.g.) the workspace was transcript-only or the runtime wasn't ready.

## Background

The Run icon (`<Play>` at `MarkdownComponents.tsx:255`) calls `messageListContext.openTerminal({ initialCommand })`, which threads through `ChatPane → WorkspaceShell.handleOpenTerminal → RightSidebar.handleAddTerminal`, and finally `TerminalService.create` + `sendInput(initialCommand + "\n")`. Two latent failure modes along this chain produce the same user-visible symptom of "the app froze / crashed" after clicking Play:

| Path | Cause | Symptom |
|---|---|---|
| Backend | `TerminalService.emitOutput` writes PTY bytes via `headless.write(data)` and `emitter.emit("data", data)` with no try/catch. The frontend already wraps `term.write` (`TerminalView.tsx:351-355`) because *"xterm WASM can throw 'memory access out of bounds' intermittently"*; the headless build has the same risk. Because `emitOutput` runs from `node-pty`'s `onData` on the libuv tick, an uncaught throw escalates to `process.uncaughtException` in main, which calls `dialog.showErrorBox("Application Error", …)` (`src/desktop/main.ts:151-166`) — the literal blocking crash. | Modal "Application Error" dialog. |
| Frontend | `RightSidebar.handleAddTerminal` (`RightSidebar.tsx:1181-1186`) and `useOpenTerminal` fired `void createTerminalSession(...).then(...)` with no `.catch`. Backend rejections (transcript-only workspaces with no `projectPath`, archived worktrees, disconnected SSH/Devcontainer) became unhandled promise rejections; the sidebar already ran `setCollapsed(false)` before the await, so it visibly twitches open with no new tab. | Sidebar half-expands, no new tab, no error toast → reads as a freeze. |

## Implementation

**Backend — `src/node/services/terminalService.ts`**

`emitOutput` now wraps both the headless write and the listener emission in `try`/`log.warn`. Mirrors the renderer-side defense and isolates a single misbehaving subscriber from taking down the Electron main process. No behavior change on the happy path.

**Frontend — `src/browser/features/RightSidebar/RightSidebar.tsx`**

`handleAddTerminal` chains `.catch` on the create promise and surfaces the message via a new `usePopoverError` (the same pattern `WorkspaceMenuBar` already uses for archive/fork/stop errors), rendered through the existing `<PopoverError>` component. The wrapper is kept non-async (`(): void`) so the existing `addTerminalRef`/`onAddTerminal` callsites still satisfy `@typescript-eslint/no-misused-promises`.

**Frontend — `src/browser/hooks/useOpenTerminal.ts`**

Wrapped both branches (`createTerminalSession` + `openTerminalPopout` for browser/SSH/devcontainer; `api.terminal.openNative` for Electron-local) in a single `try/catch` so the `WorkspaceMenuBar`'s `void openTerminalPopout(...)` call cannot become an unhandled rejection either.

## Validation

- New regression test `TerminalService > isolates emitOutput from throwing output subscribers` registers a deliberately-throwing `onOutput` listener and asserts the captured `onData` callback does not propagate the throw — without the defense this would crash the host process. Tests behavior, not literals.
- `make static-check`, `make typecheck`, `make lint` all green.
- `bun test src/node/services/terminalService.test.ts` — 41/41 pass.
- `bun test src/browser/features/Messages/MarkdownComponents.test.tsx` — 12/12 pass.
- `bun test src/browser/components/WorkspaceShell/WorkspaceShell.test.tsx` — 10/10 pass.

## Risks

Low. The backend change is a pure widening of a previously-uncaught throw path; the only observable behavioral change on a malformed escape sequence is that the host process logs a warning instead of crashing, and the data emission still reaches subscribers (verified by the regression test). The frontend changes are additive — they convert silent failures into a 5-second `PopoverError` toast and don't alter the success path.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$15.61`_

### DIFF
--- a/src/browser/features/RightSidebar/RightSidebar.tsx
+++ b/src/browser/features/RightSidebar/RightSidebar.tsx
@@ -18,6 +18,9 @@ import {
 import { useAPI } from "@/browser/contexts/API";
 import { StatsContainer } from "@/browser/features/RightSidebar/StatsContainer";
 import { ErrorBoundary } from "@/browser/components/ErrorBoundary/ErrorBoundary";
+import { usePopoverError } from "@/browser/hooks/usePopoverError";
+import { PopoverError } from "@/browser/components/PopoverError/PopoverError";
+import { getErrorMessage } from "@/common/utils/errors";
 
 import { ReviewPanel } from "@/browser/features/RightSidebar/CodeReview/ReviewPanel";
 import { OutputTab } from "@/browser/components/OutputTab/OutputTab";
@@ -610,6 +613,12 @@ const RightSidebarComponent: React.FC<RightSidebarProps> = ({
     null
   );
 
+  // Surface backend failures from terminal creation (e.g., the markdown Run button kicking off
+  // a session against a transcript-only workspace, or a runtime that isn't connected). Without
+  // this the click silently expands the sidebar with no new tab, which users perceive as a hang
+  // or app crash.
+  const terminalCreateError = usePopoverError();
+
   // Manual collapse state (persisted globally)
   const [collapsed, setCollapsed] = usePersistedState<boolean>(RIGHT_SIDEBAR_COLLAPSED_KEY, false, {
     listener: true,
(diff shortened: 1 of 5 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review
Please take another look — addressed the popout-rejection gap by making `openTerminalPopout` return `Promise<void>` and updating both raw callsites (`useOpenTerminal`, `RightSidebar.handlePopOutTerminal`) to observe the rejection.

ammar-agent:
@codex review
Friendly nudge — looks like the previous request may have stalled (only an `eyes` reaction so far). The fix-up commit (`820ebf430`) addresses your popout-rejection feedback by making `openTerminalPopout` return `Promise<void>` and updating both raw callsites to observe the rejection.

ammar-agent:
@codex review
Please take another look — addressed the `as any` concern by capturing the `onData` callback from `createSessionMock.mock.calls` instead of swapping the mock implementation. The test is now fully type-safe.

ammar-agent:
@codex review
Please take another look — addressed the listener-isolation concern by moving the per-listener try/catch into `onOutput`'s handler wrapper, so a throwing earlier subscriber no longer aborts EventEmitter's synchronous broadcast and starves later subscribers. Test now asserts a healthy listener still receives every chunk when a throwing listener is registered first.

PR #3252, opened 2026-05-07
🤖 fix: show preserved subagents in sidebar

coadler:
## Summary

Shows completed subagent rows in the sidebar by default when “Preserve subagents until the workspace gets archived” is enabled, so preserved subagents no longer appear to disappear after completion.

## Background

The backend now preserves completed subagent workspaces correctly, but the sidebar still used the existing completed-subagent collapse behavior. That meant the workspace remained in config but was hidden unless the parent row was manually expanded, making preservation look broken.

## Implementation

- Subscribe `ProjectSidebar` to config task settings.
- Default completed-subagent expansion to `preserveSubagentsUntilArchive`.
- Keep per-parent manual toggles, so users can still hide/show completed subagents for individual parents.
- Guard overlapping config refreshes so stale responses cannot overwrite the latest preserve setting.
- Add a regression test that proves a reported child row is visible by default when preservation is enabled and can still be toggled hidden.

## Validation

- Confirmed the new sidebar test failed before the fix and passed after the fix.
- `bun test src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx -t "shows completed child rows by default"`
- `bun test src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx -t "completed-subagent toggles"`
- `bun test src/browser/utils/ui/workspaceFiltering.test.ts -t "reported children|interrupted children"`
- `make typecheck`
- `make lint`
- `make fmt-check`
- `make static-check`

## Risks

Low-to-moderate. The change is scoped to sidebar visibility and config subscription behavior. Users with preservation disabled keep the existing collapsed-by-default behavior; users with preservation enabled get completed subagents shown by default, with manual per-parent toggles still honored.

---

### DIFF
--- a/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
+++ b/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
@@ -11,6 +11,7 @@ import { EXPANDED_PROJECTS_KEY } from "@/common/constants/storage";
 import { getDraftScopeId, getInputKey } from "@/common/constants/storage";
 import { MULTI_PROJECT_SIDEBAR_SECTION_ID } from "@/common/constants/multiProject";
 import { DEFAULT_RUNTIME_CONFIG } from "@/common/constants/workspace";
+import { DEFAULT_TASK_SETTINGS } from "@/common/types/tasks";
 import type { AgentRowRenderMeta } from "@/browser/utils/ui/workspaceFiltering";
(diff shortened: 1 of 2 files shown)


coadler:
@codex review

coadler:
@codex review

Please take another look.

coadler:
@codex review

Please take another look.

PR #3257, opened 2026-05-09
🤖 fix: clarify math delimiter instructions

ammar-agent:
Summary

Clarifies Mux's base markdown instructions so agents emit KaTeX math with double-dollar delimiters, including an explicit inline example and a warning that single-dollar `$...$` delimiters are treated as plain text or currency.

Background

Mux intentionally keeps broad single-dollar math parsing disabled because it conflicts with regular currency text and Mux's `$skill` inline references. This avoids adding heuristic markdown parsing while steering future model output toward the delimiter format the renderer already supports reliably.

Validation

- `bun test src/browser/features/Messages/MarkdownStyles.test.ts src/node/services/systemMessage.test.ts`
- `make static-check`

Risks

Low. This only changes prompt text and regenerated documentation/skill snapshot content; it does not alter markdown rendering behavior.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$4.24`_

### DIFF
--- a/docs/agents/system-prompt.mdx
+++ b/docs/agents/system-prompt.mdx
@@ -26,7 +26,7 @@ Always verify repo facts before making correctness claims; trusted tool output a
   
 <markdown>
 Your Assistant messages display in Markdown with extensions for mermaidjs and katex.
-For math expressions, prefer \`$$...$$\` delimiters for the most reliable rendering.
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

PR #3261, opened 2026-05-09
🤖 fix: suppress post-compaction continue notifications

ammar-agent:
## Summary

Suppress notify-on-response for source-marked internal post-compaction resume turns. The policy is now owned by the transcript aggregator/response-completion layer instead of being threaded through backend stream/activity schemas. User-authored follow-up text after `/compact` (even exactly `Continue`) still notifies normally.

## Background

I found the prior notification fixes, especially #2959 (`🤖 fix: suppress notifications for auto-follow-up handoffs`), plus the earlier compaction-specific fixes #1892 and #2914. That work was not regressed: the current test suite still encoded its intended behavior, namely “compaction with continue message should fire only ONE notification (for continue response)”. The remaining gap is narrower: internal mid-stream compaction resumes are implementation-detail follow-ups, but if the workspace is backgrounded during the handoff, the activity generation update clears stale active-stream context before the final background completion is inferred.

## Implementation

- Mark only mid-stream auto-compaction follow-ups with a source marker (`dispatchOptions.source = "internal-resume"`). On-send compaction preserves user-authored prompts, including prompts whose text is exactly `Continue`.
- Keep notification suppression policy localized to `StreamingMessageAggregator` and response-completion metadata.
- Let the aggregator preserve only explicitly suppressed completion metadata across a background streaming-generation handoff, while still dropping stale queued follow-up metadata for normal background turns.
- Avoid adding notification-specific fields to backend stream events, workspace activity snapshots, persisted message metadata, or extension metadata.

## Validation

- `bun test src/browser/utils/messages/StreamingMessageAggregator.test.ts src/browser/stores/WorkspaceStore.test.ts`
- `make typecheck`
- `make static-check`

## Risks

Low-to-medium. This touches the response completion notification policy and background activity handoff behavior, but the new suppression path is scoped to an explicit internal-resume source marker. User-authored compaction follow-ups remain notifying and are covered by tests.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `high` • Cost: `n/a`_

### DIFF
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -2720,6 +2720,122 @@ describe("WorkspaceStore", () => {
       });
     });
 
+    it("preserves internal resume metadata across background handoffs", async () => {
+      const activeWorkspaceId = "active-workspace-internal-resume-background";
+      const backgroundWorkspaceId = "background-workspace-internal-resume-background";
+      const initialRecency = new Date("2024-01-06T00:00:00.000Z").getTime();
+
+      const backgroundStreamingSnapshot: WorkspaceActivitySnapshot = {
+        recency: initialRecency,
+        streaming: true,
+        streamingGeneration: 1,
+        lastModel: "claude-sonnet-4",
+        lastThinkingLevel: null,
+      };
+
+      let releaseBackgroundCompletion!: () => void;
+      const backgroundCompletionReady = new Promise<void>((resolve) => {
+        releaseBackgroundCompletion = resolve;
+      });
+
+      mockActivityList.mockResolvedValue({
+        [backgroundWorkspaceId]: backgroundStreamingSnapshot,
+      });
+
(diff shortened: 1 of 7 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the auto-retry resume gap by preserving `suppressResponseNotification` in the same-session retry resume state and passing it through `resumeStream()` retries.

ammar-agent:
@codex review

Addressed the source-specificity issue by marking only mid-stream auto-compaction follow-ups with `dispatchOptions.suppressResponseNotification`; user-authored on-send compaction prompts, including exact `Continue`, are covered by a non-suppression regression test.

ammar-agent:
@codex review

Refactored per feedback: notification-specific suppression is no longer threaded through backend stream/activity schemas or persisted metadata. The backend only marks the compaction follow-up source as `internal-resume`; `StreamingMessageAggregator` owns the response-completion notification policy and background handoff preservation.

PR #3263, opened 2026-05-10
🤖 fix: stabilize chat input auto-resize height

ammar-agent:
Summary
- Fixes the workspace chat input collapsing to a tiny height while editing large drafts by making the auto-resize hook verify the DOM inline height before it skips a resize write.

Background
- The recent typing-performance optimization cached the last applied textarea height. On shrink-capable updates, the hook can temporarily set the textarea height to `auto`; if the measured height matches the cache, the old guard skipped restoring the pixel height. Large inputs then fell back to the CSS minimum until another resize path repaired them.

Implementation
- Keep the no-op fast path for normal insertions, but also compare the actual inline height against the desired pixel height before skipping the write.
- Added regression coverage for capped large-text deletion and callers clearing the inline height outside the hook.

Validation
- `bun test src/browser/hooks/useAutoResizeTextarea.test.tsx`
- `make static-check`

Risks
- Low. The change is scoped to the textarea resize hook and preserves the existing optimization when the DOM already has the expected inline height.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `1787684{MUX_COSTS_USD:-n/a}`_

### DIFF
--- a/src/browser/hooks/useAutoResizeTextarea.test.tsx
+++ b/src/browser/hooks/useAutoResizeTextarea.test.tsx
@@ -8,6 +8,7 @@ function createFakeTextarea(initialScrollHeight: number): {
   ref: RefObject<HTMLTextAreaElement>;
   assignments: string[];
   setScrollHeight: (value: number) => void;
+  setInlineHeight: (value: string) => void;
 } {
   let height = "";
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3265, opened 2026-05-10
🤖 fix: pre-fill sub-project on Ctrl+N from sub-project workspace

ammar-agent:
## Summary

Fix Ctrl/Cmd+N (new chat) so it pre-fills the focused workspace's sub-project instead of falling back to the parent project.

## Background

When a workspace lives in a sub-project, pressing Ctrl+N would open the new-chat screen with the **parent** project selected in the picker rather than the sub-project. Users then had to manually re-pick the sub-project every time, even though clicking the sub-project section's `+` button already does the right thing.

Sub-project workspaces share the parent's worktree, so `WorkspaceSelection.projectPath` is always the parent path. The discriminator (`subProjectPath`) lives on the workspace metadata, but the global Ctrl+N handler never looked it up — it just called `handleAddWorkspace(selectedWorkspace.projectPath)`, persisting the draft with `subProjectPath: null`. From there the existing `pendingSubProjectPath` → `creationSubProjectPath` pipeline in `ChatInput` had nothing to work with and defaulted to the parent.

## Implementation

In the Ctrl+N branch of the keydown handler in `ProjectSidebar.tsx`, look up the focused workspace's metadata via `workspaceStore.getWorkspaceMetadata(...)` and forward its `subProjectPath` to `handleAddWorkspace`. This mirrors what the sub-project section header's `+` button already does (`handleAddWorkspace(projectPath, section.id)`) and rides on the existing draft → creation-screen pre-fill logic.

## Risks

Very low. One-line change scoped to a single keybind branch; metadata lookup falls back to `undefined` for non-sub-project workspaces, preserving prior behavior. No schema, IPC, or persisted-state changes.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `high` • Cost: `$1.84`_

### DIFF
--- a/src/browser/components/ProjectSidebar/ProjectSidebar.tsx
+++ b/src/browser/components/ProjectSidebar/ProjectSidebar.tsx
@@ -56,6 +56,7 @@ import {
   getTierKey,
   getSectionExpandedKey,
   getSectionTierKey,
+  resolveEffectiveSectionId,
   type AgentRowRenderMeta,
 } from "@/browser/utils/ui/workspaceFiltering";
 import { Tooltip, TooltipTrigger, TooltipContent } from "../Tooltip/Tooltip";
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the P2 staleness concern: switched the Ctrl+N handler to read `subProjectPath` from `sortedWorkspacesByProject` (the live React prop the sidebar already renders from) instead of `WorkspaceStore.getWorkspaceMetadata` (whose backing Map is insert-only via `syncWorkspaces` and can serve a stale value after a section drop). This mirrors the existing pattern in `SidebarTitleEditKeybinds` (`EDIT_WORKSPACE_TITLE`, lines 561–567). Please take another look.

ammar-agent:
@codex review

Addressed the deleted-sub-project edge case: before forwarding `meta.subProjectPath`, the handler now validates it against the current `getSubProjectsForParent(parentPath, userProjects)` set, matching how `partitionWorkspacesBySection` decides what to render. If the section has been removed from the project config, the draft falls back to the parent (matching the sidebar's visible "unsectioned" placement) instead of forwarding a stale path that `workspace.create` would reject.

ammar-agent:
@codex review

Addressed the inherited-section case: extracted the renderer's parent-walk into an exported `resolveEffectiveSectionId(workspace, byId, sectionIds)` in `workspaceFiltering.ts`, reused by `partitionWorkspacesBySection` and by the Ctrl+N keybind handler. The keybind now resolves the same effective section the sidebar renders — including for sub-agent rows that inherit `subProjectPath` via `parentWorkspaceId`, and unsectioned fallback when the section has been deleted from the project config.

ammar-agent:
@codex review

Rebased onto latest `origin/main` and re-resolved the ProjectSidebar conflict, keeping the shared `resolveEffectiveSectionId` path so Ctrl+N matches the visible sub-project section. Local `make static-check` and `bun test src/browser/utils/ui/workspaceFiltering.test.ts` both pass after the rebase.

PR #3267, opened 2026-05-11
🤖 fix: branch SSH workspaces from upstream source

ethanndickson:
## Summary
Fresh SSH/Coder workspaces now treat the selected Source Branch as the upstream remote-tracking branch when it is available, instead of letting the synced local snapshot in `refs/mux-bundle/*` override it.

## Issue
A newly-created Coder workspace on an SSH runtime could immediately show as hundreds of commits behind `origin/main` even when the UI's Source Branch was set to `main`. In the reported case, the workspace was created on `pog2.coder` with Source Branch `main`, and the UI showed it as 236 commits behind `origin/main` right after creation. That is surprising because selecting `main` should mean the fresh workspace starts from upstream `origin/main`, so the initial behind count should be zero.

## Why it happened
The SSH worktree refactor split the remote VM's shared bare repo into two different kinds of refs: `refs/mux-bundle/main` is the user's local laptop `main` uploaded as a transport snapshot, while `refs/remotes/origin/main` is the actual upstream branch fetched on the SSH host. Fresh workspace creation still treated the uploaded local snapshot as eligible to define the new workspace base. Specifically, it only chose `origin/<source>` if the local snapshot ref could fast-forward to upstream; if local `main` was stale, ahead, diverged, or the origin fetch path failed, mux fell back to `refs/mux-bundle/main`. A background fetch could then update `origin/main` to the real upstream tip, making the newly-created workspace appear immediately behind even though the user explicitly selected Source Branch `main`.

## How this fixes it
Fresh SSH checkout now uses a remote-first source selection policy. It fetches the selected source branch into an explicit remote-tracking ref, verifies that `refs/remotes/origin/<source>` exists, and bases the new worktree on `origin/<source>` whenever that upstream source is available. `refs/mux-bundle/*` remains as a transport/local-snapshot cache and is used only as an explicit fallback when the upstream source branch cannot be fetched. The existing fast-forward safety check is preserved for existing workspace re-init paths, where protecting local workspace state still matters.

## Implementation
The fresh SSH checkout path now fetches with an explicit refspec like `+refs/heads/main:refs/remotes/origin/main`, resolves the fresh workspace base with `origin/<source>` first, and logs when it has to fall back to the local snapshot. Tests cover the upstream-first selection, explicit fallback behavior, and the new fetch refspec.

## Risks
This changes fresh SSH/Coder workspace initialization semantics when a local branch is ahead of or diverged from upstream. That is intentional for the Source Branch selector, but users relying on local `main` drift as the implicit fresh-workspace base will now need to create from that local branch explicitly or rely on the logged fallback when upstream is unavailable.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$13.34`_

### DIFF
--- a/src/node/runtime/SSHRuntime.syncContract.test.ts
+++ b/src/node/runtime/SSHRuntime.syncContract.test.ts
@@ -133,6 +133,24 @@ interface SnapshotPrivateApi {
   resolveLocalSyncRefManifest(projectPath: string): Promise<string | null>;
 }
 
+interface FreshWorkspaceSourcePrivateApi {
+  resolveFreshWorkspaceSourceBase(
+    baseRepoPathArg: string,
+    trunkBranch: string,
+    fetchedOrigin: boolean,
+    fallbackRef: string | null,
+    initLogger: InitLogger,
+    abortSignal?: AbortSignal
+  ): Promise<string>;
+  fetchOriginTrunk(
+    workspacePath: string,
+    trunkBranch: string,
+    initLogger: InitLogger,
+    abortSignal?: AbortSignal,
+    nhp?: string
+  ): Promise<boolean>;
+}
+
 function createLayout(): RemoteProjectLayout {
   return {
     projectId: "project-id",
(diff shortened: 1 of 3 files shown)


ethanndickson:
@codex review

PR #3268, opened 2026-05-12
🤖 feat: add bash monitor wake events

ammar-agent:
## Summary

Adds Claude-Code-style wake-on-match monitoring to background `bash` commands. A background bash can now include a `monitor` block with a regex filter; matching output lines are coalesced and delivered to the agent as synthetic queued wake messages while the existing `task_await` harvest path continues to work.

## Background

Long-running dev servers, watch tests, and log tails previously required polling with `task_await` to notice important output. This change lets the process itself wake the agent when relevant output appears, without introducing a new tool family or task-ID namespace.

## Implementation

- Extends the `bash` tool schema/result with an optional monitor block for background commands.
- Adds a write-time monitor tail loop in `BackgroundProcessManager` with independent cursors, cooldown coalescing, match caps, dropped-line accounting, and cleanup on process termination.
- Emits renderer-only `monitor-match` events while waking the model through `AgentSession.queueMessage(..., { synthetic: true, agentInitiated: true })`.
- Surfaces monitor counters/last matches in background process API/store/UI plumbing.
- Updates generated tool docs and built-in skill content.

## Validation

- `bun test src/node/services/backgroundProcessManager.test.ts src/node/services/tools/bash.test.ts`
- `make static-check`

## Risks

This touches background process lifecycle and queued-message wake plumbing. The implementation keeps monitors optional, preserves existing `bash:<id>` task behavior, and stops only the monitor—not the process—when `max_events` is reached.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `2255777{MUX_COSTS_USD:-unknown}`_


ammar-agent:
@codex review

ammar-agent:
@codex review
Please take another look.

ammar-agent:
@codex review
Please take another look after the CI fix.

ammar-agent:
@codex review
Please take another look after the monitor wake send-options fix.

ammar-agent:
@codex review
Please take another look after bounding incomplete monitor buffers.

ammar-agent:
@codex review
Please take another look after coalescing queued monitor wakes.

ammar-agent:
@codex review
Please take another look at the monitor UI polish (background-row "quiet" state + inline rendering of synthetic monitor wake messages).

ammar-agent:
@codex review
Please take another look after splitting batched monitor wakes into separate cards.

ammar-agent:
@codex review
Please take another look after switching to an extractor that handles monitor blocks mixed with user text.

ammar-agent:
@codex review
Please take another look after stripping monitor XML from the edit path.

ammar-agent:
@codex review
Please take another look after preserving queued user options on wakes and humanizing the Copy text on pure monitor wake messages.

ammar-agent:
@codex review
Please take another look after fixing the queue dispatch-mode collapse on synthetic appends.

ammar-agent:
@codex review
Please take another look after narrowing dispatch-mode preservation to the options-omitted append case.

ammar-agent:
@codex review
Please take another look after backing off the monitor poll interval for remote runtimes.

ammar-agent:
@codex review
Please take another look after switching to a LocalRuntime/WorktreeRuntime whitelist for the fast monitor poll path.

ammar-agent:
@codex review
Please take another look after gating monitor extraction on a persisted containsMonitorEvents metadata flag.

ammar-agent:
@codex review
Please take another look after declaring containsMonitorEvents in the MuxMessage ORPC schema.

ammar-agent:
@codex review

Followup: the queued-message banner (and the composer on interrupt/edit) no longer surface raw `<monitor-event …>` XML.

- Added a `containsMonitorEvents` flag to the `queued-message-changed` chat event and `QueuedMessage` type, driven by a new `MessageQueue.containsMonitorEvents()` getter.
- `QueuedMessage.tsx` parses wake blocks via `extractMonitorWakeEvents` when flagged and renders `MonitorWakeMessage` cards in place of the XML. Pure-monitor queues hide Edit (no user text to edit); mixed queues keep Edit and open the composer with monitor blocks stripped.
- `normalizeQueuedMessage` (Edit path) and `AgentSession.restoreQueueToInput` (interrupt path) both strip monitor XML via a new shared `stripMonitorWakeXml` helper in `@/common/utils/monitorWake`. When interrupt yields nothing user-authored, the restore-to-input emit is skipped entirely so the composer stays clean.
- Defensive: parsing is gated on the backend flag, so user-pasted similar-looking XML is still rendered/edited verbatim.

New behavioral tests cover the strip helper, `normalizeQueuedMessage` gating, and three banner cases (pure monitor, mixed user-text + monitor, unflagged passthrough).

ammar-agent:
@codex review

Addressed the P2 about queue-level vs block-level scoping: backend wakes now carry a `source="mux"` sentinel attribute, and both `extractMonitorWakeEvents`/`stripMonitorWakeXml` only match sentinel-bearing blocks. The `containsMonitorEvents` flag still gates whether parsing happens at all, but inside that gate user-pasted `<monitor-event …>` XML survives verbatim through banner render, Edit restore, and interrupt restore-to-input — even when mixed with a real wake in the same queued message. Added regression tests in both `MonitorWakeMessage.test.ts` and `monitorWake.test.ts` for the mixed case.

ammar-agent:
@codex review

Addressed: added `isPureMonitorWakeQueue` and gated both `restoreQueuedDraft` and `handleEditLastUserMessage` (the editLast keyboard shortcut path) on it. When the queue is *only* a backend wake — synthetic `<monitor-event source="mux" …>` with no surviving user text, fileParts, or reviews — we bail without clearing the queue so the pending wake reaches the agent on the next dispatch boundary; mixed user-text + wake queues still take the edit branch and strip XML from the editor draft. Tests in `chatEditing.test.ts` cover all four cases.

ammar-agent:
@codex review

Addressed both P2s:

- `AgentSession.restoreQueueToInput` now bails (no `clearQueue`, no `restore-to-input` emit) when the queue contains only a backend monitor wake — the wake stays scheduled and reaches the agent on the next dispatch boundary. Mixed queues continue to clear + restore the surviving user text/attachments/reviews.
- `QueuedMessage` banner Edit visibility now uses the shared `isPureMonitorWakeQueue` helper, matching the editLast keyboard shortcut: Edit stays visible for image-only/review-only queues that received a wake, hidden only when nothing user-authored survives.

Added two behavioral tests: `disposeRace` covers the agentSession bail + delayed dispatch, `queuedMessageBanner` covers the mixed wake+attachment Edit case.

PR #3277, opened 2026-05-12
🤖 fix: show finalized completed goal accounting

ThomasK33:
## Summary

Fix the completed goal tool call card so it displays finalized goal accounting from the live workspace goal snapshot when it matches the completed goal. This makes the transcript card agree with the Goal sidebar after the completing stream's cost and turn accounting lands.

## Background

`complete_goal` returns before final stream accounting is recorded. The completed tool card rendered that stale result snapshot, so it could show `$0.00 / budget` and `0 turns` while the sidebar correctly showed the finalized cost and turn count.

## Implementation

- Added a small display-goal merge helper for `CompleteGoalToolCall`.
- When the live sidebar goal has the same `goalId` as the tool result, the card uses live `costCents`, `turnsUsed`, limits, status, objective, and start time.
- If the live goal is absent or belongs to a different goal, the card falls back to the immutable tool result so historical transcript rendering remains safe.
- Added regression tests for same-goal live accounting and different-goal fallback behavior.

## Validation

- `bun test src/browser/features/Tools/CompleteGoalToolCall.test.ts`
- `make typecheck`
- `make fmt-check`
- `make lint`
- `make static-check`
- Installed `agent-browser` and dogfooded the Storybook goal tool card:
  - captured collapsed and expanded screenshots
  - recorded `/tmp/mux-goal-dogfood/complete-goal-ui.webm`
  - verified expanded text includes `COST` and `$0.47 / $1.00`

## Risks

Low. The live snapshot is only trusted when `goalId` matches the completed tool result; otherwise existing result-based rendering is preserved.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `high` • Cost: `$19.50`_

### DIFF
--- a/src/browser/features/Tools/CompleteGoalToolCall.test.ts
+++ b/src/browser/features/Tools/CompleteGoalToolCall.test.ts
@@ -1,0 +1,85 @@
+import { describe, expect, test } from "bun:test";
+import type { GoalRecordV1, GoalSnapshot } from "@/common/types/goal";
+import { getCompleteGoalDisplayGoal } from "./CompleteGoalToolCall";
+
+const goalId = "00000000-0000-4000-8000-000000000001";
+
+function resultGoal(overrides: Partial<GoalRecordV1> = {}): GoalRecordV1 {
+  return {
+    version: 1,
+    goalId,
+    objective: "Ship goal accounting UI",
+    status: "complete",
+    budgetCents: 100_000,
+    turnCap: 3,
+    costCents: 0,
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed the stale-regression case by retaining the last same-goal live snapshot while the transcript card is mounted, so clearing/replacing the current goal no longer falls back to the pre-accounting tool result.

PR #3279, opened 2026-05-13
🤖 fix: gate imagegen skill behind experiment

ThomasK33:
Summary

Gate the built-in imagegen skill behind the Image Generation Tool experiment so agents only see or load it when the corresponding image_generate tool is available.

Background

The imagegen skill instructs agents to call image_generate, which is confusing when the experiment is off and the tool is not registered for the turn.

Implementation

Filtered the built-in imagegen skill from stream tool descriptions and agent skill discovery when image generation is unavailable. Direct built-in imagegen reads through skill APIs/tools are blocked, and persisted inline skill refs no longer materialize built-in imagegen snapshots after the experiment is disabled; user-provided project/global skills are left untouched.

Validation

- `make static-check`
- `bun test src/node/services/agentSession.agentSkillSnapshot.test.ts`
- `bun test src/node/services/streamContextBuilder.test.ts`
- `bun test src/node/services/tools/agent_skill_read.test.ts src/node/services/tools/agent_skill_read_file.test.ts`
- `bun run typecheck`
- `make fmt-check`
- `make lint`

Risks

Low risk: the gate only filters the built-in imagegen skill and preserves project/global skill shadowing behavior.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `high` • Cost: `$18.28`_

### DIFF
--- a/src/node/orpc/router.ts
+++ b/src/node/orpc/router.ts
@@ -63,7 +63,11 @@ import {
 import {
   discoverAgentSkills,
   discoverAgentSkillsDiagnostics,
+  filterUnavailableImagegenSkills,
+  IMAGEGEN_SKILL_DISABLED_MESSAGE,
+  isBuiltInImagegenSkillPackage,
   readAgentSkill,
+  type ResolvedAgentSkill,
 } from "@/node/services/agentSkills/agentSkillsService";
 import {
   discoverAgentDefinitions,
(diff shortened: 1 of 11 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Please take another look.

PR #3292, opened 2026-05-15
🤖 fix: parse mermaid SVG as HTML so wrapped foreignObject labels render

ammar-agent:
## Summary

Every Mermaid diagram whose node labels wrapped (`label<br/>second line`) was failing in the renderer with `Mermaid Error: Mermaid returned invalid SVG output`. The fix switches our SVG sanitizer from a strict XML parser to an HTML parser, matching the sink we actually serialize into (`innerHTML`). Adds direct regression tests against `sanitizeMermaidSvg` so this can't silently regress.

## Background

Reproduced with a real chart the user's agent produced: Mermaid 11.12.2 successfully rendered a ~76KB SVG, but our `sanitizeMermaidSvg` returned `null`, so `<Mermaid>` threw "Mermaid returned invalid SVG output".

Root cause: Mermaid emits HTML labels inside `<foreignObject>` whenever `wrap`/`markdownAutoWrap` is enabled (we set both globally). That HTML legitimately contains void elements with no self-closing slash:

```html
<foreignObject><div xmlns="http://www.w3.org/1999/xhtml"><p>line1<br>line2</p></div></foreignObject>
```

We were parsing the full document with `DOMParser.parseFromString(svg, "image/svg+xml")` — strict XML. Strict XML rejects bare `<br>`, raises `parsererror`, and we returned `null`. **Any diagram with wrapped labels hit this**, which explained why Mermaid "hadn't worked in a while".

## Implementation

- Parse as `text/html` and locate the `<svg>` via `querySelector("svg")`. The HTML parser handles SVG + embedded HTML the same way the browser does when we later inject sanitized markup via `innerHTML`, so parse model now matches sink model.
- Scoped the attribute walk to descendants of the SVG root (not the whole HTML doc the parser materializes around it).
- Sanitization guarantees unchanged: still strips `script/iframe/object/embed`, `on*` handlers, and URL-attribute schemes (`javascript:`, `vbscript:`, `data:text/html`). The `canonicalizeUrlForSchemeCheck` decoding pass is untouched.

## Validation

- Reproduced the user's failing chart end-to-end: ran it through Mermaid 11.12.2 in real Chromium (Playwright), then fed the resulting 76KB SVG through the patched sanitizer — now parses cleanly. Before the fix it returned `null` with `parsererror: Opening and ending tag mismatch: br line 1 and p`.
- Added 4 unit tests directly on `sanitizeMermaidSvg` (the boundary that actually broke), in addition to the existing component-level tests:
  - accepts SVG whose `<foreignObject>` HTML labels use bare `<br>` (the exact regression)
  - still strips `<script>` nested inside `<foreignObject>`
  - still strips `on*` handlers and `javascript:` hrefs
  - rejects input that contains no `<svg>` root
- `bun test src/browser/features/Messages/Mermaid.test.tsx` → 8/8 pass
- `make typecheck`, `make lint`, `make fmt-check` all pass.

## Risks

Low. The sanitizer's allow/deny lists are unchanged; only the parsing strategy moved from strict XML to HTML. The HTML parser is also what we ultimately rely on when the sanitized output is injected via `innerHTML`, so the new parse semantics are strictly closer to what the browser does at the actual render sink. Tests cover the security-relevant cases (script removal, `on*` removal, `javascript:` scheme removal).

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `high` • Cost: `$1.62`_

### DIFF
--- a/src/browser/features/Messages/Mermaid.test.tsx
+++ b/src/browser/features/Messages/Mermaid.test.tsx
@@ -2,7 +2,7 @@ import { cleanup, render, waitFor } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 import { GlobalWindow } from "happy-dom";
 import { StreamingContext } from "./StreamingContext";
-import { Mermaid } from "./Mermaid";
+import { Mermaid, sanitizeMermaidSvg } from "./Mermaid";
 
 const DEFAULT_SVG =
   '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10"><rect width="10" height="10" /></svg>';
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed in 3d3ed74b5: parse strictly as `image/svg+xml` first and only fall back to the `text/html` parser when the input contains `<foreignObject>` (the known case where Mermaid's wrapped-label HTML breaks XML parsing). Other malformed inputs — including your `<svg><g></svg>` example — still return `null`. Added explicit regression tests for those rejection cases.

ammar-agent:
@codex review

Addressed in a7f01aa4b. The fallback now requires structural proof that the only XML issue lives inside `<foreignObject>`:

1. Strict `image/svg+xml` parse on the original; if it parses, done.
2. Otherwise strip each `<foreignObject>...</foreignObject>` subtree to a self-closed placeholder and re-parse strictly.
3. Only when the stripped form is well-formed XML (with `<svg>` as the documentElement) do we relax to the HTML parser to materialize the original tree.

Your example `<svg><foreignObject>…</foreignObject><g></svg>` now correctly fails closed: stripping leaves `<svg><foreignObject/><g></svg>`, which still has an unbalanced `<g>` and re-raises `parsererror`. Added explicit regression tests for both that case and the positive complement (valid wrapped-label SVG continues to pass).

PR #3293, opened 2026-05-16
🤖 feat: add /btw side-question command

ammar-agent:
## Summary

Adds the `/btw` slash command — a forked, single-turn, read-only side question over the current conversation. Tools are denied on both the prompt and schema sides; the question and answer are persisted to chat history with distinct metadata and stream through the existing `TypewriterMarkdown` path so the side branch reads as one offset Q/A cycle in the transcript.

`/btw` is modeled on Claude Code 2.1.73's command of the same name: a quick "by the way, what was that file you mentioned?" question that runs against the current transcript without invoking tools and without re-anchoring the main agent loop.

## Implementation

**Backend (`src/node/services/sideQuestionService.ts`)**

- Builds a side-question transcript from history (latest context boundary forward), merges any in-flight live-stream snapshot, filters prior `/btw` rows so the model never grounds on its own earlier answers, and feeds it to `streamText` with **no `tools` argument**.
- Persists a placeholder assistant row first (so `historyService` allocates a real `historySequence`), emits it as a `message` chat event with `side-question-answer` muxMetadata (including `questionMessageId`), then drives `stream-start`/`stream-delta`/`stream-end` through `session.emitChatEvent`. The frontend aggregator handles these identically to a regular agent stream — `TypewriterMarkdown`, smooth-text animation, and replay all just work.
- Tries the live-parent-stream model, then the configured chat/agent model, then `NAME_GEN_PREFERRED_MODELS`. `maxAttempts` reaches at least the first preferred fallback so three stale configured candidates can't disable the feature.
- `aiService` filters side-question rows out of provider context via `filterSideQuestionMessages`, so `/btw` never pollutes the main agent's prompt.

**Frontend rendering: inline side branch, not below**

- The `/btw` user row stamps `interruptedMessageId` + `interruptedTextLength` + `interruptedPartIndex` against whatever main-agent stream was live when it fired. `StreamingMessageAggregator.getDisplayedMessages()` pre-walks history to build an `interruptionsByInterruptedId` map and, for any interrupted main message, calls `buildInterruptedMessageDisplay()` which splits the main reply at the captured offset(s) and inserts the Q/A pair between segments. The last segment keeps the original message id so the streaming-indicator anchor and Start-Here/Fork/etc. action handlers still target the right row.
- `splitMessagePartsAtTextLengths` honors `partIndex` so non-text parts (reasoning, tool calls) that were already visible at the same text offset stay on the pre-aside side after reload — text length alone misordered them.
- Side-question user rows render with a "Side question" header and a continuous left accent stripe; side-answer rows share the stripe but skip the redundant label. Side answers expose only Copy in their action set — Start Here / Fork / Share / Show Text don't make sense for an off-main-thread message.
- `findActiveSideQuestionScrollHoldTarget` keeps the side branch in view while its answer streams and the main reply continues underneath; user wheel/touch/key/mouse and Jump-to-bottom clear the hold.

**Frontend lifecycle: `/btw` doesn't hijack main state**

- `aggregator.hasInterruptibleActiveStream()`, `getActiveStreamMessageId()`, `getActiveStreamTimingStats()`, `getCurrentModel()`, and `getCurrentThinkingLevel()` skip side-answer streams via a `getActiveMainStreamEntry()` helper — so interrupt UI, live stats, the workspace model display, and `setInterrupting()` all target the main stream when both are active.
- `handleMessage`/`handleStreamStart`/`handleStreamEnd` detect side-answer events and skip the bits that would clobber the main turn: don't clear pending main startup state, don't clear `hasQueuedFollowUp` or `backgroundHandoffCompletion`, don't fire `onResponseComplete`, and don't replace main `lastCompletedStreamStats`. Main completion is computed as final via `getActiveMainStreamEntry() === undefined`, so the main reply can finalize while a side answer is still streaming.
- `WorkspaceStore.handleChatMessage` uses `getBufferedActiveStreamStart(... isSideQuestionAnswerMessageId)` so side-only replay never looks like an active main stream, and the `stream-end` side-effect path skips `collapsePinnedTodoOnStreamStop` for side answers (Codex caught the pinned-TODO collapse regression).

**Slash command (`/btw <question>`)**

- `processSlashCommand` fire-and-forgets the `workspace.sideQuestion` RPC, clears input + attachments + attached reviews immediately. On RPC failure it restores the raw command text **only** if the command is still current (`asyncCommandToken`/`isAsyncCommandCurrent`) and the input hasn't been replaced by a newer draft.
- `placeholderTips.ts` gains one entry — `Try /btw <question> to ask a side question without nudging the agent` — surfacing the new command through main's existing wall-clock tip carousel. (The carousel itself already lives on main; this PR just adds a tip.)

## Validation

- `make static-check` ✅
- Targeted tests: `aggregator-side`, `chat-commands-side`, `scroll-hold`, `workspace-store-side` — all green.
- Notable coverage: `StreamingMessageAggregator.sideQuestion.test.ts` (split ordering with same-offset non-text parts, pairing by `questionMessageId`, main-stream-aware active stream selection, main `isFinal` while side answer streams), `WorkspaceStore.test.ts` (replayed side streams don't set interrupt state, pinned TODO panel doesn't collapse on side-answer end), `sideQuestionScrollHold.test.ts` (standalone /btw with only its answer below stays bottom-locked, hold realigns while the answer grows), `chatCommands.test.ts` (RPC failure restores `/btw …` but never over a newer draft, stale failures suppressed), `sideQuestionService.test.ts` (no tools passed to `streamText`, prior side questions filtered, fallback model selection after stale candidates).

## Risks

`/btw` deliberately bypasses the main `streamMessage` pipeline — it emits chat events through `session.emitChatEvent` directly so the workspace's main-stream lifecycle, idle-compaction recency, and post-compaction attachments all skip it. That's the intentional product behavior, but it's worth a careful eye on any `StreamingMessageAggregator` interaction that assumes one active stream per workspace.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `high` • Cost: `$291.60`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -85,6 +85,11 @@ import {
   normalizeQueuedMessage,
   type EditingMessageState,
 } from "@/browser/utils/chatEditing";
+import {
+  findActiveSideQuestionScrollHoldTarget,
+  findSideQuestionScrollHoldTarget,
+  type SideQuestionScrollHoldState,
+} from "./sideQuestionScrollHold";
 import { recordSyntheticReactRenderSample } from "@/browser/utils/perf/reactProfileCollector";
 
 // Perf e2e runs load the production bundle where React's onRender profiler callbacks may not
@@ -154,6 +159,15 @@ interface ChatPaneProps {
 type ReviewsState = ReturnType<typeof useReviews>;
 
 const AUTO_SCROLL_TRANSCRIPT_STYLE = { overflowAnchor: "none" } as const;
+
+function findTranscriptMessageElement(
+  scrollContainer: HTMLElement,
+  historyId: string
+): HTMLElement | undefined {
+  return Array.from(scrollContainer.querySelectorAll<HTMLElement>("[data-message-id]")).find(
+    (element) => element.getAttribute("data-message-id") === historyId
+  );
+}
 
 export const ChatPane: React.FC<ChatPaneProps> = (props) => {
   const {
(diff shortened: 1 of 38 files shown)


ammar-agent:
@codex review

Substantial follow-up: the /btw UX moved from "inline pair with dismiss" to a single visually-distinct Q/A block, and the main chat now suspends while a side question streams so concurrent main-agent tokens can't shove the side branch around (root cause of the prior layout flashes). The tip carousel is also rebucketed onto a wall-clock 20-min rotation so switching workspaces inside the same bucket no longer reshuffles the visible tip. PR body refreshed accordingly.

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the latest P2 findings:
- `/btw` model candidates now prefer the live parent stream model before persisted settings.
- side-answer history update failures still emit the live `stream-end`, then return an error so the caller can report the persistence failure.

Validation:
- `bun test src/node/services/sideQuestionService.test.ts`
- `bun test src/node/services/workspaceService.test.ts --test-name-pattern "getSideQuestionModelCandidates"`
- `bun test src/node/services/aiService.test.ts src/browser/utils/messages/StreamingMessageAggregator.sideQuestion.test.ts src/browser/stores/WorkspaceStore.test.ts src/browser/components/ChatPane/sideQuestionScrollHold.test.ts`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Addressed the latest P2: side-question answer streams still render through `StreamingMessageAggregator`, but they no longer count as interruptible/busy streams for `WorkspaceStore` or sidebar working-state derivation.

Validation:
- `bun test src/browser/stores/WorkspaceStore.test.ts --test-name-pattern "/btw side-question tail flow"`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Addressed the buffered replay issue: side-question stream starts are now ignored (not reset) in the pending active-stream fallback, so an in-flight main stream remains interruptible/working during hydration even if `/btw` starts afterward.

Validation:
- `bun test src/browser/stores/WorkspaceStore.test.ts --test-name-pattern "/btw side-question tail flow"`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Addressed the context reset issue: `/btw` transcript construction now filters prior side-question rows and then applies the same latest context-boundary slicing used by normal provider requests, so soft-cleared content is not sent to the side-question model.

Validation:
- `bun test src/node/services/sideQuestionService.test.ts`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Addressed the transcript cap issue: `/btw` now reads the active history window, filters prior side-question rows and context-boundary slices, then applies the 200-message cap to the remaining main-context rows so previous `/btw` exchanges cannot crowd out recent main chat.

Validation:
- `bun test src/node/services/sideQuestionService.test.ts`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Addressed the live stream context issue: `/btw` now passes the synchronous `AIService.getStreamInfo()` snapshot into transcript construction and upserts those live main-agent parts over any throttled partial, so the side-question prompt includes text already visible in the UI.

Validation:
- `bun test src/node/services/sideQuestionService.test.ts`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Addressed the live stream snapshot race: `/btw` now shallow-clones the live main-agent parts synchronously with `getStreamInfo()` before any awaited history I/O, and uses that frozen snapshot for both interruption metadata and transcript construction.

Validation:
- `bun test src/node/services/sideQuestionService.test.ts`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Addressed the pre-candidate race: `WorkspaceService.askSideQuestion()` now snapshots and clones the live stream synchronously before awaiting model candidates, passes that snapshot into `sideQuestionService`, and reuses its model for candidate ordering.

Validation:
- `bun test src/node/services/sideQuestionService.test.ts`
- `bun test src/node/services/workspaceService.test.ts --test-name-pattern "getSideQuestionModelCandidates"`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Addressed the live partial duplication issue: when a live main stream snapshot is merged into the `/btw` transcript, older partial/history copies with the same id/historySequence are removed so the model sees one current version of the interrupted assistant turn.

Validation:
- `bun test src/node/services/sideQuestionService.test.ts`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Addressed the latest findings and the reported bottom-sticking behavior:
- `/btw` side branches now split into the interrupted main stream as soon as the side answer exists, so they no longer stay pinned at the bottom for the stream duration.
- Side-question user rows no longer reset main-agent lifecycle/status.
- Fire-and-forget `/btw` failures are ignored when their originating command/workspace is stale.

Validation:
- `bun test src/browser/utils/messages/StreamingMessageAggregator.sideQuestion.test.ts src/browser/stores/WorkspaceStore.test.ts --test-name-pattern "/btw side-question|lifecycle|interrupt state"`
- `bun test src/browser/utils/chatCommands.test.ts --test-name-pattern "side-question"`
- `make typecheck`
- `make static-check`

ammar-agent:
@codex review

Updated /btw scroll handling after the streaming-order fix: side questions now anchor at the interruption point immediately and release bottom-lock once main output appears below the side branch, so the user can read the aside before opting back into the live tail.

ammar-agent:
@codex review

Addressed the replay hydration side-answer marker issue as well: buffered stream-start detection now includes historical side-answer placeholders, so replayed /btw answer streams do not briefly look like interruptible main-agent streams.

ammar-agent:
@codex review

Addressed the concurrent side-question pairing issue: side-answer metadata now stores questionMessageId and the renderer pairs by that stable link, falling back to adjacency only for legacy rows.

ammar-agent:
@codex review

Addressed the current-model and /btw composer cleanup findings, plus added a TDD scroll-hold fix so the side branch remains anchored while its answer/main-tail content grows.

ammar-agent:
@codex review

Rebased onto latest main and force-pushed the side-question scroll/model/composer fixes after local validation passed.

ammar-agent:
@codex review

Fixed the post-rebase command gating conflict, re-ran local static-check, and force-pushed the corrected branch.

ammar-agent:
@codex review

Addressed the new fallback-model and side-question edit findings, with focused regression tests, then re-ran local targeted tests and `make static-check`.

ammar-agent:
@codex review

Preserved side-question answer metadata on stream-end replay, added coverage for the terminal metadata, and re-ran focused tests plus `make static-check`.

ammar-agent:
@codex review

Updated active-stream selection to ignore /btw side-answer streams for interrupt/live usage/stats callers, added regression coverage for side-answer-first ordering, and re-ran focused tests plus `make static-check`.

ammar-agent:
@codex review

Updated caught-up buffered stream classification to use the historical-message-aware side-answer detector, re-ran workspace /btw tests, focused side-question regressions, and `make static-check`.

ammar-agent:
@codex review

Rebased on latest main and force-pushed after `make static-check` passed locally.

ammar-agent:
@codex review

Addressed the side-answer stream-start pending-startup regression and the zero-offset non-text split regression with focused tests; `make static-check` passed locally.

ammar-agent:
@codex review

Addressed the side-answer terminal handling: main completions are final even while /btw streams overlap, and side-answer stream-end no longer clears pending main startup. Added focused regressions and `make static-check` passed locally.

ammar-agent:
@codex review

Rebased onto latest main and squashed the branch to a single commit. Local targeted tests and `make static-check` passed.

ammar-agent:
@codex review

Rebased on latest main, squashed the branch to one commit, and fixed the queued-follow-up side-answer stream-start regression. Local targeted tests and `make static-check` passed.

ammar-agent:
@codex review

Restored /btw command text on pre-accept RPC failures and removed the tautological placeholder copy test. Local targeted tests and `make static-check` passed; branch remains squashed to one commit.

ammar-agent:
@codex review

Guarded the /btw restore so it only repopulates the command while the composer is still empty, avoiding overwriting newer drafts; targeted tests and `make static-check` passed. Branch remains a single squashed commit.

ammar-agent:
@codex review

Fixed standalone /btw scroll holding so a side answer alone does not release bottom-lock; focused tests and `make static-check` passed. Branch remains rebased and squashed to one commit.

ammar-agent:
@codex review

Fixed the hydration-time scroll-hold initialization and captured part index metadata so pre-existing non-text parts stay before the /btw split. Focused tests plus `make static-check` passed; branch remains one squashed commit on latest main.

ammar-agent:
@codex review

Skipped main-stream pinned-todo collapse side effects for side-answer stream-end events and added a regression for overlapping /btw completion while main todos remain active. Targeted tests and `make static-check` passed; branch remains a single squashed commit on main.

ammar-agent:
@codex review

Invalidated pending fire-and-forget slash command tokens when a normal workspace send proceeds, so stale /btw failures cannot restore command text after a newer send. Targeted tests and `make static-check` passed; branch remains one squashed commit.

ammar-agent:
@codex review

ammar-agent:
@codex review

PR #3294, opened 2026-05-16
🤖 fix: ground sidebar status tense in tool lifecycle + streaming state

ammar-agent:
## Summary

Fix the workspace-status sidebar feature getting tense wrong (e.g. emitting "Deployed service" while the service is still being deployed). The small model wasn't actually wrong in isolation — the *input* we fed it had no signal for distinguishing in-progress from completed work, so any plausible past-tense caption looked correct. This PR shifts the fix from prompt fiddling to data quality: lifecycle markers on tool parts, an explicit marker on the in-flight partial assistant message, and the live streaming bit that `AgentStatusService` already reads for cadence plumbed into the prompt.

## Background

`AgentStatusService` builds a trailing transcript from chat history + the in-flight partial, hashes it for dedup, and asks a small model to caption "what the agent is doing right now." Until this change, every observable distinction between *in progress* and *completed* was being erased before the model ever saw the transcript:

- `summarizeToolPart` emitted `[tool bash]` whether the part was in `input-available` (call sent, no result yet) or `output-available` (returned). A deploy that just kicked off and one that just finished were byte-identical in the transcript.
- The in-flight partial assistant message rendered with the same `Assistant:` prefix as a finalized turn — there was no way for the model to tell that the last turn wasn't actually over.
- `dispatch()` already reads `snapshot.streaming` from `ExtensionMetadataService` (it uses it to pick the active vs idle cadence), but `runForWorkspace` dropped it on the floor and never forwarded it to the prompt.

With those three signals missing, the prompt's tense rule ("use past tense if the last assistant turn looks complete") collapses to "does the prose end with a period," which is just as true of "Deploying service…" as "Service deployed." Small models naturally default to the more grammatical past-tense caption.

## Implementation

Three composable, low-cost fixes — data-quality first, prompt last:

1. **Tool parts carry a lifecycle phase.** `summarizeToolPart` now emits `[tool bash running]` for `input-available`, `[tool bash done]` for `output-available` / `output-redacted`. One word per marker; no measurable token increase.
2. **Partial assistant messages are tagged.** `formatMessageForTranscript` accepts an `opts.partial` flag and renders the partial with `Assistant (in progress):` instead of plain `Assistant:`. `buildTrailingTranscript` is the only caller that sets `partial: true`, and only for the `readPartial(...)` entry.
3. **`streaming` is plumbed through dispatch → `runForWorkspace` → `generateWorkspaceStatus` → `buildWorkspaceStatusPrompt`.** The prompt now opens with one of two leading hints (`actively streaming` vs `finished streaming`) and warns explicitly in the latter branch that "finished streaming" ≠ "activity complete."

The tense rule itself is rewritten to require evidence: past tense only when every relevant tool call shows `[tool … done]` *and* the assistant has handed back control. Default is present-progressive when uncertain. A concrete counter-example ("if the transcript shows `[tool bash running]` for a deploy, write 'Deploying service', not 'Deployed service'") is included in the prompt so the small model has a worked example aimed at exactly the failure mode this PR fixes.

## Validation

- `make static-check` — passes.
- `bun test src/node/services/agentStatusService.test.ts src/node/services/workspaceStatusGenerator.test.ts` — passes.
- New behavioral tests, none of them tautological:
  - `transcript tags in-flight tool calls 'running' and completed ones 'done'` — asserts both phase suffixes appear from real `state` values, which is the marker the prompt now keys off.
  - `forwards the live streaming bit to the prompt builder` — asserts `runForWorkspace(..., streaming=true)` reaches the generator as `{ streaming: true }`, guarding the dispatch boundary.
  - `switches the liveness hint based on the streaming flag` — asserts the prompt's two branches diverge (and each branch *lacks* the other branch's hint) so a future refactor can't silently collapse them.
  - `documents tool-call lifecycle markers so the model can read them` — asserts the prompt teaches the marker semantics, since the markers are noise to the model without it.
- The existing partial-assistant test was updated from `Assistant: Reading config files` to `Assistant (in progress): Reading config files`, since that prefix change *is* part of the behavioral fix.

## Risks

Low. All three changes are additive to the transcript/prompt; the underlying generator API, retry/dedup logic, and cadence are unchanged. Worst case if a model ignores the new markers, behavior degrades back to the previous "guess from prose" mode — which is the current production behavior. The streaming-branch hint is the only place where the prompt actively forbids past tense, and it only activates when the provider says the assistant is mid-response, which is exactly the case where past tense is wrong by construction.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `high` • Cost: `$0.60`_

### DIFF
--- a/src/node/services/agentStatusService.test.ts
+++ b/src/node/services/agentStatusService.test.ts
@@ -17,7 +17,11 @@ import { createTestHistoryService } from "./testHistoryService";
 
 interface AgentStatusServiceInternals {
   runTick(): Promise<void>;
-  runForWorkspace(workspaceId: string, observedRecency?: number | null): Promise<void>;
+  runForWorkspace(
+    workspaceId: string,
+    observedRecency?: number | null,
+    streaming?: boolean
+  ): Promise<void>;
 }
 
 interface ActivitySnapshotForTest {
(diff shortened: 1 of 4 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed: folded `streaming` into `computeInputHash` (single-byte prefix) and added a regression test that asserts flipping `streaming` against an identical transcript triggers re-generation. Please take another look.

ammar-agent:
@codex review

Addressed: split into two hashes — `transcriptHash` (transcript bytes only) drives the history-catch-up guard / `lastSeenInputHash`, and `dedupHash = transcriptHash + streaming` drives the "settled, skip regeneration" branch / `lastInputHash`. Added a regression test that flips `streaming` on the second tick with a fresh recency bump but unchanged history and asserts the guard still defers (no generation, recency not consumed). Please take another look.

PR #3296, opened 2026-05-16
🤖 perf: speed up local static-check

ammar-agent:
Summary
- Speeds up the local static-check path by choosing an ESLint concurrency level from available CPU count, capped to avoid ESLint auto-concurrency overcommit.
- Keeps `MUX_ESLINT_CONCURRENCY` as an explicit override for CI or local tuning.
- Shortens shellcheck/hadolint output so static-check logs stay readable.

Background
- The local static-check workflow was bottlenecked by ESLint running at a fixed concurrency of 2 on larger agent/dev machines.
- Baseline measurement in this workspace was 85s before optimization; after the change, the same `make --no-print-directory static-check` path completed in 10s.

Implementation
- `scripts/lint.sh` now derives the default ESLint concurrency from `getconf`, `nproc`, or `sysctl`, clamps it between 2 and 8, and leaves the environment override unchanged.
- `Makefile` now prints concise shellcheck/hadolint summaries instead of echoing long full command lines.

Validation
- `shellcheck scripts/lint.sh`
- `MUX_ESLINT_CONCURRENCY=4 ./scripts/lint.sh` override check
- `make --no-print-directory static-check`

Risks
- Low. The validation coverage is unchanged; this only changes default local ESLint parallelism and log verbosity. Machines that need a different concurrency can continue using `MUX_ESLINT_CONCURRENCY`.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `off` • Cost: `$4.67`_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -352,13 +352,15 @@ lint-zizmor: ## Run zizmor security analysis on GitHub Actions workflows
 SHELL_SRC_FILES := $(shell find . -not \( -path '*/.git/*' -o -path './node_modules/*' -o -path './mobile/node_modules/*' -o -path './build/*' -o -path './dist/*' -o -path './release/*' -o -path './benchmarks/terminal_bench/.leaderboard_cache/*' \) -type f -name '*.sh' 2>/dev/null)
 
 lint-shellcheck: ## Run shellcheck on shell scripts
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3298, opened 2026-05-16
🤖 fix: show mid-stream goals in panel

ammar-agent:
## Summary

Shows goals set during an active stream in the Goal panel immediately, while keeping the durable `goal.json` write ordered behind stream-end accounting. The optimistic goal is now a transient activity-only snapshot and is read-only until it is persisted.

## Background

When `/goal` was run while the first chat turn was still streaming, the backend queued the goal write and returned success, but the Goal panel read workspace activity snapshots rather than the returned record. Because no snapshot was emitted until stream-end persistence, the panel opened with “No goal is set.”

## Implementation

- Publish a transient, non-persisted pending goal snapshot for mid-stream objective changes so the panel can render immediately without writing crash-unsafe optimistic state to extension metadata.
- Mark pending snapshots with `pendingPersistence`, centralize that check in shared goal helpers, and keep pending goals read-only in the Goal tab, command palette, and active-goal counting.
- Merge transient goal-only activity events into the frontend store without replaying stale streaming/recency fields.
- Preserve the pending snapshot during live cost previews and goal reads, clear it after all stream-end drain outcomes, and reject status/budget-only follow-up mutations while a queued goal replacement is still pending.

## Validation

- `bun test src/node/services/workspaceGoalService.test.ts src/browser/stores/WorkspaceStore.test.ts src/browser/utils/commands/sources.test.ts src/browser/features/RightSidebar/GoalTab.test.tsx`
- `git diff --check`
- `make static-check`

## Risks

Medium regression risk in goal lifecycle/activity synchronization: the change touches backend goal deferral, frontend activity merging, Goal panel controls, and command palette visibility. Tests cover the pending, drain, no-op, preview/read, stale-field merge, and read-only UI paths.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$27.66`_

### DIFF
--- a/src/browser/features/RightSidebar/GoalTab.test.tsx
+++ b/src/browser/features/RightSidebar/GoalTab.test.tsx
@@ -96,6 +96,22 @@ describe("GoalTab", () => {
     expect(getByText("3 / 10")).toBeTruthy();
   });
 
+  test("renders pending goals read-only until they are saved", () => {
+    const { queryByLabelText } = render(
+      <GoalTab
+        goal={goal({ pendingPersistence: true, budgetCents: 500, turnCap: 10 })}
+        onSetStatus={mock()}
+        onClear={mock()}
+      />
+    );
+
+    expect(queryByLabelText("Pause goal")).toBeNull();
+    expect(queryByLabelText("Mark goal complete")).toBeNull();
+    expect(queryByLabelText("Clear goal")).toBeNull();
+    expect(queryByLabelText("Edit goal budget")).toBeNull();
+    expect(queryByLabelText("Edit goal turn cap")).toBeNull();
+  });
+
   test("edits budget inline and restores focus", async () => {
     const onUpdateBudget = mock(() => Promise.resolve(undefined));
     const { getByLabelText, getByText } = render(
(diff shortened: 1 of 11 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Please take another look.

ammar-agent:
@codex review

Rebased on `origin/main` and pushed the pending-goal cleanup/follow-up mutation guard. I also refreshed the PR body with the current implementation and validation.

ammar-agent:
@codex review

Addressed the read-path clobber case by re-emitting the pending transient goal snapshot from goal reads while a queued mid-stream goal exists, with a regression assertion added to the mid-stream queueing test.

PR #3301, opened 2026-05-16
🤖 feat: iterate goals UI with completed-goals history and editable objective

ammar-agent:
## Summary

Iterates the right-sidebar Goal tab into the long-lived home for goal
configuration. Builds on the original "completed-goals history + in-place
objective editing" change with three follow-on iterations: always-visible
tab, active-goal accent, in-tab create form, and a new defaults panel
that owns both the per-workspace override *and* the global defaults that
used to live in Settings → Experiments.

## Background

After the first round, the Goal tab still:

- Hid itself when there was no current goal or history, so a fresh
  workspace had to learn `/goal` to start one.
- Looked identical whether or not the workspace had a live goal.
- Required users to go to Settings → Experiments to configure budget /
  turn-cap defaults, and offered no way to keep different defaults per
  workspace (e.g., higher budget for the workspace handling a tough
  refactor, low budget for everything else).

The user asked us to consolidate goal configuration into the Goal tab
itself — the natural home for goal lifecycle.

### Original change set (kept)

- **`goal-history.jsonl`** — Append-only per-workspace history file.
  `workspace.getGoalHistory` returns newest-first, capped at 200, with
  corrupt-line tolerance.
- **`WorkspaceGoalService` hooks** — `clearGoal` and the archive+replace
  branch of `setGoalImmediately` write history under the file lock before
  the canonical `goal.json` is rewritten.
- **`editInPlace` flag** on `GoalSetInputSchema` so the objective can be
  renamed in place (preserves `goalId` + accounting). Inline objective
  editor on the active-goal card.
- **Clear demotion** — Pause / Resume / Mark complete remain primary;
  Clear is a small underlined link, relabeled to "Archive this goal" for
  completed goals.
- **`GoalHistorySection`** under the present goal (or under the empty
  state), expandable read-only rows with end-reason + accounting +
  completion summary.

### Iteration 1: Always-visible Goal tab + active-goal accent + in-tab create form

- Goal tab is now visible whenever the `goals` experiment is enabled — no
  longer gated on `goal != null || goalHistory.length > 0`. Surfaces an
  empty-state with a create form so users don't need `/goal` to begin.
- `GoalTabLabel` subscribes to the workspace sidebar state and applies
  the canonical `text-success` accent when a live (`status === "active"`,
  non-pending) goal exists — mirrors the `useActiveGoalCount` predicate.
- **`GoalCreateForm`** in the empty state with slash-command parity
  (objective + optional budget + optional turn cap). Cmd/Ctrl+Enter
  submits. Blank budget/turn-cap defers to defaults so `/goal`, palette,
  and tab paths produce identical goals.

### Iteration 2: Per-workspace overrides + Goal-tab defaults panel

- **Sparse override schema** — `WorkspaceGoalDefaultsOverrideSchema` on
  both `WorkspaceConfigSchema` and `WorkspaceMetadataSchema`. Each field
  (`defaultBudgetCents`, `defaultTurnCap`, `alwaysRequireExplicitBudget`)
  is independently nullable; `null` means "inherit from global".
- **`workspace.goalDefaults.{get,set}` oRPC** modeled on
  `workspace.heartbeat.{get,set}`. The service drops the whole record
  when every field is null, keeping `~/.mux/config.json` minimal.
- **`mergeGoalDefaults(global, override)`** layers the override on top
  of the global. `loadGoalDefaults(api, workspaceId?)` now accepts an
  optional workspace; the slash command (`chatCommands.ts`), command
  palette (`sources.ts`), and in-tab create form (`RightSidebar.tsx`)
  all pass `workspaceId` so workspace rules win over the global default.
- **`GoalDefaultsSection`** at the bottom of the Goal tab (empty + active
  states). Collapsible `<details>` with summary showing the effective
  budget + turn cap and a "Workspace" badge when any override is active.
  Two stacked sub-panels:
  - **This workspace** — three rows with Inherit/Override toggles. When
    inheriting, copy reads "Inherits $X from All workspaces". Toggle on
    seeds the input with the inherited value; toggle off persists null.
  - **All workspaces** — reuses `GoalDefaultsControls` from Settings
    (now with an `onPersist` callback so the workspace panel's inherit
    labels re-pull on every global edit).
- Settings → Experiments no longer mounts `GoalDefaultsControls`; it
  emits a small pointer toward the Goal tab as the new home.

## Validation

`make static-check` clean (lint + format + typecheck + docs link check).

Targeted tests (all pass):

- `src/browser/utils/goals/resolveGoalSetIntent.test.ts` — **15 pass**,
  including 7 new `mergeGoalDefaults` precedence tests.
- `src/node/services/workspaceService.goalDefaults.test.ts` — **8 pass**
  (round-trip, partial, all-null clears, no-op short-circuit,
  validation, not-found).
- `src/node/services/workspaceGoalService.test.ts` — 91 pass.
- `src/browser/features/RightSidebar/GoalTab.test.tsx` — **17 pass**,
  including 5 create-form tests.
- `src/browser/features/RightSidebar/Tabs/TabLabels.goal.test.tsx` —
  **5 pass** (label accent matrix).
- `src/browser/features/Settings/Sections/ExperimentsSection.test.tsx` —
  **9 pass** (updated to assert goal defaults moved out of the panel).
- `tests/ui/layout/rightSidebar.test.ts` — 2 new layout-visibility
  guards (tab hidden when GOALS off; always present when GOALS on).

## Risks

- New write path on `setGoal` replace + `clearGoal` is best-effort: a
  history write failure logs but never aborts the lifecycle mutation.
- `editInPlace` is opt-in; existing `/goal`, palette, and kickoff paths
  continue to take archive+recreate unchanged.
- Per-workspace override is sparse and all-null clears: a workspace
  with no override is indistinguishable from one whose override was
  never set, so reverting an override never strands stale data on disk.
- The `loadGoalDefaults(workspaceId?)` signature change is backwards
  compatible — callers that don't yet know the workspace (Settings
  global panel) still get global-only defaults.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `high` • Cost: `$16.93`_


ammar-agent:
@codex review

Please review this iteration of the workspace Goal UI:
- backend: append-only goal-history JSONL + getGoalHistory IPC + editInPlace setGoal rename
- frontend: editable objective, de-prominent Clear, expandable completed-goals list under the present goal

Reads + writes happen under the per-workspace file lock; history is best-effort (failures logged, lifecycle mutation never aborted). Curious for feedback on UX wording ("Archive this goal" vs "Clear goal") and the editInPlace flag design vs a dedicated endpoint.

ammar-agent:
@codex review

Pushed a fix: added `getGoal` + `getGoalHistory` mocks to `src/browser/stories/mocks/orpc.ts` so the RightSidebar mount-time history fetch doesn't throw against the storybook/integration-test mock client (this was the root cause of the Storybook + Integration CI failures on the previous push).

ammar-agent:
@codex review

Addressed your two pieces of feedback:
- P2: forwarded `editInPlace` into `PendingGoalMutation` and the mid-stream deferral path so an inline rename during streaming still takes the rename branch when it drains. Added a regression test.
- P3: wrapped `getGoalHistory` in `fileLocks.withLock` so a concurrent append cannot leave a partial line that the reader silently skips.

ammar-agent:
@codex review

Pushed two more fixes addressing your latest feedback:
- P2 (Honor editInPlace while streams are active): already addressed in f4bfc593d before the latest review snapshot — `PendingGoalMutation` carries `editInPlace` and `applyPendingAfterStreamEnd` replays it through the in-place branch. Resolved that thread.
- P2 (Preserve the Goal tab for history-only workspaces): updated the right-sidebar layout effect to keep the Goal tab visible whenever `goalHistory.length > 0` as well, with `goalHistory.length` added to the dependency array so the effect re-runs once the history fetch populates. This makes the new empty-state + history branch reachable after clearing the last goal or on a fresh launch of a history-only workspace.

ammar-agent:
@codex review

Addressed your tie-breaker feedback: `getGoalHistory` now retains the JSONL append index and uses it as the DESC tie-breaker when `endedAtMs` is equal, so the latest line in the file always wins regardless of how fast the appends happened. Added a regression test that pins `Date.now` to force same-ms timestamps across two consecutive clears.

ammar-agent:
@codex review

Three new iterations on the Goal UX:

1. Goal tab is now always visible when the GOALS experiment is enabled (no longer gated on `goal != null || goalHistory.length > 0`).
2. Goal tab label uses `text-success` accent when the workspace has a live active goal (mirrors `useActiveGoalCount` predicate).
3. Empty-state GoalTab now exposes an in-tab create form with full slash-command parity (objective + optional budget + optional turn cap), routed through the same `loadGoalDefaults` + `resolveGoalSetIntent` + unpriced-model gate as the palette and slash paths.

Plus 12 new tests (5 create-form, 5 label-accent, 2 layout-visibility integration).

ammar-agent:
@codex review

Iteration 3 just landed (commit `ddcf9d811`): consolidated goal defaults into the Goal tab.

Key things to look at:

1. **Schema layering** — new `WorkspaceGoalDefaultsOverrideSchema` is sparse-nullable (`{ defaultBudgetCents: nullable, defaultTurnCap: nullable, alwaysRequireExplicitBudget: nullable }`). Carried on both `WorkspaceConfigSchema` and `WorkspaceMetadataSchema`. Persisted via `workspaceService.setWorkspaceGoalDefaults` which drops the record when every field is null (modeled on `setHeartbeatSettings`).

2. **Resolver precedence** — `mergeGoalDefaults(global, override)` in `resolveGoalSetIntent.ts`. Used `??` for inherit fallback (null/undefined override field → global). The tricky case is `defaultTurnCap` which is itself nullable (null = "no cap" at the global layer); the `??` short-circuit correctly treats override `null` as "inherit", not as "explicit no cap", because the override schema has no way to express "explicit no cap" — that's the global-layer responsibility.

3. **Three create surfaces still in lockstep** — `loadGoalDefaults(api, workspaceId?)` now layers the workspace override. Callers updated: `chatCommands.ts` (slash), `commands/sources.ts` (palette), `RightSidebar.tsx` (in-tab create form). All three pass `workspaceId` so workspace rules win.

4. **UI** — new `GoalDefaultsSection.tsx` at the bottom of the Goal tab. Collapsible `<details>` with summary badge. Per-field inherit/override toggles. Inherit labels re-pull when global is edited via the new `GoalDefaultsControls#onPersist` callback.

5. **Settings cleanup** — `ExperimentsSection` no longer mounts `GoalDefaultsControls`; replaced with a small pointer to the Goal tab. Tests updated.

Risks called out in the PR body.

ammar-agent:
@codex review

Addressed P2 in commit `b19d5b4ec`: `handleGoalCreate` now fetches `api.providers.getConfig()` at submit time instead of reading the `useProvidersConfig()` hook value, matching the slash command + palette pattern. Falls back to the hook value on fetch failure so transient network errors don't leak through.

Iteration 3 (goal defaults consolidation) is also up for review on top of the previous always-visible/accent/create-form work — please cover the new code as well:

- Sparse `WorkspaceGoalDefaultsOverrideSchema` on `WorkspaceConfigSchema` + `WorkspaceMetadataSchema`
- `workspace.goalDefaults.{get,set}` oRPC + `workspaceService.set/getWorkspaceGoalDefaults` (modeled on heartbeat; drops record when all-null)
- `mergeGoalDefaults(global, override)` + `loadGoalDefaults(api, workspaceId?)` precedence
- `GoalDefaultsSection` (collapsible per-workspace overrides + reused global controls)
- Settings → Experiments now just points to the Goal tab

ammar-agent:
@codex review

Addressed your P2 in `18e48a939`: `GoalDefaultsSection` now tracks an explicit `isLoading` flag.

- `persistOverride()` is a no-op while loading (would have otherwise written the synthesized all-null shape on top of saved overrides).
- The override panel shows a `Loading…` label and visually + a11y-disables the toggles (`opacity-60`, `pointer-events-none`, `aria-busy`, `<button disabled>`).
- `isLoading` resets to `true` whenever `workspaceId` changes so workspace switches can't let new-workspace edits race the old read.

Please re-review iteration 3 (goal defaults consolidation).

ammar-agent:
@codex review

Addressed your latest P2 in `36343404b`:

- `persistGoalDefaults` (Settings GoalsSection) now awaits `updateGoalDefaults` before firing `onPersist`.
- `GoalDefaultsSection` dropped the `globalReloadKey` + refetch pattern entirely. `onPersist(next)` already receives the freshly normalized `GoalDefaults`; we push that straight into local state, so the race window is closed.

Tests still pass:

- `GoalsSection.test.tsx` — 2 pass
- `resolveGoalSetIntent.test.ts` — 15 pass
- `workspaceService.goalDefaults.test.ts` — 8 pass

ammar-agent:
@codex review

ammar-agent:
@codex review

P1 (mid-stream promote) addressed by rejecting promote when `isWorkspaceStreaming` is true, with a typed `WorkspaceGoalTransitionError`. The UI surfaces the message in the Upcoming section.

P2 (queued-goal defaults) addressed by applying `resolveGoalSetIntent` in `UpcomingAdder.submit` before the IPC call, matching the create-form / slash command / palette paths.

New test `promoteUpcomingGoal rejects mid-stream` covers P1.

Please take another look.

ammar-agent:
@codex review

Both P2s addressed:

- **Refresh board after auto-promotion** — `useGoalBoard` now takes an `activeGoalKey` parameter and the GoalTab passes `${goalId}:${status}`. Setting/clearing the active goal re-fetches the board.
- **Avoid demoting completed goals into Upcoming** — `promoteUpcomingGoal` now archives a completed previous-active to history under "completed" instead of pushing it back into the queue.

New test: `promoteUpcomingGoal archives a completed active goal instead of demoting to upcoming`.

Please take another look.

ammar-agent:
@codex review

P2 (gate promoted budgeted goals on pricing data) addressed by adding `canRunBudgetedGoalOnKickoffModel` checks to both promote paths:

- `promoteUpcomingGoal` throws `WorkspaceGoalTransitionError(UNPRICED_TARGET_MODEL_GOAL_MESSAGE)` — board UI surfaces the message via its existing catch.
- `promoteNextUpcomingUnlocked` (auto on complete) logs + returns null, keeping the upcoming head in place so the user can resolve and retry.

Please take another look.

ammar-agent:
@codex review

P2 (double-archive on skipped auto-promote) fixed by hoisting the pricing gate check above the history append in `maybeAutoPromoteOnComplete`. When the next queued goal is budgeted but the kickoff model is unpriced, we now skip both the history append AND the goal.json write, leaving the completed goal in place so a later manual promote archives it exactly once.

Please take another look.

ammar-agent:
@codex review

P2 (mock missing APIContext export) fixed by adding `APIContext: createContext(null)` to the mock module, mirroring the runtime hook short-circuit.

ammar-agent:
@codex review

Both P2s fixed:
- **Revived completed dedup**: `getGoalBoard` now dedupes completed-history rows against the upcoming list too (in addition to active + archived).
- **History cap**: `readHistoryUnlocked` slices to `GOAL_HISTORY_RENDER_CAP` (200) before returning, matching the public `getGoalHistory`.

ammar-agent:
@codex review

Re-requesting review after 862d006d3 (latest commit fixes both P2 items from previous round: dedupe revived completed against Upcoming + cap board history reads to GOAL_HISTORY_RENDER_CAP).

ammar-agent:
@codex review

P2 (empty-state scroll) fixed by adding `overflow-y-auto` to the empty-state section, mirroring the active-goal branch.

ammar-agent:
@codex review

P1 (auto-promote mid-stream) addressed by adding the same `isWorkspaceStreaming` guard to both `maybeAutoPromoteOnComplete` and `promoteNextUpcomingUnlocked`. Skip is logged; the upcoming head stays intact so a later mutation picks it up cleanly.

ammar-agent:
@codex review

P1 (deferred promote) addressed by hooking the auto-promote into `applyPendingAfterStreamEnd` (stream-end). The new helper runs at the same lifecycle point as the cost-accounting drain, under the same workspace lock. Runs before draining pending setGoal mutations so the order is: stream cleanly ends → completed goal archived → next goal promoted → any deferred setGoal sees the new active.

ammar-agent:
@codex review

Both P1 + P2 addressed in b4a804bc2-fix:

- **P1 (history before streaming gate)**: Added explicit `isWorkspaceStreaming` check inside `runDeferredAutoPromoteAfterStreamEnd` BEFORE any history mutation. If `stopStreamingStatus` hasnt fired yet, bail cleanly.
- **P2 (stale expectedGoalId)**: Inverted order — pending setGoal drain runs FIRST, then the deferred auto-promote. The drained setGoal now targets the same goal it was queued against.

ammar-agent:
@codex review

Both P2s addressed:
- **Re-arm deferred promotion**: `runDeferredAutoPromoteAfterStreamEnd` now polls `isWorkspaceStreaming` up to ~600ms (5 retries with 0/50/100/200/250ms backoff) before bailing. Resolves the stream-end race without needing event plumbing into WorkspaceService.
- **Load defaults at submit**: `UpcomingAdder.submit` calls `loadGoalDefaults(api, workspaceId)` at submit time instead of relying on hook state, falling back to the hook value on fetch failure.

ammar-agent:
@codex review

Latest is 39d99a779. CI all green (Required gate passed). Earlier rounds addressed: lifecycle/mode refactor, multi-goal queue, mid-stream guards, pricing gates, dedup, bounded retry on stream-end race, defaults at submit time. All previous Codex P1+P2 threads resolved.

ammar-agent:
@codex review

P2 (exhausted-limits on promote) addressed by running `applyBudgetDrivenStatus` on the activated record before writing. Re-promoted goals with cost >= budget or turnsUsed >= turnCap correctly land in `budget_limited` instead of `active`.

ammar-agent:
@codex review

P3 (Archive link routing) addressed by routing the click to `workspace.archiveGoal` when the goal is lifecycle-complete. The board now correctly surfaces archived completed goals in the Archived section. Test updated.

ammar-agent:
@codex review

P2 (stale completionSummary on promote) addressed by clearing it explicitly in both `promoteUpcomingGoal` and `promoteNextUpcomingUnlocked`. Matches the existing `completionSummaryPatch` invariant.

ammar-agent:
@codex review

Latest commit 40d75e80e:
- af6342c8b: GoalTab now reads APIContext via useContext (storybook-tolerant, same pattern as the hooks).
- 40d75e80e: clear `completionSummary` in both promote paths so revived/promoted goals do not carry stale "done" messages.

All previous threads resolved.

ammar-agent:
@codex review

P2 (Goal tab on child workspaces) addressed by gating the Goal tab injection on `parentWorkspaceId == null` via `useWorkspaceMetadata`.

ammar-agent:
@codex review

P2 (active slot vs stale history in archiveGoal) fixed by promoting the active-slot branch to the top of the source-priority order. The currently-active record always wins, so revived+re-completed goals archive cleanly without duplicates.

ammar-agent:
@codex review

P2 (duplicate completed-history rows) fixed by adding a `seen` Set keyed on goalId inside the Completed loop in `getGoalBoard`.

ammar-agent:
@codex review

ammar-agent:
@codex review

Follow-up to address three reviewer points:

1. **Promote mid-stream**: removed the `WorkspaceGoalTransitionError` throw and instead interrupt the active stream (via wired `interruptStream` callback) before promoting. Preserves cost attribution since the in-flight turn's tokens stay with the goal that ran them. Falls open if interrupter is missing or errors.

2. **Frontend error surface**: router translates `WorkspaceGoalTransitionError` / `WorkspaceGoalChildWorkspaceError` to `ORPCError("BAD_REQUEST", { message })` so the real reason reaches the user instead of "Internal server error".

3. **DnD reorder + inline edit**: replaced up/down arrows with `@dnd-kit/sortable` grip handle (mirrors `ProvidersSection`). New `workspace.updateUpcomingGoal` endpoint backs an inline editor on Upcoming rows. "Archive" relabeled "Remove".

5 new service tests (116 total pass).

ammar-agent:
@codex review

Addressed the P2 on mid-stream editInPlace optimistic snapshot:

`setGoalInternal` now overlays the rename onto the current record (preserving goalId, costCents, turnsUsed, attributedChildren, budgetCents, turnCap) for the editInPlace branch. Non-editInPlace branches still take the `createGoal` path. New test guards the projection.

ammar-agent:
@codex review

Addressed both items in one push:

1. **P2 mutable-field normalization on editInPlace optimistic snapshot**: projection now flows through `applyMutableFields` (→ `applyBudgetDrivenStatus`), mirroring the drain. New test asserts that a mid-stream rename + tightening `budgetCents` projects directly to `budget_limited` instead of a stale `active`.

2. **Chat-transcript noise from synthetic "Goal cleared" messages**: `appendClearSummary` no longer sets `uiVisible: true`, so the persisted summary still reaches the model (for "what did we just finish" follow-ups) but no longer renders as a full assistant chat bubble. The right-sidebar Goal Board's Completed/Archived sections already surface the same info visually. User noticed this as "random goal updates in chat transcript".

118 service tests pass; static-check green.

ammar-agent:
@codex review

Addressed P2 on budget placeholder/helper text. Both `GoalTab` create form and `GoalBoardSections` UpcomingAdder now gate placeholder + helper text on `defaults.alwaysRequireExplicitBudget`:
- ON → `$X (default)` placeholder + "Leave Budget / Turn cap blank to use the defaults shown above"
- OFF → `no budget (default)` placeholder + "Leave Budget blank to create an unbudgeted goal; Turn cap blank uses the default shown above"

This matches `resolveGoalSetIntent`'s actual behavior.

ammar-agent:
@codex review

Addressed P2 on hidden default turn cap. Added an explicit turn-cap field to both the queue Adder (`UpcomingAdder`) and the inline editor (`UpcomingRowEditor`), with default-aware placeholders matching the budget field's pattern. Blank in the Adder still falls through `resolveGoalSetIntent` (applies default); blank in the Editor means "no cap" (the default was already resolved at queue time, identical to the budget field's existing semantics).

ammar-agent:
@codex review

Addressed both P2s:

1. **Reject partial turn-cap strings**: extracted `parseGoalTurnCapInput` (full-string `Number` + `Number.isSafeInteger`) to `src/common/utils/goals/budgetParser.ts` and routed both the Adder and the inline editor through it. `GoalTab.tsx`'s `parseTurnCapInput` is now an alias for the same shared parser. `1.5` / `12abc` now correctly fail.

2. **Serialize workspace-default saves**: chain saves through a `useRef<Promise<void>>` in `GoalDefaultsSection.tsx` so each `set` request awaits the previous before issuing its own write. Local optimistic state still updates immediately so the UI feels snappy; on-disk state is guaranteed to match the final edit.

PR #3302, opened 2026-05-16
🤖 perf: cut SSH workspace startup ~9× on the warm path via fused single-RTT materialization

ammar-agent:
## Summary

Optimize SSH workspace startup using `mux run` as a debug + timing harness, cutting cold-path SSH workspace creation by ~24% and warm-path creation by **~9× (≈1.8s → ~205ms median)** against `ovh-1`. The warm path collapses what used to be ~9 sequential SSH round-trips into a single fused command, bringing subsequent `mux run` invocations within ~2.2× of one raw SSH-command RTT on this link.

## Background

The objective was to instrument `mux run --runtime "ssh <host>"` so it doubles as a startup-perf harness, then drive both cold and warm SSH workspace creation down as far as possible. The starting point for cold workspace startup against `ovh-1` was ~2.92s (empty repo, fresh remote), and "warm" subsequent runs were also ~1.8s because the CLI was using an ephemeral `srcBaseDir` per invocation — so every run looked cold to the remote.

## Implementation

This PR contains three commits:

### `perf: cut SSH cold-workspace startup ~24% via batched ensureBaseRepo + skipped no-op probes`

Cold-path wins (baseline 2.92s → 2.21s, –24%):

1. **Skip `prefetchOriginOnRemote` when local has no `origin`** — hoist the origin-URL check; the fetch dance is a guaranteed no-op for local-only projects (~500ms).
2. **Collapse `ensureBaseRepo` into one SSH round-trip** — single batched shell script with status sentinels (`STATUS_CREATED`, `STATUS_CORE_BARE`); falls through to slow-path retry helper only on real config-lock conflicts (~300ms).
3. **Drop trailing `refreshBaseRepoOrigin`** when prefetch already pushed it (~100ms).
4. **Skip `ensureHealthyBaseRepoForSync`** when `ensureBaseRepo` just created the bare repo — zero packs by definition (~95ms).

Also extends `mux run` itself to provision SSH workspaces (previously only Docker) and adds a timed init logger (`makeTimedCliInitLogger`) so `--verbose` annotates every step with `[+<total> Δ<delta>]` — this is the harness used to identify all subsequent wins.

### `perf: collapse warm SSH workspace creates to one fused round-trip (~9× faster)`

Warm-path wins (1.8s → ~205ms median):

1. **Stable remote `srcBaseDir` for the CLI** — `mux run` was using `tempDir/src` per invocation, guaranteeing a fresh remote base repo every time. Anchor to `~/mux` (matching the desktop app) and resolve the tilde to an absolute remote path once up-front so `path.resolve()` in `session.ensureMetadata()` doesn't mangle `~/mux/...` into local nonsense.
2. **Single-SSH-command warm fast-path (`tryWarmWorktreeAdd`)** — when the remote already has a healthy shared base repo for this project AND the staged `refs/mux-bundle/*` tips already match local heads, skip the entire slow pipeline (~9 sequential SSH round-trips) and run a single tight shell pipeline that probes + materializes + reports `.gitmodules` in one round-trip. Misses fall through with a structured `WARM_MISS:<reason>` log line.
3. **Fold `.gitmodules` probe into the warm command** so the caller can skip the post-worktree submodule-sync SSH RTT when no submodules exist.
4. **Eliminated double `git show-ref`** — added `fastReadGitHeadsRefs()` that walks `.git/refs/heads/` + `.git/packed-refs` directly (~2ms vs ~70ms), with graceful fallback to `git show-ref` for worktree-gitdir or reftable-based repos.
5. **Dropped the `createWorkspace` mkdir SSH RTT** — both materialization paths create their own parents already.
6. **Snapshot-digest-only gate** — the digest is `sha256(git show-ref --heads)`, so a digest match implies a manifest match. Skip the parallel manifest verification.

### `perf: preserve slow-path origin freshness in SSH warm fast-path`

Addresses a P1 correctness concern from Codex review on this PR. The original warm-path implementation always based the new workspace on `refs/mux-bundle/<trunk>` and returned before the slow path's `fetchOriginTrunk()` / `resolveFreshWorkspaceSourceBase()` logic could fetch and prefer `origin/<trunkBranch>`. When the remote cache was warm but `origin/main` had advanced, the user silently got a stale checkout.

Fix: fold the same origin fetch into the fused warm SSH script. When the local project has an `origin` URL we now also:

1. `git remote set-url origin <local-url>` on the remote base repo (idempotent; protects against URL drift between syncs).
2. `git fetch --quiet origin +refs/heads/<trunk>:refs/remotes/origin/<trunk>` on the host. Failure is tolerated (`fo=0`).
3. Prefer `refs/remotes/origin/<trunk>` as the worktree-add base whenever the fetch succeeded; fall back to the bundle ref otherwise. This matches `resolveFreshWorkspaceSourceBase()`'s fallback semantics exactly.

The fetch runs on the SSH host (datacenter↔upstream link), so it adds no client-side round-trip and the warm path stays inside a single client→host RTT envelope. The wall-time impact for repos with origin is on the order of the host↔upstream fetch handshake (typically fast when origin hasn't moved); for local-only repos (no origin URL) behavior is unchanged.

The warm path also now emits `"Reusing existing remote project snapshot"` on hit so existing lifecycle tests (e.g. `initWorkspace reuses snapshots and preserves remote-only tags across later resyncs`) continue to observe a consistent step.

## Validation

Measured on ovh-1 (SSH multiplexed RTT median ~92ms, 2× RTT = ~184ms):

| Run kind | Before | After |
|---|---|---|
| Cold (empty repo, fresh remote) | 2.92s | 2.21s (–24%) |
| Warm (subsequent run, local-only repo) | 1.80s | **205ms median (–89%)** |

15 consecutive warm runs after the patch: `196 198 199 199 200 203 205 205 205 206 207 209 212 214 232` ms
→ min 196, P10 198, median 205, P90 214, max 232, mean 206ms (~2.24× one SSH-command RTT).

The narrow spread (σ≈8ms) shows variance is dominated by raw SSH-link jitter — the bare `ssh true` RTT itself varies 88–140ms on this link. The fundamental floor is `1× SSH RTT + ~10ms remote git worktree add + ~100ms ssh client/server overhead for a non-trivial script`. Repos with an `origin` URL will incur an additional host↔upstream fetch handshake inside the same SSH RTT envelope.

All **444 runtime + CLI unit tests** continue to pass, including the SSH retry orchestration suite that exercises the promisor-cleanup / repair paths preserved here. The full `tests/runtime/runtime.test.ts` "SSHRuntime sync does not import stale remote tracking refs" integration suite (5/5) passes locally against the Docker SSH fixture — including the `initWorkspace reuses snapshots and preserves remote-only tags across later resyncs` case that surfaced the original correctness concern. Local `make static-check` is green (typecheck, lint, prettier, docs link check).

## Risks

Touched logic: SSH workspace creation. The warm fast-path is gated by strict miss conditions and falls through to the existing slow path on any miss; the slow path is unchanged in behavior aside from the cold-path round-trip reductions. The promisor-cleanup, partial-clone, and lock-conflict retry paths are preserved verbatim — the batched `ensureBaseRepo` falls back to the original `normalizeBaseRepoSharedConfig` retry helper whenever it sees a non-trivial config-lock error from its inline normalize attempt.

Regression risk is concentrated in:

- **Cold-path `ensureBaseRepo`**: the batched command now does test/mkdir/init/unset-core.bare/unset-promisor in one shell pipeline; behavior should be identical to the original sequential calls for all currently-tested states.
- **Warm fast-path correctness**: gated by snapshot-digest match (sha256 of `git show-ref --heads`) so a stale or mismatched local repo cannot be silently materialized against the wrong base — any mismatch falls to the slow path which independently re-verifies. The origin-fetch preamble now also matches the slow path's `resolveFreshWorkspaceSourceBase` semantics so upstream advances are picked up even on warm hits.
- **`mux run` SSH provisioning**: this is a new code path in `src/cli/run.ts` and only affects the CLI; the desktop app's `WORKSPACE_CREATE` flow is unchanged.

## Pains

The biggest debugging hiccup was a metadata-path bug: persisting `~/mux/...` through `session.ensureMetadata()` (which runs `path.resolve()`) silently mangled it to `/Users/.../~/mux/...`, which the runtime then sent verbatim to the remote — making the workspace exist on the remote but fail the subsequent `ensureReady` check with "repository not found". The fix (resolve tilde once via `runtime.resolvePath` before persisting) mirrors how the desktop app's `WORKSPACE_CREATE` IPC handler already handles this.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$27.24`_

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -71,7 +71,8 @@ import { log, type LogLevel } from "../node/services/log";
 import chalk from "chalk";
 import type { InitLogger, WorkspaceInitResult } from "../node/runtime/Runtime";
 import { DockerRuntime } from "../node/runtime/DockerRuntime";
-import { runFullInit } from "../node/runtime/runtimeFactory";
+import { createRuntime, runFullInit } from "../node/runtime/runtimeFactory";
+import type { Runtime } from "../node/runtime/Runtime";
 import { execSync } from "child_process";
 import { getParseOptions } from "./argv";
 import { EXPERIMENT_IDS } from "../common/constants/experiments";
@@ -138,13 +139,40 @@ function generateWorkspaceId(): string {
   return `run-${timestamp}-${random}`;
 }
 
-function makeCliInitLogger(writeHumanLine: (text?: string) => void): InitLogger {
+/**
+ * Init logger that prepends elapsed/delta timestamps to each step so `mux run`
+ * can double as a startup-timing harness (especially useful for SSH workspaces
+ * where the slowest phases are hidden inside multi-second remote operations).
+ *
+ * Format: `[+12.3s +4.5s] message`
+ *   - first number: ms since the timed logger was created
+ *   - second number: ms since the previous logStep call
+ */
+function makeTimedCliInitLogger(writeHumanLine: (text?: string) => void): InitLogger {
+  const start = Date.now();
+  let lastStep = start;
+  const fmt = (ms: number): string => {
+    if (ms >= 10_000) return `${(ms / 1000).toFixed(1)}s`;
+    if (ms >= 1_000) return `${(ms / 1000).toFixed(2)}s`;
+    return `${ms}ms`;
+  };
   return {
-    logStep: (msg) => writeHumanLine(`  ${msg}`),
+    logStep: (msg) => {
+      const now = Date.now();
+      const stepMs = now - lastStep;
+      const totalMs = now - start;
+      lastStep = now;
+      writeHumanLine(`  [+${fmt(totalMs)} Δ${fmt(stepMs)}] ${msg}`);
+    },
     logStdout: (line) => writeHumanLine(`  ${line}`),
     logStderr: (line) => writeHumanLine(`  [stderr] ${line}`),
     logComplete: (exitCode) => {
-      if (exitCode !== 0) writeHumanLine(`  Init completed with exit code ${exitCode}`);
+      const totalMs = Date.now() - start;
+      if (exitCode !== 0) {
+        writeHumanLine(`  [+${fmt(totalMs)}] Init completed with exit code ${exitCode}`);
+      } else {
+        writeHumanLine(`  [+${fmt(totalMs)}] Init complete`);
+      }
     },
   };
 }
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

Addressed the P1 origin-freshness issue in commit 9ac2c32b7: the warm fast-path now folds `git fetch origin <trunk>` into the fused SSH script and prefers `refs/remotes/origin/<trunkBranch>` over `refs/mux-bundle/*`, exactly mirroring `resolveFreshWorkspaceSourceBase()`'s semantics. Fetch is best-effort with bundle-ref fallback. Local validation: integration test "initWorkspace reuses snapshots and preserves remote-only tags across later resyncs" now passes (was the failing test that surfaced this issue), full SSH sync test suite (5/5) green, 444/444 unit tests green.

PR #3303, opened 2026-05-17
🤖 fix: stage SSH forks under a unique scratch path to prevent rm -rf collisions

ammar-agent:
## Summary

Hardens `SSHRuntime.forkWorkspace` against a class of silent SSH-workspace deletions where a routine fork (or aborted fork attempt) could `rm -rf` the destination directory of an *unrelated* workspace. The fix is structural — forks now land at the canonical project root by construction and all I/O is staged under a per-attempt scratch directory whose name carries 96 bits of crypto entropy, so cleanup is provably constrained to paths this fork attempt uniquely owns.

## Background

In production we observed an active SSH workspace at `<canonical-root>/<workspaceName>/` get silently reduced to an empty directory: same birth-time as a failed fork attempt elsewhere, no `goals-ui-qfdv` ever logged through `deleteWorkspace`, no `git worktree list` entry. Investigation traced this to two interacting bugs in `forkWorkspace`:

1. **Legacy-path inheritance.** The new workspace path was `path.posix.join(path.posix.dirname(sourceWorkspacePath), newWorkspaceName)`. Workspaces persisted before #3125 still carry `<srcBaseDir>/<basename>/<name>`, while the shared base repo (used by `git worktree add`) lives at `<srcBaseDir>/<basename>-<12hex>/.mux-base.git`. Forking a legacy workspace would point `worktree add` at the canonical base repo but try to materialize the worktree at the legacy parent — `git worktree add` fails with `fatal: invalid reference: <branch-only-in-legacy>` (logged 11× in the dirty log window).

2. **Unsafe `rm -rf` cleanup.** Both the worktree-add failure path and the cp-fallback failure path issued `rm -rf <newWorkspacePath>` against the *final* destination. Combined with (1), or with two concurrent forks racing on the same `newWorkspaceName`, that `rm -rf` could destroy a sibling workspace's directory. The empty-stub pattern observed on disk (canonical-path directory with `Links: 2`, mtime == birth, never written to since) is exactly the signature of the cp fallback being killed mid-create or a sibling worktree being wiped after a race.

### Canonical-layout invariant

`forkWorkspace` now computes `newWorkspacePath` from `getProjectLayout(projectPath)` instead of inheriting `dirname(sourceWorkspacePath)`. The new workspace ALWAYS lands at `<srcBaseDir>/<projectId>/<newWorkspaceName>`, where `projectId = sanitize(basename) + "-" + sha256(projectPath)[:12]`. The source workspace's persisted path is intentionally left untouched — legacy records migrate naturally as users archive + re-fork them, and they no longer pose a destruction risk because forks no longer touch their parent directory.

### Per-attempt staging directory

Every byte of fork I/O lands in `<canonical-root>/.mux-fork-staging-<12 hex>/` first. The 12 hex chars are 96 bits of `crypto.randomBytes(6)` entropy, and the `.mux-fork-staging-` prefix is used nowhere else in the codebase, so any `rm -rf` against the staging path is **provably** unable to target a real workspace. The canonical destination is materialized via:

* **`git worktree move <staging> <final>`** for the worktree-add path. This updates the bare repo's `worktrees/<stagingName>/gitdir` back-reference atomically with the on-disk rename — using a plain `mv` would leave the bare repo pointing at the staging path.
* **`mv <staging> <final>`** for the cp-fallback path (no gitdir indirection to maintain).

Both finalize commands embed a `test -e <final>` collision guard immediately before the rename, so a concurrent fork that landed at the destination between our initial `test -e` and this point causes us to surface a `created by another process during fork; pick a different name` error and clean up our staging dir — instead of overwriting.

### Defense-in-depth

* Every `rm -rf` in the fork path now logs `(project, source, target, reason)` via `log.info` so workspace-loss investigations can correlate cleanups with the fork attempt that produced them.
* The outer `try/catch` runs a final cleanup pass that explicitly unregisters the staging worktree (`git worktree remove --force`) before `rm -rf`'ing the staging path, so an aborted fork can never leave a dangling `worktrees/<stagingName>/gitdir` entry in the bare repo.

## Validation

* New test file `src/node/runtime/SSHRuntime.fork.test.ts` (7 tests) pins the safety invariants:
  - new path is canonical even when source is at a legacy `<basename>/` path
  - worktree-add targets the canonical staging path (never the legacy parent)
  - no `rm -rf` ever targets the final workspace path on a failed worktree add
  - no `rm -rf` ever targets the final workspace path on a failed cp -R -P
  - finalize uses `git worktree move` (not raw `mv`) so the bare repo's gitdir stays in sync
  - collision at finalize aborts the fork **without** destroying the destination
  - staging IDs differ across attempts (mitigates concurrent-fork same-name races)
* `make static-check` clean (typecheck, lint, prettier, eslint, code-to-docs).
* `bun test src/node/runtime/` — 322 tests pass, no regressions.
* `bun test src/node/services/utils/forkOrchestrator*.test.ts` — 22 tests pass.

## Risks

* **Behavior change for legacy-path workspaces:** forks off a legacy `/root/<basename>/<src>` workspace now produce a *new* workspace at the canonical root, so users will see two layouts coexisting in `git worktree list`. This is strictly safer than the previous behavior (which could destroy data) and the source record is unaffected. Long-running legacy workspaces continue to work; they only migrate when forked.
* **`git worktree move` requires git ≥ 2.17:** that's been GA since 2018. All Mux-supported SSH hosts ship newer git.
* **Extra SSH round-trips:** the staging→final finalize adds 1 SSH RTT vs. the previous direct-write. The worktree-add path is unchanged in steady state; the cp-fallback path was already several RTTs and one more is in the noise.
* **Storage during the fork window:** a fork in progress temporarily occupies disk under `<canonical-root>/.mux-fork-staging-<hex>/`. On any failure the catch-all removes it; in the rare case mux is killed mid-fork, the next `ls <canonical-root>` will show the staging dir — safe to delete manually because the `.mux-fork-staging-` prefix and 96-bit suffix uniquely mark it.

## Pains

The root-cause investigation required cross-referencing four sources (config.json, remote `find -newer`, the local log, and `git worktree list` on the bare repo) because none of the relevant `rm -rf` invocations previously logged their target path. The defensive logging added in this PR (project/source/target/reason on every cleanup) is specifically to short-circuit future investigations of this shape.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$4.14`_

### DIFF
--- a/src/node/runtime/SSHRuntime.fork.test.ts
+++ b/src/node/runtime/SSHRuntime.fork.test.ts
@@ -1,0 +1,398 @@
+/**
+ * Regression tests for SSHRuntime.forkWorkspace.
+ *
+ * These tests exist to lock in two safety invariants that, when broken, have
+ * caused live SSH workspaces to be silently wiped on the remote host:
+ *
+ *   1. Forks ALWAYS land at the canonical project root (`<srcBaseDir>/<projectId>/<name>`),
+ *      regardless of where the source workspace's persisted path lives. Legacy
+ *      workspaces from before #3125 sit at `<srcBaseDir>/<basename>/<name>` —
+ *      inheriting their parent dir is what triggered the worktree-add failure
+ *      cleanup that wiped sibling workspaces.
+ *
+ *   2. Every `rm -rf` issued by forkWorkspace targets a per-attempt staging
+ *      directory (`.mux-fork-staging-<12 hex>`), never the final workspace
+ *      path. Even when `git worktree add` or `cp -R -P` leaves a partial,
+ *      cleanup operates exclusively on the unique staging name and therefore
+ *      cannot destroy a sibling workspace.
+ *
+ * The tests drive a real SSHRuntime instance through a mocked `exec()` that
+ * dispatches on substring patterns. We assert on the *exact* command strings
+ * issued, because the safety property we care about is "no rm -rf ever runs
+ * against the final workspace path".
+ */
+
+import { describe, expect, it } from "bun:test";
+import type { ExecOptions, ExecStream, InitLogger, WorkspaceForkParams } from "./Runtime";
+import { SSHRuntime } from "./SSHRuntime";
+import type { SSHRuntimeConfig } from "./sshConnectionPool";
+import type { PtyHandle, PtySessionParams, SSHTransport } from "./transports";
+
+const noop = (): void => undefined;
+const noopAsync = (): Promise<void> => Promise.resolve();
+
+const noopInitLogger: InitLogger = {
+  logStep: noop,
+  logStdout: noop,
+  logStderr: noop,
+  logComplete: noop,
+};
+
+function createMockTransport(config: SSHRuntimeConfig): SSHTransport {
+  return {
+    spawnRemoteProcess() {
+      return Promise.reject(new Error("Unexpected transport use in SSHRuntime fork test"));
+    },
+    isConnectionFailure() {
+      return false;
+    },
+    acquireConnection() {
+      return Promise.resolve();
+    },
+    getConfig() {
+      return config;
+    },
+    createPtySession(_params: PtySessionParams): Promise<PtyHandle> {
+      return Promise.reject(new Error("Unexpected PTY creation in SSHRuntime fork test"));
+    },
+  };
+}
+
+function createTextStream(text: string): ReadableStream<Uint8Array> {
+  const encoded = new TextEncoder().encode(text);
+  return new ReadableStream<Uint8Array>({
+    start(controller) {
+      if (encoded.byteLength > 0) {
+        controller.enqueue(encoded);
+      }
+      controller.close();
+    },
+  });
+}
+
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the cp-fallback finalize nesting race:

- Added a post-hoc nesting detector inside the finalize script. After `mv staging dest`, the script checks `if [ -d <dest>/<stagingName> ]` and, if true (meaning another fork won the race and our `mv` nested into the existing dir instead of failing), removes ONLY the nested `<dest>/<stagingName>` path and reports `MUX_FORK_COLLISION` with exit code 7. The winning fork's destination contents are never touched.
- Fixed a bash-syntax bug from the prior commit: the multi-line `if/then/fi` block was `&&`-joined, which isn't valid. Switched to `;` separation and inlined the body onto a single statement.
- New test `cp-fallback finalize race` pins the structural guarantee that the cleanup `rm -rf` in the finalize script can ONLY target `<dest>/<stagingName>` (the nested form), never `<dest>` itself.

PR #3307, opened 2026-05-18
🤖 fix: preserve plan on soft clear

ThomasK33:
Summary
- Preserve plan files when a workspace context reset is triggered through /clear --soft.
- Keep full history clears responsible for deleting plan files.

Background
Soft clear resets provider-visible context without destroying chat history. Plan files are user-authored working artifacts, so preserving them keeps the plan available after a context reset.

Implementation
- Removed resetContext's plan-file deletion call.
- Added regression coverage proving a context reset leaves the plan file present.
- Updated the test helper to use the shared plan path utility.

Validation
- bun test src/node/services/workspaceService.test.ts -t "context reset"
- make typecheck
- make fmt-check
- make static-check

Risks
Low. The change is scoped to context reset cleanup; full hard clears and explicit replace-history plan deletion still use the existing deletion path.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `1432327{MUX_COSTS_USD:-unknown}`_


ThomasK33:
@codex review

PR #3309, opened 2026-05-18
🤖 fix: enable advisor for default subagents

ThomasK33:
## Summary
Enable the advisor tool by default for Exec and Plan when the Advisor experiment is enabled, including sub-agent runs, while preserving explicit per-agent advisor overrides.

## Background
The Advisor experiment previously required per-agent opt-in before Exec or Plan sub-agents could see the advisor tool. This makes the experiment behave as expected for the default coding/planning agents without changing override semantics.

## Implementation
- Added shared advisor default resolution for Exec and Plan.
- Reused that resolution in agent tool-policy gating, advisor runtime eligibility, and Tasks settings UI state.
- Persisted inherited task experiment state now includes the Advisor flag for queued/restarted sub-agent workspaces.
- Added backend and UI coverage for default-on behavior and explicit disable overrides.

## Validation
- `make static-check`
- `bun test src/node/services/agentResolution.test.ts`
- `bun test src/browser/features/Settings/Sections/TasksSection.ui.test.tsx`
- `bun test src/node/services/aiService.test.ts --test-name-pattern "rebuilds the stream system context when policy removes advisor guidance"`
- `make typecheck`
- `make fmt-check`

## Risks
Low. The change is gated by the Advisor experiment and explicit per-agent overrides still take precedence.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: $24.57_

### DIFF
--- a/src/browser/features/Settings/Sections/TasksSection.tsx
+++ b/src/browser/features/Settings/Sections/TasksSection.tsx
@@ -17,6 +17,7 @@ import { copyToClipboard } from "@/browser/utils/clipboard";
 import { useExperimentValue } from "@/browser/hooks/useExperiments";
 import { getDefaultModel, useModelsFromSettings } from "@/browser/hooks/useModelsFromSettings";
 import { updatePersistedState, usePersistedState } from "@/browser/hooks/usePersistedState";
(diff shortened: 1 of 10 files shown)


ThomasK33:
@codex review

PR #3318, opened 2026-05-18
🤖 feat: preview markdown attach_file outputs

ThomasK33:
## Summary

Adds display-only markdown support to `attach_file`, so generated `.md` files can be previewed and downloaded in chat without being sent back to the model as file attachments.

## Background

Agents can already use file read tools when they need markdown contents. This keeps that boundary intact while making agent-generated markdown files easy for users to inspect and save.

## Implementation

- Treat `.md`, `.markdown`, `.mdown`, and explicit `text/markdown` overrides as display-only attach_file outputs.
- Render display-only markdown with the existing sanitized markdown renderer and keep the existing download affordance.
- Document the attach_file tool behavior as model-attachment support for images/PDFs plus display-only markdown.

## Validation

- `bun test src/node/services/tools/attach_file.test.ts src/browser/features/Tools/AttachFileToolCall.test.tsx`
- `make static-check`
- Dogfooded via Storybook with `agent-browser`: opened `app-chat-tools-attachfile--display-only-markdown`, verified the markdown preview and Download link, and captured screenshot/video evidence in chat.

## Risks

Low. The new path reuses the existing display-only file mechanism, so markdown bytes are stripped from provider requests like other display-only files. Non-markdown text-like files still fail and direct the agent toward file_read.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$10.87`_

### DIFF
--- a/src/browser/features/Tools/AttachFileToolCall.stories.tsx
+++ b/src/browser/features/Tools/AttachFileToolCall.stories.tsx
@@ -103,6 +103,26 @@ export const DisplayOnlyAudio: Story = {
   ),
 };
 
+export const DisplayOnlyMarkdown: Story = {
+  render: () => (
+    <ToolStoryShell>
+      <AttachFileToolCall
+        toolName="attach_file"
+        args={{ path: "release-notes.md" }}
+        result={createAttachFileResult(
+          createDisplayOnlyFilePart({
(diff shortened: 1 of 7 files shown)


ThomasK33:
@codex review

PR #3324, opened 2026-05-18
🤖 fix: harden /btw side-question retry state

ammar-agent:
## Summary

Small follow-up on the `/btw` side-question feature merged in #3293. This keeps the implementation easier to maintain and fixes an idle-chat UI flash where sending `/btw` could momentarily show the main-agent `Stream interrupted` retry banner.

## Background

#3293 landed `/btw` across the backend side-question service, frontend streaming aggregator, transcript rendering, and tests. A focused post-merge cleanup found a few low-risk maintainability wins. While validating the idle `/btw` path, we also found that retry eligibility treated a trailing `/btw` user row like a normal main-agent send with no response yet, which briefly mounted `RetryBarrier` before the side-answer placeholder arrived.

## Implementation

- Inline the dead `WorkspaceStore.processStreamEvent` → `dispatchProcessStreamEvent` wrapper.
- Drop tautological prompt-copy assertions in `sideQuestionService.test.ts`; keep behavioral assertions and rely on the `streamText` test for the tools-denied contract.
- Name `SIDE_QUESTION_MIN_FALLBACK_ATTEMPTS` instead of a bare `3` in the candidate-attempt loop.
- Narrow `/btw`'s AI dependency to `SideQuestionAIService = Pick<AIService, "createModel" | "getStreamInfo">`, removing the test fake's `as unknown as AIService` cast.
- Centralize the `"/btw"` literal as `SIDE_QUESTION_COMMAND` for frontend draft restoration and backend persisted command metadata.
- Realign stale lifecycle comments in `sideQuestionService.ts` with the actual side-answer-aware aggregator/store lookups.
- Keep `/btw` transcript extraction tolerant of persisted history: malformed parts are skipped, current `dynamic-tool` parts are summarized, and legacy AI SDK `tool-<name>` rows from older `chat.jsonl` histories are preserved as `[tool name]` context.
- Add `getLastMainRetryCandidateMessage()` so retry eligibility, `ChatPane` suppression, and `RetryBarrier` copy all skip `/btw` rows consistently while still evaluating any earlier main-agent interruption or stream error.

## Validation

- `make static-check`
- `bun test src/node/services/sideQuestionService.test.ts`
- `bun test src/common/utils/messages/retryEligibility.test.ts`
- `bun test src/browser/features/Messages/ChatBarrier/RetryBarrier.test.tsx`
- `bun run typecheck`

## Risks

Low. The retry behavior change is constrained to retry-barrier eligibility. `/btw` rows are a read-only side branch, not resumable main-agent turns, so they should be skipped when selecting the main retry candidate. The tests cover both sides: idle `/btw` no longer shows retry UI, and an earlier interrupted main response or context-exceeded stream error remains classified through the previous main-agent row even if `/btw` rows appear after it.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$307.29`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -35,6 +35,7 @@ import { computeTaskReportLinking } from "@/browser/utils/messages/taskReportLin
 import { BashOutputCollapsedIndicator } from "@/browser/features/Tools/BashOutputCollapsedIndicator";
 import {
   getInterruptionContext,
+  getLastMainRetryCandidateMessage,
   getLastNonDecorativeMessage,
 } from "@/common/utils/messages/retryEligibility";
 import { TooltipIfPresent } from "@/browser/components/Tooltip/Tooltip";
@@ -818,10 +819,10 @@ export const ChatPane: React.FC<ChatPaneProps> = (props) => {
     workspaceState.autoRetryStatus?.type === "auto-retry-scheduled" ||
     workspaceState.autoRetryStatus?.type === "auto-retry-starting";
 
-  const lastActionableMessage = getLastNonDecorativeMessage(workspaceState.messages);
+  const lastRetryCandidateMessage = getLastMainRetryCandidateMessage(workspaceState.messages);
   const suppressRetryBarrier =
-    lastActionableMessage?.type === "stream-error" &&
-    lastActionableMessage.errorType === "context_exceeded";
+    lastRetryCandidateMessage?.type === "stream-error" &&
+    lastRetryCandidateMessage.errorType === "context_exceeded";
   const shouldMountRetryBarrier = !suppressRetryBarrier;
   const showRetryBarrierUI = showRetryBarrier && !suppressRetryBarrier;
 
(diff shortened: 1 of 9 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Amended to also realign two stale comments in `sideQuestionService.ts` (around the placeholder-before-stream-start emit and the stream-end-before-delete ordering). The old comments referenced a "WorkspaceStore main-agent buffering layer" that doesn't exist; replaced with accurate descriptions of what the side-answer-aware lookups actually guard.

ammar-agent:
@codex review

Updated the PR with a small /btw retry-state fix: trailing side-question user rows and side-answer rows are now excluded from main-agent retry eligibility, preventing an idle-chat `Stream interrupted` flash when `/btw` starts. Local `make static-check` is green.

ammar-agent:
@codex review

Addressed the retry-candidate issue: retry eligibility now skips trailing `/btw` rows and continues evaluating the previous main-agent row, so idle `/btw` stays quiet while an earlier interrupted main stream remains retryable. Added tests for both cases; `make static-check` is green.

ammar-agent:
@codex review

Addressed the context-error suppression issue by exporting `getLastMainRetryCandidateMessage()` and using it consistently in retry eligibility, ChatPane suppression, and RetryBarrier copy. Validation is green locally (`make static-check`, targeted retry tests, targeted RetryBarrier tests, typecheck).

ammar-agent:
@codex review

Addressed the legacy tool-part concern by preserving `tool-<name>` transcript summaries alongside current `dynamic-tool` parts, with a sideQuestionService regression test. Also cleaned up the remaining stale side-question placeholder comment. Local `make static-check` is green.

ammar-agent:
@codex review

Addressed the tolerance concern: `/btw` transcript extraction now guards malformed parts in both text extraction and tool summarization, while preserving legacy `tool-<name>` summaries. Added regression coverage with null/malformed parts plus a legacy tool row. Local `make static-check` is green.

PR #3319, opened 2026-05-18
🤖 feat: add goal intervention policy

ThomasK33:
## Summary

Normal user messages now steer active goals by default instead of auto-pausing them, while keeping an explicit **Send and pause goal** path for intentional pauses.

## Background

Running goals previously treated any manual user message as an interruption and auto-paused the goal. This change lets users steer a goal at the next send boundary without disabling goal continuation, while preserving conservative safety behavior for rejected sends, edits, and explicit pause intent.

## Implementation

- Added a typed `goalInterventionPolicy` send option with `"steer"` and `"pause"` policies.
- Default accepted non-edit manual sends to steering; edit sends and rejected pricing-gate sends remain pause/gated.
- Clear stale pending continuation candidates when accepting a manual intervention so old continuations cannot race ahead of the steering turn.
- Preserve sticky explicit pause policy through queued message batching.
- Added the goal-aware **Send and pause goal** menu action in ChatInput.
- Kept policy metadata out of provider/retry stream options.

## Validation

- `bun test src/node/services/agentSession.goalAutoPause.test.ts src/node/services/messageQueue.test.ts src/node/services/agentSession.budgetGate.test.ts --timeout 30000`
- `TEST_INTEGRATION=1 bun x jest tests/ui/chat/sendModeDropdown.test.ts --runInBand`
- `make static-check`

## Dogfooding

Evidence was captured under `dogfood-output/goal-intervention-policy/` in this workspace, including:

- `screenshots/mock-default-steering-sent.png`
- `screenshots/final-explicit-pause-menu-open-2.png`
- `screenshots/final-explicit-pause-sent-paused.png`
- `videos/final-explicit-pause-success.webm`
- `report.md`

The mock-AI dogfood verified normal steering keeps the goal active and **Send and pause goal** transitions the goal to paused.

## Risks

Goal continuation control is safety-sensitive. The main mitigations are backend defaults as the source of truth, explicit tests for accepted steering versus rejected sends, stale-candidate clearing, and not propagating the policy into provider/retry metadata.

---

<details>
<summary>📋 Implementation Plan</summary>

# Implementation plan: explicit goal intervention policy

## Goal

Let a user steer a running goal by sending a normal chat message **without auto-pausing the goal**, while retaining an explicit way to send a message that intentionally pauses the goal.

This implements Option 2 from the investigation: add an explicit goal-intervention policy to send-message options, with two policies:

[...]

PR #3326, opened 2026-05-18
🤖 fix(goals): treat text-only continuation turns as goal completion

ammar-agent:
## Summary

Fixes a bug in the Goals continuation loop where the agent could finish a `goal_continuation` turn with a plain-text "looks done" reply and no tool calls — leaving the goal in `active`, so the continuation loop would re-fire the same prompt against an idle agent until budget or cooldown gates intervened. We now interpret a tool-less goal-continuation turn as an implicit `complete_goal` call.

## Background

The continuation prompt explicitly asks the agent to call `complete_goal` after a completion audit, but in practice models occasionally just narrate that they're done without invoking the tool. From the workspace's perspective the loop has no signal to stop — `requestContinuationAfterStreamEnd` reads the still-`active` goal and arms another continuation. The user-visible failure mode is repeated, no-op assistant turns that burn budget without making progress.

## Implementation

The hook lives in `AgentSession`'s `stream-end` forward handler (mirroring the existing `extractSwitchAgentResult` precedent for inspecting parts at the same site):

1. After `applyPendingAfterStreamEnd` settles any mid-stream goal mutations, and inside the same guard that arms `goalContinuationRequest`, check whether the just-finished turn was a goal continuation (`activeStreamContext.goalKind === GOAL_CONTINUATION_KIND`).
2. Inspect `streamEndPayload.parts` for any `dynamic-tool` entry. If none, treat the turn as silent.
3. Synthesize a completion summary by lifting the last non-empty text part (trimmed, capped at 500 chars, ellipsised if truncated), falling back to a constant when no usable text exists.
4. Delegate to a new `WorkspaceGoalService.completeGoalFromSilentContinuation({ workspaceId, completionSummary })` which:
   - Re-reads goal state and short-circuits unless status is `active` (paused / complete / missing / budget-limited all fall through).
   - Calls `setGoal({ status: "complete", initiator: "model", expectedGoalId })` so a concurrent clear/replace surfaces as a typed `goal_conflict` rather than a confusing validation error.
   - Logs and swallows typed `Result` errors so the stream-end cleanup path never throws on this best-effort hook.

The subsequent `requestContinuationAfterStreamEnd` call now no-ops on the completed goal (`status !== "active" && status !== "budget_limited"` early-return), so the continuation loop stops without additional plumbing.

### Scope decisions

- **Continuation kind only.** Restricts to `GOAL_CONTINUATION_KIND` so a user's first manual turn answered with text is never mistaken for completion (the manual auto-pause hook already covers that case).
- **`budget_limited` is out-of-scope.** That status owns a one-shot wrap-up message; collapsing it here would race with `tryMarkBudgetLimitInjected`.
- **`initiator: "model"`.** Matches the explicit `complete_goal` tool path so analytics treats both completions as model-driven.

## Validation

- `bun test src/node/services/agentSession.goalAutoPause.test.ts` (16/16 pass, including 4 new tests).
- `bun test` over related goal suites (`workspaceGoalService`, `switchAgent`, `tools/goal`, `idleDispatcher`) — 163/163 pass.
- `make static-check` — green (typecheck + lint + fmt + docs link check).

New tests cover:
- Text-only continuation → goal transitions to `complete` with the synthesized summary and a `goal_completed` analytics event tagged `initiator: "model"`.
- Tool-call continuation (`dynamic-tool` part present) → goal stays `active`.
- Manual user message text-only finish → goal stays `paused` (auto-pause wins; auto-complete does not fire).
- Continuation against a `paused` goal → status unchanged (service helper short-circuits on non-active status).

## Risks

Low-to-moderate, scoped to the goal continuation flow:

- The hook only fires for synthetic `GOAL_CONTINUATION_KIND` turns, so user-driven turns and budget-limit wrap-ups are unaffected.
- Optimistic concurrency via `expectedGoalId` prevents a race with concurrent clear/replace from producing a stale completion.
- All error paths are swallowed (typed Result errors via `setGoal`, plus a defensive `try/catch` around the helper), so a bad goal state cannot break stream-end cleanup.
- Behavioral change: previously a silent continuation looped; now it terminates. If a model emits text-only as an intentional "still thinking" signal, the goal will be marked complete instead of continuing — but the existing continuation prompt explicitly requires tool calls, so any such pattern was already misuse.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `n/a`_

### DIFF
--- a/src/constants/goals.ts
+++ b/src/constants/goals.ts
@@ -7,6 +7,20 @@ export const GOAL_OBJECTIVE_OPEN_TAG = "<untrusted_objective>";
 export const GOAL_OBJECTIVE_CLOSE_TAG = "</untrusted_objective>";
 
 /**
+ * Synthesized completion summary used when the agent ends a goal-continuation
+ * turn with a text-only response (no tool calls). Real models occasionally
+ * finish a continuation with a plain "looks done" reply instead of calling
+ * `complete_goal`; without an implicit completion the continuation loop
+ * would re-fire on the same idle output until budget/cooldown gates
+ * intervene. Prefer the last text part from the turn (truncated to
+ * {@link SILENT_CONTINUATION_COMPLETION_SUMMARY_MAX_LENGTH}); fall back to
+ * this constant when no usable text exists.
+ */
+export const SILENT_CONTINUATION_COMPLETION_SUMMARY_FALLBACK =
+  "Agent ended the goal-continuation turn without calling complete_goal — treated as goal completion.";
+export const SILENT_CONTINUATION_COMPLETION_SUMMARY_MAX_LENGTH = 500;
+
+/**
  * Placeholder shown wherever a user is asked to write a goal objective
  * (right-sidebar form, command palette, queued-goal input). The text is
  * deliberately educational: it teaches what a good agent goal looks like
--- a/src/node/services/agentSession.goalAutoPause.test.ts
+++ b/src/node/services/agentSession.goalAutoPause.test.ts
@@ -398,4 +398,165 @@ describe("AgentSession goal safety hooks", () => {
     expect(execute).not.toHaveBeenCalled();
     session.dispose();
   });
+
+  // Auto-completion fallback for the "agent ended a goal-continuation turn
+  // with a text-only response (no tool calls)" bug. The continuation prompt
+  // asks the agent to call `complete_goal`, but real models sometimes finish
+  // with a plain "looks done" reply instead — without this fallback the
+  // continuation loop would re-fire on the same idle output until budget
+  // or cooldown gates intervene.
+
+  function emitStreamEnd(
+    aiService: AIService & EventEmitter,
+    workspaceId: string,
+    messageId: string,
+    parts: unknown[]
+  ): void {
+    aiService.emit("stream-end", {
+      type: "stream-end",
+      workspaceId,
+      messageId,
+      parts,
+      metadata: {
+        model: "openai:gpt-4o",
+        contextUsage: { inputTokens: 5, outputTokens: 5, totalTokens: 10 },
+        providerMetadata: {},
+      },
+    });
+  }
+
+  test("text-only stream-end during a goal_continuation turn auto-completes the goal", async () => {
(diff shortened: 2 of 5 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the P2 finishReason feedback in 585434704. The silent-completion fallback now requires `metadata.finishReason === "stop"` (matching the same conservatism `TaskService` uses for implicit task-report finalization), so length-truncated / content-filtered / unknown finishes keep the goal active and resume on the next continuation. Added a `length`-truncated test case alongside the existing scenarios.

ammario (approved):

ammar-agent:
@codex review

Rebased onto current `main` to pick up #3319 (goal intervention policy) and the rest of the queue. Conflict was a clean co-location of two adjacent helper methods on `WorkspaceGoalService`; resolved by keeping both `clearPendingContinuationForManualUserMessage` (from main) and `completeGoalFromSilentContinuation` (from this PR).

Adapted one test ("text-only stream-end on a manual user message does not auto-complete the goal") to the new steer-by-default behavior — the goal now stays `active` (not `paused`) on a manual send, and the test still proves the right invariant: the silent-completion gate (`goalKind === GOAL_CONTINUATION_KIND`) short-circuits on manual turns where `goalKind` is undefined.

Local validation: 18/18 tests in `agentSession.goalAutoPause.test.ts` pass, `make static-check` green.

ammar-agent:
@codex review

Addressed the P2 auto-promotion feedback in the latest commit. After a successful `completeGoalFromSilentContinuation`, we now invoke `runDeferredAutoPromoteAfterStreamEnd` — the same hook `applyPendingAfterStreamEnd` already uses for the parallel "agent called `complete_goal` mid-stream" case. It polls `isWorkspaceStreaming` with a bounded backoff and performs the archive + promote when streaming settles.

Added a service-level regression test (`completeGoalFromSilentContinuation promotes the next upcoming goal`) covering the seed-active+upcoming → silent-complete → upcoming-becomes-active transition.

PR #3327, opened 2026-05-19
🤖 fix: prevent SSH base repo runaway tmp packs

ethanndickson:
## Summary

This PR hardens the plain SSH runtime's shared `.mux-base.git` cache so interrupted remote git object transfers cannot silently accumulate unbounded temporary pack files. It adds explicit base-repo health telemetry, age-gated stale `tmp_pack_*` / `tmp_idx_*` cleanup, low-disk headroom checks before object transfer, and compensating cleanup when SSH workspace init is canceled or fails partway through.

## Background

The issue presented as a remote SSH workspace host filling its disk even though the user-facing workspaces were much smaller. The large directory was the SSH runtime's per-project shared bare repository, `<srcBaseDir>/<projectId>/.mux-base.git`, which backs all remote worktrees for a project. In the observed incident, `.mux-base.git` reached roughly 104G, with `objects/pack` containing thousands of Git temporary pack/index files: 2,234 `tmp_pack_*` files and 18 `tmp_idx_*` files totaling about 72G. `git count-objects -vH` emitted repeated `warning: garbage found: ./objects/pack/tmp_pack_*` messages, and the host filesystem reached 100% usage.

The failure mode was especially bad because those files are not normal refs, worktrees, or packs. They are partial files left behind by interrupted `git fetch`, `git push`, or `index-pack` work against the shared bare repo. Once the remote host is low or out of disk, subsequent sync attempts become more likely to fail and leave additional partial packs, creating a feedback loop. Canceling workspace creation can also contribute: init may already be syncing or adding a worktree when the user cancels, and the old cleanup path did not consistently prune base-repo state when the workspace directory was absent or only partially materialized.

This is intentionally fixed in `SSHRuntime`, not in the Coder-specific runtime. A plain SSH runtime pointed at `<name>.coder` uses `SSHRuntime` directly, and the Coder wrapper inherits the same SSH sync/init machinery for shared base-repo operations.

## Implementation

`SSHRuntime` now treats the shared bare repo as a managed cache with lightweight health signals. The health probe still reads Git's pack count, but now also reports pack bytes, temporary pack/index counts and bytes, stale temporary pack/index counts and bytes, and available filesystem bytes. These values are logged around maintenance and transfer decisions so future incidents have direct evidence in normal mux logs instead of requiring manual shell forensics.

Maintenance now handles stale temporary pack files explicitly. `repairBaseRepoForSync()` still strips legacy promisor config and removes `.promisor` marker files before `git gc`, but it now also sweeps `tmp_pack_*` and `tmp_idx_*` files older than a safety age gate. The age gate is important: active Git processes legitimately use these names while fetching or indexing, so Mux only removes stale files that are old enough to be orphaned by an interrupted transfer.

The proactive maintenance trigger now includes stale temporary packs, not just fragmented pack count. Previously the code only repaired when `packs >= 25`, which meant a repo could contain tens of gigabytes of orphaned `tmp_pack_*` files while still looking healthy if the normal pack count was low. Now stale tmp packs are a first-class health signal.

Object transfer paths now check remote disk headroom before doing more work that can write into `.mux-base.git`. Best-effort origin prefetch is skipped when headroom is low, while required git-push and bundle sync paths fail fast with a clear error before starting large transfers. This prevents the low-disk feedback loop from continuing to create more partial pack files.

Workspace init/cancel cleanup is now compensating and generic for SSH. `Runtime` exposes an optional `cleanupFailedInit()` hook, `runBackgroundInit()` invokes it when init fails, and `SSHRuntime` implements it by sweeping stale tmp packs, pruning worktrees, removing the failed workspace branch when safe, and deleting the partial workspace path. `SSHRuntime.deleteWorkspace(force=true)` also invokes this cleanup path even when the workspace directory is already missing, because that is exactly when hidden base-repo state could otherwise be left behind after a canceled create.

## Validation

The focused tests cover the new incident-specific behavior: stale temporary packs trigger maintenance even when pack count is low, tmp-pack cleanup runs before `git gc`, low disk headroom blocks git-push and bundle transfers before upload/push starts, partial SSH init cleanup prunes base-repo/worktree state, and background init failures invoke the runtime cleanup hook. I also ran the required local static checks before pushing.

## Risks

The main risk is deleting a Git temporary pack that is still active. The implementation mitigates that by only deleting `tmp_pack_*` / `tmp_idx_*` files older than the configured safety window, and by limiting this cleanup to maintenance/cancel/failure paths instead of sweeping on every command. Required transfer paths fail fast on low headroom rather than attempting risky cleanup of young files.

Another risk is cleanup removing a useful branch after a failed init. The cleanup only targets the workspace branch name and skips protected trunk-like names. Normal force deletion after an existing worktree was successfully removed prunes worktree metadata and stale tmp packs but does not delete the branch a second time.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$N/A`_

### DIFF
--- a/src/node/runtime/Runtime.ts
+++ b/src/node/runtime/Runtime.ts
@@ -530,14 +530,6 @@ export interface Runtime {
   postCreateSetup?(params: WorkspaceInitParams): Promise<void>;
 
   /**
-   * Optional compensating cleanup after workspace initialization fails or is aborted.
-   * Used by runtimes that create shared cache state during init (for example SSH
-   * base repos and remote worktrees) so canceling creation does not leave hidden
-   * partial state behind.
-   */
-  cleanupFailedInit?(params: WorkspaceInitParams, reason: string): Promise<void>;
-
-  /**
    * Initialize workspace asynchronously (may be slow, streams progress)
    * - LocalRuntime: Runs init hook if present
    * - SSHRuntime: Syncs files, checks out branch, runs init hook
--- a/src/node/runtime/Runtime.ts
+++ b/src/node/runtime/Runtime.ts
@@ -530,6 +530,14 @@ export interface Runtime {
   postCreateSetup?(params: WorkspaceInitParams): Promise<void>;
 
   /**
+   * Optional compensating cleanup after workspace initialization fails or is aborted.
+   * Used by runtimes that create shared cache state during init (for example SSH
+   * base repos and remote worktrees) so canceling creation does not leave hidden
+   * partial state behind.
+   */
+  cleanupFailedInit?(params: WorkspaceInitParams, reason: string): Promise<void>;
+
+  /**
    * Initialize workspace asynchronously (may be slow, streams progress)
    * - LocalRuntime: Runs init hook if present
    * - SSHRuntime: Syncs files, checks out branch, runs init hook
--- a/src/node/runtime/SSHRuntime.retry.test.ts
+++ b/src/node/runtime/SSHRuntime.retry.test.ts
@@ -318,6 +318,20 @@ function expectCommandMatching(
   return index;
 }
 
+function expectPartialWorkspaceCleanupOrder(
+  command: string,
+  workspacePath: string,
+  workspaceName: string
+): void {
+  const removeIndex = command.indexOf(`rm -rf "${workspacePath}"`);
+  const pruneIndex = command.indexOf("worktree prune");
+  const branchIndex = command.indexOf(`branch -D -- '${workspaceName}'`);
+
+  expect(removeIndex).toBeGreaterThanOrEqual(0);
+  expect(pruneIndex).toBeGreaterThan(removeIndex);
+  expect(branchIndex).toBeGreaterThan(pruneIndex);
+}
+
 function isTmpPackCleanupCommand(command: string): boolean {
   return command.startsWith("pack_dir=") && command.includes("tmp_pack_*");
 }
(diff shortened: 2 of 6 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed the P1 by removing generic failed-init cleanup from runtimeFactory and moving SSH cleanup to the materialization boundary, so submodule/init-hook failures preserve already-materialized workspaces.

ethanndickson:
@codex review

Addressed the stale-worktree cleanup ordering issue. Partial SSH materialization cleanup now removes the owned workspace path before `git worktree prune`, then deletes the branch only after Git has had a chance to clear the checked-out worktree registration. Added assertions covering that ordering.

ethanndickson:
@codex review

Addressed the warm-path low-disk gap. The fused warm SSH script now treats remote object transfer as a gated primitive: it realigns origin, checks remote free space inline before `git fetch --quiet origin`, skips that best-effort fetch when headroom is below the shared threshold, and falls back to the cached bundle ref without creating new pack temp files. Added focused coverage asserting the warm origin fetch is behind the disk gate.

ethanndickson:
@codex review

Addressed the post-sync origin fetch headroom gap. `fetchOriginTrunk()` now resolves the remote Git common dir and gates every best-effort origin fetch on the shared object store's free-space policy before running `git fetch origin`, so both fresh post-sync materialization and existing-workspace refreshes skip the pack-writing fetch when disk headroom is below threshold. Added focused coverage for the low-disk post-sync path.

PR #3329, opened 2026-05-19
🤖 tests: fix ExperimentsSection storybook flake

ammar-agent:
## Summary

Fix a flaky `userEvent.clear` + `userEvent.type` sequence in the `ExperimentsSection.ImageGenerationEnabled` Storybook play function, and introduce a shared `replaceInputValue` helper so this class of flake is harder to reintroduce.

## Background

`ExperimentsSection.stories.tsx` exercises the image-generation settings UI by clearing and retyping the **Max images per call** input twice:

```ts
await userEvent.clear(maxImagesInput);
await userEvent.type(maxImagesInput, "11");
// ...
await userEvent.clear(maxImagesInput);
await userEvent.type(maxImagesInput, "2");
await waitFor(() => expect(maxImagesInput).toHaveValue("2"));
```

That input has an `onBlur` handler (`handleMaxImagesBlur` in `ImageGenerationExperimentConfig.tsx`) that normalizes invalid drafts back to `DEFAULT_IMAGE_GENERATION_MAX_IMAGES` (4). Any blur that lands between `clear` (which transiently sets the value to `""`) and the subsequent `type` will reset the draft to `"4"` — so `type("2")` appends to that and you get `"42"`, failing `toHaveValue("2")` non-deterministically.

## Implementation

- Add `replaceInputValue(input, value, expectedValue?)` to `src/browser/stories/storyPlayHelpers.ts`. It focuses the input, selects its current contents, types the new value, and waits for the DOM to settle on the expected value. Because the existing contents are replaced atomically by the next keystroke, there is no transient empty/invalid draft for an interleaved blur to clobber.
- Switch `ExperimentsSection.stories.tsx` to use `replaceInputValue` for both max-images mutations. Drop the dangling `waitFor`/`userEvent.clear`/`userEvent.type` lines those replaced.
- Add a `NOTE` comment beside `handleMaxImagesBlur` pointing future story authors at `replaceInputValue`, so the next person editing this file does not unintentionally reintroduce the racy pattern.

## Validation

- `make static-check` — passes locally.
- `bun test src/browser/features/Settings/Sections/ExperimentsSection.test.tsx` — 9/9 pass.

## Risks

Very low. Production code is touched only by a comment in `ImageGenerationExperimentConfig.tsx`; the rest is Storybook test-only.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$2.21`_

### DIFF
--- a/src/browser/features/Settings/Sections/ExperimentsSection.stories.tsx
+++ b/src/browser/features/Settings/Sections/ExperimentsSection.stories.tsx
@@ -1,5 +1,6 @@
 import { expect, userEvent, waitFor, within } from "@storybook/test";
 import { lightweightMeta } from "@/browser/stories/meta.js";
+import { replaceInputValue } from "@/browser/stories/storyPlayHelpers.js";
 import { EXPERIMENT_IDS } from "@/common/constants/experiments";
 import { DEFAULT_IMAGE_GENERATION_MODEL } from "@/common/types/imageGeneration";
 import { DEFAULT_GOAL_DEFAULTS } from "@/constants/goals";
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the selection-collapse issue you flagged. `replaceInputValue` now uses explicit `userEvent.click` → `setSelectionRange(0, length)` → `userEvent.keyboard(value)` so the selection is set AFTER the only click and is not collapsed by an implicit one. `userEvent.keyboard` dispatches to the focused element without performing its own click, so the selection persists and the typed value replaces the current contents in one shot.

ammar-agent:
@codex review

Both prior threads resolved — the helper now uses explicit `click` → `setSelectionRange(0, length)` → `userEvent.keyboard(value)`. `userEvent.keyboard` dispatches to the focused element without performing its own click, so the selection set on the line before persists and the typed value replaces the entire current contents atomically.

PR #3331, opened 2026-05-19
🤖 feat: add review_pane_update tool + Assisted review toggle

ammar-agent:
## Summary

Adds two new agent tools — `review_pane_update` and `review_pane_get` — that let an agent flag specific code regions in the Review pane for the user to review next. The Review pane gains an "Assisted" toggle that filters to those hunks; flagged hunks are always pinned at the top in agent-declared order, the agent's optional comment is rendered above each hunk, and the hunk body is trimmed to just the flagged new-side line range with the surrounding diff hidden behind a "Show N lines …" affordance. Both new tools get purpose-built transcript renderers instead of the generic JSON-dump fallback. The Review tab auto-focuses on agent-flagged hunks and shows a pulsing Sparkles indicator while there's unread agent focus pending. Tool state is persisted under the workspace session dir so `add` semantics survive backend restarts.

## Background

When an agent makes a large or risky change, the user often wants to know *where to look first*. Today there's no signal from the agent into the review surface — every hunk has equal weight, and a 500-line diff buries the 20 lines that actually matter. This feature gives the agent a first-class way to say "review these regions, in this order, with this context."

## Implementation

**Tools (single source of truth in `toolDefinitions.ts`):**

- `review_pane_update({ operation: "add" | "replace", hunks: [{ path, comment? }] })` — paths use a familiar `path[:start-end]` syntax (`"src/foo.ts"`, `"src/foo.ts:42"`, or `"src/foo.ts:10-20"`, new-side line numbers). `replace` overwrites the set; `add` appends with dedup-by-`path:range` (preferring the latest comment).
- `review_pane_get()` — returns the current set in declared order so the agent can inspect before adding more.

**State model:**

- Backend tool state is persisted under `<workspaceSessionDir>/assistedReview.json` (parallel to `todos.json`), with writes serialized via `workspaceFileLocks` and `writeFileAtomic`. The file is unlinked on a clear so a stale file can't shadow a fresh start. `coerceAssistedReviewHunks` survives corrupted JSON per the self-healing doctrine — bad data is treated as empty rather than thrown.
- `StreamingMessageAggregator.processToolResult` parses each successful `review_pane_update` output and rebuilds `assistedReviewHunks` on the frontend. Because the tool returns the *resulting* list directly, replaying history naturally reconstructs final state on reload.
- Exposed via `WorkspaceStore.getAssistedReviewHunks(workspaceId)`, consumed by `ReviewPanel` through `useSyncExternalStore`.

**Review pane UI:**

- One shared module `src/common/utils/review/assistedReview.ts` owns parsing, matching, and formatting — reused by the tool handler, aggregator, and panel.
- `applyFrontendFilters` gains an optional `isAssisted` predicate so the existing read/search pipeline is untouched.
- `ReviewPanel`'s `filteredHunks` memo applies the existing sort, then partitions into `[assistedBucket, rest]` preserving agent-declared index, so assisted hunks are always pinned first regardless of the toggle.
- `ReviewControls` shows an "Assisted ✦ (N)" toggle only when `assistedCount > 0`, keeping the toolbar compact for normal sessions.
- `HunkViewer` renders the comment in an accent-bordered row above the diff header; a hunk flagged without a comment gets a subtle left-border accent.

**Range trimming + show-more:**

- New helper `sliceHunkByNewLineRange` (`src/browser/utils/review/sliceHunkContent.ts`) walks the diff body using the same algorithm as `groupDiffLines`, partitions lines into `before` / `inside` / `after` (attaching `-` lines to the next visible region; trailing `-` lines to the previous), and returns `null` for no-op cases (full coverage, out-of-bounds, pure deletions, pure renames) so the existing single-renderer path still applies.
- `HunkViewer` now renders the inside slice always; before/after stay hidden behind a `ContextCollapseIndicator` that gained a `mode: "expand" | "collapse"` prop, so "Show N lines above" and "Collapse N lines above" share the same squiggle pattern.
- Agent-flagged hunks override the auto-collapse behavior that kicks in for large hunks / heavy reviews — otherwise the assistance signal got buried under "Click to expand".

**Chat-transcript renderers:**

- `ReviewPaneUpdateToolCall.tsx` (≈120 LOC) — collapsed header shows the operation (Add / Replace / Cleared) + "N hunks pinned" + rejected count when present. Expanded view lists each path:range with a Sparkles-accented comment row. Prefers the *post-merge* result list so dedup outcomes are visible.
- `ReviewPaneGetToolCall.tsx` (≈45 LOC) — single-line preview ("Inspected N pinned hunks"); no expand affordance because the full list lives in the Review pane.
- Both are registered in `TOOL_REGISTRY` (`getToolComponent.ts`); `TOOL_NAME_TO_ICON` maps `review_pane_update` → `Sparkles` and `review_pane_get` → `ScanEye`.

**Focus mode + Review tab pizzazz:**

- The Assisted toggle now auto-flips ON the first time the agent flags hunks in a session, so the Review pane lands directly on the critical changes instead of the full diff. A ref-guarded latch makes the user's subsequent manual toggles stick; a drop-to-zero (agent cleared its hint set) re-arms the latch so the next batch of flagged hunks gets focus again.
- `ReviewStats` gains an `unreadAssisted` field — the count of agent-flagged hunks intersecting the current diff that the user hasn't acked. Computed in `ReviewPanel` alongside the existing `readHunkCount` summary and pushed through the unchanged `onStatsChange` callback.
- `ReviewTabLabel` renders a Sparkles pill in `--color-review-accent` with `animate-pulse` when `unreadAssisted > 0`, alongside the existing `read/total` badge. Stops pulsing once everything assisted is marked read. Tooltip explains "N agent-flagged hunk(s) pending review."

Matching is exact on workspace-relative paths (with `oldPath` fallback for renames), and overlap-based on new-side line numbers (old-side fallback for pure deletions), so the syntax does what an agent would naturally expect.

## Validation

- 42 new unit tests:
  - 12 for the parser/matcher (`assistedReview.test.ts`)
  - 6 for `applyReviewPaneUpdate` semantics (add/replace, dedup, comment normalization, rejection of malformed filters, clear-via-empty-replace)
  - 7 for `sliceHunkByNewLineRange` (whole-hunk coverage, out of bounds, pure deletions, context-only split, deletion grouping, empty-inside short-circuit, boundary cases)
  - 5 for `ReviewTabLabel` pizzazz indicator (presence/absence gating, accent + pulse classes, singular/plural aria-labels, null-stats safety)
  - 4 for `coerceAssistedReviewHunks` (non-arrays, missing/invalid path, malformed ranges, empty-string comment)
  - 4 for end-to-end persistence (restart-survives-state, clear-unlinks-file, per-workspace isolation, corrupted-file self-heal)
- Full `make test` deltas: +42 passing tests, no new failures. The 21 `GitStatusStore` and 1 `NestedToolRenderer` cross-test interference failures observed in full-suite runs are **pre-existing on baseline `main`** and reproduce without these changes.

## Risks

Low. The new state is opt-in and entirely additive:

- The Assisted toggle, per-hunk comment row, range trimming, auto-focus latch, and tab pizzazz pill are all conditional on the agent having flagged something, so users who never see the tool fire get the unchanged Review pane and tab strip.
- The slicer is a pure transform and returns `null` for any edge case it doesn't fully handle, falling back to the unchanged full-hunk render path.
- Ordering changes only when at least one hunk matches — the existing file-order / last-edit sort still owns the rest of the list.
- The tools only persist a small JSON file under the workspace session dir; no git or other filesystem side effects.
- `ContextCollapseIndicator`'s new `mode` prop defaults to `"collapse"`, preserving existing call-site behavior.
- The auto-on Assisted effect is ref-latched so it won't churn against user toggles; it only re-arms when the agent clears the hint set entirely.
- File reads self-heal from corruption (treated as empty), so a hand-edited / partial-write `assistedReview.json` can't brick the tool.

Touched files outside the feature scope: `StreamingMessageAggregator.processToolResult` (added a guarded branch alongside `todo_write`), `applyFrontendFilters` (added an optional predicate), `ContextCollapseIndicator` (additive `mode` prop), and `ReviewStats` (additive `unreadAssisted` field). All are narrowly scoped and covered by existing or new tests.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$26.13`_


ammar-agent:
@codex review

ammar-agent:
@codex review

Two new changes pushed in `0665c04be`:
1. Custom transcript renderers for `review_pane_update` (collapsed: operation + N pinned; expanded: list with agent comments) and `review_pane_get` (single-line preview).
2. Review pane now trims agent-flagged hunks to just the new-side range the agent declared, hiding the rest behind a "Show N lines above/below" affordance (the existing `ContextCollapseIndicator` gained a `mode: "expand"` variant).

New helper `sliceHunkByNewLineRange` walks the diff body and returns `null` for no-op cases (full coverage, pure deletions, pure renames, out-of-bounds) — those still render the full hunk unchanged. 7 new unit tests cover the slicer.

ammar-agent:
@codex review

Addressed the previous comment in `e3036bf36`: the Assisted predicate is now applied whenever the toggle is on, even with an empty match map, so the mismatch case produces the empty state instead of "show everything". Added tailored copy distinguishing "no pinned hunks yet" from "pinned set doesn't match current diff".

ammar-agent:
@codex review

Latest commit (`73ee25850`) adds two focus affordances on top of the previous review-pane work:

1. **Auto-focus the Assisted toggle** in `ReviewPanel` the first time the agent flags hunks in a session (ref-latched so manual toggles stick; re-arms when the agent clears).
2. **Pizzazz Sparkles pill on the Review tab** when there are unread agent-flagged hunks (`unreadAssisted` count on `ReviewStats`, animate-pulse + `text-review-accent`).

5 new tests in `TabLabels.review.test.tsx` cover presence/absence gating, accent/pulse classes, singular/plural aria-labels, and null-stats safety.

ammar-agent:
@codex review

Addressed the P2 finding in `2ef04f61d`:

- New `<workspaceSessionDir>/assistedReview.json` (parallel to `todos.json`) persists the agent's pinned set across backend restarts.
- `createReviewPaneUpdateTool` reads prior state from disk before applying add/replace and persists the result before returning so a successful tool result is always disk-backed.
- The file is unlinked on clear so a stale file can't shadow a fresh start.
- Defensive `coerceAssistedReviewHunks` survives corrupted JSON per the self-healing doctrine in AGENTS.md.

8 new tests (4 coerce cases + 4 end-to-end: restart-survives-state, clear-unlinks-file, per-workspace isolation, corrupted-file self-heal). Frontend transcript-replay path is unchanged — the new disk persistence is purely a backend-tool consistency fix.

Resolved thread `PRRT_kwDOPxxmWM6DNiRa`.

ammar-agent:
@codex review

Latest commit (`b723453fe`) refines the Review tab indicator based on user feedback:

- Dropped `animate-pulse` — too noisy next to other tab labels.
- Dropped the pill background + padding (`bg-review-accent/15` + `py-px` + `rounded-sm`) — those broke the parent strip's `items-baseline gap-1.5` rhythm and visibly offset the indicator from siblings.
- Now: the "Review" word itself is tinted `text-review-accent` when there are unread agent-flagged hunks (mirrors the Goal tab's `text-success`/`text-warning` pattern), and the Sparkles + count renders as plain inline-flex baseline-aligned text in the same accent.

Updated the relevant test to explicitly guard against regressions to `animate-pulse` / `bg-review-accent` / `rounded` classes.

ammar-agent:
@codex review

Latest commit (`f5a8c9c1e`) addresses two UX issues the user reported on the previous tab-label iteration:

1. **Digit misaligned with Sparkles icon.** Root cause: three separately-baselined siblings (`Review` word, Sparkles+count pill, `4/10` badge) all running through the parent strip's `items-baseline gap-1.5` layout. The icon's box-center and the digit's text baseline are different vertical anchors, so they visibly drifted. Fix: collapse the whole label into one `inline-flex items-center` group (Review · Sparkles · count) — same composition pattern as the Goal tab's `<Target>` + label — so icon + digit + word share one alignment context.

2. **Hide the `read/total` backup badge while the assisted indicator is showing.** Two adjacent numbers (`Review ✦ 3 4/10`) were hard to parse and the assisted count is the user's primary cue. The badge comes back once everything assisted is read.

Tests updated to assert `items-center` + `inline-flex` on the indicator (regression guard for the alignment fix) and `queryByText("4/10")` is null in the assisted state.

ammar-agent:
@codex review

Addressed the P1 race in `70af6aeb1`:

- The full `review_pane_update` read/apply/write sequence now runs inside one `workspaceFileLocks.withLock(...)` critical section, not just the write.
- The low-level file write helper is intentionally lock-free to avoid nested same-key lock deadlocks.
- Added a concurrent-add regression test: 12 simultaneous `operation: "add"` calls against the same workspace/session dir, then `review_pane_get` must return all 12 pinned hunks.

Local validation: `bun test src/node/services/tools/review_pane.test.ts` (15 pass) and `make static-check` passed.

Resolved thread `PRRT_kwDOPxxmWM6DOHbf`.

ammar-agent:
@codex review

Addressed the typography issue in `f02b60f15`:

- Replaced the Review tab assisted unread count's `tabular-nums` with the semantic `counter-nums` utility, per AGENTS.md's numeric UI rule.
- Extended `TabLabels.review.test.tsx` to assert `counter-nums` is present and `tabular-nums` is absent on that changing count.

Local validation: `bun test src/browser/features/RightSidebar/Tabs/TabLabels.review.test.tsx` and `make static-check` passed.

Resolved thread `PRRT_kwDOPxxmWM6DOcpy`.

ammar-agent:
@codex review

Addressed the stale assisted-badge issue in `8633228de`:

- Added `ReviewAssistedStatsReporter`, mounted from `RightSidebar` outside the active-tab panel tree, so the Review tab attention badge updates even when `ReviewPanel` is not selected/mounted.
- The reporter subscribes to `WorkspaceStore.getAssistedReviewHunks(workspaceId)` and `useReviewState(workspaceId)`, then loads/matches the current full diff only when assisted hunks exist.
- Shared `parseReviewDiffCacheValue` between the panel and reporter so path reprojection semantics stay identical.
- Shared `countUnreadAssistedHunks` between the panel and reporter so unmatched assisted filters and read hunks are handled the same way in both paths.
- Changed `useReviewState` to `{ listener: true }` so the always-mounted reporter observes read-state updates made by the panel.
- Added `ReviewPanel.assistedStats.test.ts` for count semantics (matched unread only, range overlap only).

Local validation: `bun test src/browser/features/RightSidebar/CodeReview/ReviewPanel.assistedStats.test.ts`, `bun test src/browser/features/RightSidebar/Tabs/TabLabels.review.test.tsx src/node/services/tools/review_pane.test.ts`, and `make static-check` passed.

Resolved thread `PRRT_kwDOPxxmWM6DOkhr`.

PR #3340, opened 2026-05-20
🤖 tests: stabilize flaky signing and best-of task tests

ammar-agent:
## Summary

Two unit tests have been intermittently timing out at exactly the `bun:test` default 5000ms budget in CI, even though both pass in well under 1s locally. Bumps each test's timeout to 15s with an explanatory comment about the source of the slowness. Pure tests-only change.

- `SigningService > with Ed25519 key > should load key and return capabilities`
- `TaskService > agent_report finalizes interrupted best-of parent output after partial best-of spawn failure`

## Background

`Test / Unit` started failing on unrelated PRs with these two tests reporting `[5001.02ms]` and `[5002.02ms]` — i.e. they were actively running when bun's per-test 5s deadline fired. Local runs of either test individually finish in 250ms–1s.

### `SigningService` test

The test only asserts that `capabilities.publicKey` starts with `"ssh-ed25519 "`, but `getCapabilities()` internally calls `detectIdentity()`, which shells out to `gh auth status 2>&1` via `execAsync` (`src/node/services/signingService.ts` line 533). I confirmed on this machine that `gh auth status` alone takes ~4.7s, which leaves no headroom under CI load.

A real production-side fix would be to add a timeout to that subprocess (so a slow `gh` never blocks the app), but that touches the shared `execAsync` helper and is a behavior change. I'm deliberately deferring that to a follow-up so this PR stays tests-only and low-risk.

### `TaskService` best-of test

The test body exercises real config, history, and partial-on-disk I/O across many sequential `await`s (`saveWorkspaces`, `writePartial`, `finalizeReportedChildTaskForTest`, multiple `readPartial`s, etc.). Locally it lands ~250ms; in parallel CI runs, disk contention from the other ~7800 tests blows past the 5s budget. No single slow call jumped out from profiling locally, so the pragmatic fix is to give this test enough slack.

## Implementation

- `signingService.test.ts`: convert the single failing `it(...)` to the three-argument form with `{ timeout: 15_000 }` and a comment pointing at `gh auth status` as the culprit.
- `taskService.test.ts`: same treatment for the single failing `test(...)`, with a comment about CI disk contention.

Both use the `{ timeout: 15_000 }` option form already established in `backgroundProcessManager.test.ts` and the code-execution integration tests.

## Validation

- `bun test src/node/services/signingService.test.ts src/node/services/taskService.test.ts` — 137/137 pass locally.
- `make static-check` — passes.

## Risks

Tests-only change with the maximum possible scope of a longer wait when the test would have failed under the old budget. The two adjusted tests still complete in well under 1s under normal conditions, so the new headroom only matters when CI is genuinely slow; in steady state nothing changes.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `max` • Cost: `$10.11`_

### DIFF
--- a/src/node/services/signingService.test.ts
+++ b/src/node/services/signingService.test.ts
@@ -97,14 +97,23 @@ describe("SigningService", () => {
   });
 
   describe("with Ed25519 key", () => {
-    it("should load key and return capabilities", () =>
-      withoutSshAgent(async () => {
-        const service = new SigningService([ed25519KeyPath]);
-        const capabilities = await service.getCapabilities();
+    it(
+      "should load key and return capabilities",
+      () =>
+        withoutSshAgent(async () => {
+          const service = new SigningService([ed25519KeyPath]);
+          const capabilities = await service.getCapabilities();
 
-        expect(capabilities.publicKey).toBeDefined();
-        expect(capabilities.publicKey).toStartWith("ssh-ed25519 ");
-      }));
+          expect(capabilities.publicKey).toBeDefined();
+          expect(capabilities.publicKey).toStartWith("ssh-ed25519 ");
+        }),
+      // SigningService.getCapabilities() shells out to `gh auth status` to detect
+      // GitHub identity. `gh` can take several seconds (network/credential lookup),
+      // so under CI load this comfortably exceeds Bun's default 5s per-test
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3341, opened 2026-05-20
🤖 fix: retry agent sidebar status updates

ammar-agent:
## Summary
Fix flaky agent sidebar status updates by giving provider-side status-generation misses immediate same-transcript retries, followed by bounded cooldown retries so unchanged transcripts can still recover after transient provider/model issues.

## Background
The sidebar status generator previously deduped a transcript immediately after any post-provider failure, including cases where the small model simply missed the required `propose_status` call once. That made status updates appear flaky because the workspace would not retry until the chat transcript changed. The first patch added bounded immediate retries; this update closes the remaining recovery gap by continuing to retry on cooldown instead of permanently settling provider-side failures.

## Implementation
- Added `AGENT_STATUS_PROVIDER_FAILURE_RETRY_ATTEMPTS` for immediate retries on unchanged transcripts.
- Added active/idle provider-failure cooldown constants so active streams recover faster while idle chats back off harder for cost control.
- Tracked provider-side failure counts and retry-after timestamps per transcript+streaming hash in `AgentStatusService`.
- Reset provider failure tracking on successful settle paths and config/pre-provider paths so stale retry counters do not leak across inputs.
- Updated regression coverage to prove provider failures retry first, pause during cooldown, then recover on the same unchanged transcript without a new user turn.

## Validation
- `bun test src/node/services/agentStatusService.test.ts --timeout 20000`
- `make typecheck`
- `make lint`
- `make static-check`
- Reproduced unrelated transient CI failure locally: `TEST_INTEGRATION=1 bun x jest tests/ui/chat/sendModeDropdown.test.ts --runInBand --detectOpenHandles`

## Risks
Low-to-medium: this only changes the background AI-generated sidebar status path, but repeated provider-side failures can now retry indefinitely on cooldown. The retry budget, active/idle cooldown split, exponential backoff, and max cooldown bound provider cost while preserving eventual recovery.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `off` • Cost: `1050807{MUX_COSTS_USD:-unknown}`_

### DIFF
--- a/src/constants/agentStatus.ts
+++ b/src/constants/agentStatus.ts
@@ -28,6 +28,24 @@ export const AGENT_STATUS_IDLE_UNFOCUSED_INTERVAL_MS = 2 * 60 * 1000;
  */
 export const AGENT_STATUS_TICK_INTERVAL_MS = 10 * 1000;
 
+/**
+ * Retry budget for provider-side status generation failures on an unchanged
+ * transcript. A single missed tool call or transient provider hiccup should not
+ * freeze the sidebar until the next chat turn. After that, retry on a bounded
+ * cooldown so transient provider outages still recover without hammering the
+ * small-model path indefinitely.
+ */
+export const AGENT_STATUS_PROVIDER_FAILURE_RETRY_ATTEMPTS = 2;
+
+/**
+ * Cooldowns after the immediate provider-failure retry budget is exhausted.
+ * Active streams recover faster because visible status freshness matters most;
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Updated the patch so provider-side status failures do not permanently settle after the immediate retry budget. They now pause on an active/idle cooldown and can recover on the same unchanged transcript.

PR #3344, opened 2026-05-20
🤖 fix: surface assisted review hunks reliably

ammar-agent:
## Summary

Makes Assisted Review visibility parity-by-construction: once `review_pane_update` accepts a pin, the Review sidebar fetches the pinned paths from their owning repo roots and ignores stale user filters that could otherwise hide those hunks.

## Background

Agents can flag specific review hunks, but the sidebar could still render an empty Assisted view when unrelated UI state was active: a previously-selected file path, hidden read hunks, a search term, an off include-uncommitted toggle, or a multi-project workspace rooted to a different repository. That made successful tool calls appear as if they had no sidebar effect.

## Implementation

- Builds Review diff pathspecs from the assisted pin set while Assisted mode is active instead of using a stale selected-file path.
- Groups assisted pathspecs by owning project repo root in multi-project workspaces, so accepted pins in secondary projects are fetched from the right checkout.
- Forces Assisted diff loads to include uncommitted changes, since agent pins commonly target latest working-tree edits.
- Bypasses search and read filters while Assisted mode is active, preserving those filters outside Assisted mode.
- Keeps agent-pinned hunks expanded even if they are already read.
- Adds focused tests for Assisted pathspec, multi-project repo-root, include-uncommitted, and frontend-filter behavior.

## Validation

- `bun test src/browser/features/RightSidebar/CodeReview/ReviewPanel.assistedStats.test.ts src/common/utils/review/assistedReview.test.ts src/node/services/tools/review_pane.test.ts`
- `make typecheck`
- `make lint`
- `make static-check`

## Risks

Low-to-moderate: the change is scoped to Review sidebar Assisted mode, but it intentionally overrides user visibility filters while Assisted is enabled so accepted agent pins cannot be hidden.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$6.49`_

### DIFF
--- a/src/browser/features/RightSidebar/CodeReview/HunkViewer.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/HunkViewer.tsx
@@ -247,11 +247,11 @@ export const HunkViewer = React.memo<HunkViewerProps>(
     const hasManualState = hunkId in expandStateMap;
     const manualExpandState = expandStateMap[hunkId];
 
-    // Agent-flagged hunks should default to expanded even when they're "large"
-    // or in a heavy review where everything else is collapsed — otherwise the
-    // assisted-review focus signal gets buried.
+    // Agent-flagged hunks should default to expanded even when they're already
+    // read, "large", or in a heavy review where everything else is collapsed —
+    // otherwise the assisted-review focus signal gets buried.
     const shouldAutoExpand =
-      !isRead &&
+      (!isRead || isAssisted) &&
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Please take another look. I grouped Assisted-mode diff requests by each pinned path's owning repo root so multi-project pins are fetched from the correct checkout.

PR #3345, opened 2026-05-20
🤖 fix: release settled /btw transcript holds

ammar-agent:
## Summary

Fixes `/btw` side-question scroll holds so the side Q&A can no longer remain stuck to the bottom of the transcript as permanent visual clutter.

## Background

`/btw` intentionally hands scroll ownership from bottom-lock to the side branch when main-agent output continues below the aside. The previous hold kept re-aligning through the answer lifecycle, but did not distinguish a real start-aligned branch from a browser bottom-clamp caused by insufficient content below it.

## Implementation

- Detect whether `scrollIntoView({ block: "start" })` is still bottom-clamped using transcript/target geometry.
- Keep the temporary side-question hold only while the side answer is streaming or while the attempted start alignment is still clamped to the bottom.
- Release settled holds once the branch can align at the transcript viewport start.
- Preserve fallback answer-row handling for side answers whose user row is unavailable.

## Validation

- `bun test src/browser/components/ChatPane/sideQuestionScrollHold.test.ts`
- `bun test src/browser/utils/messages/StreamingMessageAggregator.sideQuestion.test.ts src/browser/stores/WorkspaceStore.test.ts --test-name-pattern "/btw|side-question|Side question|side answer"`
- `bun run typecheck`
- `make fmt-check`
- `make lint`
- `make static-check`

## Risks

Low-to-moderate. The change touches transcript scroll ownership, but it is limited to `/btw` side-question hold release logic and covered by targeted unit tests for streaming, settled, fallback-row, and geometry cases.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$7.88`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -89,6 +89,8 @@ import {
 import {
   findActiveSideQuestionScrollHoldTarget,
   findSideQuestionScrollHoldTarget,
+  getSideQuestionScrollHoldScrollportStartTop,
+  isSideQuestionScrollHoldBottomClamped,
   type SideQuestionScrollHoldState,
 } from "./sideQuestionScrollHold";
 import { recordSyntheticReactRenderSample } from "@/browser/utils/perf/reactProfileCollector";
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review
Please take another look. I updated the clamp check to compare against the transcript scrollport padding edge instead of the container border top, and added coverage for padded alignment.

PR #3346, opened 2026-05-21
🤖 fix: detach SSH base repo gc

ethanndickson:
## Summary

Detach shared SSH base-repo gc from mux's SSH command timeout so maintenance can keep long-lived bare repos defragmented without mux killing Git mid-pack-finalization.

## Background

The existing base repo maintenance path ran `git gc --prune=now` through mux's timed SSH command wrapper. On large shared bare repos, that timeout can fire while Git is finalizing packs, leaving a `.pack` without its matching `.idx`. Git then cannot see the objects in that pack, and later sync pushes can fail with unresolved remote deltas.

## Implementation

This keeps the existing maintenance triggers and retry flow, but runs gc through `setsid` so the actual Git process is in its own session/process group. Proactive preflight uses `setsid -f` and only waits for the remote spawn because that maintenance is advisory. Retry cleanup uses `setsid -f -w`, which waits for the detached child to exit and returns its status, so the next retry can benefit from completed maintenance without putting gc itself under mux's kill timeout. Missing `setsid` is checked before redirecting gc output, so local logs preserve actionable spawn diagnostics; gc's own output still goes to `/tmp/.mux-base-repo-gc.log` on the SSH host.

## Risks

The behavior depends on util-linux `setsid` with `-f`/`-w` support on the SSH host. If it is missing, maintenance logs a warning and sync proceeds through the existing best-effort path; correctness is preserved, but base-repo defragmentation may not run on that host.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$29.50`_

### DIFF
--- a/src/node/runtime/SSHRuntime.retry.test.ts
+++ b/src/node/runtime/SSHRuntime.retry.test.ts
@@ -206,6 +206,9 @@ class CleanupCommandSSHRuntime extends SSHRuntime {
       return Promise.resolve(createExecStream(this.promisorStdout));
     }
     if (command.includes(" gc --prune=now")) {
+      // Spawn-only call: production code returns as soon as the remote shell
+      // has detached `git gc`. The mock mirrors that shape (stdout/stderr/exit
+      // here represent the detached spawn, not gc itself).
       return Promise.resolve(createExecStream(this.gcStdout, this.gcStderr, this.gcExitCode));
     }
     return Promise.resolve(createExecStream(""));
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
Addressed Codex feedback:

- `PRRT_kwDOPxxmWM6DrCEC`: retry cleanup now uses `setsid -f -w`, so it waits for the detached gc child to finish before retrying while still keeping the gc process outside mux's kill-timeout process group.
- `PRRT_kwDOPxxmWM6DrCEF`: the command now checks `command -v setsid` before redirecting gc output to the remote log, preserving actionable spawn diagnostics in mux logs.

@codex review

PR #3351, opened 2026-05-21
🤖 fix(goals): keep live goal costs in sync with Stats

ammar-agent:
## Summary

Goal-tab cost (and other surfaces reading `goal.costCents`) now tracks the Stats/Costs tabs mid-stream. The fix removes the normal per-`usage-delta` disk write from the live cost preview path, adds a safe fallback for missing activity baselines, preserves the live "budget used" overlay when the user edits only the budget/turn limit mid-stream, and serializes failed-stream preview cleanup before restoring durable state.

## Background

Stats/Costs and Goal cost both ultimately derive from the same `usage-delta` events, but their delivery paths diverged:

- **Stats / Costs tab (instant):** `usage-delta` → frontend `StreamingMessageAggregator.handleUsageDelta` → bumps `usageStore` → `CostsTab` reads `liveCostUsage` straight from the aggregator.
- **Goal cost (slow round-trip):** `usage-delta` → backend `previewGoalAccountingFromUsage` → `workspaceGoalService.previewStreamAccounting` → `pushSnapshot` → `extensionMetadata.setGoal` → `writeFileAtomic` on the shared `extensionMetadata.json` (serialized through `withSerializedMutation`) → ORPC `workspace.activity` event → renderer `WorkspaceStore.applyWorkspaceActivitySnapshot` updates `goal.costCents`.

Mid-stream the Goal tile lagged Stats by the round-trip cost. Every workspace's preview also contended on the same global mutation lock on `extensionMetadata.json`.

Separately, editing only the budget limit during a live stream could make "budget used" appear to reset to the pre-stream durable value. That happened because the budget edit correctly wrote the durable `goal.json` value (before the in-flight stream is finally accounted), but the UI activity snapshot lost the live preview overlay until the next preview/final accounting event.

## Implementation

- `previewStreamAccounting` now prefers `pushTransientGoalSnapshot`. The renderer already handles `transientGoalOnly: true` snapshots via `WorkspaceStore.applyWorkspaceActivitySnapshot`, merging the preview `goal` into the cached activity without persisting it.
- If there is no baseline activity snapshot (metadata reset/corruption or stream-start metadata write race), preview delivery falls back to `pushSnapshot` so the first preview still creates and emits a workspace activity entry rather than being dropped.
- `WorkspaceGoalService` tracks the latest live preview snapshot in memory and re-emits it as a transient overlay after in-place mutable edits (budget / turn cap / inline rename). This keeps the UI's "used" amount aligned with Stats while leaving `goal.json` at the pre-stream durable base so stream-end accounting cannot double-count.
- Terminal stream errors call `restoreGoalAccountingSnapshot` with the active stream start time. That method runs under the same workspace lock as preview accounting, clears the cached preview, marks that failed stream as recorded, and re-emits the durable goal snapshot. Late queued usage deltas from the failed stream are ignored instead of repopulating stale preview state.
- No change to `recordStreamAccounting`: stream-end still writes both `goal.json` and `extensionMetadata.json` and emits the durable activity event.

Tests updated to observe normal previews through the activity listener queue (`captureGoalActivity` / `setOnActivityChange`) instead of inspecting `extensionMetadata.getSnapshot`. The normal contract is: **preview is transient; durable record stays at `goal.json`**. The fallback test explicitly covers the no-baseline case where a one-time activity write is still needed for delivery.

## Validation

- `bun test src/node/services/workspaceGoalService.test.ts src/node/services/agentSession.goalAutoPause.test.ts src/browser/features/RightSidebar/GoalTab.test.tsx src/common/utils/goals/budgetPricing.test.ts src/browser/utils/goals/resolveGoalSetIntent.test.ts` — 189 pass / 0 fail.
- `bun test src/node/services/agentSession` (15 files) — 158 pass / 0 fail.
- `make typecheck` — green.
- `make lint` — green.
- `make fmt-check` — green.
- `make static-check` — green.

## Risks

Low-to-medium. The change is scoped to the preview/activity emit path:

- Final accounting (`recordStreamAccounting`) is unchanged, so durable budget enforcement, budget-limited gating, and restart recovery remain anchored on `goal.json`.
- The transient-activity-emit pattern is the same one already used by `publishPendingGoalSnapshot` for queued goal replacements, so the renderer's `applyWorkspaceActivitySnapshot` path is already well-exercised.
- A fallback still uses the durable activity path when no baseline activity snapshot exists, addressing the race where stream-start metadata has not populated `extensionMetadata` yet.
- In-memory preview overlays are cleared on stream accounting, stop/abort cleanup, terminal stream-error restore, replacement, clear, and persisted-snapshot restore to avoid stale overlays after the stream has settled.

---

### DIFF
--- a/src/node/services/agentSession.goalAutoPause.test.ts
+++ b/src/node/services/agentSession.goalAutoPause.test.ts
@@ -268,16 +268,40 @@ describe("AgentSession goal safety hooks", () => {
       objective: "Restore preview on error",
       budgetCents: 1_000,
     });
+    // Capture goal-related activity snapshots. `previewStreamAccounting`
+    // is transient and intentionally does NOT touch
+    // extensionMetadata.json or goal.json — the UI receives the preview
+    // through the activity stream instead, which we observe here.
+    const activitySnapshots: Array<{
+      goal?: { costCents?: number } | null;
+      transientGoalOnly?: boolean;
+    }> = [];
+    goalService.setOnActivityChange((_workspaceId, snapshot) => {
+      activitySnapshots.push(snapshot);
+    });
+    const failedStreamStartedAtMs = created.createdAtMs + 1;
     await goalService.previewStreamAccounting({
       workspaceId,
       costUsd: 1.25,
-      streamStartedAtMs: created.createdAtMs + 1,
+      streamStartedAtMs: failedStreamStartedAtMs,
     });
-    expect(await extensionMetadata.getSnapshot(workspaceId)).toMatchObject({
+    expect(activitySnapshots.at(-1)).toMatchObject({
+      transientGoalOnly: true,
       goal: { costCents: 125 },
+    });
+    // The durable record stays untouched by the preview.
+    expect(await extensionMetadata.getSnapshot(workspaceId)).toMatchObject({
+      goal: { costCents: 0, budgetCents: 1_000 },
     });
 
     aiService.streamMessage = mock(() => {
+      aiService.emit("stream-start", {
+        type: "stream-start",
+        workspaceId,
+        messageId: "assistant-stream-error",
+        model: SEND_OPTIONS.model,
+        startTime: failedStreamStartedAtMs,
+      });
       aiService.emit("error", {
         workspaceId,
         messageId: "assistant-stream-error",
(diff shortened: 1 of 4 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the missing-baseline fallback by making `pushTransientGoalSnapshot` return whether it emitted and falling back to `pushSnapshot` for previews when no activity entry exists. I also added regression coverage for that path and for the budget-limit edit resetting live budget-used UI.

ammar-agent:
@codex review

Addressed the stale failed-stream preview cache by adding `discardLiveGoalPreview()` and calling it before AgentSession restores durable goal accounting on terminal stream errors. Added a regression assertion that a budget edit after a failed preview stream does not re-emit the old transient cost.

ammar-agent:
@codex review

Addressed the serialization race by replacing the unlocked discard with `WorkspaceGoalService.restoreGoalAccountingSnapshot(workspaceId, streamStartedAtMs)`, which clears the live preview and marks the failed stream as recorded under the workspace file lock before re-emitting the durable snapshot. Extended the regression test so a late preview for the failed stream returns null before the post-failure budget edit assertion.

PR #3352, opened 2026-05-21
🤖 fix: bound Shiki worker highlighting

ammar-agent:
## Summary

Keep syntax highlighting optimistic while preventing pathological Shiki/TextMate inputs from wedging the renderer. Highlighting still attempts human-scale diffs/code blocks, but worker calls now have bounded runtime, oversized generated/minified hunks fall back before expensive renderer-thread preprocessing, and recovery no longer leaves stale or dead workers in the queue.

## Background

A live Mux instance had a `shiki-highlighter` worker pegging one renderer thread after it became unresponsive inside native Oniguruma matching. Because `highlightCode()` awaited the Comlink call without a timeout, the worker singleton stayed wedged forever and every later highlight request queued behind it.

## Implementation

- Add a 5s budget around worker `highlight()` calls.
- Serialize highlight calls on the client so a queued request's timeout starts only when that payload reaches the front of the queue.
- On timeout, terminate/recycle the worker and remember the input fingerprint so re-renders do not repeatedly kill fresh workers for the same payload.
- Resolve the worker proxy when each queued job starts, so jobs queued behind a recycled worker use the fresh worker instead of a terminated Comlink proxy.
- Guard `worker.onerror` recycling against stale workers and release the Comlink proxy before termination to avoid unhandled `InvalidStateError` in Bun.
- Avoid main-thread Shiki fallback for worker runtime failures, since it uses the same matcher and could freeze the UI.
- Replace the old 32KB diff cap with a pre-worker sync budget that allows 10k-line human-scale chunks while skipping generated/minified payloads before `join`/structured-clone work can stall the renderer.

## Validation

- `make static-check`
- `bun test src/browser/utils/highlighting/`
- `make typecheck`
- CI: Codex approved; GitHub CI checks pass. Chromatic's optional UI baseline checks are pending human acceptance for the intended visual update.

## Risks

Low-to-medium, limited to syntax highlighting. Pathological or generated/minified inputs now render as plain text after the sync/runtime budget is exceeded, and the fingerprint cache can conservatively skip highlighting for a rare collision, but the underlying code/diff text remains visible and copyable.

## Pains

`git push`/`git fetch` to GitHub hung repeatedly in this workspace, so branch updates after the initial push were applied through the GitHub Git Database API while keeping the local branch ref synchronized to the API-created commits.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$5.98`_

### DIFF
--- a/src/browser/utils/highlighting/highlightDiffChunk.test.ts
+++ b/src/browser/utils/highlighting/highlightDiffChunk.test.ts
@@ -1,4 +1,4 @@
-import { highlightDiffChunk } from "./highlightDiffChunk";
+import { highlightDiffChunk, isWithinDiffHighlightSyncBudget } from "./highlightDiffChunk";
 import type { DiffChunk } from "./diffChunking";
 
 /**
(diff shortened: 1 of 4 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Please take another look. I addressed the queued-request timeout poisoning by serializing worker highlight calls on the client side so the timeout clock starts only when a payload reaches the front of the queue.

ammar-agent:
@codex review

Please take another look. I now resolve/revalidate the worker proxy when each queued highlight job starts, so jobs queued behind a recycled worker talk to the fresh worker instead of a terminated Comlink proxy.

ammar-agent:
@codex review

Please take another look on the updated branch. The latest head preserves the queued-worker revalidation fix and keeps the local/remote branch in sync for readiness checks.

ammar-agent:
@codex review

Please take another look. I added an active-worker guard so stale error events from replaced highlighter workers cannot recycle the current healthy worker.

ammar-agent:
@codex review

Please take another look. The branch now includes the stale-worker guard: worker creation is captured in `createdWorker`, and `recycleWorker(createdWorker)` returns early if the errored worker is no longer active.

ammar-agent:
@codex review

Please take another look. I added a pre-worker sync budget for oversized/generated diff payloads while explicitly allowing 10k-line human-scale chunks through that budget.

ammar-agent:
@codex review

Please take another look. I added a Comlink proxy release before worker termination to avoid Bun surfacing unhandled InvalidStateError after worker recycling.

PR #3353, opened 2026-05-21
🤖 fix: continue promoted upcoming goals

ammar-agent:
## Summary

Fixes promoted upcoming goals so they start automatically on idle workspaces instead of requiring a manual pause/unpause nudge.

## Background

New goals and explicit paused-goal resumes already arm a kickoff continuation. Upcoming-goal promotion bypassed that path, so manual promotion and auto-promotion could leave the workspace with an active goal but no queued synthetic send. If the prior active turn had been stopped, its stale stop/ack gates could also suppress a promoted goal's continuation.

## Implementation

- Arm kickoff continuations or budget-limit wrap-ups after upcoming goals are promoted into the active slot.
- Clear stale stop/ack gates during activation so older stopped turns cannot strand newly promoted goals.
- Added regression coverage for auto-promotion after completion and manual promotion after a stopped turn, including re-promoting a goal that carried an acknowledgment gate.

## Validation

- `bun test src/node/services/workspaceGoalService.test.ts`
- `make static-check`

## Risks

Low-to-moderate. The change is scoped to goal promotion paths and reuses the existing idle-dispatch eligibility checks, so busy/streaming/queued-message gates still prevent unsafe sends.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `unknown`_

### DIFF
--- a/src/node/services/workspaceGoalService.test.ts
+++ b/src/node/services/workspaceGoalService.test.ts
@@ -2839,10 +2839,22 @@ describe("WorkspaceGoalService", () => {
     test("auto-promotes the next upcoming goal when the active goal completes", async () => {
       const active = await setGoalOk(service, { workspaceId, objective: "First" });
       const queued = await service.addUpcomingGoal({ workspaceId, objective: "Second" });
+      const dispatcher = new IdleDispatcher();
+      const executed: string[] = [];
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3336, opened 2026-05-20
🤖 fix: disable ask_user_question in mux run

ThomasK33:
## Summary

Disable `ask_user_question` for headless `mux run` sessions so CLI runs cannot stall waiting for a desktop/mobile answer UI.

## Background

`mux run` already disables UI-only tools such as TODO/status/notify. `ask_user_question` is also UI-dependent unless prefilled answers are supplied, so exposing it in headless CLI runs can leave the session waiting for an answer path the CLI does not provide.

## Implementation

Adds `ask_user_question` to the CLI `toolPolicy` disable list used by `mux run`, with an inline comment explaining the headless interaction constraint.

## Validation

- `bun run typecheck`
- `bun test src/cli/run.test.ts`
- `make fmt-check`
- `make lint`
- `make static-check`

## Risks

Low. This only narrows the tool set for `mux run`; desktop/mobile and normal plan-mode UI flows are unchanged.

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$9.14`_

PR #3316, opened 2026-05-18
🤖 feat: enable ask_user_question for exec agent

ThomasK33:
Summary

Enable the built-in Exec agent to call `ask_user_question` by removing it from Exec's tool deny list and regenerating synced agent/docs artifacts.

Implementation

- Removed `ask_user_question` from `src/node/builtinAgents/exec.md` `tools.remove`.
- Regenerated built-in agent content, synced docs, and built-in skill content.
- Added a regression test that confirms the built-in Exec definition no longer removes `ask_user_question`.

Validation

- `make static-check`
- `bun test src/node/services/agentDefinitions/builtInAgentDefinitions.test.ts src/node/services/agentDefinitions/resolveToolPolicy.test.ts src/common/utils/agentTools.test.ts`
- `make typecheck`
- `make fmt-check`

Risks

Low. This changes the top-level built-in Exec agent configuration only; existing subagent runtime hard-deny behavior for `ask_user_question` is unchanged.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$2.92`_

PR #3354, opened 2026-05-21
🤖 revert: remove ask_user_question from exec agent (#3316)

ammar-agent:
## Summary

Reverts #3316, which removed `ask_user_question` from the built-in Exec
agent's tool deny list. Exec runs are meant to be autonomous; exposing
an interactive question tool there lets agents stall waiting on a
desktop/mobile answer UI and prevents long autonomous runs.

## Background

Plan mode is the right place for clarifying questions — it already has a
human-in-the-loop interaction model. Exec mode is meant to keep working
through the contract laid out in an accepted `<plan>` or the user's
instructions without blocking on UI input. After #3316 shipped, Exec
agents could call `ask_user_question` and block indefinitely; #3336
already had to add a CLI-only carve-out for `mux run` for the same
reason. The right fix is to put `ask_user_question` back behind the
plan-mode wall.

## Implementation

- `git revert 2a5add6b2` of #3316 (Exec deny list + matching test + docs comment).
- Regenerated `builtInAgentContent.generated.ts` and
  `builtInSkillContent.generated.ts` so the synced artifacts match the
  reverted `exec.md`.
- The `mux run` disable from #3336 is intentionally left in place as
  defense in depth — `mux run` still has no UI to answer the question
  even via a non-Exec agent.

## Validation

- `bun test src/node/services/agentDefinitions/builtInAgentDefinitions.test.ts src/node/services/agentDefinitions/resolveToolPolicy.test.ts src/common/utils/agentTools.test.ts`
- `make typecheck`
- `make static-check`

## Risks

Low. The revert is mechanical and restores the pre-#3316 state of the
Exec deny list. No downstream callers depended on Exec having
`ask_user_question`; the tool still exists and remains available to
plan-mode / `ask` agent chains where it was always intended to be used.

---


ammar-agent:
@codex review

PR #3337, opened 2026-05-20
🤖 feat: add intent-aware bash summaries

ibetitsmike:
> Mux is acting on Mike's behalf.

## Summary

Adds intent-aware collapsed summaries for bash tool calls. Bash inputs can now include optional `model_intent`, and users can choose whether collapsed bash tools show only the command or the intent plus command.

## Background

Collapsed bash tool rows previously showed only the raw script. That made long-running or setup commands harder to scan in chat. This keeps the command visible while allowing the model to provide a short purpose for the command.

## Implementation

- Added optional `model_intent` to the bash tool schema.
- Added a global `toolCollapsedDisplayMode` preference with `command` and `intent-command` modes.
- Added a bash summary helper that sanitizes redundant `using <command>` and `for <duration>` text before rendering.
- Keeps duration metadata visible outside the truncated intent text, keeps active timeout metadata visible while a command is running, and truncates intent and command separately so both remain visible.
- Updated `BashToolCall` to render the configured collapsed summary and keep raw scripts in expanded details.
- Added the `Collapsed bash summaries` control to General settings.
- Added focused tests for schema parsing, summary building, settings selection and fallback behavior, and Storybook coverage for command, completed intent, and executing intent display modes.

## Validation

- `bun test src/common/utils/tools/toolDefinitions.test.ts`
- `bun test src/browser/features/Tools/bashCollapsedSummary.test.ts`
- `bun test src/browser/features/Settings/Sections/GeneralSection.test.tsx`
- `make static-check`
- `bun test --coverage --coverage-reporter=lcov src`
- `TEST_INTEGRATION=1 bun x jest tests/ui/chat/newChatStreamingFlash.test.ts --runInBand`
- Validated in a `dev-server-sandbox` instance with `agent-browser`, including screenshots for intent mode, command mode, and the settings control.

## Risks

Low to moderate. This changes collapsed bash tool rendering and a new localStorage preference. Existing history without `model_intent` falls back to command-only display, and corrupted preference values fall back to the default.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$66.14`_

PR #3355, opened 2026-05-21
🤖 feat: stack bash intent + command in collapsed header

ammar-agent:
## Summary

Replaces the inline "intent **using** command" header that #3337 introduced with a stacked two-line collapsed bash summary (intent on top, command below), surfaces the intent in the expanded details, and removes the **Collapsed bash summaries** General setting so the experience is consistent for everyone.

## Background

`model_intent` was added in #3337 to give scannable context for long-running bash calls, but the collapsed header rendered intent + command as a single inline row joined by the word "using". Both halves were truncated to ~48ch and the connector pushed real content off-screen. The behavior was also gated behind a General-section toggle, which added configuration surface without a strong reason to keep both modes around.

This PR keeps the underlying `model_intent` tool field and the sanitizer intact — it only changes how the value is presented and removes the toggle.

## Implementation

- `BashToolCall.tsx`: collapsed intent-command rows now render a stacked block — primary intent on top, muted mono `command` (10px) beneath — inside a `max-w-[28rem]` container, so both lines truncate independently. Right-side duration/status chips stay vertically centered next to the block. When `model_intent` is absent or equals the command, the single-line mono command renders as before.
- `BashToolCall.tsx` expanded view: when `model_intent` is a non-blank string, a new `Intent` `DetailSection` renders above `Script` using `whitespace-pre-wrap` so longer intents wrap cleanly.
- `bashCollapsedSummary.ts`: dropped the `displayMode` parameter; the helper now always attempts the intent-command summary and falls back to the command kind when intent is missing/blank/equal-to-command. Sanitization logic is unchanged.
- `common/constants/storage.ts`: removed `TOOL_COLLAPSED_DISPLAY_MODE_KEY`, `ToolCollapsedDisplayMode`, `DEFAULT_TOOL_COLLAPSED_DISPLAY_MODE`, and `isToolCollapsedDisplayMode`.
- `GeneralSection.tsx`: removed the Collapsed bash summaries select, its options table, change handler, persisted-state hook, and corresponding imports.
- Tests: dropped the two display-mode tests and the orphan `waitForArchiveSettingsLoad` helper from `GeneralSection.test.tsx`; refreshed `bashCollapsedSummary.test.ts` to the new signature. `BashToolCall.stories.tsx`: removed the `setCollapsedDisplayMode` helper and renamed `CommandCollapsedSummary` → `NoIntentSummary` (model omits `model_intent`) to still exercise the command-only fallback.

## Validation

- `bun test src/browser/features/Tools/bashCollapsedSummary.test.ts` ✅
- `bun test src/browser/features/Settings/Sections/GeneralSection.test.tsx` ✅
- `bun test src/browser/features/Tools/ src/browser/features/Settings/ src/common/constants/` ✅
- `make typecheck` ✅
- `make lint` ✅
- `make static-check` ✅

## Risks

Low. Surface area is limited to bash tool call rendering, one removed General setting, and a few storage exports nothing else consumes (verified via repo-wide grep). Stale `toolCollapsedDisplayMode` localStorage values are simply ignored on the next load. No backend or IPC changes.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `xhigh` • Cost: `$6.53`_

### DIFF
--- a/src/browser/features/Settings/Sections/GeneralSection.test.tsx
+++ b/src/browser/features/Settings/Sections/GeneralSection.test.tsx
@@ -12,7 +12,6 @@ import {
   DEFAULT_WORKTREE_ARCHIVE_BEHAVIOR,
   type WorktreeArchiveBehavior,
 } from "@/common/config/worktreeArchiveBehavior";
-import { TOOL_COLLAPSED_DISPLAY_MODE_KEY } from "@/common/constants/storage";
 
 interface MockConfig {
   coderWorkspaceArchiveBehavior: CoderWorkspaceArchiveBehavior;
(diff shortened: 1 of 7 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

PR #3356, opened 2026-05-21
🤖 fix(ssh): harden bare base repo against receive-pack thin-pack failures

ammar-agent:
## Summary

Hardens the SSH runtime's shared bare base repo (`.mux-base.git`) against the receive-pack `unresolved deltas left after unpacking` / `unpack-objects abnormal exit` failure that aborts workspace creation. The fix layers four independent defenses so that any one of them is enough to keep the slow-path push healthy on both new and legacy remote repos.

## Background

A user hit this during SSH workspace init:

```
Pushing to remote...
remote: fatal: unresolved deltas left after unpacking
error: remote unpack failed: unpack-objects abnormal exit
 ! [remote rejected]     <branch> -> refs/mux-bundle/<branch> (unpacker error)
error: failed to push some refs to 'ssh://.../.mux-base.git'
Initialization failed: Failed to sync project: Failed to push to remote: ...
```

Two interacting root causes:

1. **Receive-pack's small-push fast path.** When the incoming pack contains fewer than `transfer.unpackLimit` objects (default `100`), receive-pack routes objects through `unpack-objects` instead of `index-pack`. `unpack-objects` cannot resolve thin-pack delta bases that live in *other* on-disk packs, so a small thin push to a populated bare repo can fail with `unresolved deltas left after unpacking` even when every base object is technically present.
2. **Orphan packs on disk.** A `git gc` that gets SIGKILLed between writing `pack-<sha>.pack` and `pack-<sha>.idx` leaves the pack with no index. Git's negotiation surface ignores indexless packs, so the *server* doesn't advertise objects it actually has on disk → the client builds a thin pack assuming those bases are missing → receive-pack fails to resolve them.

The pre-fetch from origin (which would have populated the bases) had also failed, exposing the underlying weakness.

The existing retry classifier only matched generic transport errors (`pack-objects died`, `Connection reset`, …), so a thin-pack failure surfaced immediately as a fatal `Failed to sync project`.

## Implementation

Four layered defenses in `src/node/runtime/SSHRuntime.ts`:

1. **Prevent the failure mode** — `ensureBaseRepo`'s batched setup script now also runs `git config --local receive.unpackLimit 1`. Receive-pack always routes incoming pushes through `index-pack` (which calls `--fix-thin` and resolves missing delta bases from local objects) instead of `unpack-objects`, regardless of object count. Idempotent — runs every sync, so legacy repos heal on the next workspace creation. Trade-off: more small packs, but the existing fragmented-pack maintenance (`ensureHealthyBaseRepoForSync`) already coalesces them on a threshold.
2. **Heal pre-existing on-disk damage** — `repairBaseRepoForSync` gets a new step before `gc`: scan `objects/pack/pack-*.pack` for orphans missing `.idx` and run `git index-pack` to rebuild the index from the pack. The existing code comment already noted this exact SIGKILL failure mode but didn't act on it.
3. **Retry the failure** — added `unresolved deltas`, `unpacker error`, `unpack-objects abnormal exit`, `remote unpack failed` to `PROJECT_SYNC_RETRYABLE_ERRORS` so the existing retry loop kicks in (which runs the maintenance from #1 + #2 before re-attempting).
4. **Defang the retry push** — when a retryable failure matches the unresolved-deltas pattern, the retry loop latches a `forceNoThinNextAttempt` flag and threads it into `syncProjectSnapshotViaGitPush` as `{ forceNoThin: true }`. The next push then includes `--no-thin` on both the branch and tag pushes, sending a self-contained pack the receiver can unpack without any delta-base lookup. Unrelated retryable failures (connection reset, killed by signal, …) leave the flag off — no bandwidth penalty for problems `--no-thin` wouldn't fix.

Also surfaces `Pre-fetch from origin skipped (fetch failed)` cause via `log.debug` so we can diagnose recurrent prefetch misses without spamming the init logger on the happy path.

## Validation

- `make static-check` passes (typecheck + lint + format + docs links).
- `bun test src/node/runtime/` passes (all SSH runtime / sync contract / retry orchestration suites).
- New retry tests cover: thin-pack failure → next attempt flips `--no-thin`; connection-reset → flag stays off.
- New contract tests cover: `--no-thin` appears on both branch and tag pushes when `forceNoThin: true`; default sync path omits `--no-thin`.
- Updated existing maintenance-command expectation tests for the new orphan-pack reindex step.

## Risks

Touches the SSH workspace init slow path, which runs on every cold SSH workspace creation and every snapshot-drift sync.

- `receive.unpackLimit=1` is unconditionally applied to every shared bare repo via `ensureBaseRepo`. The trade-off is more small packs vs. fewer thin-pack-unpack failures; existing maintenance already collapses fragmented packs so the steady-state is unchanged.
- The orphan-pack reindex runs `git index-pack` against any `.pack` lacking an `.idx`. A corrupt `.pack` will fail `index-pack` (best-effort, logged), then fall through to `gc` and the `--no-thin` retry push.
- `--no-thin` is only applied on retry after a confirmed thin-pack failure, so the happy path is unchanged.
- Retry classification is purely additive — existing retryable patterns and orchestration are untouched.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `high` • Cost: `$6.77`_

### DIFF
--- a/src/node/runtime/SSHRuntime.retry.test.ts
+++ b/src/node/runtime/SSHRuntime.retry.test.ts
@@ -86,6 +86,9 @@ class TestSSHRuntime extends SSHRuntime {
   readonly callOrder: string[] = [];
   readonly cleanupCalls: string[] = [];
   readonly backoffCalls: number[] = [];
+  /** Records `options.forceNoThin` per attempt so tests can assert the retry
+   *  loop opts the next push into `--no-thin` after a thin-pack failure. */
+  readonly forceNoThinPerAttempt: boolean[] = [];
 
   private readonly actions: SyncAction[] = [];
   private cleanupHook?: () => Promise<void>;
@@ -114,7 +117,8 @@ class TestSSHRuntime extends SSHRuntime {
     _projectPath: string,
     _layout: RemoteProjectLayout,
     _initLogger: InitLogger,
-    _abortSignal?: AbortSignal
+    _abortSignal?: AbortSignal,
+    options: { forceNoThin?: boolean } = {}
   ): Promise<void> {
     const action = this.actions.shift();
     if (!action) {
@@ -122,6 +126,7 @@ class TestSSHRuntime extends SSHRuntime {
     }
 
     this.callOrder.push(action.label);
+    this.forceNoThinPerAttempt.push(options.forceNoThin === true);
     action.abortController?.abort();
     if (action.error) {
       return Promise.reject(action.error);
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

PR #3357, opened 2026-05-21
🤖 feat(goals): per-goal auto-compaction threshold override

ammar-agent:
## Summary

Adds an optional per-goal auto-compaction threshold override. Each goal can pin its own auto-compaction behavior (compact aggressively for cost, or disable for fidelity) independently of the workspace's per-model slider. Tri-state matching `budgetCents` / `turnCap`: omitted = no override (the workspace setting applies), `null` = explicit clear, integer `0–100` = explicit percent (`100` = compaction disabled for this goal). Surfaced as an embedded **slider tile** in the GoalTab — Default / Customize / drag-to-Off — and as `/goal --compact <value>` on the slash command.

## Background

The right-sidebar auto-compact slider is keyed per-model, not per-goal — see `getAutoCompactionThresholdKey(model)` in `src/common/constants/storage.ts`. That works for the steady-state default, but goal-driven workflows often have different cost-vs-fidelity preferences than the workspace default:

- A long research goal might want **aggressive** compaction (40–50%) for cost control while the workspace is otherwise fine at the 70% default.
- A short refactor or migration goal might want **no** compaction so the agent keeps the full plan in view (drag the slider all the way right → "Off").
- Or a single goal might want a different threshold than the rest of that workspace's flow.

Defaults to off (the field is absent on legacy/new goal records → no override → existing behavior preserved), so users have to actively opt in to change anything.

## Implementation

**Data model** (`src/common/orpc/schemas/goal.ts`, `src/common/types/goal.ts`)
- New `autoCompactionThresholdPct: z.number().int().min(0).max(100).nullable().optional()` on `GoalRecordV1Schema`, `GoalSnapshotSchema`, `GoalSetInputSchema`, `GoalBoardAddUpcomingInputSchema`, `GoalBoardUpdateUpcomingInputSchema`. Optional + nullable so legacy records load without migration.
- `toGoalSnapshot` carries the field through to the renderer-visible snapshot using the same `Object.hasOwn`-style tri-state preservation pattern as the existing `completionSummary` handling.

**Backend service** (`src/node/services/workspaceGoalService.ts`)
- `SetGoalInput` and `PendingGoalMutation` get the field; `createGoal`, `applyMutableFields`, `setGoalInternal` (mid-stream + projected record), `setGoalImmediately` (replace branch), `addUpcomingGoal`, `updateUpcomingGoal` each thread it through.
- `applyMutableFields` uses `Object.hasOwn` so `null` (explicit clear) survives the patch while `undefined` (no opinion) preserves the existing value — same pattern as `budgetCents` / `turnCap`.
- The same-objective shortcut in `setGoalImmediately` adds `autoCompactionThresholdPct` to its `hasMutableChange` predicate so a threshold-only edit triggers the apply branch instead of being treated as a no-op.

**Compaction consumer** (`src/node/services/compactionMonitor.ts`, `src/node/services/agentSession.ts`)
- `CompactionMonitor.checkBeforeSend` / `checkMidStream` accept an optional `thresholdOverride` (decimal 0–1). When finite, it replaces `this.threshold` for the single call — the monitor's persistent state is untouched, so the override layer is per-call and never leaks across goals.
- Override `>= 1` clamps to compaction-disabled for that call (matches the slider's "off" sentinel).
- Override `0` is **honored** as the aggressive extreme (Codex P2 fix — see the dedicated "override of 0% is honored" test). Only `null` / `undefined` / `NaN` / `Infinity` / negative values fall back to `this.threshold`.
- `AgentSession.resolveGoalAutoCompactionThresholdOverride()` fetches the active goal via the existing injected `workspaceGoalService` and converts the persisted percent → decimal. Called from both compaction call sites (on-send and mid-stream usage-delta handler). Errors are swallowed silently because the resolver runs on the hot path of every send; a corrupt persisted goal record must never interrupt streaming.

**Renderer banner** (`src/browser/components/ChatPane/ChatPane.tsx`)
- `effectiveAutoCompactionThreshold` consults `workspaceState.goal?.autoCompactionThresholdPct` first, falling back to the per-model slider value, so the warning banner stays in lockstep with what the backend will actually do.

**UI — `GoalTab` slider tile** (`src/browser/features/RightSidebar/GoalTab.tsx`)

Replaces the previous text-input editor with an embedded slider tile that has two visual modes:

- **Default mode** (`value == null`) — workspace per-model slider governs. Tile renders "Default — workspace setting" plus a `Customize` button. Slider is hidden because we don't know the workspace per-model threshold at this layer; a ghost thumb would imply a value that isn't really in effect.
- **Override mode** (`value` is a number) — native range slider with thumb at the override value. Drag commits on **release** (`onMouseUp` / `onTouchEnd` / `onKeyUp`), not on every drag step, so backend mutation rate stays bounded by user gestures. The right end is labeled "Off" mirroring the workspace per-model slider; dragging there commits `100` (per-goal disabled). A `Use default` button clears back to `null`.

`Customize` seeds the override at `DEFAULT_AUTO_COMPACTION_THRESHOLD_PERCENT` (70) so the slider has a concrete value to drag from. Cost of an accidental Customize is one click on `Use default`.

The slider uses `onInput` + `onChange` together: happy-dom (the `tests/ui` runtime) only synthesizes the native `input` event for range inputs, so `onInput` is required for deterministic tests, and `onChange` satisfies React's controlled-input contract. The two handlers route to the same idempotent draft updater.

**UI — responsive grid layout** (`src/browser/features/RightSidebar/GoalTab.tsx`)

The active-goal stat grid is now container-query-driven:

- Narrow (default, `grid-cols-2`): Budget col-span-2 (full row); Turns + Elapsed half-half; Auto-compact col-span-2 (full row, slider has room to breathe).
- Wider (`@md:grid-cols-3`, triggered when the dynamically-resizable right sidebar passes the `@md` container width): Budget + Turns + Auto-compact share row 1 (each col-span-1); Elapsed becomes a thin full-width row below. BudgetTile's internal copy stacks vertically at the wider breakpoint so its cost / remaining pair still fits in ~1/3 width.

Container queries (`[container-type:inline-size]` + `@md:`) are used instead of viewport breakpoints because the right sidebar is dynamically resizable per AGENTS.md, so `sm:` / `md:` wouldn't have responded to sidebar resizing.

**UI — create form** (`src/browser/features/RightSidebar/GoalTab.tsx`)
- `GoalCreateForm` reuses the same slider tile with local state instead of the previous text input + parser. `null` (Default mode) means the create intent omits the field entirely; a number is sent verbatim.

**UI — slash command + handler chain**
- `/goal --compact <value>` parses with the same vocabulary as the slider's Off/Default sentinels; composes with `-b` and `--turns` in the documented order; non-leading occurrences stay as goal text. The parser flows through `resolveGoalSetIntent` (verbatim pass-through, since there's no workspace-level default for this field yet) into the existing `setGoal` IPC.
- `RightSidebar.handleGoalUpdateAutoCompactionThresholdPct` mirrors `handleGoalUpdateBudget` / `handleGoalUpdateTurnCap`, wiring through `setGoalWithSingleConflictRetry` so the optimistic-concurrency retry loop covers all three.
- Tab registry context picks up `onUpdateAutoCompactionThresholdPct` so the prop is typed end-to-end.

## Validation

- `make typecheck` ✅
- `make lint` ✅
- `make fmt-check` ✅
- `make storybook-build` ✅
- `bun test src/node/services/compactionMonitor.test.ts` — **13 pass** (5 override-lane tests + 1 dedicated Codex P2 test pinning `0%` as honored)
- `bun test src/node/services/workspaceGoalService.test.ts` — **129 pass** (6 threshold-pinning tests including legacy-record back-compat)
- `bun test src/browser/utils/slashCommands/goal.test.ts` — **22 pass** (7 `--compact` parser tests)
- `bun test src/browser/features/RightSidebar/GoalTab.test.tsx` — **32 pass** (9 slider-tile tests: Default/Off/numeric display; Customize commits 70; drag commits snapped value on mouseUp; drag-to-100 commits 100; mouseUp without change does NOT commit; Use default clears to null; canEdit gating)
- Broader `bun test src/node/services/ src/browser/features/RightSidebar/ src/browser/utils/slashCommands/ src/browser/utils/goals/` — **3554 pass / 0 fail**

## Risks

Low–medium, scoped to the auto-compaction subsystem.

- **Legacy goal records** without the field continue to behave exactly as before — the field is `.optional()` + `.nullable()` and `applyMutableFields` only touches it via `Object.hasOwn`. A dedicated test (`workspaceGoalService.test.ts`) writes a pre-field record to disk and asserts `getGoal` returns it parsed with the field undefined.
- **Mid-stream + on-send compaction paths** are on the hot path for every send + every usage delta. The override resolver swallows errors silently so a corrupt persisted goal record never interrupts streaming; the monitor's `resolveEffectiveThreshold` does the same for malformed values it receives.
- **Renderer / backend agreement** on the effective threshold is pinned by both consulting the same goal snapshot field, so the warning banner and the actual compaction trigger stay in sync. The Codex P2 fix (honor `0%` as the aggressive extreme rather than silently falling back) closed the last gap here.
- **Slider drag → backend** is gated to the release event, not the per-pixel drag step, so the backend mutation rate is bounded by user gestures (mouseUp / touchEnd / keyUp) and the `draft !== value` guard inside `commit()` further suppresses no-op releases.

## Follow-ups (deferred)

- Workspace-level default for the per-goal override (would let users say "all my goals in this workspace start with aggressive compaction").
- Upcoming-row inline editor for the threshold (currently editable via `updateUpcomingGoal` API but not surfaced in the UI's row editor; can be set via `--compact` on the slash command or after promotion).

## Pains

- happy-dom's range-input event synthesis differs from real browsers — it fires the native `input` event but does NOT fire the `change` event when `fireEvent.input` is dispatched, so a slider driven only by React's `onChange` is read-only in tests. Took a debug-test pass to isolate (initially looked like a stale-closure / React state-flush issue). Resolved by routing drag updates through both `onInput` and `onChange`; the duplicate handler is documented inline.
- Two iterations on layout: the first pass kept Budget full-width at all widths, but per advisor guidance switched to a 3-col layout at the `@md` container width so Budget / Turns / Auto-compact line up on one row exactly as requested. BudgetTile's internal copy had to gain a vertical stacking variant for the narrower tile size.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `xhigh` • Cost: `$53.97`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -376,17 +376,28 @@ export const ChatPane: React.FC<ChatPaneProps> = (props) => {
     [deferredMessages]
   );
 
+  // Resolve the effective threshold for the warning banner: a per-goal
+  // override (set on the active goal's `autoCompactionThresholdPct`)
+  // wins over the workspace-level slider for both the backend and the
+  // renderer. Keeping the two in lockstep matters because the banner
+  // says "compaction is imminent" — if the renderer used the slider
+  // value but the backend used the goal override, the user would see
+  // either spurious or missing warnings around the threshold boundary.
+  const effectiveAutoCompactionThreshold =
+    workspaceState.goal?.autoCompactionThresholdPct != null
+      ? workspaceState.goal.autoCompactionThresholdPct / 100
+      : autoCompactionThreshold / 100;
   const autoCompactionResult = useMemo(
     () =>
       checkAutoCompaction(
         workspaceUsage,
         pendingModel,
         use1M,
-        autoCompactionThreshold / 100,
+        effectiveAutoCompactionThreshold,
         undefined,
         providersConfig
       ),
-    [workspaceUsage, pendingModel, use1M, providersConfig, autoCompactionThreshold]
+    [workspaceUsage, pendingModel, use1M, providersConfig, effectiveAutoCompactionThreshold]
   );
 
   // Show warning when: shouldShowWarning flag is true AND not currently compacting.
--- a/src/browser/features/RightSidebar/GoalTab.test.tsx
+++ b/src/browser/features/RightSidebar/GoalTab.test.tsx
@@ -403,6 +403,157 @@ describe("GoalTab", () => {
     await waitFor(() => expect(onUpdateTurnCap).toHaveBeenCalledWith(15));
   });
 
+  test("auto-compact tile shows 'Default' when no per-goal override is set", () => {
+    const { getByText, queryByText } = render(
+      <GoalTab
+        goal={goal({ autoCompactionThresholdPct: null })}
+        onSetStatus={mock()}
+        onClear={mock()}
+        onUpdateAutoCompactionThresholdPct={mock()}
+      />
+    );
+
+    // Null override → fall back to the workspace setting. The tile
+    // must distinguish "no override" from "explicit value" so users
+    // can tell at a glance whether the model-level slider is in effect.
+    expect(getByText("Auto-compact")).toBeTruthy();
+    expect(getByText("Default")).toBeTruthy();
+    expect(getByText("workspace setting")).toBeTruthy();
+    expect(queryByText("Off")).toBeNull();
+  });
+
+  test("auto-compact tile shows 'Off' at the per-goal-disabled sentinel", () => {
+    const { getByText, queryByText } = render(
+      <GoalTab
+        goal={goal({ autoCompactionThresholdPct: 100 })}
+        onSetStatus={mock()}
+        onClear={mock()}
+        onUpdateAutoCompactionThresholdPct={mock()}
+      />
+    );
+
+    // 100 = explicit per-goal disable. Distinct from null (default)
+    // and from 0–99 (numeric threshold). The "Off" label is the
+    // user-visible cue that this specific goal won't auto-compact.
+    expect(getByText("Off")).toBeTruthy();
+    expect(getByText("compaction disabled")).toBeTruthy();
+    expect(queryByText("Default")).toBeNull();
+  });
+
+  test("auto-compact tile renders an explicit numeric override", () => {
+    const { getByText } = render(
+      <GoalTab
+        goal={goal({ autoCompactionThresholdPct: 50 })}
+        onSetStatus={mock()}
+        onClear={mock()}
+        onUpdateAutoCompactionThresholdPct={mock()}
+      />
+    );
+
+    expect(getByText("50%")).toBeTruthy();
+    expect(getByText("of context window")).toBeTruthy();
+  });
+
+  test("edits auto-compact threshold inline and persists the percent", async () => {
+    const onUpdateAutoCompactionThresholdPct = mock(() => Promise.resolve(undefined));
+    const { getByLabelText, getByText } = render(
+      <GoalTab
+        goal={goal({ autoCompactionThresholdPct: null })}
+        onSetStatus={mock()}
+        onClear={mock()}
+        onUpdateAutoCompactionThresholdPct={onUpdateAutoCompactionThresholdPct}
+      />
+    );
+
+    fireEvent.click(getByLabelText("Edit goal auto-compact threshold"));
+    const input = getByLabelText("Goal auto-compact threshold percent");
+    await waitFor(() => expect(document.activeElement).toBe(input));
+    fireEvent.input(input, { target: { value: "60" } });
+    fireEvent.click(getByText("Save threshold"));
+
+    await waitFor(() => expect(onUpdateAutoCompactionThresholdPct).toHaveBeenCalledWith(60));
+  });
+
+  test("editing auto-compact with 'off' submits the per-goal-disabled sentinel (100)", async () => {
+    const onUpdateAutoCompactionThresholdPct = mock(() => Promise.resolve(undefined));
+    const { getByLabelText, getByText } = render(
+      <GoalTab
+        goal={goal()}
+        onSetStatus={mock()}
+        onClear={mock()}
+        onUpdateAutoCompactionThresholdPct={onUpdateAutoCompactionThresholdPct}
+      />
+    );
+
+    fireEvent.click(getByLabelText("Edit goal auto-compact threshold"));
+    const input = getByLabelText("Goal auto-compact threshold percent");
+    fireEvent.input(input, { target: { value: "off" } });
+    fireEvent.click(getByText("Save threshold"));
+
+    // "off" must round-trip to 100, not to null — these are different
+    // states (explicit disable vs. clear override). The parser path
+    // and the test path must agree.
+    await waitFor(() => expect(onUpdateAutoCompactionThresholdPct).toHaveBeenCalledWith(100));
+  });
+
+  test("editing auto-compact with a blank value clears the override (null)", async () => {
+    const onUpdateAutoCompactionThresholdPct = mock(() => Promise.resolve(undefined));
+    const { getByLabelText, getByText } = render(
+      <GoalTab
+        goal={goal({ autoCompactionThresholdPct: 50 })}
+        onSetStatus={mock()}
+        onClear={mock()}
+        onUpdateAutoCompactionThresholdPct={onUpdateAutoCompactionThresholdPct}
+      />
+    );
+
+    fireEvent.click(getByLabelText("Edit goal auto-compact threshold"));
+    const input = getByLabelText("Goal auto-compact threshold percent");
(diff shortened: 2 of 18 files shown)


ammar-agent:
@codex review

Per-goal auto-compaction threshold override. Tri-state mirroring `budgetCents`/`turnCap`: omitted = no override, `null` = clear, integer 0–100 = explicit percent (`100` = disabled for that goal). Threads through schemas, `WorkspaceGoalService` mutators, `CompactionMonitor` (new per-call `thresholdOverride` lane), `AgentSession` on-send + mid-stream call sites, `ChatPane` warning banner, `GoalTab` UI tile + inline editor + `GoalCreateForm`, `/goal --compact` slash command, and the `RightSidebar` handler chain.

Particular attention welcome on:

1. **Override resolution** in `CompactionMonitor.resolveEffectiveThreshold` — the defensive fallback for malformed values (`null` / NaN / non-positive / non-finite). Compaction is on the hot path; thrown errors there would interrupt streams.
2. **Persistence tri-state** in `applyMutableFields` and the upcoming-goal mutators — `Object.hasOwn` so `null` survives a patch while omitted preserves the existing value. Same pattern as `budgetCents`/`turnCap`.
3. **Legacy record loading** — schema is `.optional().nullable()`. There's a dedicated back-compat test that writes a pre-field goal.json and asserts it parses.

ammar-agent:
@codex review

Addressed your P2:

* `CompactionMonitor.resolveEffectiveThreshold` now treats `0` as a valid aggressive-extreme override and only rejects truly corrupt values (`null` / `NaN` / `±Infinity` / negative). Comment in the source spells out why renderer/backend agreement on the `0` extreme matters.
* New dedicated test `override of 0% is honored (Codex P2: renderer/backend must agree)` in `compactionMonitor.test.ts` pins both `checkBeforeSend` and `checkMidStream` for that contract.
* Updated the negative test list to remove `0` from the "should fall back" cases and document why.

All checks still green locally (`make static-check`, `make storybook-build`, full bun test suite — 3553 pass / 0 fail).

ammar-agent:
@codex review

Slider-driven aesthetic refinement on the per-goal auto-compact tile. Worth a fresh pass — the previous review approved the text-input + inline editor design, but this commit replaces it with a slider tile (Default / Customize / drag / Use default) and reshapes the active-goal grid to be container-query responsive (Budget + Turns + Auto-compact share a row at the wider @md breakpoint; Elapsed becomes a thin full-width row below).

The schema, backend, monitor, and slash-command paths are unchanged from the previous round you approved — only `GoalTab.tsx` and its tests moved.

Specific things to look at:
- `GoalAutoCompactSlider` / `GoalAutoCompactSliderOverride` — the two-mode split + the "prevValue" sync pattern instead of `useEffect` (per the `react-effects` skill).
- The `onInput` + `onChange` duplicate handler for happy-dom compatibility — comments explain why.
- The container-query layout (`[container-type:inline-size]` + `@md:` variants) on the grid + BudgetTile.
- `parseGoalAutoCompactionThresholdInput` was dropped from `GoalTab.tsx` since the slider commits typed values directly; the slash-command parser is unaffected.

Validation: typecheck / lint / fmt-check / storybook-build all green; 3554 tests pass.

ammar-agent:
@codex review

Addressed the P2 finding (`PRRT_kwDOPxxmWM6D9ZRp`):

> The range input wires both `onTouchEnd` and `onMouseUp` to `commit`, but many touch browsers emit a synthetic mouse-up after touch-end; in that sequence `commit()` runs twice before `value` is refreshed from the async mutation, so the `draft !== value` guard passes both times and `onChange(draft)` is sent twice.

Fixed in 30774e50f with a `justCommittedRef` flag on `GoalAutoCompactSliderOverride`:

- `commit()` bails if the flag is set, and sets the flag immediately before calling `onChange` so the synthetic mouseup that follows on the same tick is short-circuited.
- `updateDraftFromEvent` (the onInput/onChange path) clears the flag whenever a fresh drag/keyboard gesture starts, so subsequent edits commit normally.
- Used a ref (not state) because we explicitly don't want a re-render between commit() and the synthetic mouseup. The flag lifetime is tied to user intent (the next `input` event), not a setTimeout — per AGENTS.md's "avoid timing-based coordination" rule.

Two new tests pin the contract:
- `touch + synthesized mouseup from the same gesture commits only once (Codex P2)` — touchEnd + mouseUp dispatched sequentially must result in exactly one `setGoal` call. Without the fix this fails with two calls.
- `a fresh drag after a release commits again (per-gesture flag resets)` — confirms the flag resets correctly so the slider doesn't lock up after one edit.

Validation: typecheck / lint / fmt-check all green; GoalTab tests 34 pass.

PR #3362, opened 2026-05-22
🤖 fix: keep terminal subscription stable

ethanndickson:
## Summary

Keep the integrated terminal's router subscription stable when parent renders recreate the `onExit` callback, preventing chat streaming updates from causing a terminal unsubscribe/resubscribe cycle that clears the visible screen.

## Background

The web terminal flashed because `TerminalView` cleared the terminal every time its router subscription effect re-ran. Chat streaming can rerender the right sidebar and recreate inline terminal callbacks, which changed `handleExit`, retriggered the subscription effect, and briefly showed a blank terminal before cached screen state replayed.

## Implementation

`TerminalView` now stores the latest committed `onExit` callback in a ref synchronized from a layout effect, then keeps `handleExit` stable. That preserves current exit behavior while ensuring callback identity churn does not tear down the terminal router subscription, and it avoids making render-time callback props observable before React commits them. The new regression test renders `TerminalView`, rerenders it with a different `onExit`, and asserts the router subscription, unsubscribe, and terminal clear counts remain unchanged while exit still invokes the latest committed callback.

## Validation

Verified the regression test fails without the `TerminalView` fix by temporarily restoring that file from `origin/main`; it fails with `Expected number of calls: 1 / Received number of calls: 2` for `mockRouter.subscribe`. Reapplied the fix and ran the regression test successfully, then ran the regression test 10 times in a loop to check consistency and ran `make static-check` with local ESLint concurrency reduced to 2.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$17.02`_

### DIFF
--- a/src/browser/components/TerminalView/TerminalView.test.tsx
+++ b/src/browser/components/TerminalView/TerminalView.test.tsx
@@ -1,0 +1,207 @@
+import "../../../../tests/ui/dom";
+
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import { act, cleanup, render, waitFor } from "@testing-library/react";
+import { installDom } from "../../../../tests/ui/dom";
+
+interface TerminalSubscribeCallbacks {
+  onOutput: (data: string) => void;
+  onScreenState: (state: string) => void;
+  onExit: (code: number) => void;
+}
+
+interface MockTerminalOptions {
+  fontSize: number;
+  fontFamily: string;
+  cursorBlink: boolean;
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed the committed-callback concern by moving the `onExitRef` sync into a layout effect, so the stable terminal subscription callback only observes committed `onExit` props.

PR #3363, opened 2026-05-22
🤖 refactor(review): drop dismiss + jump-to-source from Assisted pins

ammar-agent:
## Summary

Follow-up to the assisted-review UX pass (#3358). Removes the per-pin
"dismiss" and "jump to source" actions plus all of their supporting
state, since neither was carrying its weight: dismiss did not survive
user testing as useful, and jump-to-source only worked when the chat
transcript happened to be mounted with the matching `data-message-id`
in the DOM — it silently no-op'd in immersive review and in many
split-view layouts.

## Background

Direct feedback after the UX pass shipped:

> The dismiss action is not useful, and the jump to source doesn't
> even work so should be removed unless it can easily be made more
> reliable.

The jump-to-source mechanism relied on a global `document.querySelector`
hitting the transcript boundary. That assumption breaks any time the
right sidebar isn't on the Chat tab, the immersive review overlay is
covering the transcript, or virtualization clipping is in play. Making
it reliable in all those cases would require routing the jump through a
chat-pane scroll API and reasoning about which tab/route is currently
in front — well beyond a minor UX touch-up, especially given the action
wasn't earning attention in practice.

## Implementation

- `HunkViewer`: drops `assistedKey`, `assistedSourceMessageId`,
  `onDismissAssisted`, and `onJumpToAssistedSource` props plus the
  entire actions strip below the assisted comment.
- `ReviewPanel`: drops the dismissed-pin `usePersistedState` (in both
  the panel and the always-mounted stats reporter), the self-healing
  prune effect, the `handleDismissAssistedPin` /
  `handleRestoreDismissedAssisted` / `handleJumpToAssistedSource`
  callbacks, the `assistedKeyByHunkId` /
  `assistedSourceMessageIdByHunkId` maps, and the "Restore N dismissed"
  links in the banner / empty-state. The keybind effect no longer needs
  `dismissedAssistedKeys.length` to decide whether to no-op the toggle.
- `ReviewControls`: drops the `assistedDismissedCount` /
  `onRestoreDismissedAssisted` props and the "+N dismissed" tooltip
  entry in the control bar.
- `STORAGE_KEYS.reviewAssistedDismissed`: deleted (no other consumer).
- `AssistedReviewHunk.sourceMessageId`: deleted; it was only consumed
  by the removed jump button. `StreamingMessageAggregator` no longer
  threads `messageId` into `processToolResult`; the live-vs-replay
  distinction is now carried entirely by the existing `addedAt`
  timestamp, which still drives the transient "new" badge.

## Validation

- `make static-check`
- `bun test src/browser/utils/messages/StreamingMessageAggregator.test.ts src/browser/features/RightSidebar/CodeReview/ReviewPanel.assistedStats.test.ts`
- `make test` — same 5 pre-existing failures as on `origin/main`
  (`useChatTranscriptFullWidth`, `GeneralSection`); unrelated to this
  change.

## Risks

Low. The change is a pure removal of two surfaced affordances and their
supporting localStorage. Existing `review-assisted-dismissed:<workspaceId>`
entries become harmless orphans in localStorage — they're per-workspace
and small, and self-heal next time the user clears site data; no
migration is warranted given the feature shipped only days ago in
#3358. The remaining assisted-review behavior (matching, "new" badge,
worklist toggle, filter persistence) is unchanged.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `xhigh` • Cost: `$?`_

### DIFF
--- a/src/browser/features/RightSidebar/CodeReview/HunkViewer.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/HunkViewer.tsx
@@ -4,7 +4,7 @@
  */
 
 import React, { useState, useMemo } from "react";
-import { Check, Circle, MessageCircle, Sparkles, X } from "lucide-react";
+import { Check, Circle, Sparkles } from "lucide-react";
 import type { DiffHunk, Review, ReviewNoteData } from "@/common/types/review";
 import { SelectableDiffRenderer } from "../../Shared/DiffRenderer";
 import type { ReviewActionCallbacks } from "../../Shared/InlineReviewNote";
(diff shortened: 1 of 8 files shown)


ammar-agent:
@codex review

PR #3370, opened 2026-05-23
🤖 fix: repair review paths for sub-projects

ammar-agent:
Summary
- Fixes Review/Assisted Review path handling for sub-project workspaces by keeping repo-oriented review commands at the checkout root and keeping Review pins in project-relative coordinates.

Background
- Sub-project workspaces intentionally run normal agent/tool commands from a scoped cwd. The Review pane, however, builds repo-root git pathspecs and renders project-relative diff paths, so running review git commands from the scoped cwd could make selected files or assisted pins disappear.
- Agents also naturally call `review_pane_update` with paths like `src/file.ts`; those can be ambiguous between a project-relative path and a path relative to the current tool cwd.

Implementation
- `executeBash({ cwdMode: "repo-root" })` and git command mode now use the single-project checkout root while preserving default scoped execution cwd.
- Assisted review now treats project-relative paths as canonical. Ambiguous plain paths keep their primary project-relative form, while Review diff fetching/matching can try an execution-root fallback only when the primary path has no matching hunk.
- Explicit `./` and `../` tool paths resolve from the current tool cwd; root files like `README.md` and sibling paths like `packages/shared.ts` are not blindly prefixed.
- Assisted dedupe keeps exact keys separate from fallback keys so root/scoped sibling pins can coexist; fallback refinement only matches entries that existed before the current update, avoiding order-dependent collapse within a single tool call.
- New tool writes normalize explicit cwd-relative input at the backend boundary; ReviewPanel also resolves candidates when consuming stored/transcript pins so older cwd-relative pins self-heal without corrupting canonical paths.
- Updated review tool docs/generated skill content to say paths are project-relative and `./`/`../` should be used when cwd-relative resolution is required.

Validation
- `bun test src/common/utils/review/assistedReview.test.ts src/node/services/tools/review_pane.test.ts src/browser/features/RightSidebar/CodeReview/ReviewPanel.assistedStats.test.ts src/node/services/workspaceService.test.ts -t "executeBash workspace path resolution|project-relative path normalization|applyReviewPaneUpdate|normalizeReviewPanelAssistedHunks|countUnreadAssistedHunks|buildReviewDiffPathFilter|buildReviewDiffPathFilterSpecs"`
- `make static-check`

Risks
- Low-to-moderate. The change touches shared review path resolution and executeBash cwd selection for repo-root/git command mode; tests cover default scoped cwd, repo-root cwd, canonical project-relative paths, execution-root fallbacks, assisted pin dedupe, and ReviewPanel matching.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$37.97`_


ammar-agent:
@codex review

ammar-agent:
@codex review

I folded in the cleanup to make Review's canonical path contract project-relative instead of introducing sub-project-relative terminology.

ammar-agent:
@codex review

Addressed the path-candidate issue: project-relative paths remain canonical, ambiguous plain paths keep exact primary matching, and execution-root fallbacks are only used for fetching/fallback matching.

ammar-agent:
@codex review

Addressed the dedupe ordering issue by separating exact assisted keys from execution-root fallback keys, so exact canonical updates win before fallback matching.

ammar-agent:
@codex review

Addressed the fallback dedupe miss by checking fallback mappings for incoming exact canonical keys before appending, with a regression test for the duplicate case.

ammar-agent:
@codex review

Updated dedupe semantics to keep incoming exact canonical paths distinct from ambiguous fallback pins, while still allowing incoming ambiguous paths to refine an existing exact fallback target. This preserves both files when root and scoped paths can coexist.

ammar-agent:
@codex review

Addressed the order-dependent fallback dedupe issue: fallback matching is now limited to entries that existed before the current update, so root/scoped sibling pins added in one call can coexist regardless of order.

PR #3372, opened 2026-05-23
🤖 fix: isolate fork partial snapshots

ammar-agent:
## Summary

Fixes `/fork` while the parent is mid-stream (including `task_await`) so forking is read-only with respect to the source workspace. Live partial assistant state is now materialized only into the forked workspace snapshot, and live `partial.json` state is no longer copied as active child state.

## Background

`WorkspaceService.fork()` previously called `historyService.commitPartial(sourceWorkspaceId)` when the source was streaming. During a long-running tool call like `task_await`, that could delete or rewrite the parent workspace's live `partial.json`, making the parent turn appear interrupted.

## Implementation

- Removed source partial commits from the fork path.
- Stopped copying `partial.json` as a raw session file into forks.
- Added a fork-only partial snapshot helper that reads the source partial, writes it to the new workspace, commits it there, and cleans up only the fork's partial file on failure.
- Added a regression test proving the source partial remains intact while the fork receives a finalized snapshot.

## Validation

- `bun test src/node/services/workspaceService.test.ts --test-name-pattern "WorkspaceService fork"`
- `make typecheck`
- `make static-check`

## Risks

Low-to-medium. This changes fork snapshot semantics for in-flight streams: forks no longer inherit a live `partial.json`; they inherit a finalized snapshot when a live partial exists. That is intentional to prevent child or parent disruption from shared live-stream state.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `high` • Cost: `unknown`_

### DIFF
--- a/src/node/services/workspaceService.test.ts
+++ b/src/node/services/workspaceService.test.ts
@@ -7411,6 +7411,124 @@ describe("WorkspaceService fork", () => {
       generateStableIdSpy.mockRestore();
     }
   });
+  test("fork snapshots persisted partials without mutating the source workspace", async () => {
+    const sourceWorkspaceId = "source-workspace";
+    const newWorkspaceId = "forked-workspace";
+    const sourceProjectPath = path.join(tempDir, "project");
+    const forkedWorkspacePath = path.join(sourceProjectPath, "fork-child");
+    const sourceMetadata: FrontendWorkspaceMetadata = {
+      id: sourceWorkspaceId,
+      name: "source-branch",
+      projectPath: sourceProjectPath,
+      projectName: "project",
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Please take another look. I now capture the source partial snapshot before fork checkout/copy I/O so a parent stream finishing mid-fork cannot make the child miss the latest visible assistant state.

ammar-agent:
@codex review

Please take another look. Persisted partial snapshots are now captured for normal forks regardless of active stream state, so interrupted/restarted workspaces keep the latest visible assistant state without copying live partial state into the child.

PR #3373, opened 2026-05-23
🤖 perf: enable React profiling for dev launches

ammar-agent:
## Summary

Default desktop dev launches now enable Mux's lightweight React render sampler so already-running dev instances can be inspected with component-level profiling data. Developers can still opt out with `MUX_PROFILE_REACT=0` when they need an uninstrumented baseline.

## Background

The live profiling probe showed CDP CPU/trace capture was available, but Mux's component-level React samples were disabled because `MUX_PROFILE_REACT` is latched by preload startup. Enabling it by default for standard dev desktop launches makes the usual workflow profile-ready without a restart cycle after discovering a performance issue.

## Implementation

- Added a Makefile default of `MUX_PROFILE_REACT ?= 1` and passed it through `make start`.
- Defaulted `dev-desktop-sandbox` Electron launches to `MUX_PROFILE_REACT=1` while preserving explicit opt-out.
- Updated the global API comment to cover dev profiling as well as automated perf tests.

## Validation

- `make -n start MUX_PROFILE_REACT=0`
- `make -n dev-desktop-sandbox`
- `make typecheck`
- `make static-check`

## Risks

Low risk. The sampler is capped and limited to selected render markers, but it can slightly perturb micro-benchmarks; developers can set `MUX_PROFILE_REACT=0` to compare against an uninstrumented renderer.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `13655{MUX_COSTS_USD:-unknown}`_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,12 @@
 #   It is automatically disabled in CI, test environments, and automation contexts.
 #   To manually disable telemetry, set MUX_DISABLE_TELEMETRY=1.
 
+# React profiling in development:
+# Default desktop dev launches opt into Mux's lightweight component render sampler so
+# already-running dev instances can be inspected without a restart. Override with
+# `make start MUX_PROFILE_REACT=0` when measuring an uninstrumented baseline.
+MUX_PROFILE_REACT ?= 1
+
 # Use PATH-resolved bash for portability across different systems.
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

PR #3374, opened 2026-05-23
🤖 perf: split chat pane workspace state

ammar-agent:
## Summary

Splits the chat pane's broad workspace-state dependency so transcript/message deltas no longer cascade through `WorkspaceShell` and the workspace header. The shell now subscribes to a stable shell-only status snapshot, while the full `WorkspaceState` subscription lives inside the chat content component that actually renders transcript/input state.

## Background

Live component profiling showed full-pane update bursts where `chat-pane`, `chat-pane.header`, `chat-pane.transcript`, and `chat-pane.input` all sampled together. The first optimization target was to stop routing every workspace-state bump through the shell/header boundary before attacking row-level transcript work.

## Implementation

- Added `WorkspaceShellStatus` plus `useWorkspaceShellStatus`, a stable selector for loading/hydration/starting gates.
- Updated `WorkspaceShell` to use that narrow selector instead of full `useWorkspaceState`.
- Split `ChatPane` into an outer shell/header wrapper and `ChatPaneContent`, which owns the full workspace-state subscription.
- Updated WorkspaceShell tests to mock the new shell-status selector.

## Validation

- `bun test src/browser/components/WorkspaceShell/WorkspaceShell.test.tsx`
- `make typecheck`
- `make static-check`

## Risks

Moderate-low. This is a structural render-boundary refactor with no intended UI behavior change. The main risk is divergence between the shell-only status derivation and the full `WorkspaceState` loading/hydration logic; the selector intentionally mirrors the existing gates and keeps the displayed-message scan only for init/hydration cases.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `41375{MUX_COSTS_USD:-unknown}`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -46,9 +46,9 @@ import { useOpenInEditor } from "@/browser/hooks/useOpenInEditor";
 import { usePersistedState } from "@/browser/hooks/usePersistedState";
 import {
   useWorkspaceAggregator,
+  useWorkspaceState,
   useWorkspaceUsage,
   useWorkspaceStoreRaw,
-  type WorkspaceState,
 } from "@/browser/stores/WorkspaceStore";
 import { WorkspaceMenuBar } from "../WorkspaceMenuBar/WorkspaceMenuBar";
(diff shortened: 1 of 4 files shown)


ammar-agent:
@codex review

PR #3375, opened 2026-05-23
🤖 fix: repair corrupt SSH worktree base repos

ammar-agent:
Summary
- Repair corrupt SSH shared base repos when snapshot refs are reusable but the object database is missing blobs/trees.
- Harden warm and slow worktree creation so missing-object checkout failures self-heal instead of failing workspace initialization.

Background
A recent SSH workspace path/layout change can leave a reusable remote project snapshot pointing at a shared bare repo whose refs and snapshot marker look valid, but whose object store is incomplete. Git then fails `worktree add` with errors such as `unable to read sha1 file ...` and `Could not reset index file to revision 'HEAD'`.

Implementation
- Validate `refs/mux-bundle/*` connectivity before reusing a remote snapshot.
- Repair missing objects additively by streaming a complete local pack into the remote base repo via `git unpack-objects -r`, avoiding deletion/recreation of `.mux-base.git` because existing worktrees share that gitdir.
- Classify warm-path missing-object failures and fall through to the repairing slow path.
- Retry slow-path `git worktree add` once after repair, with a fallback from corrupt `origin/<trunk>` to the local snapshot ref when available.

Validation
- `bun test src/node/runtime/SSHRuntime.retry.test.ts src/node/runtime/SSHRuntime.syncContract.test.ts`
- `TEST_INTEGRATION=1 bun x jest tests/runtime/runtime.test.ts -t "missing objects" --runInBand`
- `make typecheck`
- `make lint`
- `make fmt-check`
- `make static-check`

Risks
- Touches SSH workspace initialization, especially the warm fast path and slow worktree materialization path. The common healthy snapshot path pays one scoped connectivity check only when the slow path is already considering snapshot reuse; the warm fast path remains a single probe and only falls back on worktree-add failure.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$16.67`_

### DIFF
--- a/src/node/runtime/SSHRuntime.syncContract.test.ts
+++ b/src/node/runtime/SSHRuntime.syncContract.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, describe, expect, it, mock, spyOn } from "bun:test";
-import { execSync } from "node:child_process";
+import { execSync, spawnSync } from "node:child_process";
 import { mkdtemp, rm } from "node:fs/promises";
 import * as os from "node:os";
 import * as path from "node:path";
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

PR #3376, opened 2026-05-23
🤖 fix: scope review pane to sub-projects

ammar-agent:
Summary
- Scope the Review pane to a workspace's configured sub-project by default while keeping git commands rooted at the parent repository.
- Keep Assisted review as the explicit exception: agent-pinned hunks still fetch their own pathspecs even when they point outside the sub-project.
- Scope the file tree and untracked-files banner to the same sub-project path, and ignore stale selected-file filters outside the scope.

Background
- Sub-project workspaces run in a scoped cwd but share the parent repository checkout. The Review pane needs to diff from the git root for correct base/ref behavior, but showing every parent-repo change by default adds noise when the user is working in a sub-project.

Implementation
- Derive a repo-root pathspec from `subProjectPath` relative to `projectPath` and apply it to broad, non-assisted Review diff/file-tree loads.
- Preserve selected-file pathspec behavior only when the selection is inside the sub-project; otherwise fall back to the scoped pathspec.
- Leave Assisted Review pathspec grouping untouched so pinned hunks outside the sub-project can still surface.

Validation
- `bun test src/browser/features/RightSidebar/CodeReview/ReviewPanel.assistedStats.test.ts`
- `bun run typecheck`
- `make static-check`

Risks
- Low. The change is isolated to Review pane git pathspec construction and the untracked-files banner, with unit coverage for sub-project pathspec derivation, scoped broad diffs, stale parent file selections, and Assisted Review exceptions.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$22.46`_

### DIFF
--- a/src/browser/features/RightSidebar/CodeReview/ReviewPanel.assistedStats.test.ts
+++ b/src/browser/features/RightSidebar/CodeReview/ReviewPanel.assistedStats.test.ts
@@ -7,6 +7,7 @@ import {
   countUnreadAssistedHunks,
   getEffectiveReviewFrontendFilters,
   getEffectiveReviewIncludeUncommitted,
+  getReviewSubProjectPathspec,
   normalizeReviewPanelAssistedHunks,
 } from "./ReviewPanel";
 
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Please take another look. I now compare scoped selected-file filters against the repo-relative `selectedDiffPath`, so multi-project workspace-relative selections like `project-a/packages/api/...` still pass the `packages/api` sub-project scope check.

ammar-agent:
@codex review

Please take another look after the repo-relative selected-path fix.

PR #3379, opened 2026-05-24
🤖 feat: coalesce consecutive file_read/file_edit tool calls in transcript

ammar-agent:
## Summary

Add automatic coalescing for consecutive `file_read` and `file_edit_*` tool calls in the transcript. The first call in a run renders normally; when a second arrives, the slot converts in-place to a **Coalesce** row that summarizes the burst (e.g. _Read files src/App.tsx, src/main.ts_). Clicking the row expands the group and reveals each underlying tool call exactly as before.

## Background

Long agent turns frequently emit bursts of file reads/edits that push more useful content (assistant prose, terminal output) off-screen. The existing `bash_output` collapsing pattern already does this for repeated polls of the same process; this PR extends the same idea to file IO bursts behind a rule-based registry so additional kinds (web fetches, advisor, …) can opt in later.

## Implementation

- `src/browser/utils/messages/toolCoalescing.ts` — `computeToolCoalesceInfos(messages)` walks the transcript in one O(n) pass and returns per-index `{ kind, position: "head" | "member", filePaths, headIndex, totalCount }`. Groups of size ≥ 2 with the same `ToolCoalesceKind` (`file_read` or `file_edit`) qualify.
- `src/browser/features/Tools/CoalescedToolCall.tsx` — presentational summary row that mirrors a collapsed `FileReadToolCall`'s dimensions (same `ToolContainer` padding, same row height). That keeps the 1 → N transition a content swap rather than a row insertion, so users do not see a layout flash when the second member streams in mid-turn.
- `src/browser/components/ChatPane/ChatPane.tsx` — wired alongside the existing `bashOutputGroupInfos` flow:
  - precomputes `toolCoalesceInfos` once per snapshot,
  - hides members when the group is collapsed,
  - replaces the head's `MessageRenderer` with `CoalescedToolCall` when collapsed,
  - renders both the summary row and the head's normal tool call when expanded so the summary doubles as the collapse toggle.
  - owns the `expandedToolCoalesceGroups` `Set<headId>` and resets it on workspace switch (mirroring `expandedBashGroups`).

The summary uses past-tense copy matching the request (`Read files …` / `Wrote files …`). All `file_edit_*` variants (`replace_string`, `replace_lines`, `insert`) collapse under a single `file_edit` kind so a mixed burst still forms one block; the active registry only routes two of those today but the legacy variant is included for historical transcripts.

## Validation

- `bun test src/browser/utils/messages/toolCoalescing.test.ts` — 10 cases covering head/member layout, mixed file_edit variants, kind boundaries, intervening tool/user messages, multi-group transcripts, and unknown path fallback.
- `bun test src/browser/features/Tools/CoalescedToolCall.test.tsx` — 4 happy-dom cases for collapsed/expanded copy, `onToggle` wiring, and `aria-expanded` state.
- `bun test src/browser/features/Tools src/browser/features/Messages src/browser/utils/messages` — 612 pass, 0 fail.
- `make static-check` — clean.
- Two new Storybook stories (`TwoReads`, `ManyReads`, `TwoEdits`, `ExpandedReads`) under `App/Chat/Tools/CoalescedToolCall` for visual review.

## Risks

Render-path change in `ChatPane.tsx` map block — the same hot path that handles `bash_output` collapsing. Mitigations:

- The new branch is additive (early-returns null for hidden members, swaps the head's element when collapsed) and short-circuits to the existing render when no coalesce info is present, so non-file_read/file_edit traffic is unaffected.
- Expansion state is keyed by the head message ID, so it survives later additions to the same group without changing identity.
- `findTranscriptMessageElement` selects on `[data-message-id]`. The summary row deliberately omits that attribute so scroll-into-view targeting still resolves to real messages (head MessageRenderer when expanded; the user/assistant rows in all other cases).
- Existing tests that count `[data-testid="chat-message"]` (truncation, fork, task report relocation) all use bash/task/assistant fixtures, not file_read/file_edit, so they are unaffected.

---

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -32,8 +32,10 @@ import {
   shouldBypassDeferredMessages,
 } from "@/browser/utils/messages/messageUtils";
 import { computeTaskReportLinking } from "@/browser/utils/messages/taskReportLinking";
+import { computeToolCoalesceInfos } from "@/browser/utils/messages/toolCoalescing";
 import { BashCollapsedSummaryModeProvider } from "@/browser/features/Tools/BashCollapsedSummaryModeContext";
 import { BashOutputCollapsedIndicator } from "@/browser/features/Tools/BashOutputCollapsedIndicator";
+import { CoalescedToolCall } from "@/browser/features/Tools/CoalescedToolCall";
 import {
   getInterruptionContext,
   getLastMainRetryCandidateMessage,
@@ -338,6 +340,13 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
   // Track which bash_output groups are expanded (keyed by first message ID)
   const [expandedBashGroups, setExpandedBashGroups] = useState<Set<string>>(new Set());
 
+  // Track which tool-coalesce groups (file_read / file_edit bursts) the user
+  // has expanded. Keyed by the head message ID so the entry survives later
+  // additions to the same group without changing identity.
+  const [expandedToolCoalesceGroups, setExpandedToolCoalesceGroups] = useState<Set<string>>(
+    new Set()
+  );
+
   // Extract state from workspace state
 
   // Keep a ref to the latest workspace state so event handlers (passed to memoized children)
@@ -426,6 +435,12 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
   // Precompute bash_output grouping once per message snapshot so row rendering stays O(n).
   const bashOutputGroupInfos = useMemo(
     () => computeBashOutputGroupInfos(deferredMessages),
+    [deferredMessages]
+  );
+
+  // Precompute tool-coalesce metadata (file_read / file_edit bursts) once per snapshot.
+  const toolCoalesceInfos = useMemo(
+    () => computeToolCoalesceInfos(deferredMessages),
     [deferredMessages]
   );
 
@@ -681,6 +696,7 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
   useEffect(() => {
     setEditingState({ workspaceId, message: undefined });
     setExpandedBashGroups(new Set());
+    setExpandedToolCoalesceGroups(new Set());
   }, [workspaceId]);
 
   const handleChatInputReady = useCallback((api: ChatInputAPI) => {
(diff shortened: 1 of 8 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Pushed a follow-up that dedupes repeated file paths in the coalesced summary so a burst that touches `a.ts, a.ts, b.ts` displays as `a.ts, b.ts`. Dedupe is display-only; the underlying `filePaths` array still reflects every coalesced tool call.

ammar-agent:
@codex review

Addressed the P1 by refusing to coalesce a group that contains an interrupted (`isPartial`) member, so the InterruptedBarrier still renders exactly where it would have without coalescing. Added a regression test in `toolCoalescing.test.ts`.

ammar-agent:
@codex review

Addressed the P2 by generalising the per-member check: a group only coalesces when every member is `!isPartial` and `status === "completed"`. Pending/executing/failed/interrupted/redacted all force normal rendering so the per-row status/error UI stays visible. New regression tests cover failed and still-running members.

ammar-agent:
@codex review

The previous review was against the partial-only commit. The current head (`c48ea47fa`) generalises the per-member check to `!isPartial && status === "completed"`, so failed/executing/pending/interrupted/redacted members all force the group back to normal rendering. Regression tests for failed and executing members were added in the same commit.

ammar-agent:
Added a single Chromatic baseline for the new coalesced-tool-call row in `CoalescedToolCall.visual.stories.tsx` (`CollapsedReadFiles`). The original `CoalescedToolCall.stories.tsx` stays Chromatic-disabled and continues to own all the `play`-driven interaction coverage (click-to-expand, dedupe, `file_edit` verb, initial-expanded state).

To keep the global snapshot budget (`tests/ui/storybook/budget.test.ts`, ≤ 300) green, I opted `ProvidersSection.stories.tsx` out of Chromatic at the meta level. Every story in that file already drives a `play` function that asserts the relevant settings-panel behavior, and the layout has no animation/scroll-fade/state nuances worth a pixel baseline — so disabling visual snapshots there costs nothing in real coverage while freeing four snapshots.

Net delta: +1 (new visual story) − 4 (ProvidersSection) = −3 ⇒ estimated count goes from 302 (pre-existing main overage) to 299, comfortably under the 300 cap.

@codex review

ammario (approved):

PR #3383, opened 2026-05-25
🤖 fix: keep concurrent warning out of transcript tail

ammar-agent:
## Summary

Moves the concurrent local-agent warning out of the transcript tail stack and into the composer decoration lane. This keeps the warning pinned above the input while new transcript messages append, preventing the bottom-lock correction flash that happened when the warning was a live tail row. Adds Storybook coverage for the new composer-lane placement.

## Background

These flashes keep happening because the chat view has two competing jobs: preserve the user-visible scroll position, and continuously insert/mutate layout-affecting UI during streaming. When non-message chrome lives inside the transcript scroll area—especially at the tail—new messages insert above it. That changes `scrollHeight`, lets browser scroll anchoring pick its own anchor, and then our ResizeObserver/rAF bottom-lock corrects the scroll position. If the browser paints between those steps, users see a one-frame jump/flash.

Categorically, we should prevent this by making transcript layout ownership explicit: message rows belong in the transcript; persistent warnings, queues, todos, background process banners, and other composer-adjacent state belong in stable decoration lanes outside the transcript; overlays should not affect scroll height; and transient tail barriers should reserve/measure height before they appear during hydration/stream-start. New layout-affecting chat chrome should pick one of those lanes up front instead of ad hoc mounting inside the scrollport.

## Implementation

- Added `ConcurrentLocalWarningDecoration`, which wraps the existing warning copy with the same composer-decoration gutter and border treatment used by the input-adjacent stack.
- Passed the detected concurrent local streaming workspace name into `ChatInputPane`.
- Removed the concurrent warning from `transcriptTailItems`, so appended messages no longer insert above it while streaming.
- Added a Storybook story showing the warning pinned above the composer and outside the transcript flow.
- Left an inline rationale comment near the decoration placement to document the flash mechanism.

## Validation

- `make fmt`
- `bun test src/browser/components/ChatPane/LayoutStackLane.test.tsx`
- `make typecheck`
- `git diff --check`
- `env STORYBOOK_PORT=6017 make storybook-run CMD='bun x test-storybook --url http://127.0.0.1:6017 --testTimeout 30000 --includeTags concurrent-local-warning --no-cache'`
- `make static-check`

## Risks

Low. This changes placement of one warning from the transcript tail to the composer decoration stack. The warning condition and copy are unchanged; the main user-visible difference is that it is pinned above the composer instead of scrolling with the transcript tail.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$9.46`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -59,7 +59,7 @@ import { QueuedMessage } from "@/browser/features/Messages/QueuedMessage";
 import { CompactionWarning } from "../CompactionWarning/CompactionWarning";
 import { ContextSwitchWarning as ContextSwitchWarningBanner } from "../ContextSwitchWarning/ContextSwitchWarning";
 import {
-  ConcurrentLocalWarningView,
+  ConcurrentLocalWarningDecoration,
   useConcurrentLocalStreamingWorkspaceName,
 } from "../ConcurrentLocalWarning/ConcurrentLocalWarning";
 import { BackgroundProcessesBanner } from "../BackgroundProcessesBanner/BackgroundProcessesBanner";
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

PR #3389, opened 2026-05-26
🤖 fix: prevent compaction handoff notification pings

ammar-agent:
## Summary

Simplifies notify-on-response policy so compaction completions are treated as context-management handoffs instead of user-visible assistant responses. This prevents notification pings when compaction is followed by an auto-continue message, including background/unsubscribed workspace paths.

## Background

The prior policy split notification suppression across renderer stream state, compaction/follow-up message metadata, and workspace activity snapshots. When a compaction completed in the background, the renderer could receive only a generic `streaming: false` activity update and fall back to `completion: undefined`, which is notify-eligible.

## Implementation

- Makes `kind: "compaction"` completions never notify-eligible.
- Adds transient `isCompaction` activity snapshot classification from backend stream generations.
- Uses that backend classification for background completion metadata instead of relying on replayed chat state.
- Removes fragile compaction follow-up fields from the renderer notification metadata path while preserving synthetic internal-resume suppression for follow-up handoffs.
- Updates UI integration coverage to assert standalone compaction does not produce a notification.

## Validation

- `bun test src/browser/utils/messages/StreamingMessageAggregator.test.ts src/browser/utils/messages/StreamingMessageAggregator.sideQuestion.test.ts src/browser/stores/WorkspaceStore.test.ts src/node/services/workspaceService.test.ts`
- `TEST_INTEGRATION=1 bun x jest tests/ui/compaction/compaction.test.ts -t "compaction without continue message" --runInBand`
- `make typecheck`
- `make fmt-check`
- `make lint`
- `make static-check`

## Risks

Compaction completion notifications are now suppressed universally, including manual `/compact` without a follow-up. This is intentional to keep notification semantics simple and durable: only assistant response completions should ping.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$40.83`_

### DIFF
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -3144,6 +3144,39 @@ describe("WorkspaceStore", () => {
       });
     });
 
+    it("marks background compaction stops from activity snapshots as non-notifying completions", async () => {
+      const activeWorkspaceId = "active-workspace-compaction-snapshot";
+      const backgroundWorkspaceId = "background-workspace-compaction-snapshot";
+      const initialRecency = new Date("2024-01-05T12:00:00.000Z").getTime();
+      const initialSnapshot = createActivitySnapshot(initialRecency);
+      const releaseBackgroundCompletion = mockBackgroundActivityTransition(
+        backgroundWorkspaceId,
+        initialSnapshot,
+        [
+          {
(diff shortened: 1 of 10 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

PR #3390, opened 2026-05-26
🤖 fix: stabilize coalesced transcript tool rows

ammar-agent:
## Summary

Stabilizes the coalesced file-tool transcript rows introduced in #3379 so file-read/file-edit bursts keep a one-row baseline while streaming, and adds a real-browser Storybook layout contract to catch future row-height regressions.

## Background

The original coalescing path waited until every member was completed before hiding follow-up rows. That preserved non-completed status UI, but it meant a second `file_read` could briefly render as its own row and then disappear into the summary when it completed, causing exactly the layout flash the coalesce row was meant to avoid.

## Implementation

- Coalescing now includes `pending`, `executing`, and `completed` file-tool rows, with an aggregate status rendered on the summary row.
- Failed, interrupted, redacted, and partial rows still break coalescing so error details and interruption barriers remain visible without expansion.
- Tool containers no longer use `transition-all`; they only transition colors so transcript row size changes settle immediately for scroll pinning.
- Successful file-edit rows now default to the same collapsed one-line baseline as the coalesced edit summary, with stories that inspect diffs opting into `initialExpanded` explicitly.
- Coalesced edit summaries reserve the head row's trailing action slot when replacing a completed edit row, preventing the summary from shrinking relative to the row it replaced.
- Added a Chromatic-disabled Storybook layout-contract story that checks coalesced rows match single tool-row heights and reject layout-affecting transitions.

## Validation

- `bun test ./src/browser/utils/messages/toolCoalescing.test.ts ./src/browser/features/Tools/CoalescedToolCall.test.tsx ./src/browser/features/Tools/FileEditToolCall.test.ts`
- `bun test ./tests/ui/storybook/`
- `make storybook-build`
- `make test-storybook`
- `make static-check`

## Risks

Low-to-medium UX risk in the transcript tool rows: single successful file edits now start collapsed by default to maintain the same stable baseline used for coalescing. The diff remains one click away, and existing diff-focused Storybook coverage opts back into expanded rendering.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$33.38`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -1213,6 +1213,8 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
                           {renderCoalesceSummary && toolCoalesceGroup && toggleCoalesceGroup && (
                             <CoalescedToolCall
                               kind={toolCoalesceGroup.kind}
+                              reserveActionSlot={toolCoalesceGroup.reserveActionSlot}
+                              status={toolCoalesceGroup.status}
                               filePaths={toolCoalesceGroup.filePaths}
                               expanded={isToolCoalesceExpanded}
                               onToggle={toggleCoalesceGroup}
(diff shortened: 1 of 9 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

PR #3391, opened 2026-05-26
🤖 feat: show active mode instructions in Instructions pane

ammar-agent:
## Summary

The Instructions pane now leads with the **active mode's system prompt**, rendered with the same `uiColor` that drives the mode picker pill and the chat-input focus ring. Switching modes (e.g. plan ↔ exec) re-fetches the agent body and visibly retints the panel — so the Instructions tab now tracks the mode you're in.

## Background

The Instructions tab already exposes the AGENTS.md files and Chat Instructions that get injected into the system prompt, but the **mode prompt itself** — usually the largest part of the system message — was invisible. Users couldn't easily inspect what "Plan" vs "Exec" actually tells the model, and the tab gave no visual signal that the active mode matters.

This PR fills that gap by rendering the resolved agent body inside the same tab, styled with the mode's own color so the affordance is unmistakable.

## Implementation

**`ModeInstructionsPanel`** (`src/browser/components/InstructionsTab/ModeInstructionsPanel.tsx`)

- Subscribes to `useAgent()` and fetches the agent package via `api.agents.get`, keyed on the agent id so switching modes triggers a fresh fetch and a smooth color transition.
- The mode's `uiColor` (already inheritance-resolved on the backend) drives a 3px left accent bar, a subtle background tint, the icon swatch, the mode-name heading color, and the collapsed-preview pill. All tints use `color-mix(in srgb, <color> N%, transparent)` so they degrade gracefully across themes.
- Body renders through the shared `MarkdownRenderer` (same security-hardened pipeline as chat), capped at `max-h-[40vh]` with its own scrollbar so a multi-KB prompt never pushes the rest of the panel off-screen.
- Collapsible (expanded by default); collapsed state shows a one-line tinted preview of the first non-blank line. Empty/loading/error states all degrade gracefully.

**Shared icon helper** (`src/browser/utils/agentIcons.ts`)

- Extracted the agent → lucide icon map out of `AgentModePicker` so the picker pill and the new panel always show the same glyph for a given agent (Route for plan, SquareCode for exec, Bot fallback).

**Tab wiring** — `ModeInstructionsPanel` sits above `ChatInstructionsPanel`, which sits above the AGENTS.md file list. Reading the tab top-to-bottom now mirrors the order of contributions to the system prompt.

**Storybook + mock plumbing** — Extended `createMockORPCClient` with an `agentBodies` option so `agents.get` can return real markdown in stories. Added two Chromatic-snapshotted stories (`ExecMode`, `PlanMode`) covering the two built-in mode colors so visual regressions are caught.

## Validation

- `make static-check` (typecheck + lint + fmt-check + docs link check) ✅
- Snapshot-budget guard (`tests/ui/storybook/budget.test.ts`) updated implicitly — verified the two new snapshots fit under the cap (305 → would have blown the 303 cap, then trimmed by marking the non-color-relevant variants `disableSnapshot: true`).
- `bun x storybook build` ✅ — confirms the new story file is wired end-to-end.
- Regression checks: `AgentModePicker.test.tsx` (5/5), `agents.test.ts` (4/4).

## Risks

Low. The new panel is additive (sits above existing sections), reads from an IPC method that's already used by Settings → Tasks, and falls back gracefully when the agent has an empty body. The shared `getAgentIcon` extraction is a pure refactor with the same lookup as before; the picker's behavior is unchanged.

The only behavioral nuance worth flagging: `agents.get` returns the raw body of the agent's `.md` file, not the inheritance-resolved prompt that's actually injected into the system message. For built-in `plan`/`exec` they're identical; for custom agents with `prompt.append: true`, the user sees only the additions on top of the base. Surfacing the fully-resolved prompt would require a new IPC and is out of scope.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `xhigh` • Cost: `$10.76`_

### DIFF
--- a/src/browser/components/AgentModePicker/AgentModePicker.tsx
+++ b/src/browser/components/AgentModePicker/AgentModePicker.tsx
@@ -1,8 +1,8 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import { Bot, ChevronDown, Route, SquareCode } from "lucide-react";
-import type { LucideIcon } from "lucide-react";
+import { ChevronDown } from "lucide-react";
 
 import { useAgent } from "@/browser/contexts/AgentContext";
+import { getAgentIcon } from "@/browser/utils/agentIcons";
 import { CUSTOM_EVENTS } from "@/common/constants/events";
 import type { AgentDefinitionDescriptor } from "@/common/types/agentDefinition";
 import { normalizeAgentId as normalizeStoredAgentId } from "@/common/utils/agentIds";
@@ -41,17 +41,6 @@ interface AgentOption {
   aiDefaults?: { model?: string; thinkingLevel?: string };
   /** Whether this agent can be spawned as a subagent */
   subagentRunnable: boolean;
-}
-
-/** Maps well-known agent IDs to lucide icons for the dropdown */
-const AGENT_ICONS: Record<string, LucideIcon> = {
-  plan: Route,
-  exec: SquareCode,
-};
-const DEFAULT_AGENT_ICON: LucideIcon = Bot;
-
-function getAgentIcon(agentId: string): LucideIcon {
-  return AGENT_ICONS[agentId] ?? DEFAULT_AGENT_ICON;
 }
 
 export function formatAgentIdLabel(agentId: string): string {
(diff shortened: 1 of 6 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed in bb196b61: track the agent id of the loaded `pkg` and treat any mismatch as "not loaded yet" so a stale prompt body / token count is never shown under the wrong mode's color and name. The loading state now also re-engages on mode switch, while manual refresh of the same agent still keeps the existing body visible (no flash).

ammar-agent:
@codex review

Addressed in 0191593c: cache the `(agentId, workspaceId)` pair the loaded `pkg` belongs to and require *both* to match the current context before treating the body as fresh. A workspace switch with the same mode id now correctly falls back to the loading state instead of showing the previous workspace's prompt.

PR #3392, opened 2026-05-26
🤖 fix: make /btw side question stickiness dismissable

ammar-agent:
## Summary

Add an explicit "Dismiss sticky position" control to the `/btw` side-question header so users have a deterministic escape hatch from the transcript scroll-hold ("stickiness") that pins a side question to the top of the viewport.

## Background

The side-question feature uses a complex `useLayoutEffect`-based scroll-hold (see `sideQuestionScrollHold.ts` and the layout effect in `ChatPane.tsx`) that hands ownership of `scrollTop` between bottom-lock and the side branch as the side answer streams and as new main-stream rows accumulate below the branch. Recovery from that hold relies on a handful of heuristic release paths — wheel, `mousedown`, key press, touch, jump-to-bottom — each of which has been the source of repeated bugs (the side Q/A sticking at the bottom edge of the viewport, the hold re-arming on subsequent stream updates, the alignment fighting browser scroll anchoring, etc.).

Rather than try to make those heuristics perfect, this PR ships a user-controlled override.

## Implementation

- `ChatPane` keeps a `dismissedSideQuestionStickyIdsRef = useRef<Set<string>>` keyed by the side-question user message's `historyId`. The set is reset on workspace switch alongside the existing scroll-hold ref reset.
- The scroll-hold `useLayoutEffect` now consults that set:
  - if the currently-active hold target has been dismissed, the active ref is cleared immediately,
  - `shouldStartHold` is gated on `!isTargetDismissed(detectedTargetHistoryId)`, so detection cannot re-arm the hold for a dismissed branch on subsequent stream updates,
  - dismissal of the question row also covers the paired side-answer row (the existing hold logic uses either historyId as the target depending on which row is visible).
- A small `PinOff` icon button is rendered in the existing side-question header next to the "Side question" label, wrapped in the shared `TooltipIfPresent` surface (per the repo's `no-native-interactive-tooltips` lint rule). The button is only rendered when an `onDismissSticky` handler is supplied — so non-`/btw` user rows are unaffected, and existing stories/tests that render `UserMessage` standalone don't grow chrome.
- The handler is plumbed `ChatPane → MessageRenderer → UserMessage` as an optional callback, conditional on `msg.isSideQuestion === true`, so the dismiss UI only appears on `/btw` user rows.

## Validation

- `bun test src/browser/features/Messages/MessageRenderer.test.tsx src/browser/components/ChatPane/sideQuestionScrollHold.test.ts` — 31 pass, including 3 new tests covering (a) no button without handler, (b) button click forwards the `historyId`, (c) button never appears on non-side-question user rows.
- `make static-check` — passes (caught and resolved one lint hit on a native `title` attribute by switching to `TooltipIfPresent`).

## Risks

Low. The added state lives in a `ref`, is reset on workspace switch, and only gates the existing scroll-hold paths (it cannot start a hold that wouldn't have started before). The new UI control renders only inside the `/btw` header for messages where `isSideQuestion === true` and an opt-in callback is provided, so other user-message rendering paths (regular chat, goal-continuation, budget-limit, synthetic-auto, edited messages, queued previews, stories) are untouched.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-7` • Thinking: `high` • Cost: `$`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -494,7 +494,22 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
 
   const activeSideQuestionScrollHoldTargetRef = useRef<string | null>(null);
 
+  // User-dismissed side question stickiness: any historyId in this set is
+  // permanently exempt from the scroll-hold for the lifetime of the workspace
+  // session. This is the deterministic escape hatch from the various heuristic
+  // hold-release paths (wheel / mousedown / key / touch / jump-to-bottom), which
+  // have been the source of recurring bugs. The set is keyed by the side
+  // question user-message historyId, but is also consulted when the associated
+  // side-answer is the active target (the hold logic aligns to the question
+  // row in both cases).
+  const dismissedSideQuestionStickyIdsRef = useRef<Set<string>>(new Set());
+
   const clearActiveSideQuestionScrollHold = useCallback(() => {
+    activeSideQuestionScrollHoldTargetRef.current = null;
+  }, []);
+
+  const handleDismissSideQuestionSticky = useCallback((historyId: string) => {
+    dismissedSideQuestionStickyIdsRef.current.add(historyId);
     activeSideQuestionScrollHoldTargetRef.current = null;
   }, []);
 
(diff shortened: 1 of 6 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

PR #3393, opened 2026-05-26
🤖 fix: stabilize /btw transcript placement

ammar-agent:
Summary
- Fixes recurring `/btw` transcript tail stickiness by making side-question display placement explicit and by limiting scroll ownership to valid interrupted branches.

Background
- `/btw` rows are durable history, but interrupted `/btw` branches are projected into the interrupted assistant message by the renderer. The previous display walk decided inline ownership during rendering, and scroll hold logic only looked at broad `isSideQuestion` / `isSideAnswer` flags. That made standalone or stale side-question rows eligible for sticky scroll behavior.

Implementation
- Adds display-only `sideQuestionBranch` placement metadata for `/btw` user and assistant rows.
- Precomputes a side-question display plan before rendering transcript rows so anchored `/btw` rows are emitted exactly once inside the interrupted assistant split.
- Validates `interruptedHistorySequence` when present to avoid splitting stale anchors.
- Makes side-question scroll hold require interrupted placement and expire once the side answer and interrupted main message have both settled.

Validation
- `bun test src/browser/utils/messages/StreamingMessageAggregator.sideQuestion.test.ts`
- `bun test src/browser/components/ChatPane/sideQuestionScrollHold.test.ts`
- `bun test src/browser/stores/WorkspaceStore.test.ts --test-name-pattern "/btw|side-question|side answer"`
- `bun run typecheck`
- `make static-check`

Risks
- Moderate localized risk around transcript projection and `/btw` scroll behavior. Legacy side-question answers without `questionMessageId` still use the existing adjacency fallback; stale or mismatched anchors now render chronologically instead of splitting an unrelated assistant row.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$31.98`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -98,8 +98,6 @@ import {
 import {
   findActiveSideQuestionScrollHoldTarget,
   findSideQuestionScrollHoldTarget,
-  getSideQuestionScrollHoldScrollportStartTop,
-  isSideQuestionScrollHoldBottomClamped,
   type SideQuestionScrollHoldState,
 } from "./sideQuestionScrollHold";
 import { recordSyntheticReactRenderSample } from "@/browser/utils/perf/reactProfileCollector";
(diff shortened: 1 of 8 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the scroll-hold initialization issue by tracking standalone/stale /btw question IDs as seen, and added a regression test for rows that first render standalone and later become interrupted.

ammar-agent:
@codex review

Addressed the hidden-owner case by excluding hidden synthetic assistant anchors from the inline /btw display plan, so side-question rows remain visible chronologically when their split owner will not render. Added regression coverage.

PR #3395, opened 2026-05-27
🤖 fix: preserve browser file URL navigation

### LINKED ISSUE #3382: Browser pane prepends https:// to file:// URLs, breaking local file navigation
## Summary

When I navigate the cmux browser pane to a `file://` URL (e.g. by clicking a link in a terminal pane, or pasting into the address bar), cmux strips the `file://` scheme and prepends `https://` instead. The browser then tries to resolve the path's first segment as a hostname, which fails.

## Repro

1. In a terminal pane, output a `file://` URL such as `file:///Users/me/some-file.html` (e.g. an `Open it at file:///…` message from a tool).
2. Click the URL, or copy it and paste into the browser pane's address bar.

## Expected

Browser loads the local file at `/Users/me/some-file.html`.

## Actual

Address bar shows `https://Users/me/some-file.html` (or, in my case, `https://raw/pipeline/observability-diagrams/exec-summary.html` — the leading `/Users/admin/berkshire-vault/` was also stripped along with the scheme).

Page shows: **"Can't reach this page — `https://…` refused to connect. Check that a server is running on this address."**

## Why this matters

Local HTML previews are a common workflow — agents generate static HTML reports/dashboards into the workspace and the natural next step is to open them in the cmux browser pane. With this bug, the only workaround is to open them in an external browser, which loses the in-cmux workflow.

## Environment

- cmux desktop app (recent build, May 2026)
- macOS (Darwin 24.1.0)

coadler:
Summary
- Fixes browser pane navigation for explicit `file://` URLs by preserving the local-file target end-to-end.

Background
- Closes #3382. The browser pane could route local file URLs through `agent-browser open`, where older CLI URL parsing normalized `file:///...` into an `https://...` URL before the daemon saw it.

Implementation
- Routes explicit `file:` opens directly to the selected agent-browser daemon with a `navigate` command, bypassing CLI URL normalization only for local-file URLs.
- Retries transient direct-daemon transport failures so file URL navigation keeps the same stale-socket/restarting-daemon resilience as the agent-browser CLI path.
- Shares agent-browser socket path resolution between session discovery and the new daemon command client.
- Keeps existing CLI behavior for non-file opens plus back/forward/reload/get-url commands.

Validation
- `bun test src/node/services/browser/BrowserControlService.test.ts`
- `bun test src/browser/features/RightSidebar/BrowserTab/BrowserToolbar.test.tsx`
- `bun test src/node/services/browser/agentBrowserCommandClient.test.ts`
- `bun test src/node/services/browser/AgentBrowserSessionDiscoveryService.test.ts`
- `make typecheck`
- `make lint`
- `make fmt-check`
- `make static-check`

Risks
- Low to moderate: the change adds a direct daemon protocol path, but scopes it to explicit `file:` opens and leaves existing browser control commands on their prior CLI path.

---

### DIFF
--- a/src/browser/features/RightSidebar/BrowserTab/BrowserToolbar.test.tsx
+++ b/src/browser/features/RightSidebar/BrowserTab/BrowserToolbar.test.tsx
@@ -135,6 +135,26 @@ describe("BrowserToolbar", () => {
     });
   });
 
+  test("preserves file URLs when submitting navigation", async () => {
+    const { onSetPendingUrl, getByLabelText } = renderToolbar({
+      pendingUrl: "file:///Users/me/report.html",
+    });
+    const input = getByLabelText("Browser URL") as HTMLInputElement;
+
+    input.focus();
+    fireEvent.keyDown(input, { key: "Enter" });
+
+    expect(onSetPendingUrl).toHaveBeenCalledWith("file:///Users/me/report.html");
+    await waitFor(() => {
+      expect(controlMock).toHaveBeenCalledWith({
+        workspaceId: "workspace-1",
+        sessionName: "session-a",
+        action: "open",
+        url: "file:///Users/me/report.html",
+      });
+    });
+  });
+
   test("shows open command errors returned by the browser control API", async () => {
     controlMock.mockResolvedValueOnce({ success: false, error: "Navigation failed" });
     const { onSetPendingUrl, getByLabelText, getByText } = renderToolbar({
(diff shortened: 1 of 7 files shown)


coadler:
@codex review

coadler:
@codex review

Addressed the retry feedback by adding bounded retries for transient direct daemon transport failures and a regression test that exercises a closed-before-response retry.

PR #3397, opened 2026-05-27
🤖 fix: correct browser preview click offset

coadler:
Summary
- Fix Browser tab pointer coordinate mapping when agent-browser frame images are shorter than Chrome's reported outer viewport.
- Add regression coverage for decoded frame dimensions driving click coordinates.

Background
- During live testing, clicks in the Browser preview landed below the cursor because stream metadata reported an outer height (for example 720px) while the displayed JPEG frame represented the page viewport (for example 633px).

Implementation
- Continue using decoded frame image dimensions for rendered hit testing, and derive the Y coordinate space from that frame aspect when a frame image size is available.
- Preserve existing capped-stream behavior while correcting the outer-height/page-frame mismatch.

Validation
- `bun test src/browser/features/RightSidebar/BrowserTab/BrowserTab.test.tsx src/browser/features/RightSidebar/BrowserTab/BrowserToolbar.test.tsx src/browser/features/RightSidebar/BrowserTab/BrowserViewport.test.tsx src/browser/features/RightSidebar/BrowserTab/useBrowserBridgeConnection.test.tsx`
- `nix develop --command make static-check`

Risks
- Low. The change is scoped to Browser preview pointer coordinate mapping and keeps the existing fallback for frames without decoded image dimensions.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `108995{MUX_COSTS_USD:-0.00}`_

### DIFF
--- a/src/browser/features/RightSidebar/BrowserTab/BrowserViewport.test.tsx
+++ b/src/browser/features/RightSidebar/BrowserTab/BrowserViewport.test.tsx
@@ -101,6 +101,72 @@ describe("BrowserViewport", () => {
     ).toBeNull();
   });
 
+  test("maps decoded frame height when Chrome outer height differs from the page viewport", () => {
+    expect(
+      mapDomPointToViewport(
+        640,
+        317,
+        { left: 0, top: 0, width: 1280, height: 633 },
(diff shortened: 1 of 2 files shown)


coadler:
@codex review

PR #3398, opened 2026-05-27
🤖 fix: stop forwarding OpenRouter model catalog

coadler:
Summary
- Fixes #3119 by keeping Mux's configured OpenRouter model catalog out of request-level `extraBody`.

Background
- OpenRouter caps request-level model fallback arrays at three entries, while Mux provider config can store longer model catalogs for routing/UI metadata.

Implementation
- Destructure `models` out of the OpenRouter provider config before converting passthrough options into `extraBody`.
- Preserve supported OpenRouter routing options under `extraBody.provider`.
- Add regression coverage for a four-model OpenRouter config.

Validation
- `bun test src/node/services/providerModelFactory.test.ts`
- `nix develop --command make static-check`

Risks
- Low: scoped to OpenRouter provider option assembly. Existing OpenRouter routing options remain covered by the regression assertion.

---

### DIFF
--- a/src/node/services/providerModelFactory.test.ts
+++ b/src/node/services/providerModelFactory.test.ts
@@ -1225,6 +1225,45 @@ describe("ProviderModelFactory routing", () => {
     });
   });
 
+  it("omits configured model catalog from OpenRouter request body", async () => {
+    await withTempConfig(async (config, factory) => {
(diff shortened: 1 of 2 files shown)


coadler:
@codex review

PR #3402, opened 2026-05-27
🤖 fix: confirm pre-boundary message edits

ethanndickson:
## Summary
Allow editing user messages from before the latest compaction or context-reset boundary again, while requiring a confirmation before the edited send proceeds.

## Background
The edit pipeline already truncates chat history from the edited message, which naturally discards any later compaction/reset boundary and summary. The recent editability gate prevented that older behavior entirely, so users could no longer revise pre-compaction turns.

## Implementation
Pre-boundary user messages are editable again, and the edit state carries whether the source row was before the latest context boundary. When sending one of those edits, the composer opens a confirmation dialog explaining that the send will discard the latest compaction/reset summary and every message after the edited one; confirming reruns the normal send path with an internal confirmation bypass.

The backend edit truncation target lookup now falls back to full-history search when the edited message is outside the active context window, so immediately preceding file/skill snapshot rows are still removed before appending the replacement message.

## Risks
The intended destructive behavior is restored for users who confirm the dialog. The main regression risk is around the send affordance for pre-boundary edits; normal edits, side-question rows, goal continuation rows, local command output edit gates, and snapshot cleanup for post-boundary edits remain covered.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$9.77`_

### DIFF
--- a/src/browser/features/ChatInput/index.tsx
+++ b/src/browser/features/ChatInput/index.tsx
@@ -222,6 +222,15 @@ const MAX_PERSISTED_ATTACHMENT_DRAFT_CHARS = 4_000_000;
 
 export type { ChatInputProps, ChatInputAPI };
 
+interface SendOverrides {
+  queueDispatchMode?: QueueDispatchMode;
+  goalInterventionPolicy?: GoalInterventionPolicy;
+}
+
+interface InternalSendOverrides extends SendOverrides {
+  skipBoundaryEditConfirmation?: boolean;
+}
+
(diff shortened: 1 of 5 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review
Please take another look.

ethanndickson:
@codex review
Please take another look.

ethanndickson:
@codex review

Rebased onto origin/main and squashed the branch into one commit.

PR #3403, opened 2026-05-27
Prevent XSS via unsafe href protocols in WebFetchToolCall

ThomasK33:
## Summary
Add URL validation to prevent javascript: and other unsafe protocol XSS attacks when rendering fetched URLs as clickable links in the WebFetchToolCall component.

## Changes
- **New `isSafeHref()` function**: Validates that URLs use only http: or https: protocols before rendering as clickable `<a>` tags. Uses `URL` constructor for robust parsing and safely catches malformed URLs.
- **Conditional link rendering**: URLs that fail the safety check are now rendered as plain `<span>` elements instead of clickable links, preventing potential XSS while still displaying the URL text.

## Implementation Details
The validation happens at render time in the URL display section. Safe URLs render as before with full link functionality; unsafe URLs (including malformed ones) degrade gracefully to plain text, maintaining readability without the security risk.

https://claude.ai/code/session_015vY8c55jcKPoPNKm2264uC

### DIFF
--- a/src/browser/features/Tools/WebFetchToolCall.tsx
+++ b/src/browser/features/Tools/WebFetchToolCall.tsx
@@ -36,7 +36,7 @@ interface NormalizedResult {
 }
 
 /**
- * Only allow http/https URLs as clickable hrefs to prevent javascript: XSS.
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

PR #3405, opened 2026-05-27
🤖 fix: guard direnv allow in init hook

coadler:
## Summary

Guard the `.mux/init` `direnv allow` step so workspace bootstrap only runs it when `direnv` is installed.

## Background

The init hook should support contributors who use direnv-managed toolchains without breaking environments where `direnv` is not present.

## Validation

- `bash -n .mux/init`

---

_Generated with `mux` • Model: `unknown` • Thinking: `unknown` • Cost: `$0.00`_

### DIFF
--- a/.mux/init
+++ b/.mux/init
@@ -2,7 +2,11 @@
 set -euo pipefail
 
 echo "Installing dependencies with bun..."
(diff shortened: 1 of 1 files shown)


coadler:
@codex review

PR #3406, opened 2026-05-27
🤖 fix: repair Nix native rebuild tooling

coadler:
Summary
- Add Python to the Nix dev shell so node-gyp can rebuild Electron native modules.
- Make `make rebuild-native` recover when a previous failed optional install left `node-pty` missing from `node_modules`.
- Harden native-module presence checks and Nix format checking in temporary directories.

Background
This fixes the NixOS native rebuild path where `node-gyp` first failed without Python and then `@electron/rebuild` failed with ENOENT for `node_modules/node-pty/package.json` after the optional dependency was left missing.

Validation
- `make fmt-nix-check`
- `nix develop --command make rebuild-native`
- `make static-check`

---


coadler:
@codex review

PR #3412, opened 2026-05-28
🤖 refactor: scope ask_user_question to genuinely open decisions

ammar-agent:
## Summary

Tightens guidance for the `ask_user_question` tool so it is reserved for genuinely balanced decisions, not used as a confirmation step when the agent already has a recommended answer.

## Background

`ask_user_question` was being invoked too aggressively — it would surface choices even when the agent already had a clear recommendation. If the agent has a recommended option, asking the question defeats its purpose: it should just proceed (stating the assumption) and reserve the tool for decisions that actually require user-specific context the agent cannot derive itself.

## Implementation

Updated the two sources of truth and regenerated all derived artifacts:

- **Tool description** (`src/common/utils/tools/toolDefinitions.ts`): replaced the old `"MUST be used when you need user clarification"` framing with explicit gating — use only for genuinely balanced decisions hinging on user-specific context/preference/info; do **not** use when a reasonable recommendation already exists; keep options genuinely open rather than steering toward a "recommended" choice.
- **Plan-mode prompt** (`src/node/builtinAgents/plan.md`): the agent now resolves uncertainty itself when it has a reasonable default/recommendation (adopt + note the assumption), and only reaches for `ask_user_question` on genuinely open decisions.

The tool is already removed from exec/desktop agents, so it is effectively plan-mode-only.

Regenerated (auto-generated, kept in sync):

- `builtInAgentContent.generated.ts` (from `plan.md`)
- `docs/agents/index.mdx` (from `plan.md`)
- `builtInSkillContent.generated.ts` (mux-docs offline snapshot embedding the docs)

## Validation

- `bun scripts/gen_docs.ts check` — generated artifacts in sync
- `make static-check` — passed

No tests were added: the change is prose-only and any assertion on the wording would be tautological.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$1.80`_


ammar-agent:
@codex review

PR #3413, opened 2026-05-28
🤖 fix: disable native spellchecker in Electron windows

ammar-agent:
## Summary

Disable Electron's native spellchecker across the renderer windows. mux is a coding tool whose text inputs are full of code, file paths, and identifiers, so the OS spellchecker just produces noisy red squiggles with no value.

## Implementation

Added `spellcheck: false` to `webPreferences` for:

- The main app window (`src/desktop/main.ts`) — where all chat/text input happens.
- The popped-out terminal window (`src/desktop/terminalWindowManager.ts`) — content is code.

The splash window has no text inputs, so it was left untouched. Per-element `spellCheck={false}` attributes that already existed on individual React inputs remain (now redundant but harmless).

## Validation

- `make static-check` passes (eslint, typecheck for both app + main tsconfig projects, formatting, docs links).

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$0.43`_

### DIFF
--- a/src/desktop/main.ts
+++ b/src/desktop/main.ts
@@ -883,6 +883,9 @@ function createWindow() {
       nodeIntegration: false,
       contextIsolation: true,
       preload: path.join(__dirname, "../preload.js"),
+      // Disable the native spellchecker: mux is a coding tool where inputs are
+      // full of code, paths, and identifiers that trigger noisy red squiggles.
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3415, opened 2026-05-29
🤖 fix: require provider stream terminal events

ethanndickson:
Summary

Requires provider stream completion to be witnessed by the AI SDK `finish` part before Mux finalizes an assistant message. If a non-empty stream closes before that terminal event, Mux now persists a retryable `stream_truncated` partial instead of committing partial text to `chat.jsonl` as a successful response.

Background

This mirrors the stream-terminal invariant from coder/coder#25074 and coder/fantasy#33: Anthropic streams must reach `message_stop`, and OpenAI Responses streams must reach a terminal lifecycle event (`response.completed`, `response.incomplete`, or `response.failed`) before the response is considered complete. A clean network EOF can be a proxy or provider drop and is not enough to prove semantic completion.

Implementation

`StreamManager` now records the terminal `finish` part as the proof required for the success path and classifies non-empty missing-terminal completions as retryable `stream_truncated` errors. Empty completions keep the existing safe internal retry path. The Copilot Responses adapter now enforces the same Responses lifecycle locally, maps `response.incomplete` to a length finish, and surfaces `response.failed` as an error instead of falling through to EOF handling.

Already-streamed content is preserved, not discarded

If the stream produced output before closing without a terminal event, that output is **not** thrown away. The final `flushPartialWrite` runs before the truncation guard fires, and `persistStreamError` writes a `MuxMessage` to `partial.json` whose `parts` array contains every streamed part accumulated so far (`metadata.partial = true`, `metadata.errorType = "stream_truncated"`). The success-path call to `historyService.updateHistory` is unreachable in this branch, so nothing partial leaks into `chat.jsonl`. The frontend surfaces the partial turn with a retry barrier; the user sees the streamed text and can retry, and because `stream_truncated` is omitted from `NON_RETRYABLE_STREAM_ERRORS`, retry is enabled by default. The new `streamManager.test.ts` regression test pins this: after a truncated stream, `partial.parts` still contains the streamed text-delta and `partial.metadata.errorType === "stream_truncated"`.

Risks

This makes the success path stricter for every provider routed through `streamText`. If a provider adapter fails to emit a `finish` part for a genuinely complete response, Mux will now surface a retryable partial instead of committing it. That is intentional for Anthropic and OpenAI Responses and safer than silently finalizing truncated output.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$3.95`_

### DIFF
--- a/src/common/orpc/schemas/errors.ts
+++ b/src/common/orpc/schemas/errors.ts
@@ -40,6 +40,7 @@ export const StreamErrorTypeSchema = z.enum([
   "runtime_not_ready", // Container/runtime doesn't exist or failed to start (permanent)
   "runtime_start_failed", // Runtime is starting or temporarily unavailable (retryable)
   "empty_output", // Provider ended the stream without any assistant-visible output
+  "stream_truncated", // Provider stream closed before its terminal finish event
   "max_output_tokens", // Provider truncated the response at max_tokens (finishReason: "length")
   "unknown", // Catch-all
 ]);
--- a/src/node/services/copilot/copilotResponsesLanguageModel.test.ts
+++ b/src/node/services/copilot/copilotResponsesLanguageModel.test.ts
@@ -378,6 +378,98 @@ describe("CopilotResponsesLanguageModel", () => {
     });
   });
 
+  it("emits an error when the SSE stream closes before a terminal event", async () => {
+    restoreFetchers.push(
+      mockFetch(() =>
+        Promise.resolve(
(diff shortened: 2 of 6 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Pushed a follow-up test-only fix for the unit-test mocks that needed to emit the terminal finish event under the new invariant.

PR #3431, opened 2026-05-30
🤖 feat: add dynamic workflows

ThomasK33:
## Summary

Adds first-class dynamic workflows behind the experiment gate: reusable workflow definitions, durable workflow runs, sandboxed JavaScript orchestration, structured subagent report validation, workflow slash invocation, chat run cards, scratch workflow promotion, and the built-in `deep-research` workflow.

## Background

Workflows give Mux a durable orchestration layer distinct from agent skills: skills remain guidance/reference, while workflow runs own control flow, replay, child task fan-out, and resumable state.

## Implementation

- Added workflow domain schemas/types, discovery precedence, project-trust gating, and durable run storage.
- Added a QuickJS-backed workflow runner with conductor-only primitives, replay keys, structured output validation, parallel agent fan-out, interrupt/resume, and crash auto-resume.
- Integrated workflow invocation through slash suggestions and the `workflow_run` tool, including durable run-card projection on workspace reload.
- Added scratch workflow promotion and the built-in `deep-research` showcase workflow.

## Validation

- `all-workflow-tests` passed for workflow schemas, JSON schema validation, slash suggestions, run-card UI, workflow tool/router/store/runner/service/task adapter, built-in Deep Research, subagent report artifacts, agent report, and task service coverage.
- `make static-check` passed after rebasing on latest `origin/main`.
- Dogfooding captured workflow suggestions, run-card creation/reload/interrupt/resume, scratch promotion, and trusted-vs-untrusted workflow suggestions in local `dogfood-output/workflows` evidence.

## Risks

This touches chat input, task reporting, and workflow execution paths. The feature is experiment-gated, project-local execution is guarded by Project Trust, and durable replay/validation behavior has targeted test coverage.

---

<details>

# Implementation Plan: Dynamic Workflows

## Context and target

This plan implements the RFC committed as `rfc/20260529_dynamic-workflows.md` and the glossary updates in `CONTEXT.md`. The implementation should introduce first-class **Workflow Definitions** and durable **Workflow Runs** for Mux, with conductor-only JavaScript orchestration, report-time structured task outputs, a lightweight chat run card, skill-like discovery, existing Project Trust integration, scratch workflow promotion, and built-in Deep Research as the first showcase.

Primary design thesis:

- **Goals are loops; workflows are control flow inside those loops.**
- **Workflow Run = durable orchestration; Task = delegated execution.**
- **Agent Skill = guidance/reference, not executable workflow code.**
- **Workflow Resume** should continue the same run identity from durable state; **Restart** is not the primary action.

## Recommended approach

Build a first-class `WorkflowService` layer rather than stretching `code_execution` into a product feature.

**Net product LoC estimate:** rough, low-confidence estimate of ~6,500–10,500 LoC, excluding tests, docs, stories, generated snapshots, and built-in workflow content. Durable replay, child-task resume, structured reporting, and UI integration may push this higher. Re-estimate after the thin vertical slice and durable replay spike.

### Why this approach

- It keeps PTC as sandbox substrate, not product API.
- It gives workflows their own durable run identity, journal, replay semantics, and UI events.
- It preserves the existing task/sub-agent system as the execution substrate.
- It keeps project-local executable workflows under existing **Project Trust**.
- It leaves room for later dashboard, TypeScript authoring, nested workflows, and workflow-level budgets.

## Files and subsystems likely involved

The exact file names can shift during implementation, but expect these areas:

- Common types/schemas
  - `src/common/types/workflows.ts` (new)
  - `src/common/orpc/schemas/workflow.ts` (new)
  - `src/common/orpc/schemas/api.ts`
  - `src/common/types/stream.ts`
  - `src/common/config/schemas/taskSettings.ts`
- Node workflow services
  - `src/node/services/workflows/WorkflowDefinitionStore.ts` (new)
  - `src/node/services/workflows/WorkflowRunStore.ts` (new)
  - `src/node/services/workflows/WorkflowRunner.ts` (new)
  - `src/node/services/workflows/WorkflowService.ts` (new)
  - `src/node/services/workflows/builtInWorkflowDefinitions.ts` (new)
  - `src/node/services/ptc/*` or a workflow-specific QuickJS wrapper
  - `src/node/orpc/router.ts`
- Task/report integration
  - `src/common/utils/tools/toolDefinitions.ts`
  - `src/node/services/tools/agent_report.ts`
  - `src/node/services/tools/task.ts`
  - `src/node/services/tools/task_await.ts`
  - `src/node/services/taskService.ts`
  - `src/node/services/subagentReportArtifacts.ts`
- Invocation and discovery
  - `src/browser/features/ChatInput/utils.ts`
  - `src/browser/utils/agentSkills/*` or parallel workflow-reference utilities
  - `src/browser/utils/slashCommands/*`
  - `src/node/services/agentSkills/*` only as a reference pattern, not as executable storage
- UI
  - `src/browser/features/Tools/WorkflowRunToolCall.tsx` (new)
  - `src/browser/features/Tools/shared/*`
  - `src/browser/utils/messages/StreamingMessageAggregator.ts`
  - `src/browser/utils/messages/displayedMessageBuilder.ts`
  - `src/browser/features/ChatInput/CommandSuggestions.tsx`
- Built-ins
  - `src/node/builtinWorkflows/deep-research.js` (new; exact location TBD)
  - generation script equivalent if built-in workflow content is generated like built-in skills

## Phase 0 — Safety scaffolding and contracts

**Goal:** Create the minimal type/schema foundation so later phases have a stable vocabulary.

### Work

1. Add common workflow domain types:
   - `WorkflowDefinitionDescriptor`
   - `WorkflowDefinitionScope` (`project`, `global`, `built-in`, `scratch`)
   - `WorkflowRunStatus`
   - `WorkflowRunEvent`
   - `WorkflowRunRecord`
   - `WorkflowStepRecord`
   - `WorkflowResult`
   - `StructuredTaskOutput` wrapper types
2. Add run/event status constants rather than duplicating strings.
3. Add defensive assertions for IDs, non-empty paths, run status transitions, and event ordering.
4. Add a workflow experiment/feature gate that can disable workflow tools, discovery, UI surfaces, and built-in workflow exposure while the feature is dogfooded.
5. Add schema tests for workflow types and feature-gate/default behavior.

### Acceptance criteria

- Workflow type schemas round-trip valid records and reject malformed records.
- Workflow feature gate defaults to off or experiment-controlled for developer-facing rollout.
- No runtime behavior changes yet beyond type/schema/feature-gate plumbing.

### Quality gate

- Run targeted schema/type tests.
- Run `make typecheck` after this phase.

**Phase LoC estimate:** ~250–450 product LoC.

## Phase 1 — Workflow definition discovery and storage

**Goal:** Discover reusable workflow definitions separately from skills while preserving skill-like ordering and trust intuition.

### Work

1. Implement `WorkflowDefinitionStore` with roots:
   - project-local: `.mux/workflows/<name>.js`
   - global user-private: `~/.mux/workflows/<name>.js`
   - built-in definitions shipped with Mux
   - scratch definitions referenced only by run records
2. Mirror skill naming rules where practical: `^[a-z0-9]+(?:-[a-z0-9]+)*$`, 1–64 chars.
3. Implement precedence:
   - project-local wins if project is trusted
   - global next
   - built-in fallback
4. Gate project-local discovery/execution by existing Project Trust.
5. Decide UI-facing descriptor fields:
   - name
   - description
   - scope/source
   - project/global/built-in indicator
   - whether executable in current trust state
6. Avoid `.agents/workflows` in v1 unless a future cross-tool standard exists.
7. Add read/list APIs for workflow definitions, initially internal service methods; expose oRPC later.

### Acceptance criteria

- Project-local workflow definitions are not executable for untrusted projects.
- Project-local definitions override global/built-in only when trusted.
- Global definitions override built-ins.
- Built-ins are available without project trust.
- Scratch workflow definitions do not appear in reusable discovery.

### Quality gate

- Unit tests for discovery roots, precedence, trust gating, invalid names, missing files, and duplicate names.
- Run `make typecheck` and targeted workflow definition tests.

**Phase LoC estimate:** ~600–950 product LoC.

## Phase 2 — Durable workflow run store and event journal

**Goal:** Persist enough workflow run state to render run cards, resume after restart, and avoid duplicate completed work.

### Work

1. Implement `WorkflowRunStore` under workspace/session storage.
2. Use a simple file-backed layout for v1, for example:

   ```text
   ~/.mux/sessions/<workspace-id>/workflows/<run-id>/
     run.json
     definition.js
     events.jsonl
     steps.jsonl
   ```

   Exact paths can change if implementation finds a better fit, but keep run records workspace/session-scoped.
3. Persist:
   - run ID
   - parent workspace ID
   - workflow definition identity/source
   - captured executable script content and content hash for every run, including reusable project/global/built-in definitions, so resume is not affected by later definition edits unless an explicit migration is chosen
   - args
   - status
   - phase/log/task/validation/result/error events
   - step records keyed by stable step ID + normalized input identity
   - spawned task IDs and accepted report artifacts
4. Add a per-run mutex/lease so only one runner can replay or resume a workflow run at a time; include stale lease recovery for process crashes.
5. Define definition-change behavior: editing a reusable workflow definition affects new runs only. Existing runs resume from captured script content/hash unless a future explicit migration/restart action is introduced.
6. Make writes atomic enough for crash resilience:
   - append events defensively
   - validate JSON on read
   - ignore or quarantine malformed records rather than bricking the workspace
7. Add a replay lookup API:
   - `getCompletedStep(stepId, inputHash)`
   - `recordStepStarted(...)`
   - `recordStepCompleted(...)`
   - `recordStepFailed(...)`
8. Add partial recovery behavior for malformed/missing step entries:
   - reuse intact completed records
   - rerun missing/unrecoverable records where safe
   - never duplicate a completed step solely due to process restart

### Acceptance criteria

- A run cannot be replayed by two active runners at once; stale leases can be recovered after process crash.
- A run record survives process restart.
- Malformed event/step lines do not prevent loading the rest of the run.
- Completed step lookup short-circuits by stable ID + input identity.
- Editing a reusable workflow definition affects new runs only by default.
- Saved project/global/built-in workflow runs resume from captured script content/hash, not whatever the definition file currently contains.
- Missing step records are represented as recoverable, not fatal, when the workflow can rerun them safely.

### Quality gate

- Unit tests for run store creation, append/read, corrupted JSONL lines, step lookup, and partial recovery.
- Run `make typecheck` and targeted run-store tests.

**Phase LoC estimate:** ~900–1,400 product LoC.

## Phase 3 — Workflow runner and conductor-only primitives

**Goal:** Execute plain JavaScript workflow definitions with conductor-only APIs, deterministic replay constraints, and durable replay hooks.

### Work

1. Start with a runtime/determinism spike before broad primitive work:
   - define stable input hashing/canonicalization for step replay
   - block or virtualize nondeterminism such as `Date.now()`, `Math.random()`, timers, and argless `new Date()`
   - prove replay does not depend on source location or call order
2. Implement `WorkflowRunner` using a QuickJS-based sandbox or a workflow-specific wrapper around the PTC runtime.
3. Do **not** expose `mux.*`, Node, `import`, `require`, filesystem, shell, network, timers, or direct tools.
4. Expose initial globals/primitives:
   - `args`
   - `limits` or `budget`
   - `phase(name, details?)`
   - `log(message, data?)`
   - `agent(spec)`
   - `backgroundAgent(spec)`
   - `awaitAgents(handles, opts?)`
   - `parallelAgents(specs, opts?)`
5. Enforce stable IDs for replay-boundary primitives:
   - `agent.id` required
   - `backgroundAgent.id` required
   - `parallelAgents.id` required
   - each parallel item must have stable item ID or a `key(item)` mapping
6. Make missing stable IDs fail fast as authoring errors during normal execution.
7. Implement replay behavior:
   - normalize primitive input
   - check journal for completed step
   - return recorded result when valid
   - otherwise execute and record
8. Emit workflow run events for phases/logs/status/results/errors.
9. Support foreground/background launch from the service API:
   - foreground waits for result until complete/backgrounded/interrupted
   - background returns run metadata quickly
10. Ensure abort/interrupt signals stop the coordinator cleanly without deleting durable state.
11. Explicitly keep workflow primitives and `workflow_run` out of PTC `mux.*` bridging in v1 unless a later design intentionally opts in.
12. Define task-spawn idempotency before connecting to real `TaskService`:
   - persist step-start intent and/or spawned task ID before returning from task creation
   - recover `step started, no task ID` as a safe rerun only when no task can be found
   - recover `task ID exists, no report yet` by awaiting/resuming that task, not spawning a duplicate
   - recover `report artifact exists, step completion missing` by completing the step from the artifact

### Acceptance criteria

- Stable input hashing is deterministic across process restarts.
- Nondeterministic APIs are blocked or virtualized in the coordinator runtime.
- A simple workflow can log, phase, spawn one task, await it, and return a result.
- A second replay of the same durable run does not respawn completed `agent(...)` steps.
- Missing step IDs fail with a clear authoring error.
- `phase`/`log` do not require author-provided IDs.
- Recovery handles `step started/no task ID`, `task ID/no report`, and `report artifact/no step completion` without duplicate task spawning.
- The coordinator cannot directly access general tools or filesystem APIs.

### Quality gate

- Unit tests with fake task adapter for primitive execution and replay.
- Security tests/assertions that forbidden globals are unavailable.
- Run `make typecheck` and targeted workflow runner tests.

**Phase LoC estimate:** ~1,100–1,800 product LoC.

## Integration Milestone A — Thin vertical workflow slice

**Goal:** Reduce integration risk by proving a minimal end-to-end workflow before structured output and full UI are complete.

### Work

1. Add the narrowest API/tool path that can start a hardcoded or scratch workflow while the workflow experiment gate is enabled.
2. Run a simple workflow that emits `phase`, emits `log`, spawns one fake or real task, awaits it, and returns a string result.
3. Emit minimal workflow events to the chat stream.
4. Render a skeletal workflow run card with status, phase/log list, child task ID, and result.
5. Keep this slice intentionally incomplete: no promotion, no slash discovery, no structured output, no built-in deep research.

### Acceptance criteria

- A developer can start one simple workflow through a gated tool/API and see a skeletal run card.
- The run persists enough state to reload the card after transcript/workspace reload.
- This milestone produces the first screenshots before deeper features accumulate.

### Quality gate

- One integration test for the simple workflow path.
- One UI/story or harness screenshot for the skeletal run card.

**Milestone LoC estimate:** included in Phases 3, 6, and 7; expect ~400–700 product LoC pulled forward from those phases.

## Phase 4 — Task integration and report-time structured output validation

**Goal:** Let workflow-spawned tasks return validated structured data through the final report path.

### Work

1. Extend task creation metadata for workflow context:
   - workflow run ID
   - workflow step ID
   - optional output schema
   - parent workflow phase/context if useful for UI
2. Extend final report contract for workflow-spawned tasks:

   ```ts
   {
     reportMarkdown: string;
     title?: string | null;
     structuredOutput?: unknown;
   }
   ```

3. Keep non-workflow tasks compatible with Markdown-only `agent_report`.
4. Implement a shared JSON Schema subset validator utility used by report-time validation and tests; avoid duplicating ad-hoc validators across services.
5. When `outputSchema` exists, expose an `agent_report` input shape that accepts `structuredOutput: unknown` and validate that value against the JSON Schema subset inside the report tool execution path.
6. On validation failure:
   - return a structured `success: false` validation result or tool error inside the child task
   - do not let StreamManager/task finalization treat the report as accepted
   - keep the child task active
   - allow the child to call the report tool again
7. On validation success:
   - return `success: true`
   - allow normal task finalization/stop conditions to proceed
8. Persist accepted structured output into subagent report artifacts and workflow step records.
9. Update `task`, `task_await`, and workflow primitive result shapes to carry structured output when present.
10. Add defensive assertions and schema validation at the service boundary.

### Acceptance criteria

- A workflow can request five findings from a child task and receive validated structured output.
- Invalid structured output is rejected inside the child task, and the task can resubmit.
- `success: false` validation results do not trigger final task settlement or workflow step completion.
- A workflow does not receive malformed structured output.
- Existing non-workflow task behavior remains compatible.
- Persisted report artifacts include structured output when accepted.

### Quality gate

- Unit/integration tests for valid/invalid structured outputs, retry, persistence, `task_await`, and workflow primitive results.
- Regression tests for ordinary `agent_report` without schema.
- Run `make typecheck` and targeted task/report tests.

**Phase LoC estimate:** ~900–1,500 product LoC.

## Phase 5 — Interrupt, backgrounding, resume, and recovery

**Goal:** Make workflow lifecycle match the RFC: foreground/background behavior, preserved interruption, and durable resume.

### Work

1. Implement workflow lifecycle actions:
   - start foreground
   - start background
   - move foreground run to background when the parent chat needs to continue
   - interrupt
   - resume
2. Interrupt semantics:
   - stop coordinator
   - cascade-interrupt active/queued child tasks
   - preserve durable workflow state
   - preserve interrupted task workspaces where possible
   - model after existing parent hard-interrupt behavior, not destructive `task_terminate`
3. Resume semantics:
   - keep same workflow run ID
   - rerun workflow definition against durable journal
   - short-circuit completed steps
   - resume interrupted task workspaces in place where possible
   - rerun missing/unrecoverable steps through Partial Workflow Recovery
4. Add startup discovery of resumable/interrupted runs.
5. Define crash/restart vs user-interrupt resume behavior explicitly:
   - running/background runs interrupted by process crash or app restart should auto-resume where safe
   - user-interrupted workflow runs should not auto-resume; they require explicit **Resume**
   - explicitly interrupted runs must never restart spend unexpectedly on app launch
6. Define safe auto-resume conditions for crash/restart recovery:
   - captured script exists and matches the run record hash
   - Project Trust still allows project-local workflow/task execution
   - run is not marked user-interrupted
   - no unresolved required user input is blocking the run
   - no active conflicting run lock/lease exists
   - task settings/concurrency can either run or queue the next work safely
7. Design and implement child task resume rather than assuming it is already free:
   - interrupted task workspaces should be marked runnable again when the workflow resumes
   - queued child tasks with persisted prompts should be restarted or requeued correctly
   - tasks with accepted report artifacts should complete workflow steps without resuming
8. Ensure completed child task reports remain reusable after interrupt.

### Acceptance criteria

- Interrupting a workflow does not delete run state or completed step results.
- Active child tasks are stopped/preserved rather than destructively removed.
- Resume continues the same run ID.
- Auto-resume only proceeds when all safe-resume conditions pass; otherwise the run is surfaced as resumable or blocked with a clear reason.
- Crash/restart resumes running/background runs where safe; user-interrupted runs require explicit Resume.
- Completed steps are not duplicated after resume.
- Missing/corrupt records trigger safe partial recovery.
- Foreground workflow moves to background when the user sends another message or starts another workflow.

### Quality gate

- Integration tests for foreground-to-background, interrupt, resume, crash-like reload, and partial recovery.
- Manual dogfood: start a workflow, interrupt mid-task, resume, confirm run card and task workspaces behave as expected.
- Capture screenshots/video for UI behavior once UI exists.

**Phase LoC estimate:** ~800–1,300 product LoC.

## Phase 6 — API, invocation, slash/inline discovery, and scratch promotion

**Goal:** Expose workflows to agents and users through chat/tool invocation and skill-like discovery.

### Work

1. Add oRPC/IPC/API methods for:
   - list workflow definitions
   - read workflow definition
   - start workflow run
   - interrupt workflow run
   - resume workflow run
   - promote scratch workflow definition
   - get/list workflow runs for a workspace
2. Add a model-facing workflow run tool, likely `workflow_run`, with:
   - workflow name or scratch script reference
   - args
   - `run_in_background` defaulting to `false`
3. Add a scratch workflow creation path for dynamic agent-authored scripts.
4. Add slash-command behavior with an explicit conflict policy:
   - known built-in slash commands always win
   - `/<workflow-name> ...` starts a workflow run only when the name resolves to a workflow and does not collide with an agent skill
   - if an agent skill and workflow share a name, do not silently choose; require an explicit workflow invocation such as `/workflow <name> ...` or a disambiguation UI before running executable code
5. Add `$workflow-name` inline reference behavior:
   - reference/include workflow definition
   - do not execute unless the user clearly asks to run it
6. Add suggestions UI that distinguishes skills from workflows.
7. Add promotion UI flow:
   - name
   - description
   - explicit location choice: project-local or global
   - overwrite handling
   - source preview
8. Ensure project-local promotion respects Project Trust and writes only when user explicitly chooses that location.

### Acceptance criteria

- With the workflow feature gate off, workflow APIs/tools/suggestions are hidden or disabled.
- Agent can start a named or scratch workflow from a tool call.
- Slash invocation starts a workflow run.
- `$name` reference does not run by itself.
- Skill/workflow name collisions require disambiguation or explicit `/workflow <name>` invocation before executable workflow code runs.
- `workflow_run` is not exposed through PTC `mux.*` in v1.
- Workflow and skill suggestions are visually distinguishable.
- Scratch promotion requires explicit name and location.
- Project-local workflows are unavailable or disabled for untrusted projects.

### Quality gate

- Unit/UI tests for slash parsing, inline reference resolution, suggestions, conflict behavior, and promotion form validation.
- Run `make typecheck` and targeted browser utility tests.

**Phase LoC estimate:** ~1,000–1,700 product LoC.

## Phase 7 — Workflow run card and message aggregation

**Goal:** Make workflow runs visible, understandable, and reviewable in the main chat.

### Work

1. Add stream/chat events for workflow run lifecycle:
   - run started
   - status changed
   - phase emitted
   - log emitted
   - child task linked/status changed
   - structured output validation event
   - result/error/interrupted
2. Update message aggregator/projection to include workflow run state.
3. Add `WorkflowRunToolCall` or equivalent chat card.
4. Card should render:
   - name/source/scope
   - status
   - phases
   - logs
   - child tasks with links/transcripts
   - validation events
   - result/error
   - interrupt/resume actions
   - promote button for scratch workflows
5. Use existing task cards/components where possible.
6. Avoid emoji status icons; use existing icon components/lucide/shared UI primitives.
7. Add Storybook stories and UI tests for states:
   - running
   - backgrounded
   - validation failed
   - interrupted/resumable
   - completed
   - failed
   - scratch promotion available

### Acceptance criteria

- With the workflow feature gate off, workflow run-card UI and actions are hidden or disabled except for any explicit developer/debug recovery surface.
- User can understand a running Deep Research workflow from the run card.
- Child tasks are visible and inspectable from the card.
- Logs/phases are distinguishable from task reports.
- Resume/interrupt/promote affordances appear only when valid.
- Visual states survive transcript reload.

### Quality gate

- UI tests for card rendering and event projection.
- Storybook stories for major states.
- Dogfood through `make dev-server-sandbox` and `agent-browser`:
  - open the sandboxed Vite URL with `agent-browser --session mux-workflows open <VITE_URL>`
  - run workflow card story or app route
  - capture annotated screenshots of running/completed/interrupted states
  - record short video for foreground-to-background or resume flow when implemented

**Phase LoC estimate:** ~1,400–2,300 product LoC.

## Phase 8 — Built-in Deep Research workflow

**Goal:** Ship the first built-in workflow and use it to validate the whole system.

### Work

1. Add built-in `deep-research` workflow definition.
2. Keep it conductor-only:
   - scope phase
   - source discovery fan-out
   - source reading/summarization
   - claim extraction
   - adversarial verification/refutation
   - final synthesis
3. Use structured output schemas for:
   - source summaries
   - claims
   - verification/refutation results
   - final structured result
4. Include useful logs/phases so behavior is visually understandable.
5. Ensure workflow step IDs are stable and readable.
6. Add built-in workflow content generation if needed, following the built-in skills pattern but not sharing storage semantics.

### Acceptance criteria

- `/deep-research <topic>` runs the built-in workflow.
- The run card shows distinct phases and adversarial verification/refutation work.
- Final result includes `reportMarkdown` and structured claims/sources/confidence.
- The workflow can be interrupted and resumed without duplicating completed steps.

### Quality gate

- Integration test with fake task adapter for the built-in workflow structure.
- Manual dogfood on at least two topics through the sandboxed dev server and `agent-browser`:
  - one codebase/internal topic
  - one external research topic
- Capture run-card screenshots, task list, final report, console/errors output, and a short screen recording.

**Phase LoC estimate:** ~300–700 product LoC plus built-in workflow content.

## Phase 9 — Release prep, feature gate, and global task default

**Goal:** Turn the dogfooded experiment into the intended developer-facing rollout without surprising existing users.

### Work

1. Keep workflow surfaces behind the workflow experiment/feature gate until the thin slice, durable replay, structured output, and run-card dogfood pass.
2. Raise the global `maxParallelAgentTasks` default from `3` to `16` only after workflow dogfood validates task UI, provider spend, queue behavior, and local resource usage.
3. Add release notes or in-app copy if the default task concurrency change is user-visible.
4. Confirm existing user-configured task settings are preserved and not overwritten by the new default.

### Acceptance criteria

- Existing configs with an explicit `maxParallelAgentTasks` keep their value.
- New/default configs use 16 after the rollout decision.
- Workflows can still be disabled by the experiment/feature gate.
- Dogfood evidence justifies enabling the higher default.

### Quality gate

- Config normalization tests for explicit vs default task settings.
- Manual dogfood with at least one high-fan-out Deep Research run in a `dev-server-sandbox` instance.
- Capture `agent-browser` screenshots/video showing the high-fan-out run card and task queue behavior.
- Review cost/resource behavior before enabling by default.

**Phase LoC estimate:** ~50–150 product LoC.

## Cross-phase acceptance criteria

The feature is ready for broader use when all of these hold:

1. Workflow surfaces are gated until dogfood validates the experiment.
2. Workflow definitions are separate from skills but discoverable with skill-like intuition.
3. Project-local workflow definitions are gated by Project Trust.
4. A workflow run can spawn tasks, validate structured task outputs, and return a workflow result.
5. Run state survives restart and can resume without duplicating completed steps.
6. Interrupt preserves durable state and child task workspaces where possible.
7. Foreground/background behavior matches bash/task expectations.
8. Scratch workflows are one-off by default and explicitly promotable.
9. Deep Research demonstrates real value and is visually understandable.
10. Tests cover schema, storage, runner, task integration, invocation, and UI projections.
11. Dogfooding evidence includes screenshots and videos for reviewer verification.

## Dogfooding plan

Dogfooding is mandatory before calling the implementation complete.

### Setup

1. Start an isolated Mux dev server with the `dev-server-sandbox` workflow:
   - default: `make dev-server-sandbox`
   - clean sandbox when testing onboarding/trust states: `make dev-server-sandbox DEV_SERVER_SANDBOX_ARGS="--clean-providers --clean-projects"`
   - keep the sandbox root for debugging when needed: `KEEP_SANDBOX=1 make dev-server-sandbox`
2. Note the selected backend/Vite URLs and sandbox `MUX_ROOT` from the server output.
3. Enable any required workflow experiment flag in the sandboxed Mux instance.
4. Use a trusted project so project-local workflow behavior can be tested.
5. Also test an untrusted project to verify project-local workflow definitions are disabled or omitted.
6. Use `agent-browser` directly for UI dogfooding; do not use `npx agent-browser`.
7. Initialize a dedicated browser session and evidence directory for each dogfood pass:

   ```bash
   mkdir -p dogfood-output/workflows/{screenshots,videos}
   agent-browser --session mux-workflows open <VITE_URL>
   agent-browser --session mux-workflows wait --load networkidle
   agent-browser --session mux-workflows screenshot --annotate dogfood-output/workflows/screenshots/initial.png
   agent-browser --session mux-workflows snapshot -i
   ```

8. During issue-focused dogfooding, append findings immediately to a report file and capture evidence as found rather than batching notes at the end.

### Dogfood scenarios

1. **Built-in Deep Research**
   - Invoke `/deep-research <topic>`.
   - Verify phases, logs, child tasks, validation events, and final result.
   - Capture screenshot of running and completed card.
   - Record a short video showing phase progression.

2. **Dynamic scratch workflow**
   - Ask the agent to create a one-off research workflow for a specific prompt.
   - Run it as scratch.
   - Promote it through the UI with explicit location choice.
   - Capture screenshots of scratch card and promotion modal.

3. **Structured validation failure**
   - Use a test/fake workflow or controlled task to submit invalid structured output.
   - Confirm report-time validation error is shown to the child and it can resubmit.
   - Capture task transcript and workflow validation event screenshot.

4. **Foreground to background**
   - Start a foreground workflow.
   - Send another user message or start another workflow.
   - Confirm the first run moves to background and remains observable.
   - Record a short video.

5. **Interrupt and resume**
   - Interrupt a run while child tasks are active.
   - Confirm tasks are preserved/interrupted, not destructively removed.
   - Confirm user-interrupted runs do not auto-resume on app restart.
   - Resume the run explicitly.
   - Confirm completed steps are not duplicated and missing/interrupted steps recover.
   - Capture screenshots/video and the run/task IDs used.

6. **Crash/restart auto-resume**
   - Kill/restart Mux while a background or foreground workflow is running.
   - Confirm the run resumes where safe without duplicating completed steps.
   - Confirm accepted child task reports are reused.
   - Capture screenshots/video and run/task IDs.

7. **Trust gating**
   - Add `.mux/workflows/demo.js` in a trusted project and confirm discovery/execution.
   - Revoke trust and confirm project-local workflow is disabled/omitted.
   - Confirm built-in/global workflows still behave according to the trust model.

### Evidence to attach/report

Use `agent-browser` for all browser evidence. For static states, a single annotated screenshot is enough. For interactive behavior, record a repro video and take screenshots at each meaningful step.

- Initial annotated screenshot and interactive snapshot for orientation.
- Screenshots of run card states: running, backgrounded, validation failed, interrupted/resumable, completed, failed, and scratch promotion available.
- Screen recordings for foreground-to-background, interrupt/resume, crash/restart auto-resume, and scratch promotion.
- Workflow scripts used.
- Run IDs and task IDs.
- Final reports and structured outputs.
- Any validation errors and recovery logs.
- Browser console/errors snapshots during each dogfood scenario.

Recommended command pattern for interactive evidence:

```bash
agent-browser --session mux-workflows record start dogfood-output/workflows/videos/<scenario>.webm
agent-browser --session mux-workflows screenshot dogfood-output/workflows/screenshots/<scenario>-step-1.png
# perform action with agent-browser click/type/fill
agent-browser --session mux-workflows screenshot --annotate dogfood-output/workflows/screenshots/<scenario>-result.png
agent-browser --session mux-workflows record stop
agent-browser --session mux-workflows errors
agent-browser --session mux-workflows console
```

## Validation commands

Use targeted validation after each phase, then full validation near the end.

Recommended baseline:

- `make typecheck`
- targeted Bun tests for touched service/types/UI utilities
- targeted UI tests for workflow card/suggestions/promotion
- `make lint` if lint-sensitive files are touched
- `make static-check` before final handoff, unless blocked by unrelated known issues

If the implementation touches docs navigation or public docs, also run the relevant docs checks.

### Durable replay duplicates work

Risk: a bug in step identity or input hashing respawns already-completed agents.

Mitigation: require stable IDs, persist started/completed records carefully, add replay tests, and dogfood crash/resume.

### Structured output validation traps child agents

Risk: a child cannot satisfy schema and loops forever.

Mitigation: validation errors must be clear, task/run limits still apply, and UI should show validation failure events.

### Workflow runtime becomes hidden PTC

Risk: adding broad tools to workflows undermines conductor-only safety.

Mitigation: test forbidden globals/tools and keep direct tools explicitly out of v1.

### Trust confusion with skills

Risk: users see workflows alongside skills and assume same safety model.

Mitigation: label workflows distinctly, keep storage separate, and gate project-local executable workflows by Project Trust.

### UI becomes too busy

Risk: run card overwhelms chat with logs/tasks/events.

Mitigation: use progressive disclosure, collapsed details, and reuse task cards/transcript links.

### Implementation size balloons

Risk: durable workflows are a large product layer.

Mitigation: keep v1 slices vertical and dogfoodable; avoid dashboard, TypeScript, nested workflows, direct tools, and workflow-specific caps.

## Explicit deferred work

- Full workflow dashboard.
- TypeScript authoring.
- Generic parallel over arbitrary functions.
- Nested workflows.
- Direct workflow-side tools or memory writes.
- Workflow-specific budgets/caps.
- Deep Review as built-in workflow.
- Per-workflow hash approval.
- Cross-tool `.agents/workflows` support.

</details>

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `2153402{MUX_COSTS_USD:-unknown}`_


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
Addressed Codex findings:

- `WorkflowRunner.ts`: persists task IDs as soon as workflow child tasks are created and resumes recorded in-flight tasks via `waitForAgentTask` instead of spawning duplicates.
- `WorkflowTaskServiceAdapter.ts` / `TaskService`: scopes workflow interrupts to descendants belonging to the selected workflow run.
- `jsonSchemaSubset.ts`: rejects unsupported JSON Schema `type` unions instead of skipping validation.

ThomasK33:
@codex review

ThomasK33:
Addressed the latest Codex findings:

- Active workflow leases now reject re-entry even when the owner id matches, preventing same-process duplicate runners.
- The workflow interrupt path now passes `workflowRunId` into task termination, and `TaskService` filters descendants by that workflow run (including children below workflow tasks).

ThomasK33:
@codex review

ThomasK33:
Addressed latest Codex findings:

- `WorkflowRunStore.acquireLease` now serializes lease acquisition with an exclusive mutation lock, so concurrent recovery attempts cannot both acquire the same missing/stale lease.
- `WorkflowRunner` now checks durable run status before writing failed/completed terminal events, preserving user-interrupted runs when an in-flight host call returns after interruption.

ThomasK33:
@codex review

ThomasK33:
Addressed latest Codex findings:

- `WorkflowRunner` now renews the run lease while executing so active long-running workflows are not reclaimed as stale.
- `workflows.start` now defaults omitted API args to `{}` before creating the durable run.

ThomasK33:
@codex review

ThomasK33:
Addressed latest Codex findings:

- Project-local workflow discovery now resolves against the active workspace root via `resolveWorkspaceRootPath`, while trust still uses the base project path.
- Workflow compilation errors now append error/failed events, and async default function exports are supported and awaited.

ThomasK33:
@codex review

ThomasK33:
Addressed latest Codex finding: the model-facing workflow tool path now uses the active `workspacePath` for project-local workflow definitions, matching slash/oRPC behavior.

ThomasK33:
@codex review

ThomasK33:
@codex review

Please take another look. I handled the empty workflow return case by validating normalized workflow results before recording terminal success, marking invalid/empty returns failed, and keeping async workflow exports working via returned Promise resolution in the QuickJS runtime.

ThomasK33:
@codex review

Please take another look. I guarded the built-in deep-research source and claim fan-out stages so empty discovery/claim arrays skip `parallelAgents` and still reach final synthesis with empty structured arrays.

ThomasK33:
@codex review

Please take another look. I updated project-local workflow discovery/promotion to use runtime I/O for non-local runtimes, while preserving host-local filesystem behavior for local/worktree projects, and added a runtime-backed definition store test.

ThomasK33:
@codex review

Please take another look. I pushed a test-only isolation fix for the unit failure by giving the workflow run card tests their own happy-dom localStorage.

ThomasK33:
@codex review

Please take another look. I excluded devcontainer from runtime workflow project I/O so devcontainer workflow files stay on host-local I/O, and added coverage for the runtime-mode gating.

ThomasK33:
@codex review

Please take another look. I blocked project-local workflow resume/auto-resume when current project trust is missing, and capped the built-in deep-research source/claim fan-out before spawning parallel agents.

ThomasK33:
@codex review

Please take another look. I removed the pre-lease resume status flip so interrupted runs stay interrupted if a resume runner cannot acquire the lease, and routed workflow_run background tool calls through startNamedWorkflowInBackground.

ThomasK33:
@codex review

Please take another look. I kept resumed workflow cards polling after the resume action even if the immediate getRun response still shows interrupted, and added a UI regression test for that race.

ThomasK33:
@codex review

Please take another look. I now only start resume polling after the resume request is accepted, and added a regression test that a rejected resume does not keep polling getRun.

ThomasK33:
@codex review

Please take another look. I changed background resume to wait until the runner has actually acquired the run lease before returning success; if the run is still active, resume now rejects and leaves the run interrupted.

ThomasK33:
@codex review

Please take another look. I serialized workflow event/status writes and now reject terminal runner statuses when the durable run is already interrupted, preserving resumability across concurrent interrupts.

ThomasK33:
@codex review

Please take another look. I moved appendStatus sequence allocation under the same workflow event mutation lock used for appending/writing the event, so interrupt/status updates cannot compute a stale sequence while a runner event is being recorded.

ThomasK33:
@codex review

Please take another look. I added the terminal-status transition guard inside the locked event append path, so a late interrupt/status update can no longer overwrite completed or failed runs.

ThomasK33:
@codex review

Please take another look. I also isolated the workflow card tests from APIProvider/window globals after CI exposed full-suite test contamination.

ThomasK33:
@codex review

Please take another look. I now replay run status/updatedAt from the event journal so stale run.json cannot reopen terminal runs, and I apply QuickJS sandbox limits before workflow evaluation.

ThomasK33:
@codex review

Please take another look. I now ignore started-but-incomplete task IDs when resuming from a user-interrupted workflow run, so completed boundaries still replay while interrupted in-flight tasks are restarted instead of awaited.

ThomasK33:
@codex review

Please take another look. I kept the QuickJS memory cap active but extended the workflow runtime timeout so long delegated-agent waits do not fail normal workflow runs.

ThomasK33:
@codex review

Addressed the timeout propagation finding by plumbing the workflow runtime abort signal and timeout into both newly-created child-task waits and resumed existing-task waits, with targeted coverage for the runner and task-service adapter paths.

ThomasK33:
@codex review

Addressed both new findings:
- Workflow interrupts now persist the interrupted status before descendant task waits are rejected.
- Foreground workflow_run calls now pass the tool abort signal into WorkflowService; foreground aborts interrupt the run, cascade to descendants, and abort the sandbox runtime.

ThomasK33:
@codex review

Addressed the latest two findings:
- Foreground tool aborts now use a service-owned abort controller so the interrupted status and descendant cascade are awaited before the sandbox/runtime wait is aborted.
- Foreground wait backgrounding now records a backgrounded workflow status, returns a backgrounded tool result, and resumes the same run in the background instead of marking it failed.

ThomasK33:
@codex review

Addressed the background-run wait flag finding: WorkflowRunner now carries a per-run `backgroundOnMessageQueued` wait option, `runInBackground` disables it, and WorkflowTaskServiceAdapter forwards that flag to `waitForAgentReport`. Added coverage that foreground workflow waits opt in while resumed background runs opt out.

ThomasK33:
@codex review

Addressed the JSON Schema subset finding by rejecting schema-valued `additionalProperties` as unsupported, with regression coverage to ensure those schemas are not silently ignored.

ThomasK33:
@codex review

Added workflow definition tools and session-scoped scratch workflow authoring:
- `workflow_list` and `workflow_read` expose available workflow definitions to agents.
- `workflow_write` writes scratch workflows into the workspace session's workflow area by safe workflow name; these are runnable but not promoted to project/global storage by agents.
- UI/API workflow discovery now includes session scratch workflows ahead of project/global/built-in definitions so user promotion can still happen from scratch workflow run cards.

ThomasK33:
@codex review

Addressed both findings:
- Stale crash-recovered started workflow steps now fall back to spawning a fresh child task when the recorded task has no recoverable report (`Task not found` / `Task interrupted`), with regression coverage.
- Removed the root-level `CONTEXT.md` glossary and kept the workflow glossary in the RFC instead, per repo documentation placement rules.

ThomasK33:
@codex review

Addressed workflow-run brittleness and observability:
- Workflow run cards now render invocation arguments so manual slash args and model-supplied workflow_run args are visible in the UI.
- Agent output shape/schema validation failures now record failed task events and retry the failed step with a corrective prompt, up to three attempts.
- Parallel workflow steps retry only the failed agent step instead of discarding successful sibling work.

ThomasK33:
@codex review

Addressed the duplicate workflow card finding by detecting existing persisted `workflow_run` tool calls with the same runId before projecting durable run cards on workspace load. Added regression coverage for the runId detector.

ThomasK33:
@codex review

Addressed in-flight workflow card duplication: the detector now also recognizes existing `workflow_run` parts that are still `input-available` by matching workflow name and args when no runId output exists yet. Added regression coverage for both completed and in-flight workflow tool-call detection.

ThomasK33:
@codex review

Pushed one final fix so `WorkflowRunner.run` returns the normalized durable workflow result instead of the raw sandbox value, with regression coverage.

ThomasK33:
@codex review

Resolved the normalized result review thread after pushing the fix and regression test.

ThomasK33:
@codex review

Addressed the model-alias workflow suggestion finding by hiding ambiguous top-level workflow shortcuts when a workflow name collides with a model one-shot alias, while preserving explicit `/workflow <name>` suggestions.

ThomasK33:
@codex review

Added dynamic workflow agent_report schemas and the new Workflow File Reports experiment. Local validation is green (`make static-check` plus focused report tests).

ThomasK33:
@codex review

Addressed both findings:
- forward `dynamicWorkflows` through send options alongside `workflowFileReports`
- preserve workflow file-report experiment settings in the ORPC workflow adapter used for resume/start paths

ThomasK33:
@codex review

Please review the follow-up that renames workflow file reports to subagent file reports and generalizes file-backed reports to ordinary subagents.

ThomasK33:
@codex review

Addressed the strict-provider schema issue by requiring the nullable file-report fields while preserving default-path handling for omitted/undefined/null arguments.

ThomasK33:
@codex review

Addressed the inline strict-schema finding by requiring the nullable title field when workflow structured-output schemas are embedded.

ThomasK33:
@codex review

Addressed the null structured-output finding by preserving own-property `structuredOutput: null` from inline `agent_report` inputs and adding a regression test.

ThomasK33:
@codex review

Addressed the parallelAgents sibling-spend finding by interrupting remaining workflow child tasks when any parallel child task fails, with a regression test.

ThomasK33:
@codex review

Addressed the foreground-backgrounding regression by skipping sibling interruption for ForegroundWaitBackgroundedError while still interrupting siblings on real parallel child failures. Added a regression test for the backgrounded path.

ThomasK33:
@codex review

Addressed stale-lease backgrounding by aborting/draining foreground parallel wait promises on ForegroundWaitBackgroundedError without interrupting the underlying child tasks. Added the regression coverage to ensure siblings are not terminated.

ThomasK33:
@codex review

Addressed lease renewal races by serializing renew/release through the same lease mutation lock used by acquisition, with regression coverage for locked renewal.

ThomasK33:
@codex review

Addressed release-vs-renewal races by waiting for the lease mutation lock during release instead of returning early. Added coverage that release waits through an in-flight lease mutation and then clears the lease.

ThomasK33:
@codex review

Follow-up for CI flake on the lease tests: made the lease mutation wait timeout robust under slower runners while preserving the renewal/release serialization behavior.

ThomasK33:
@codex review

Follow-up after the unit flake: lease mutation locks now use a minimum active-lock stale threshold so release cannot steal a still-running renewal lock and race with its write.

ThomasK33:
@codex review

Addressed acquire-vs-renew lock stealing by applying the mutation-lock stale floor to acquireLease as well, with regression coverage that a fresh mutation lock blocks stale lease acquisition until released.

ThomasK33:
@codex review

Addressed workflow slash-command gating by passing the dynamic-workflows experiment state into command execution, rejecting workflow-run parsed commands when disabled, adding browser-side regression coverage, and adding an explicit backend assertion on the start handler.

ThomasK33:
@codex review

Addressed the remaining schema visibility thread by including workflow `outputSchema` JSON directly in the file-backed subagent prompt and covering it with a TaskService test.

ThomasK33:
@codex review

Updated the schema prompt test to avoid matching new prompt prose. It now verifies the behavioral contract by parsing the embedded JSON schema from the generated file-report prompt and comparing it to the workflow output schema.

ThomasK33:
@codex review

Addressed the interrupted workflow status race by requiring explicit resume permission before an interrupted run can transition back to running. Stale active runners now preserve the interrupted status, while resume paths pass the explicit allowance and remain covered by store/runner/service tests.

ThomasK33:
@codex review

Pushed a follow-up fix for the unit failure: the project-trust resume test now uses the explicit interrupted-resume path when setting up a running crash-recovery state.

ThomasK33:
@codex review

Addressed the lease-owner finding. WorkflowService now treats the configured runner id as a stable prefix and appends the run id plus a random nonce for each WorkflowRunner lease owner. Added a service test that records acquireLease owners and verifies each runner gets a fresh owner id.

ThomasK33:
@codex review

Addressed the lease-loss split-brain issue. WorkflowRunner now marks the run as lease-lost when renewal fails, aborts the active QuickJS runtime/task wait signal, and checks the lease-loss guard before subsequent event/step writes so stale runners do not write terminal statuses. Added a runner test that forces renewal loss while a task is active and verifies the task is aborted and no completed/failed status is written.

ThomasK33:
@codex review

Addressed the in-flight terminal write race. WorkflowRunStore now supports expected lease-owner fencing and holds the lease mutation lock while applying fenced event/step mutations; WorkflowRunner routes its journal and step writes through that fence. Added coverage that stale owners cannot append terminal statuses or step completions after another owner acquires the lease.

ThomasK33:
@codex review

Fixed the remaining unfenced backgrounding status path. `markBackgrounded` now goes through the same `appendEvent` helper, preserving the expected lease-owner fence and checking lease loss before writing.

ThomasK33:
@codex review

Aligned the file-backed report instructions with the strict nullable schema. Subagents are now told to call `agent_report` with `reportMarkdownPath`, `structuredOutputPath`, and `title` set to `null`, including recovery/restart wording, so defaults are applied after provider validation.

ThomasK33:
@codex review

Addressed both file-report wording findings:
- `agent_report` file-backed tool description no longer says to omit args; it instructs setting reportMarkdownPath, structuredOutputPath, and title to null for defaults.
- Workflow structured-output retry prompts no longer ask for a `structuredOutput` tool argument; they ask for corrected structured output and explicitly cover the file-backed `structured-output.json` + nullable path-fields flow.

ThomasK33:
@codex review

Added command-palette access for workflow card actions. Mounted workflow run cards now register interrupt/resume/promote actions with the command registry (promote opens the promotion form), and the new UI test verifies the interrupt command action is discoverable and invokes the workflow API.

ThomasK33:
@codex review

Pushed a follow-up unit-failure fix: low stale-lease test settings now use a small renewal interval floor so renewal polling does not starve fenced workflow journal writes on the lease mutation lock. Full workflow runner + built-in workflow tests and static checks pass locally.

ThomasK33:
@codex review

Addressed the stale-window concern by restoring renewal intervals to remain below `staleLeaseMs` and avoiding lock contention at the runner instead: renewal ticks now skip when a previous renewal is still in flight, so low stale-lease tests do not pile up overlapping lease-lock attempts. Full workflow runner + built-in workflow tests and static checks pass locally.

ThomasK33:
@codex review

Changed the workflow `start` API used by slash/chat invocation to run through the foreground `startNamedWorkflow` path. The router test now verifies the API returns a completed foreground result for a simple workflow instead of immediately returning a background/running run.

ThomasK33:
@codex review

Forwarded the ORPC request `signal` into `WorkflowService.startNamedWorkflow` as `abortSignal`, so aborted foreground slash/API workflow starts now exercise the same interrupt/abort path as tool invocations.

ThomasK33:
@codex review

Please review the latest update: completed background subagent reports now inject structured output into the parent context and the parent handoff no longer tells the agent to call task_await for already-completed tasks.

ThomasK33:
@codex review

Addressed the deferred best-of fallback path so persisted structured outputs are preserved when synthetic report messages are delivered from artifacts.

ThomasK33:
@codex review

Addressed the foreground workflow abort race by aborting the runner signal before interrupt cleanup, while still waiting for the interrupt side effects before returning from the foreground start path.

ThomasK33:
@codex review

Addressed active workflow interruption by registering abort controllers for foreground/background runners and aborting the active runner immediately from interruptRun.

ThomasK33:
@codex review

Addressed the per-request service issue by moving the active workflow runner abort registry to module scope and adding a cross-service interrupt regression test.

ThomasK33:
@codex review

Addressed both latest comments: active runner controllers are only registered after lease acquisition and slash workflow starts now request background execution so the run card appears immediately.

ThomasK33:
@codex review

Moved foreground/background runner abort registration into the onLeaseAcquired path so duplicate resume attempts that fail to acquire the lease cannot abort the current owner.

ThomasK33:
@codex review

Addressed the post-interrupt mutation fence by rejecting non-resume event and step writes once a run is interrupted.

ThomasK33:
@codex review

Addressed the step/status race by serializing step records through the same workflow events lock used by status mutations and using the longer mutation-lock wait window for those journal writes.

ThomasK33:
@codex review

Tiny follow-up: reverted the Storybook-only synthetic message copy so Chromatic does not require a visual baseline update; production handoff behavior remains unchanged.

ThomasK33:
@codex review

Follow-up for CI: moved react/react-dom into runtime dependencies at the existing locked 18.3.1 version so the lockfile-free package CLI check no longer resolves the broken react@latest tarball.

ThomasK33:
@codex review

I pushed a follow-up fix so workflow definitions are discovered and invocable from the new-chat creation composer as well as existing workspaces.

ThomasK33:
@codex review

Rebased the branch onto latest main and resolved conflicts; local targeted tests, typecheck, and static-check pass after the rebase.

ThomasK33:
@codex review

Addressed both findings:
- runtime setup/create/configuration failures now append an error and failed status instead of leaving runs stuck running;
- parallel foreground backgrounding now wins over sibling abort errors after guarded waits settle.

Local validation: WorkflowRunner regression test, typecheck, and static-check all pass.

ThomasK33:
@codex review

Pushed a follow-up for slash workflow transcript behavior: workflow slash commands now persist the user-authored command and workflow_run card in history before the model follow-up, and the frontend skips the old bottom ephemeral projection when persistence succeeds.

Local validation: split targeted tests for chat commands, creation workflow, streaming aggregation, router workflow routes, workflowRunMessages; make typecheck; make static-check.

ThomasK33:
@codex review

Addressed the latest findings:
- removed manual `useCallback` memoization from the workflow run card action path
- delayed/retried crash recovery for fresh persisted workflow leases so stale post-restart leases do not leave runs stuck

ThomasK33:
@codex review

Added workflow-run support to `task_await`:
- accepts explicit `wfr_...` IDs from `workflow_run`
- discovers active workflow runs when `task_ids` is omitted
- returns the consolidated workflow report plus the full run object when complete
- snapshots `backgrounded`/`interrupted` workflow states instead of returning `not_found`

ThomasK33:
@codex review

Addressed the `task_await` workflow auto-discovery feedback by excluding workflow-owned child task subtrees from omitted-`task_ids` discovery and keeping the workflow `wfr_...` run as the awaitable item for consolidated results.

ThomasK33:
@codex review

Pushed a follow-up after CI:
- stabilized the WorkflowRunner abort-signal assertion to check the signal while the workflow child wait is active
- local validation: WorkflowRunner suite, task_await tests, taskService filter test, and `make static-check`

ThomasK33:
@codex review

ThomasK33:
@codex review

Please take another look. I addressed the workflow card refresh issue by allowing completed/updated workflow runs to re-project over an existing run card while preserving the original message metadata/order, and added regression coverage for the projection policy.

ThomasK33:
@codex review

Please take another look. I addressed the duplicate-card projection path by only refreshing synthetic projected workflow cards, and excluded interrupted background runs from the continuation hook.

ThomasK33:
@codex review

Please take another look. I added the interrupted slash-workflow guard so it does not send hidden continuation context after a user interruption, with regression coverage in chat command tests.

ThomasK33:
@codex review

Please take another look. I handled superseded workflow slash completions as stale no-ops so they do not restore old command text over newer composer state, with regression coverage.

ThomasK33:
@codex review

Please take another look. I added a final staleness check before terminal workflow slash results are sent, and made workflow failure restoration respect newer composer drafts.

ThomasK33:
@codex review

Please take another look after the CI unit-test fixes. I updated the creation workflow expectation for rawCommand and made the workflow card test wait for async markdown rendering.

ThomasK33:
@codex review

Please take another look after the CI unit-test hardening. The integration failure in the prior run was an external provider stream error; this push also avoids the full-suite markdown timing flake in the workflow card test.

ThomasK33:
@codex review

Please take another look. I moved active-runner abort behind the workspace ownership check and added coverage that a mismatched workspace cannot interrupt another workspace's workflow run.

ThomasK33:
@codex review

Rebased on main and squashed the dynamic workflows branch into a single commit.

ThomasK33:
@codex review

Pushed a single amended commit after hardening the WorkflowRunToolCall test that failed in CI.

ThomasK33:
@codex review

Addressed the workflow-owned descendant auto-resume finding by excluding workflow task descendants from parent auto-resume nudges, with coverage for the workflow-owned descendant case.

ThomasK33:
@codex review

Pushed another amended single commit with CI hardening for the WorkflowRunToolCall markdown assertion and goal auto-completion wait.

ThomasK33:
@codex review

Addressed scratch workflow trust gating by treating workspace scratch definitions as project-local code: untrusted projects no longer scan or expose `.mux/workflows/.scratch/*.js`; tests cover the trusted and untrusted paths.

ThomasK33:
@codex review

Addressed scratch trust follow-ups: scratch runs now require current project trust to resume/recover, and scratch run/definition promotion requires trust before copying repo-local scratch source elsewhere.

ThomasK33:
@codex review

Addressed the delayed crash-recovery trust finding by re-reading project trust through the live workspace context before each crash-recovery resume attempt, including scheduled retry callbacks.

ThomasK33:
@codex review

Addressed the workflow run ID path traversal finding by introducing a shared `WorkflowRunIdSchema`, applying it at ORPC/schema boundaries, and asserting the same ID shape inside `WorkflowRunStore` before any run ID is joined into filesystem paths.

ThomasK33:
@codex review

Addressed the report-handoff review feedback by suppressing generic parent report delivery and post-report parent auto-resume for workflow-owned child task reports, so WorkflowRunner remains the only path that resumes/synthesizes workflow results.

ThomasK33:
@codex review

Added the requested built-in `deep-review-workflow` while keeping the existing `/deep-review` skill. The workflow scopes review lanes, runs structured lane reviewers, dedupes candidate issues, adversarially verifies findings, and synthesizes a final review with structured output.

ThomasK33:
@codex review

Addressed the plain-text workflow target feedback by accepting `args.input` in `deep-review-workflow` normalization, matching slash-command parsing for `/workflow deep-review-workflow PR #123`.

ThomasK33:
@codex review

Added the `workflow-authoring` built-in skill and updated workflow tool descriptions to tell agents to read it before authoring workflow JS. The skill documents when to use workflows, scratch/project/global locations, available workflow globals, structured-output schema limits, and replay gotchas.

ThomasK33:
@codex review

Added the workflow `applyPatch` primitive for host-mediated patch integration. Please re-review the new mutation/replay/security path.

ThomasK33:
@codex review

Updated deep-research workflow routing so exploration/source-reading uses Explore while reasoning-heavy extraction/verification/final synthesis uses Exec with read-only prompts.

ThomasK33:
@codex review

Also updated deep-review-workflow routing: scope discovery stays on Explore while lane review, triage, adversarial verification, and final synthesis use Exec with explicit read-only prompts.

ThomasK33:
@codex review

Addressed the Explore/file-backed report compatibility issue by disabling subagentFileReports for workflow Explore tasks, while preserving file-backed reports for Exec reasoning tasks. Also kept deep-review-workflow scope on Explore and review/triage/verification/synthesis on Exec with read-only prompts.

ThomasK33:
@codex review

Addressed the slash workflow timeout feedback by removing the fixed 30-minute client-side poll timeout; slash-started workflows now keep waiting until terminal state or supersession, so long background runs still deliver their result to the agent turn.

ThomasK33:
@codex review

Addressed the composer-pinning follow-up: slash workflow commands now clear sending state after the background workflow starts, keep polling without disabling the composer, and only set sending state again while delivering the terminal workflow result to the agent.

ThomasK33:
@codex review

Addressed the renderer-loss continuation issue by passing continuation options into the workflows.start endpoint and installing an onBackgroundRunTerminal callback in the backend ORPC workflow service. Persisted slash workflow invocations now let the backend send the hidden workflow result to the agent when the durable run completes; the renderer returns after the run card is persisted instead of owning the only continuation path.

ThomasK33:
@codex review

Addressed the backend continuation follow-ups: backend workflow continuations now only send when the invocation run card was successfully persisted, and they re-check that the workflow invocation remains the latest user turn before injecting the hidden result. If the user sends a newer message, the stale workflow continuation is skipped.

ThomasK33:
@codex review

Addressed the model-started workflow continuation feedback by reusing the same current-invocation guard for AIService workflow_run continuations. The guard now recognizes both slash run-card invocations and normal workflow_run tool-call messages, and skips continuation if a newer user turn exists.

ThomasK33:
@codex review

Addressed the active-stream persistence case: model-started workflow continuations now wait while the parent stream is still active and re-check the invocation guard after the workflow_run tool call has had a chance to persist to history. If it still is not current after the parent stream settles, the continuation is skipped.

ThomasK33:
Addressed Codex thread `PRRT_kwDOPxxmWM6HA9YB`: workflow-result continuations now pass `requireIdle: true` for both model-started and slash-started backend continuations, so `WorkspaceService.sendMessage` rejects instead of queueing behind superseding user input.

ThomasK33:
@codex review

Addressed the queued workflow continuation feedback by making workflow-result continuation sends idle-only (`requireIdle: true`) in both AIService and slash backend continuation paths.

ThomasK33:
@codex review

Pushed the Unit CI test fix for the workflow creation start expectation after adding backend continuation options.

ThomasK33:
@codex review

Pushed the WorkflowService unit-test wait hardening after CI showed the crash-recovery retry assertion could miss under load.

ThomasK33:
@codex review

Pushed a Docker smoke fix for the merge-queue failure: the bundled server runtime now resolves `dist/typescript-lib` from the Docker bundle layout, and the Docker image copies those TypeScript lib files into the runtime image.

ThomasK33:
Addressed Codex thread `PRRT_kwDOPxxmWM6HDQ5-`: `workflows.resume` now mirrors foreground workflow start behavior when a child wait backgrounds the run. It starts the background continuation and returns `backgrounded` instead of letting the backgrounding error escape. Added a WorkflowService regression test for resumed workflow backgrounding.

ThomasK33:
@codex review

Addressed the resumed-backgrounded-workflow feedback and added regression coverage. Also included the earlier Docker smoke fix for bundled TypeScript libs.

ThomasK33:
Addressed Codex thread `PRRT_kwDOPxxmWM6HDWMW`: slash workflow terminal continuations now wait for invocation persistence to settle instead of permanently skipping when a foreground run backgrounds and completes quickly. Added a router regression test that blocks invocation persistence while the background continuation completes, then verifies the workflow-result send happens after persistence is released.

ThomasK33:
@codex review

Addressed the foreground slash workflow persistence race and added regression coverage. Please re-review.

Issue #3338, opened 2026-05-20
Mux 0.25 doesn't work on Mac x64


yanncabral:
<img width="927" height="810" alt="Image" src="https://github.com/user-attachments/assets/63043b24-672c-4867-9906-7c6f067d113a" />

How can I install the latest version?

PR #3434, opened 2026-06-01
🤖 fix: bundle macOS sharp runtimes for both architectures

coadler:
## Summary
Fixes the Intel macOS startup failure from #3338 by ensuring macOS release packaging installs both Darwin `sharp` runtime variants before Electron Builder copies `node_modules`, then verifies those architecture-specific assets in CI.

## Background
The published x64 macOS artifacts could contain only `@img/sharp-darwin-arm64` / `@img/sharp-libvips-darwin-arm64`, causing `sharp` to fail at startup on Intel Macs when it tried to load the missing `darwin-x64` runtime.

## Implementation
- Adds a Makefile packaging prerequisite that runs documented Bun installs for both `--cpu=x64` and `--cpu=arm64` with `--os=darwin` before every macOS Electron Builder target.
- Strengthens `checkMacAttachFileRuntime.ts` to inspect packaged app architectures with `lipo` and assert the exact `sharp` and `libvips` packages exist for each architecture.
- Runs the packaged attach_file runtime smoke test in the PR macOS build, not just release builds.

## Validation
- `git diff --check`
- `make ensure-mac-sharp-runtime-deps`
- `nix shell nixpkgs#shfmt nixpkgs#shellcheck nixpkgs#hadolint -c make static-check`
- `make lint-actionlint`

## Risks
Low-to-medium release-pipeline risk: this changes macOS packaging only, but it intentionally reruns `bun install` before Electron Builder so the packaged `node_modules` tree includes both macOS native sharp runtimes. The smoke test now catches missing architecture-specific runtime assets before artifacts are uploaded.

---

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -465,6 +465,8 @@ jobs:
       - name: Build macOS distributables
         # Retry for transient Apple timestamp-server failures during code signing.
         run: ./scripts/retry.sh 3 30 make dist-mac
+      - name: Smoke test packaged macOS attach_file runtime
+        run: make check-mac-attach-file-runtime
       - name: Upload x64 artifact
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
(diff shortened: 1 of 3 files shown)


coadler:
@codex review

coadler:
@codex review

Please take another look. I replaced the wildcard CPU selector with two documented Bun installs (`--cpu=x64` and `--cpu=arm64`) and re-ran local validation.

coadler:
@codex review

Re-requesting review after the previous feedback was resolved and CI is now green.

ibetitsmike:
Closing as superseded by #3936, which landed the same fix modernized onto current main: identical `ensure-mac-sharp-runtime-deps` flow (`bun install --os=darwin --cpu=x64/arm64` before all four dist-mac targets), the same CI smoke test and both-architecture runtime checker, plus a bun >= 1.3 guard (older bun silently ignores `--os`/`--cpu`, which would have made this fix a no-op on the 1.2.x CI image). Thanks for pioneering the approach — the shipped fix is directly descended from this PR.

PR #3444, opened 2026-06-03
🤖 feat: single-line immersive review TODO bar

ammar-agent:
## Summary

Collapse the immersive-review agent TODO bar from two stacked rows into a single line. The "TODO" label, the agent's plan (rendered as the existing horizontally-scrolling strip), and the live streaming/question chip now all share one row, and the bar is no longer collapsible.

## Background

The TODO bar at the top of immersive review previously rendered two rows: a header button (`TODO · summary` + expand chevron + streaming chip) above a separate full-width plan strip that appeared when expanded. That reserved more vertical space than necessary in the review viewport. The ask was to make it one line; per follow-up, the collapse affordance isn't needed either.

## Implementation

- `ImmersiveReviewAgentStatusBar.tsx`: replaced the toggle button + conditional strip with a single `min-h-7` flex row — a static `List` + "TODO" label, the `TodoList` (`layout="horizontal"`) strip in a `min-w-0 flex-1` container so it scrolls sideways and stays bounded, and the streaming chip pinned right via the existing `renderStatusChip`. Removed the persisted `expanded` state, the now-redundant count-summary computation, and the unused `ChevronDown`/`ChevronRight`/`usePersistedState` imports. Used `min-h-7` rather than a fixed `h-7` so the row grows to the strip's natural height instead of clipping the chips.
- `storage.ts`: removed `getImmersiveReviewAgentBarExpandedKey` (its only consumer) and its entry in the ephemeral-key cleanup list.
- Tests/stories: updated the unit test to assert the single-line plan renders with no collapse button, dropped the collapse/persistence test, and refreshed story comments. The streaming-only (no-plan) path is unchanged.

## Risks

Low. Scoped to the immersive-review status bar's presentation; no change to streaming/question state wiring or the leaf-subscription re-render isolation. Removing the persisted-expand key is safe — it was a transient, not-copied-on-fork UI preference, and stale localStorage entries are simply ignored.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$1.92`_

### DIFF
--- a/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewAgentStatusBar.test.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewAgentStatusBar.test.tsx
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, mock, spyOn, test } from "bun:test";
-import { act, cleanup, fireEvent, render, type RenderResult } from "@testing-library/react";
+import { act, cleanup, render, type RenderResult } from "@testing-library/react";
 import { Profiler } from "react";
 
 import { installDom } from "../../../../../tests/ui/dom";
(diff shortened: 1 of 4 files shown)


ammar-agent:
@codex review

PR #3445, opened 2026-06-03
🤖 feat: replace thinking cycle keybind with bracket increase/decrease

ammar-agent:
## Summary

Replaces the single "cycle thinking level" keybind with a directional **increase/decrease** pair bound to the bracket keys, while keeping the old cycle shortcut alive (deprecated) for muscle memory.

- **Increase thinking effort** — `Ctrl/Cmd+Shift+]`
- **Decrease thinking effort** — `Ctrl/Cmd+Shift+[`

Unlike the old cycle (which wrapped from `max` back to `off`), the directional keybinds **clamp** at the ends of the model's allowed range — stepping past the top/bottom is a no-op, which matches the mental model of an increase/decrease control.

## Background

`TOGGLE_THINKING` (`Ctrl/Cmd+Shift+T`) was named like a toggle but actually cycled forward through every allowed level, wrapping around. A directional pair is more predictable and mirrors the existing thinking-slider paddle buttons (which already step up/down with clamping).

## Implementation

- **`Keybind.deprecated` flag** (`src/common/types/keybind.ts`): new optional field. `isKeybindDeprecated()` helper added in `keybinds.ts`. Deprecated keybinds still match/fire, but are filtered out of generated references.
- **Keybinds** (`keybinds.ts`): added `INCREASE_THINKING` / `DECREASE_THINKING`. Both pin the physical bracket key via `code` (`BracketRight`/`BracketLeft`) so Shift producing `}`/`{` doesn't break matching. `TOGGLE_THINKING` marked `deprecated: true` and kept functional (still wraps).
- **Handler** (`ThinkingContext.tsx`): one global keydown listener now handles all three — directional steps clamp, legacy cycle wraps. No-op steps skip the state write.
- **Discoverability**: new shortcuts surface in the ChatInput **tip carousel** (`placeholderTips.ts`, rendered via `formatKeybind` for platform-correct display), the thinking-slider tooltip, the Settings → Keybinds list, and the keybinds doc page. The deprecated cycle is hidden from Settings (via the `deprecated` filter) and removed from the docs.

## Validation

- New behavioral tests in `keybinds.test.ts`: verify code+Shift matching for both directions, that Shift is required, and that there is **no collision** with `NAVIGATE_BACK/FORWARD` (`Ctrl+[` / `Ctrl+]`, which omit Shift) in both directions.
- `make static-check` green (this also regenerated the docs snapshot embedded in `builtInSkillContent.generated.ts`).

## Risks

Low. Scoped to keybind registration + the thinking adjustment handler. The legacy shortcut is preserved, so no muscle-memory regression. Main subtlety is the shifted-bracket matching, which is covered by the new no-collision tests.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$2.07`_


ammar-agent:
@codex review

ammario (approved):

PR #3442, opened 2026-06-02
🤖 fix: prevent immersive review hunk layout flashes

ammar-agent:
## Summary

Fixes layout flashes while iterating through hunks in Immersive Review by synchronizing visual hunk state before paint and keeping the agent-assisted callout from changing the diff column height mid-file.

## Background

Immersive Review hunk navigation is keyboard-first, so J/K iteration should feel stable. The previous flow corrected scroll position, cursor selection, and reveal state after paint, and the per-hunk assisted banner could mount/unmount above the diff when moving between flagged and unflagged hunks in the same file.

## Implementation

- Moved file reveal gating, selected-hunk cursor/selection sync, and scroll/outline DOM writes to layout effects so they settle before the browser paints.
- Added a fixed assisted-banner slot for files containing assisted hunks, so entering or leaving an assisted hunk does not reflow the diff column.
- Kept the assisted callout inside the diff column and rendered it as a stable one-line row.

## Validation

- `make fmt`
- `bun test src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.test.tsx`
- `make typecheck`
- `make lint`
- `make static-check`
- `git diff --check`

## Risks

Low-to-medium. The change is scoped to Immersive Review layout/navigation. The main risk is altered timing around scroll/cursor effects, covered by existing Immersive Review tests plus a new regression test for same-file assisted/non-assisted hunk iteration.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `3826192{MUX_COSTS_USD:-0}`_

PR #3446, opened 2026-06-03
🤖 fix: gate immersive review diff hydration

ammar-agent:
## Summary

Prevents Immersive Review from painting unpositioned diff geometry while compact hunk overlays hydrate into full-file context. The reveal gate now keys on the rendered overlay identity instead of file path alone, so file switches, hydration swaps, and filtered hunk-content swaps stay hidden until the selected hunk has been scrolled into place.

## Background

PR #3442 fixed hunk-to-hunk scroll thrash and assisted-banner reflow, but the remaining flash came from a different invariant: while full-file content loaded asynchronously, the visible diff could swap from a compact hunk-only overlay to a much taller full-file overlay. That changes row indices and minimap geometry even when the selected hunk is unchanged.

## Implementation

- Gate reveal by `{ filePath, overlayData.content }`, not just file path.
- Keep both the diff stage and minimap stage invisible while a geometry-changing overlay is pending.
- During pending reveal, scroll using the selected hunk's resolved line in the new overlay, ignoring stale compact-overlay cursor indices.
- Only preserve an out-of-hunk context cursor when the overlay content is unchanged; compact → full-file hydration remaps the cursor back into the selected hunk.
- Force a hidden minimap redraw after parent scroll positioning and before the reveal RAF opens the stage, so the thumb/active marker read the final scrollTop.
- Move minimap canvas redraw to a layout effect because canvas pixels are visible layout state.
- Pin the `ai` package to the already-locked 6.0.175 release so lockfile-free static checks avoid the currently broken 6.0.196 tarball.

## Validation

- `bun test src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.test.tsx src/browser/features/RightSidebar/CodeReview/ImmersiveMinimap.test.tsx`
- `make typecheck`
- `make lint`
- `git diff --check`
- `make static-check`
- `make static-check-full`
- Re-ran targeted tests + `make static-check` after rebasing onto latest `origin/main`.
- Read-only review sub-agent audited the final design and found no remaining plausible hydration/minimap flash path.

## Risks

Moderate, localized to Immersive Review diff/minimap reveal behavior. The main user-visible tradeoff is that geometry swaps may briefly show the existing loading scrim instead of painting an unanchored diff. The code intentionally keeps normal same-file hunk navigation out of the reveal gate unless the overlay content changes.

The `ai` dependency pin narrows the package range to the version already resolved in `bun.lock`; this is intended to make lockfile-free package checks deterministic while npm metadata advertises a latest tarball that returns 404.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$71.64`_

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -49,7 +49,7 @@
         "@vercel/ai-sdk-openai-websocket-fetch": "^1.0.0",
         "@xterm/addon-serialize": "^0.14.0",
         "@xterm/headless": "^6.0.0",
-        "ai": "^6.0.175",
+        "ai": "6.0.175",
         "ai-tokenizer": "^1.0.6",
         "chalk": "^5.6.2",
         "comlink": "^4.4.2",
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "@vercel/ai-sdk-openai-websocket-fetch": "^1.0.0",
     "@xterm/addon-serialize": "^0.14.0",
     "@xterm/headless": "^6.0.0",
-    "ai": "^6.0.175",
+    "ai": "6.0.175",
     "ai-tokenizer": "^1.0.6",
     "chalk": "^5.6.2",
     "comlink": "^4.4.2",
--- a/src/browser/features/RightSidebar/CodeReview/ImmersiveMinimap.test.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/ImmersiveMinimap.test.tsx
@@ -97,6 +97,7 @@ describe("ImmersiveMinimap", () => {
       />
     );
 
+    expect(view.getByTestId("immersive-minimap").className).toContain("h-full");
     // Synchronous render — no waitFor needed since parseDiffLines runs during render
     expect(view.getByTestId("immersive-minimap-canvas")).toBeTruthy();
   });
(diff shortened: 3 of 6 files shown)


ammar-agent:
@codex review

Please review the hydration reveal-gate fix, especially the overlay identity invariant, stale compact cursor remapping, and minimap redraw ordering.

ammar-agent:
@codex review

Pushed a follow-up to replace the flaky scrollIntoView-based stale-cursor test with a deterministic active-line data contract while keeping the hydration cursor remap behavior covered.

ammar-agent:
@codex review

Replaced the CI-flaky DOM/scroll stale-cursor regression with a pure cursor-preservation contract test. Local targeted tests, typecheck, lint, diff-check, and static-check pass.

ammar-agent:
@codex review

Pushed final stabilization: replaced the flaky DOM cursor test with a pure cursor-preservation helper contract, and pinned `ai` to the already-locked 6.0.175 version because remote `make static-check-full` was resolving the broken 6.0.196 tarball. Local targeted tests, static-check, and static-check-full pass.

ammar-agent:
Addressed Codex finding `PRRT_kwDOPxxmWM6G4LIX` / "Preserve the minimap's stretched height": the minimap reveal wrapper now keeps `h-full self-stretch`, the minimap root itself has `h-full`, and tests assert both the wrapper/root height contracts. Re-requesting review after this push.

@codex review

ammar-agent:
Addressed Codex finding `PRRT_kwDOPxxmWM6G4W0G` / "Update the lockfile for the pinned ai dependency": `bun.lock` now matches the exact `ai` package spec, and `bun install --frozen-lockfile` plus `make static-check-full` pass locally.

@codex review

PR #3449, opened 2026-06-04
🤖 fix: prevent discarded workflow runs from reappearing

ThomasK33:
## Summary

Fixes workflow run card projection so durable workflow records no longer reappear in chat after their transcript rows were intentionally discarded, while still allowing anchored workflow cards to refresh or repair.

## Background

A slash-invoked workflow can outlive the chat rows that launched it. After an edit or compaction removes those display-only rows from the active transcript, `listRuns()` can still return the completed durable run. The UI previously projected that unanchored run back into chat, causing old completed workflows to appear at the bottom unexpectedly.

## Implementation

- Gate workflow run projection until the active transcript catch-up marker has hydrated history.
- Treat current workflow card metadata as a repairable projection anchor.
- Treat surviving workflow trigger rows as anchors for repairing missing cards.
- Keep fully unanchored durable workflow runs suppressed.

## Validation

- `bun test src/browser/utils/workflowRunMessages.test.ts`
- `bun test src/browser/utils/workflowRunMessages.test.ts src/browser/components/PinnedTodoList/PinnedTodoList.test.tsx src/browser/features/RightSidebar/CodeReview/ImmersiveReviewAgentStatusBar.test.tsx src/browser/utils/commands/sources.test.ts`
- `make typecheck`
- `make lint`
- `git diff --check`
- `make static-check`

## Risks

Medium. The touched path is transcript projection/hydration for workflow cards. The change intentionally narrows unanchored projections, while preserving refresh for cards and triggers that are still present in the active transcript.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$71.87`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -410,6 +410,7 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
     isStreamStarting,
     loading,
     isHydratingTranscript,
+    isTranscriptCaughtUp,
     hasOlderHistory,
     loadingOlderHistory,
   } = workspaceState;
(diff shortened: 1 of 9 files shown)


ThomasK33:
@codex review

PR #3457, opened 2026-06-04
🤖 fix: render foreground workflows like backgrounded runs

ThomasK33:
## Summary

Render foreground workflow runs with the same live workflow details as backgrounded runs while they are still executing.

## Background

Foreground workflow tool calls block until completion, so the initial tool message does not include a durable `runId`. That left the UI showing only the raw arguments panel during execution. This change lets the card safely discover the matching active workflow run and render definition/details/events while the foreground tool is still running.

## Implementation

- Discover executing foreground workflow runs by matching workspace, workflow name, arguments, active status, and a freshness window.
- Refuse ambiguous same-name/same-args matches so mutating actions do not bind to the wrong run.
- Merge refreshed workflow snapshots monotonically by `updatedAt`, with event sequence as the same-millisecond tie-breaker.
- Gate mutating actions until run identity is confirmed, and keep resume state aligned with the displayed run status.
- Make the Storybook workflow API mock stateful so interrupt/resume can be dogfooded locally.

## Validation

- `make static-check`
- `bun run prettier --check src/browser/features/Tools/WorkflowRunToolCall.tsx src/browser/features/Tools/WorkflowRunToolCall.test.tsx src/browser/features/Tools/WorkflowRunToolCall.stories.tsx`
- `bun x eslint src/browser/features/Tools/WorkflowRunToolCall.tsx src/browser/features/Tools/WorkflowRunToolCall.test.tsx src/browser/features/Tools/WorkflowRunToolCall.stories.tsx`
- `make typecheck`
- Dogfooded in Storybook with `agent-browser`: discovered foreground run rendering, interrupt transition, and resume transition. Screenshots/video were captured locally for review evidence.

## Risks

Moderate UI-state risk around workflow cards only. The discovery path is intentionally conservative: it declines ambiguous matches, filters stale runs, and only enables mutations for confirmed run IDs.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$109.15`_

### DIFF
--- a/src/browser/features/Tools/WorkflowRunToolCall.stories.tsx
+++ b/src/browser/features/Tools/WorkflowRunToolCall.stories.tsx
@@ -1,6 +1,7 @@
-import type { ReactNode } from "react";
+import { useRef, useState, type ReactNode } from "react";
 import type { Meta, StoryObj } from "@storybook/react-vite";
 
+import type { WorkflowRunRecord } from "@/common/types/workflow";
 import { APIContext } from "@/browser/contexts/API";
 import { WorkflowRunToolCall } from "@/browser/features/Tools/WorkflowRunToolCall";
 import { lightweightMeta } from "@/browser/stories/meta.js";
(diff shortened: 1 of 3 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
Addressed Codex P2: uniquely discovered foreground runs are now treated as confirmed identities for actions while the foreground tool call is still executing. Ambiguous matches continue to hide actions. Added regression coverage for discovered foreground interrupt and ambiguous-match action hiding.

Validation after the fix:
- `make static-check`
- Storybook dogfood of foreground-discovered interrupt/resume flow with screenshots/video captured locally.

@codex review

PR #3460, opened 2026-06-04
🤖 fix: await background workflow runs

ThomasK33:
## Summary

Hardens background workflow turn enforcement so agent-started durable workflow runs are awaited like background sub-agent work without leaking stale continuations after supersession.

## Background

This branch extends parent/task stream-end enforcement from sub-agent tasks to background workflow runs. A foreground deep-review pass after rebasing found race and currentness issues around queue-backgrounded tasks, compaction/reset boundaries, queued user turns, and workflow run provenance; the follow-up commits address those findings.

## Implementation

- Persist agent-started background workflow run references and use them for active workflow auto-resume prompts.
- Preserve queue-backgrounded task exemptions through final active-work rechecks.
- Treat reset/manual/on-send compaction as workflow supersession, while keeping mid-stream auto-compaction current.
- Prevent stale stream-end workflow parts from bypassing newer manual/slash supersessions.
- Block task workspace report finalization while task-local workflow runs are active.
- Retry terminal workflow continuations until accepted or superseded.
- Reset auto-resume state for queued user messages and avoid fallback auto-resume queueing after a user turn races in.

## Validation

- Foreground `deep-review-workflow` against `origin/main...HEAD`.
- `bun test src/node/services/agentWorkflowRunReferences.test.ts src/node/services/tools/workflow_run.test.ts src/node/services/taskService.test.ts src/node/services/workspaceService.test.ts --timeout 30000`.
- `make typecheck`.
- `env MUX_ESLINT_CONCURRENCY=1 make static-check`.

## Risks

High-touch coordination code: TaskService stream-end enforcement, workflow currentness, and WorkspaceService queue behavior. Risk is mitigated with targeted regressions for queue-backgrounded descendants, supersession boundaries, task-local workflow blocking, queued-user resets, idle fallback races, and same-millisecond persisted reference filtering.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$489.11`_


ThomasK33:
@codex review

Please review the background workflow await enforcement changes.

ThomasK33:
@codex review

Addressed the P1 terminal task-local workflow finding by blocking task reports on terminal-but-unconsumed workflow runs and adding regression coverage.

ThomasK33:
@codex review

Pushed a small follow-up to fix the unit failure by making queued-message TaskService hook calls method-optional for partial test doubles.

PR #3461, opened 2026-06-05
🤖 fix: preserve sub-agent identity in workflows

ThomasK33:
## Summary

Preserves sub-agent identity across workflow task creation, queue recovery, stream startup, startup retry, and follow-up reporting so read-only lanes such as `explore` do not fall back to stale `exec` metadata or inherit editing-capable prompt text.

## Background

Parallel workflow sub-agents can be persisted with both modern `agentId` and legacy `agentType` metadata. When queued or recovered task workspaces had stale/restamped `agentId` values, later resolution paths could treat exploratory child tasks as `exec`, which risks using the wrong prompt, model settings, and patch/report behavior.

## Implementation

- Adds shared persisted-agent candidate resolution that prefers creation-time legacy `agentType` for child task workspaces while still preserving modern `agentId` precedence for normal workspaces.
- Resolves agent definitions project-first per candidate, then falls back to global/built-in definitions, so stale built-ins cannot mask parent-only project agents.
- Propagates the discovery runtime that successfully found a sub-agent definition into stream-context and plan-instruction compilation.
- Prevents legacy mode metadata from overriding custom/effective agent identity at stream boundaries and simulated completions.
- Aligns frontend/backend best-of and task-report matching against canonical resolved agent IDs.
- Rebuilds child startup retry options from the recovered workspace agent settings before stale retry metadata so restarted sub-agents keep the intended model and thinking level.

## Validation

- `bun test src/node/services/agentSession.startupAutoRetry.test.ts`
- `make typecheck`
- `make static-check`

## Risks

This touches intricate task recovery and agent-resolution paths. The primary regression risk is misclassifying legacy task workspaces, so the change includes targeted coverage for stale `agentId` fallbacks, project-scoped precedence, simulated stream-end mode preservation, git patch eligibility, and startup retry model/thinking recovery.

## Pains

The bug spanned workflow dispatch, persisted task metadata, stream startup, discovery runtime isolation, startup retry recovery, and UI recovery, so the fix required aligning multiple candidate-resolution call sites instead of changing a single queueing path.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `3245513{COSTS}`_


ThomasK33:
@codex review

Please review this sub-agent identity recovery fix.

ThomasK33:
@codex review

Addressed the startup retry settings feedback by preferring the child task workspace's recovered agent identity/settings over stale retry metadata, with targeted startup auto-retry coverage.

PR #3462, opened 2026-06-05
🤖 fix: prefer foreground workflow runs

ThomasK33:
Summary

Update workflow tooling guidance so agents default to foreground `workflow_run` calls and reserve background workflow runs for cases where parallel or independent work can proceed.

Background

Foreground workflow runs return the completed result directly. Recommending foreground by default avoids an unnecessary `task_await` round-trip when the agent has no useful work to do while the workflow runs.

Implementation

- Added schema-level guidance to `workflow_run.run_in_background`.
- Updated the `workflow_run` tool description and `task_await` warning text.
- Updated the workflow authoring skill with foreground-first workflow invocation guidance.
- Regenerated tool hook docs and built-in skill content.

Validation

- `make static-check`
- `bun test src/common/utils/tools/toolDefinitions.test.ts`
- `make typecheck`

Risks

Low. This changes tool and skill guidance text only; no workflow execution behavior changes.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$2.89`_


ThomasK33:
@codex review

ThomasK33:
Addressed Codex feedback:

- `src/common/utils/tools/toolDefinitions.ts:1587`: preserved guidance for the `status=backgrounded` foreground-wait fallback. The tool description, schema description, and workflow-authoring skill now say to await the returned `runId` whenever `workflow_run` returns `status: "backgrounded"` before using the result.

ThomasK33:
@codex review

Please take another look.

ThomasK33:
Addressed Codex feedback:

- `src/common/utils/tools/toolDefinitions.ts:832`: expanded the await guidance to cover both `status=running` and `status=backgrounded`, so background-mode workflow runs also tell agents to `task_await` the returned `runId` before using or reporting workflow output.

ThomasK33:
@codex review

Please take another look.

PR #3465, opened 2026-06-05
🤖 fix: keep workflow scratch drafts out of git status

ThomasK33:
## Summary

Stops workflow discovery from dirtying clean repositories with generated scratch `.gitignore` files, while keeping scratch workflow drafts hidden from Git status by default.

## Background

Workflow scratch drafts live under `.mux/workflows/.scratch/`, but the previous implementation eagerly generated `.mux/workflows/.scratch/.gitignore` during discovery. That made repos show `?? .mux/workflows/` even when users had no scratch scripts.

## Implementation

- Replaces generated scratch `.gitignore` files with repo-local `.git/info/exclude` entries.
- Prepares the local exclude during workflow discovery without creating `.mux/`, so first-time scratch authoring after workflow listing remains clean.
- Escapes Git ignore metacharacters in repo-relative scratch paths.
- Preserves existing unreadable Git excludes instead of treating every read failure as empty.
- Serializes in-process exclude updates per exclude path to avoid lost updates across multiple workspaces in the same repo.
- Adds a self-ignored `.scratch/.gitignore` fallback only when repo `.gitignore` rules re-include scratch workflow files and override `.git/info/exclude`.

## Validation

- `bun test src/node/services/workflows/WorkflowDefinitionStore.test.ts`
- `bun test src/node/services/workflows/WorkflowService.test.ts`
- `make typecheck`
- `make fmt-check`
- `bun x eslint src/node/services/workflows/WorkflowDefinitionStore.ts src/node/services/workflows/WorkflowDefinitionStore.test.ts src/node/services/workflows/WorkflowService.test.ts`
- `env MUX_ESLINT_CONCURRENCY=1 make lint`
- `env MUX_ESLINT_CONCURRENCY=1 make static-check`

## Risks

Medium. This touches workflow definition discovery and local Git metadata updates. The change is intentionally fail-soft for non-Git projects and Git errors, and regression coverage now includes nested workspace paths, `.gitignore` unignore precedence, old generated scratch `.gitignore` files, concurrent exclude updates, and runtime-backed discovery.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$3.24`_

### DIFF
--- a/src/node/services/workflows/WorkflowDefinitionStore.test.ts
+++ b/src/node/services/workflows/WorkflowDefinitionStore.test.ts
@@ -304,7 +304,7 @@ describe("WorkflowDefinitionStore", () => {
     ).toContain("packages/app0/.mux/workflows/.scratch/sibling.js");
   });
 
-  test("adds a self-ignored fallback when repo rules unignore workflow files", async () => {
+  test("adds a self-ignored fallback before repo rules can unignore the first draft", async () => {
     using tmp = new DisposableTempDir("workflow-definitions");
     const workspaceRoot = path.join(tmp.path, "project");
     const scratchRoot = path.join(workspaceRoot, ".mux", "workflows", ".scratch");
@@ -317,7 +317,6 @@ describe("WorkflowDefinitionStore", () => {
       "/.mux/\n!/.mux/\n!/.mux/workflows/\n!/.mux/workflows/**\n",
       "utf-8"
     );
-    await writeWorkflow(scratchRoot, "scratch-demo", "Workspace scratch demo");
     const store = new WorkflowDefinitionStore({
       scratchRoot,
       projectRoot,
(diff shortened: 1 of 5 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed the fallback timing issue: discovery now installs the self-ignored scratch fallback immediately when repo `.gitignore` rules override `.git/info/exclude`, before the first scratch draft exists.

ThomasK33:
@codex review

Addressed the fallback detection issue: we now only skip appending the scratch fallback when the terminal non-comment pattern is a real `*` catch-all, with a regression covering comments/negated globs ending in `*`.

ThomasK33:
@codex review

Pushed one additional CI reliability fix: workflow tests use tiny stale leases, so the renewal interval now has a 50ms floor to avoid lease mutation starvation on slower CI runners. Local targeted workflow tests and `make static-check` pass.

ThomasK33:
@codex review

Kept the production renewal invariant intact and instead raised the tiny stale lease values in workflow runner/built-in workflow tests. This reduces CI lock contention without scheduling renewals after the stale threshold.

PR #3469, opened 2026-06-05
🤖 fix: isolate workflow action exec diagnostics

ThomasK33:
## Summary

Fix workflow action diagnostics so output produced by `ctx.exec` commands stays on the returned command result instead of being echoed into the action-level `stdout`/`stderr` fields. Also suppress stderr from the internal `ps` cleanup probe so host warnings do not pollute action diagnostics.

## Background

A workflow run showed built-in Git actions with confusing action diagnostics: `stdout` contained concatenated internal Git command output and `stderr` contained host `ps` warnings such as `bad data in /proc/uptime`. The structured `output` fields were usable, but the action-level diagnostics were misleading.

## Implementation

- Stop forwarding `ctx.exec` child stdout/stderr chunks to the action child process stdout/stderr while still draining and returning those streams from `ctx.exec`.
- Run the cleanup `ps` probe with stderr ignored.
- Add regression coverage for command-output isolation and deterministic fake-`ps` cleanup warning suppression.
- Keep the Git action workflow regression assertions that distinguish `git.diff` patch output from `git.diffStat` stat output.

## Deep Review

Ran `deep-review-workflow` after the initial commit. It identified two actionable findings, both addressed:

- Added deterministic coverage for noisy cleanup `ps` stderr.
- Removed the accidental duplicate scratch workflow `.gitignore` block.

Workflow run: `wfr_4254a7721b3f70f4`.

## Validation

- `bun test src/node/services/workflows/WorkflowActionRunner.test.ts --timeout 30000 --test-name-pattern "cleanup ps warnings|ctx.exec output"`
- `bun test src/node/services/workflows/WorkflowRunner.test.ts --timeout 30000 --test-name-pattern "runs built-in Git workflow actions"`
- Isolated `git check-ignore -v --no-index` check for `.mux/workflows/.scratch/.gitignore` vs a scratch draft file
- `git diff --check origin/main..HEAD && git diff --check`
- `make static-check`

## Risks

Low-to-medium. The production change affects workflow action diagnostics only: command streams are still drained to avoid backpressure and still returned through `ctx.exec`. Users who depended on implicit live echoing from `ctx.exec` into action stdout/stderr would no longer see those chunks there; actions can still log explicitly with `console.log` when desired.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `702472{MUX_COSTS_USD:-0}`_

### DIFF
--- a/.mux/workflows/.scratch/.gitignore
+++ b/.mux/workflows/.scratch/.gitignore
@@ -1,3 +1,5 @@
 # mux: hide scratch workflow drafts when repo rules unignore workflows
 *
 !.gitignore
+# mux: hide scratch workflow drafts when repo rules unignore workflows
+*
--- a/.mux/workflows/.scratch/.gitignore
+++ b/.mux/workflows/.scratch/.gitignore
@@ -1,5 +1,3 @@
 # mux: hide scratch workflow drafts when repo rules unignore workflows
 *
 !.gitignore
-# mux: hide scratch workflow drafts when repo rules unignore workflows
-*
--- a/src/node/services/workflows/WorkflowActionRunner.test.ts
+++ b/src/node/services/workflows/WorkflowActionRunner.test.ts
@@ -228,6 +228,41 @@ describe("WorkflowActionRunner", () => {
     expect(result.stderr).toBe("");
   });
 
+  test("keeps cleanup ps warnings out of action diagnostics", async () => {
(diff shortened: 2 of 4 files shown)


ThomasK33:
@codex review

Please review the workflow action diagnostics fix.

PR #3472, opened 2026-06-05
🤖 feat: show file edit LoC delta preview

ammar-agent:
Summary
- Shows a compact line delta preview (for example `+433, -23`) in successful file edit tool headers so it remains visible when the tool call is collapsed.
- Counts unified diff payload additions/deletions while ignoring `---`/`+++` file headers.
- Adds UI coverage for the collapsed header behavior and delta-count helper coverage.

Background
- Collapsed file edit tool calls previously showed the target path but not the size of the applied change, making it harder to scan completed write/edit operations.

Validation
- `bun test src/browser/features/Tools/FileEditToolCall.ui.test.tsx`
- `make typecheck`
- `make static-check`

Risks
- Low; the change only derives a display-only summary from the existing diff already rendered by the tool call UI.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `high` • Cost: `n/a`_

### DIFF
--- a/src/browser/features/Tools/FileEditToolCall.tsx
+++ b/src/browser/features/Tools/FileEditToolCall.tsx
@@ -86,6 +86,56 @@ export function buildLargeDiffPreview(diff: string): LargeDiffPreview | null {
   };
 }
 
+interface DiffLineDeltaPreview {
+  additions: number;
+  deletions: number;
+  additionsLabel: string;
+  deletionsLabel: string;
+  title: string;
+}
+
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the diff counting issue by deriving the LoC preview from parsed hunk payload lines instead of filtering raw `+++` / `---` prefixes.

PR #3473, opened 2026-06-05
🤖 refactor: clean up immersive review reveal performance

ammar-agent:
Summary
Refactors immersive review's reveal pipeline so file-content hydration, overlay identity, highlight readiness, and reveal scheduling live behind a focused hook instead of the monolithic view, and removes the remaining flash paths. The loading cover is now reserved for genuine file switches: iterating hunks within one file — including marking hunks read while read hunks are hidden — is instant, with no loading skeleton. The branch still includes the original immersive hunk-iteration perf benchmark and real Electron coverage for minimap and same-file stability. Rebased onto `main`.

Background
Immersive review was still visibly flashy. Reveal state was scattered across file-loading effects, highlight callbacks, scroll layout effects, sticky measured overlays, and full-content string keys inside `ImmersiveReviewView`, making it easy to render or reveal intermediate geometry. A concrete symptom remained after the first cleanup pass: when read hunks are hidden, marking a hunk read removes it from the visible hunk set, which (a) invalidated the full-file content cache because its key included the hunk signature, forcing a redundant file re-read, and (b) changed the overlay `contentKey`, which the reveal cover was gated on — so every same-file mark-read replayed the full-screen loading skeleton even though the file was already on screen.

Implementation
- Extracted full-file loading, overlay construction, compact `contentKey` identity, highlight readiness, and reveal scheduling into `useImmersiveOverlay`.
- Extracted the shimmer loading cover into `ImmersiveDiffRevealLoadingState`; it is an absolute scrollport sibling (no `ResizeObserver`/measured height).
- Keyed the full-file content cache by `(workspaceId, filePath)` only. File bodies do not depend on the hunk set, so hunk-set changes (mark-read, filter toggles) no longer invalidate the cache or trigger a re-read; the overlay is rebuilt in-memory from the cached body plus the live hunks.
- Split reveal gating: the loading cover, hidden diff stage, and hidden minimap now gate on a file-path change (`isActiveFileRevealPending`). The content-key pending state is retained only to re-run the hidden scroll pass that keeps the selected hunk in view across same-file rebuilds. Same-file overlay changes reveal in place; only switching to a different file shows the cover.
- Scoped the overlay deferral (blank stage behind the cover) to genuine new-file loads and resolved already-cached bodies synchronously during render, so revisiting a file or rebuilding the same-file overlay never blanks behind the cover.
- Made `useHighlightedDiff` avoid rebuilding fallback chunks on preloaded cache hits and avoid redundant state churn.
- Added cheap immersive debug attributes used by E2E sampling instead of per-frame `textContent` reads.

Validation
- `bun test src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.test.tsx` (incl. new "keeps same-file hunk-set changes instant without re-reading or re-covering")
- `bun test src/browser/features/Shared/DiffRenderer.test.tsx src/browser/features/Shared/SelectableDiffRenderer.dragSelect.test.tsx`
- `make static-check` / `make build`
- Electron hydration suite (file-switch covers + same-file/mark-read instant), incl. new "keeps same-file mark-read iteration instant while read hunks are hidden": `xvfb-run -a env MUX_E2E_LOAD_DIST=1 MUX_E2E_SKIP_BUILD=1 PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 bun x playwright test --project=electron tests/e2e/scenarios/reviewHydration.spec.ts`
- Electron perf specs: `xvfb-run -a env MUX_E2E_RUN_PERF=1 MUX_PROFILE_REACT=1 MUX_E2E_LOAD_DIST=1 MUX_E2E_SKIP_BUILD=1 PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 bun x playwright test --project=electron tests/e2e/scenarios/perf.reviewHunkIteration.spec.ts tests/e2e/scenarios/perf.reviewMarkHunkRead.spec.ts`

Risks
- Moderate UX risk around reveal timing and scroll behavior; mitigated by Electron frame-sampling tests asserting the cover still appears for file switches / Shift+M file-read advancement / compact-to-full hydration, and is absent during same-file J/K and same-file mark-read iteration.
- Moderate shared-renderer risk from the `useHighlightedDiff` cache-hit change; covered by targeted shared diff renderer tests and static checks.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `max` • Cost: `$1352.76`_

### DIFF
--- a/src/browser/features/RightSidebar/CodeReview/ImmersiveDiffRevealLoadingState.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/ImmersiveDiffRevealLoadingState.tsx
@@ -1,0 +1,53 @@
+import { Skeleton } from "@/browser/components/Skeleton/Skeleton";
+
+const REVEAL_SKELETON_LINE_WIDTHS: readonly string[] = [
+  "w-[72%]",
+  "w-[46%]",
+  "w-[88%]",
+  "w-[61%]",
+  "w-[80%]",
+  "w-[52%]",
+  "w-[70%]",
+  "w-[92%]",
+  "w-[44%]",
+  "w-[83%]",
+  "w-[58%]",
+  "w-[76%]",
+  "w-[49%]",
+  "w-[87%]",
+  "w-[65%]",
+  "w-[79%]",
+  "w-[54%]",
+  "w-[91%]",
+  "w-[68%]",
+  "w-[42%]",
+  "w-[74%]",
+  "w-[85%]",
+  "w-[57%]",
+  "w-[78%]",
+];
+
+export function ImmersiveDiffRevealLoadingState(props: { label: string }) {
+  return (
+    <div
+      role="status"
+      aria-busy="true"
+      aria-label={props.label}
+      data-testid="immersive-diff-reveal-skeleton"
+      className="mx-auto flex h-full w-full max-w-5xl flex-col gap-3 overflow-hidden px-4 py-5 select-none"
+    >
+      {/* Match the transcript hydration shimmer instead of a centered spinner: the
+          diff-shaped placeholder keeps the hidden geometry swap feeling like code
+          is hydrating in place rather than a blocking modal flash. */}
+      <Skeleton variant="shimmer" className="mb-1 block h-3 w-40 rounded" />
+      <div className="flex min-h-0 flex-1 flex-col gap-2.5 overflow-hidden">
+        {REVEAL_SKELETON_LINE_WIDTHS.map((width, rowIndex) => (
+          <div key={rowIndex} className="grid grid-cols-[3.5rem_1fr] items-center gap-3">
+            <Skeleton variant="shimmer" className="h-2 w-8 rounded" />
+            <Skeleton variant="shimmer" className={`${width} h-2 rounded`} />
+          </div>
+        ))}
+      </div>
+    </div>
+  );
+}
--- a/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewAgentStatusBar.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewAgentStatusBar.tsx
@@ -55,28 +55,27 @@ export const ImmersiveReviewAgentStatusBar: React.FC<ImmersiveReviewAgentStatusB
   // `todos` keeps a stable reference from the aggregator (same basis as
   // PinnedTodoList reading only `.todos`).
   const workspaceStore = useWorkspaceStoreRaw();
+  const hasRegisteredWorkspace = () =>
+    // Some unit tests mock only the store selectors they exercise; keep the
+    // immersive status bar crash-safe in those partial-store environments too.
+    typeof workspaceStore.hasRegisteredWorkspace === "function" &&
+    workspaceStore.hasRegisteredWorkspace(workspaceId);
   const subscribe = (callback: () => void) =>
-    workspaceStore.hasRegisteredWorkspace(workspaceId)
-      ? workspaceStore.subscribeKey(workspaceId, callback)
-      : () => undefined;
+    hasRegisteredWorkspace() ? workspaceStore.subscribeKey(workspaceId, callback) : () => undefined;
   const todos = useSyncExternalStore(subscribe, () =>
-    workspaceStore.hasRegisteredWorkspace(workspaceId)
-      ? workspaceStore.getWorkspaceState(workspaceId).todos
-      : EMPTY_TODOS
+    hasRegisteredWorkspace() ? workspaceStore.getWorkspaceState(workspaceId).todos : EMPTY_TODOS
   );
   const canInterrupt = useSyncExternalStore(subscribe, () =>
-    workspaceStore.hasRegisteredWorkspace(workspaceId)
-      ? workspaceStore.getWorkspaceState(workspaceId).canInterrupt
-      : false
+    hasRegisteredWorkspace() ? workspaceStore.getWorkspaceState(workspaceId).canInterrupt : false
   );
   // Sidebar derives `isStarting` directly from `isStreamStarting`.
   const isStarting = useSyncExternalStore(subscribe, () =>
-    workspaceStore.hasRegisteredWorkspace(workspaceId)
+    hasRegisteredWorkspace()
       ? workspaceStore.getWorkspaceState(workspaceId).isStreamStarting
       : false
   );
   const awaitingUserQuestion = useSyncExternalStore(subscribe, () =>
-    workspaceStore.hasRegisteredWorkspace(workspaceId)
+    hasRegisteredWorkspace()
       ? workspaceStore.getWorkspaceState(workspaceId).awaitingUserQuestion
       : false
   );
--- a/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.test.tsx
+++ b/src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.test.tsx
@@ -315,6 +315,131 @@ describe("ImmersiveReviewView", () => {
     expect(view.queryByTestId("immersive-assisted-banner")).toBeNull();
   });
 
+  test("normalizes CRLF hunk rows in compact overlays", () => {
+    const crlfHunk = createHunk({
+      id: "hunk-crlf",
+      oldStart: 5000,
+      newStart: 5000,
+      header: "@@ -5000 +5000 @@",
+      content: "-old crlf\r\n+new crlf\r\n context crlf\r",
+    });
+
+    const view = renderImmersiveReview({
+      fileTree: createFileTree(crlfHunk.filePath),
+      hunks: [crlfHunk],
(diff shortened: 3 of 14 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Updated the perf scenario to wait until all rendered diff rows are syntax-highlighted before resetting profiler samples and measuring hunk iteration.

ammar-agent:
@codex review

Updated this from measurement-only to include the immersive review optimization. The new highlighted harness now reports TaskDuration 0.505s for 60 hunk iterations.

ammar-agent:
@codex review

Addressed the read-action sync finding by committing the immersive-local selected hunk back to the parent before hunk read / mark-file-read actions, then revalidated the highlighted perf scenario at TaskDuration 0.565s.

ammar-agent:
@codex review

Addressed the unmount path by committing the immersive-local selected hunk in cleanup as well as on explicit exit/read actions. Revalidated the highlighted perf scenario at TaskDuration 0.739s.

ammar-agent:
@codex review

Addressed the mark-file-read stale closure by making the parent handler consult `selectedHunkIdRef.current` and revalidated the highlighted perf scenario at TaskDuration 0.506s.

ammar-agent:
@codex review

Addressed the visible multi-line hunk range finding by applying an imperative selected-range highlight for J/K hunk jumps without re-rendering the syntax-highlighted diff tree. Revalidated at TaskDuration 0.570s.

ammar-agent:
@codex review

Addressed persisted selection during page teardown by persisting parent selected-hunk state on change, while immersive hot-path J/K iteration remains local. Revalidated at TaskDuration 0.561s.

ammar-agent:
@codex review

Addressed parent teardown by writing the immersive-local selected hunk directly to persisted state during immersive cleanup, using a shared selected-hunk storage key helper. Revalidated at TaskDuration 0.605s.

ammar-agent:
@codex review

Addressed the viewport jump feedback by using nearest scroll alignment for same-file J/K hunk iteration while keeping centering for file switches, note jumps, undo, and overlay hydration. Revalidated the highlighted perf scenario at TaskDuration 0.569s.

ammar-agent:
@codex review

Addressed the stale hunk-jump range after inline composer submit/cancel by clearing the imperative range highlight/ref during composer teardown. Revalidated the highlighted perf scenario at TaskDuration 0.532s.

ammar-agent:
@codex review

Addressed the remaining nearby-content flash by not showing the reveal/loading overlay for same-file compact→full context hydration and disabling scroll anchoring on the diff scroller. Revalidated the highlighted perf scenario at TaskDuration 0.556s.

ammar-agent:
@codex review

Updated the branch to batch syntax highlighting, preload the full-file overlay before same-file hydration swaps, and wait for worker-backed highlighting before first file reveal to avoid the visible plain-text → token-color flash.

ammar-agent:
@codex review

Updated again to address the initial immersive-review vertical jump: file-level reveal now waits for in-budget full-file context to settle before opening, so the compact hunk → full-file scroll reposition stays hidden behind the reveal overlay.

ammar-agent:
@codex review

TDD update for the mark-read regression: added a failing unit test showing immersive selection stayed on hunk 1 when read hunks remained visible, fixed immersive mark-read to advance local selection to the next hunk, and updated the mark-read perf scenario to assert it advances across files.

ammar-agent:
Addressed Codex thread `PRRT_kwDOPxxmWM6HlVRo` (`highlightDiffChunk.ts`): batching now preserves old/new diff versions separately and splits streams at line-number discontinuities, so removed-side syntax state cannot bleed into added rows. Added a regression where an unterminated removed comment is followed by an added `const` line, and the batched result must match highlighting the added chunk by itself.

ammar-agent:
@codex review

Addressed the previous highlight batching finding by preserving old/new syntax streams separately and adding a regression for removed-side syntax state bleeding into added chunks. Also kept the TDD mark-read advancement fix and updated perf assertions.

ammar-agent:
@codex review

Small follow-up: adjusted the full-file reveal unit test to assert the diff stage remains hidden while context is pending without requiring the loading overlay node to still exist after a manual rAF flush. This addresses the CI-only unit failure while preserving the actual no-jump invariant.

ammar-agent:
@codex review

Latest update adds a failing-first regression for the remaining same-file hydration layout jump and gates overlay geometry swaps until the full-file target row has been positioned.

ammar-agent:
@codex review

Latest update adds the Vercel-style shimmer reveal skeleton inspired by chat transcript hydration, plus a CI-safe status-bar fallback for partial workspace-store test mocks.

ammar-agent:
@codex review

Addressed PRRT_kwDOPxxmWM6HlrUv by ignoring the one parent selection echo produced when a read action commits a locally navigated hunk, and added a regression that would fail if the echo replays over the local advance.

ammar-agent:
@codex review

Small follow-up: made the hydration/parent-echo regression tests robust in CI while preserving the same coverage.

ammar-agent:
@codex review

Follow-up only adjusts the new regression tests to avoid happy-dom/full-suite timing assumptions; production code unchanged from the addressed review round.

ammar-agent:
@codex review

I replaced the happy-dom hydration assertions with a real Electron frame-sampling regression. It reproduces the vertical jump as overlayTopRange=7283 before the fix and now passes with the reveal overlay pinned to the scrollport while hidden hydration scrolls underneath.

ammar-agent:
@codex review

Follow-up for user-reported m-key tearing: added an Electron frame-sampling E2E for mark-read iteration overlay sizing. It failed before the fix with a 598px overlay/scrollport height mismatch and now passes after sizing the reveal overlay to the scrollport.

ammar-agent:
@codex review

Follow-up after three independent review sub-agents: addressed the remaining Shift+M stale-file flash path, updated stale reveal comments, and added a bounded fail-open for full-file hydration/highlight stalls.

ammar-agent:
@codex review

I refactored the immersive review reveal pipeline and expanded the Electron hydration coverage. Please review the updated branch.

ammar-agent:
@codex review

Addressed the same-file compact→full hydration gate: full-file-context waits now apply to any pending overlay content-key swap, not just file-path swaps, and I added a regression test for the same-file far-hunk-to-near-hunk transition.

ammar-agent:
@codex review

Pushed the same-file hydration gate fix and regression test. Please take another look.

ammar-agent:
@codex review

Addressed the CRLF overlay feedback by normalizing hunk rows in compact overlay construction and added a regression test for CRLF hunk content.

ammar-agent:
@codex review

One more small CI fix: the loading skeleton no longer exposes `data-line-index`, so happy-dom selectors cannot click skeleton rows while the hidden diff stage is still hydrating.

ammar-agent:
@codex review

Latest push rebases onto `main` and fixes excess loading states during same-file hunk iteration: the full-file content cache is now keyed by path only (so marking a hunk read no longer re-reads the file), and the loading cover/hidden-stage/hidden-minimap now gate on file-path changes rather than overlay content-key changes, so same-file hunk iteration (including mark-read while read hunks are hidden) reveals in place without the skeleton. File switches remain covered.

ammar-agent:
@codex review

Addressed the P2 (stale full-file cache when the file changes on disk).

The full-file body cache is now keyed by `(workspaceId, filePath, fileContentVersion)`. `fileContentVersion` is an FNV-1a hash of the file's **unfiltered** diff hunk content (headers + bodies), so:

- Marking a hunk read while read hunks are hidden filters the *visible* set but leaves the unfiltered diff content unchanged → version stable → cache hit → still instant, no re-read, no cover.
- A real file change (tool edits the file, diff is re-fetched) changes the diff content → version changes → cache miss → re-read with fresh context. I hash content rather than hunk ids on purpose, because ids are a hash of path + line ranges and would not change for an in-place edit that preserves line ranges.

Added unit coverage: "re-reads the full-file body when the file's diff content changes" (same hunk id, changed content → re-read) alongside the existing "keeps same-file hunk-set changes instant without re-reading or re-covering".

ammar-agent:
@codex review

Addressed the second P2 (stale paint on same-path in-place edits).

`ActiveFileContentState` now tracks the cache key (`workspace + path + content version`) the body was loaded for, and `isActiveFileContentSettled` compares that cache key against the current one instead of only the file path. So when a file is edited in place (same path, new content version):

- the cache key is busted (miss), and
- the settled check now returns false because the stored body belongs to the previous version,

which means `resolvedActiveFileContent` no longer falls back to the previous version's body. The overlay drops to the compact hunk (correct, built from the fresh hunks) for the brief window until the new full-file body loads — no stale full-file context is ever painted. Strengthened the unit test to assert the previous body's context line disappears immediately on the edit and only returns after the re-read.

PR #3477, opened 2026-06-06
🤖 feat: expand workflow task structured output inline

ThomasK33:
## Summary

Adds inline structured-output inspection for completed workflow task rows while preserving separate task workspace navigation and report viewing affordances.

## Background

Completed workflow tasks can now return structured output that is useful to inspect without opening the full task report. This change lets users expand that JSON inline from the task row while keeping the existing workflow card auto-collapse behavior from interrupting active inspection.

## Implementation

- Completed task rows with `structuredOutput` can expand/collapse a pretty JSON block inline.
- The inline structured-output panel is keyboard focusable and named for accessibility.
- Completed structured-output task rows retain a separate keyboard-accessible workspace action.
- Report dialogs use a distinct visible `Report` action to avoid overloading the `Open` label.
- Storybook task-action fixture includes structured output so the interactive layout is covered.

## Validation

- `make static-check`
- `make typecheck`
- `make lint`
- `make test-unit`
- Deep review workflow with auto-fixes: `wfr_c4c448b5fe8ed29c`
- Isolated scratch `.gitignore` semantics check verified the sentinel remains unignored while draft workflow files remain ignored.

## Risks

The main regression risk is workflow task-row interaction complexity: row expansion, workspace navigation, and report opening now coexist in one compact event-list row. Tests cover all three actions plus the workflow auto-complete preservation case.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$97.02`_

### DIFF
--- a/.mux/workflows/.scratch/.gitignore
+++ b/.mux/workflows/.scratch/.gitignore
@@ -1,3 +1,5 @@
 # mux: hide scratch workflow drafts when repo rules unignore workflows
 *
 !.gitignore
+# mux: hide scratch workflow drafts when repo rules unignore workflows
+*
--- a/.mux/workflows/.scratch/.gitignore
+++ b/.mux/workflows/.scratch/.gitignore
@@ -1,5 +1,3 @@
 # mux: hide scratch workflow drafts when repo rules unignore workflows
 *
 !.gitignore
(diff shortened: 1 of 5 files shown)


ThomasK33:
@codex review

PR #3479, opened 2026-06-06
🤖 fix: lazily create scratch workflow ignores

ThomasK33:
## Summary

Make scratch workflow ignore setup lazy so regular workflow discovery no longer creates or touches `.mux/workflows/.scratch/.gitignore` in clean worktrees.

## Background

Mux previously prepared scratch workflow ignore files even when no scratch workflow existed, which could leave an annoying modified `.mux/workflows/.scratch/.gitignore` in the user's working tree. The UX should stay read-only until actual scratch content is present, and stale generated scratch ignore files from older eager behavior should not keep dirtying upgraded workspaces.

## Implementation

- Removed the tracked scratch fallback `.gitignore` from the repo.
- Changed `WorkflowDefinitionStore` to scan scratch definitions first and only install local Git excludes or fallback `.gitignore` content when scratch content exists.
- Deletes stale generated scratch `.gitignore` files only when the scratch directory contains no other entries that rely on them.
- Preserves and repairs generated fallback ignores when non-workflow scratch files are present so those files stay hidden.
- Updated regression coverage for read-only empty scratch discovery, stale generated fallback cleanup, preservation for other scratch files, lazy fallback creation, and duplicate fallback avoidance.

## Validation

- `bun test src/node/services/workflows/WorkflowDefinitionStore.test.ts`
- `make typecheck`
- `make lint`
- `git diff --check`
- `make static-check`

## Risks

Low. The change is scoped to workflow definition discovery and scratch ignore setup. Existing scratch content still triggers local ignore installation before Git status assertions, while empty scratch directories avoid file-system side effects and clean up only generated stale fallback files.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$4.16`_

### DIFF
--- a/.mux/workflows/.scratch/.gitignore
+++ b/.mux/workflows/.scratch/.gitignore
@@ -1,3 +1,0 @@
-# mux: hide scratch workflow drafts when repo rules unignore workflows
-*
-!.gitignore
--- a/src/node/services/workflows/WorkflowDefinitionStore.test.ts
+++ b/src/node/services/workflows/WorkflowDefinitionStore.test.ts
@@ -155,7 +155,7 @@ describe("WorkflowDefinitionStore", () => {
     expect(discovered.every((candidate) => candidate.scope !== "scratch")).toBe(true);
   });
 
-  test("prepares local scratch excludes while listing without creating workspace scratch files", async () => {
+  test("keeps local scratch listing read-only until a scratch workflow exists", async () => {
     using tmp = new DisposableTempDir("workflow-definitions");
     const workspaceRoot = path.join(tmp.path, "project");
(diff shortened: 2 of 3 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Please take another look. I addressed the stale generated `.mux/workflows/.scratch/.gitignore` upgrade case by deleting generated fallback files when no scratch workflow candidate exists, while preserving lazy ignore setup once a scratch workflow is present.

ThomasK33:
@codex review

Please take another look. I updated cleanup so generated fallback files are only deleted when `.scratch` has no other entries; if other scratch files remain, Mux preserves/repairs the fallback so those files stay hidden.

PR #3482, opened 2026-06-07
🤖 feat: key tool auto-expand preference by tool name

ammar-agent:
## Summary

Makes the transcript **tool** auto-expand preference sticky **per tool name** instead of one shared per-workspace bucket. Expanding/collapsing one tool no longer changes the inherited default for unrelated tools.

## Background

PR #3452 introduced `useStickyExpand`, persisting a single per-workspace record `auto-expand:{workspaceId}` → `{ thinking?, tools? }`. All tool rows shared the one `tools` boolean, so a choice made on, say, a `bash` row became the inherited default for `file_read`, `task`, etc. The intent was for the preference to be remembered **by tool name**.

## Implementation

- New `ToolNameContext` (`src/browser/features/Messages/ToolNameContext.tsx`) exposes the current tool row's name; `ToolMessage` and `NestedToolRenderer` wrap each rendered tool in `ToolNameProvider`.
- `useStickyExpand` reads the tool name from context (mirroring how it already reads `workspaceId` from `MessageListContext`), so **no tool component call sites change** — `useToolExpansion` and the direct `useStickyExpand("tools", …)` callers (`FileEditToolCall`, `TaskToolCall`) all pick it up.
- `AutoExpandPrefs.tools` is now `Record<toolName, boolean>` instead of a single `boolean`. Thinking blocks are unchanged (`thinking?: boolean`). Read/write are funneled through small `readStoredPref`/`applyStoredPref` helpers; persistence is skipped for a tool row with no resolvable tool name (degrades to local-only, matching the no-workspace path).

All present-block invariants from #3452 are preserved: the preference is still snapshotted once at mount, so a write from one row never retroactively mutates a mounted row — only future rows of the same tool inherit it.

## Validation

- `useStickyExpand.test.tsx`: updated tool cases to the per-tool shape and added a behavioral test that a choice on one tool name does not leak to another (`{ tools: { bash: true, file_read: true } }`).
- `FileEditToolCall.ui.test.tsx`: wrapped renders in `ToolNameProvider` and keyed the seeded preference by tool name.
- `make static-check` green; Tools + Messages unit suites pass.

## Risks

Low. Behavior is isolated to the expand/collapse default for tool rows; logic stays centralized in one hook. The persisted shape changed (`tools: boolean` → `tools: Record<string, boolean>`); a stale record from a prior build is read defensively (an old boolean simply yields no per-tool match and falls back to each tool's default), with no crash.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `high` • Cost: `$`_

### DIFF
--- a/src/browser/features/Messages/ToolMessage.tsx
+++ b/src/browser/features/Messages/ToolMessage.tsx
@@ -9,6 +9,7 @@ import {
   extractHookOutput,
   extractHookDuration,
 } from "../Tools/Shared/HookOutputDisplay";
+import { ToolNameProvider } from "./ToolNameContext";
 
 interface ToolMessageProps {
   message: DisplayedMessage & { type: "tool" };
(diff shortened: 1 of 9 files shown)


ammar-agent:
@codex review

ammario (approved):

ammar-agent:
@codex review

Addressed the legacy-preference P2: `readStoredPref` now treats a legacy boolean `tools` value as the per-tool fallback (so an upgrading user's collapsed/expanded choice is preserved), and the next toggle migrates the key to the per-tool record shape. Added a regression test in `useStickyExpand.test.tsx`.

PR #3484, opened 2026-06-07
🤖 feat: allow model and thinking overrides on sub-agent launch

ammar-agent:
## Summary

Adds optional `model` and `thinking` parameters to the `task` (sub-agent launch) tool so a launching agent can override the spawned sub-agent's model and thinking level. Both default to the existing **inherit** semantics, and the param descriptions instruct the model to omit them unless explicitly told otherwise (and to never assume a particular model is available).

## Background

Previously a sub-agent's model/thinking were always inherited (parent live runtime env → parent workspace AI settings → per-agent config defaults → global default). There was no way for a launching agent to request a specific model/thinking for a delegated task. The override slots (`TaskCreateArgs.modelString` / `thinkingLevel`) already existed at the top of `resolveTaskAISettings`' precedence chain but were never wired to the tool.

## Implementation

- **Schema** (`toolDefinitions.ts`): new `model` and `thinking` fields on the task tool, both `.nullish()` (strict-mode convention). `thinking` is preprocessed to coerce a JSON number to a string before parsing.
- **Handler** (`tools/task.ts`): `parseTaskAiOverrides` parses the inputs **reusing the exact UI logic** and forwards them to `taskService.create`. Invalid input throws a descriptive tool error before any task is spawned. The parent runtime fallback hint is still forwarded so unspecified fields keep inheriting.
- **DRY model parsing**: relocated `normalizeModelInput` from `src/browser/utils/models/` to `src/common/utils/ai/` (history preserved via `git mv`; its only deps were already in `common`), updated the browser importers, and collapsed the duplicate ACP `normalizeModelForCommand` onto the shared helper.
- **DRY thinking parsing**: handler uses `parseThinkingInput` (named levels OR numeric indices). Numeric indices stay deferred as `ParsedThinkingInput` and are resolved against the **chosen** model's policy via `resolveThinkingInput` inside `resolveTaskAISettings` (so `2` means the same thing it does for `/model+level` in the UI). `TaskCreateArgs.thinkingLevel` was widened from `ThinkingLevel` to `ParsedThinkingInput`; existing concrete-level callers are unaffected (named levels pass through unchanged).

## Validation

- `make static-check` green (lint, typecheck, docs sync regenerated for the two new tool params).
- New tests: task-tool handler forwards an alias + named thinking, forwards a numeric thinking as a deferred index, and rejects an invalid model before spawning; taskService resolves a numeric thinking override against the inherited model's policy. Full `taskService.test.ts` (150) and `task.test.ts` (23) pass.

## Risks

Low. Behavior is opt-in; omitting both params preserves the prior inherit path exactly. The `thinkingLevel` type widening is backward-compatible (named levels are a subset of `ParsedThinkingInput` and pass through `resolveThinkingInput` unchanged). The ACP refactor is a straight delegation to an already-tested helper.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh`_


ammar-agent:
@codex review

PR #3486, opened 2026-06-07
🤖 feat: allow collapsing attached reviews in the chat decoration

ammar-agent:
## Summary

Make the "Attached Reviews" panel in the chat composer collapsible by rendering it through the shared `ChatInputDecoration` primitive, so a pending message with many attached reviews can be tucked away without detaching them.

## Background

Reviews attached to the pending message were always rendered fully expanded above the textarea. When several reviews are attached, the panel could consume a large portion of the composer area and push the input around. Every other composer-adjacent decoration (pending reviews banner, pinned TODO list, queued message, background processes) already uses the collapsible `ChatInputDecoration` chrome; the attached-reviews panel was the odd one out.

## Implementation

- `AttachedReviewsPanel` now wraps its contents in `ChatInputDecoration`:
  - Collapsed summary shows the review icon + attached count.
  - Expanded body keeps the existing review blocks and the "Clear all" action (shown when more than one review is attached).
- Expansion intent persists per-workspace via `usePersistedState` under a new `getAttachedReviewsExpandedKey(workspaceId)` storage key, defaulting to **expanded** to preserve the prior always-visible behavior.
- The panel renders inside the chat-input card (which already supplies the gutter), so the decoration's top border + horizontal padding are dropped and a bottom divider is kept above the textarea, matching the panel's prior placement.
- `ChatInput` passes `workspaceId` to the panel (only rendered in the `workspace` variant).

## Validation

- `make typecheck`, `make static-check`, and targeted `eslint` on the touched files all pass.

## Risks

Low. Scope is limited to the composer's attached-reviews preview; detach/clear/edit handlers are unchanged, and the default expanded state preserves existing behavior. Worst case is a per-workspace UI preference being mis-persisted, which self-heals on toggle.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `n/a`_

### DIFF
--- a/src/browser/features/ChatInput/AttachedReviewsPanel.tsx
+++ b/src/browser/features/ChatInput/AttachedReviewsPanel.tsx
@@ -1,10 +1,14 @@
 import React from "react";
-import { X } from "lucide-react";
+import { MessageSquare, X } from "lucide-react";
 import { Tooltip, TooltipTrigger, TooltipContent } from "@/browser/components/Tooltip/Tooltip";
+import { ChatInputDecoration } from "@/browser/components/ChatPane/ChatInputDecoration";
+import { usePersistedState } from "@/browser/hooks/usePersistedState";
+import { getAttachedReviewsExpandedKey } from "@/common/constants/storage";
 import { ReviewBlockFromData } from "../Shared/ReviewBlock";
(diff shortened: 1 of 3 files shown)


ammar-agent:
@codex review

PR #3488, opened 2026-06-08
🤖 fix: collapse workflow action event rows

ThomasK33:
## Summary

Collapses workflow action lifecycle events in the Workflow Run tool call UI so a started action row updates in place when its terminal event arrives, matching task-row behavior and reducing event-list noise.

## Background

Workflow action executions currently emit separate started and terminal events. Rendering those as separate UI rows splits the context for one logical action and makes workflow histories harder to scan.

## Implementation

- Adds an action display-row type alongside existing task coalescing.
- Coalesces an unambiguous action started event with its terminal event while preserving the initiating raw event sequence for tooltip/ARIA context.
- Combines started and terminal JSON details when an action row is expanded.
- Preserves raw rows for ambiguous same-key persisted streams rather than guessing and risking detail misattribution.
- Updates the completed workflow story fixture to include a representative action lifecycle pair.

## Validation

- `make static-check`
- `make typecheck`
- `make lint`
- `git diff --check`
- Dogfooded the workflow events UI with collapsed, expanded, and action-detail states; captured screenshots/video locally before PR creation.

## Risks

Low to medium. The change is scoped to workflow run event rendering, but it touches timeline grouping logic. Regression coverage now includes both the normal coalesced path and an ambiguous same-key stream that must remain unmerged.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `640148{MUX_COSTS_USD:-unknown}`_


ThomasK33:
@codex review

ThomasK33:
Addressed Codex feedback:

- `workflow-action cached rows`: cached action events now stay as standalone rows and no longer close pending `started` action rows.
- Added regression coverage for a stale `started` row followed by a same-key `cached` event.
- Re-ran `make static-check` successfully before pushing.

ThomasK33:
@codex review

Please take another look.

PR #3491, opened 2026-06-08
🤖 fix: harden deep research workflow

ThomasK33:
## Summary

Hardens the built-in `deep-research` workflow by turning the prior linear pass into a staged, bounded, adversarial research pipeline with strict structured-output schemas and targeted regression coverage.

## Background

The workflow now decomposes research topics into source-discovery angles, extracts quote-backed claims from selected sources, ranks claims, runs adversarial verification votes, and synthesizes a final report. A follow-up `deep-review-workflow` pass found four additional correctness/reliability issues, which are fixed in this PR.

## Implementation

- Adds smart/quick research configs with bounded angle/source/claim/verification fan-out.
- Uses source-key `Set` dedupe to avoid `Object.prototype` collisions while preserving meaningful URL query/path variants.
- Caps extracted claims per source before ranking/serialization and filters quote-less claims.
- Groups exact normalized duplicate claim text before applying the verification budget while retaining evidence/source references.
- Treats verifier claim-identity drift as a refutation and batches verifier agents.
- Keeps terminal structured outputs consistent, including `findings: []` on no-topic/no-claim exits.

## Validation

- Ran `deep-review-workflow` against the branch and fixed all four verified findings.
- `bun test src/node/services/workflows/builtInWorkflowDefinitions.test.ts -t "built-in deep-research workflow"`
- `bun test src/node/services/workflows/builtInWorkflowDefinitions.test.ts`
- `make static-check`
- Re-ran workflow tests and `make static-check` after rebasing onto latest `origin/main`.

## Risks

Medium: this changes a built-in workflow conductor string executed inside QuickJS. The changes are bounded to `deep-research` and covered by regression tests for no-topic/no-claim paths, strict schema shape, source dedupe, quote filtering, claim identity validation, quick-mode caps, verifier batching, per-source claim caps, and duplicate-claim grouping.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$49.29`_

### DIFF
--- a/src/node/services/workflows/builtInWorkflowDefinitions.test.ts
+++ b/src/node/services/workflows/builtInWorkflowDefinitions.test.ts
@@ -9,7 +9,7 @@ import { DisposableTempDir } from "@/node/services/tempDir";
 import { BUILT_IN_WORKFLOW_DEFINITIONS } from "./builtInWorkflowDefinitions";
 import { WorkflowActionRegistry } from "./WorkflowActionRegistry";
 import { WorkflowRunStore } from "./WorkflowRunStore";
-import { WorkflowRunner, type WorkflowAgentSpec } from "./WorkflowRunner";
+import { WorkflowRunner, type WorkflowAgentResult, type WorkflowAgentSpec } from "./WorkflowRunner";
 
 const BUILT_IN_WORKFLOW_TEST_STALE_LEASE_MS = 100;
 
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

Please review the deep-research workflow hardening and follow-up fixes from the deep-review-workflow findings.

PR #3496, opened 2026-06-09
🤖 fix: harden remote base repo normalization

ethanndickson:
## Summary

Hardens SSH/Coder remote workspace setup so Mux treats `.mux-base.git` as an internal bare common git directory rather than a checkout. Shared base repos are normalized by removing leaked `core.bare` / `core.worktree` config, keeping the base repo's own `HEAD` off user branches, and healing poisoned warm-path caches before materializing new workspaces.

## Background

A remote Coder host ended up with `.mux-base.git/config` containing shared checkout config (`core.bare=false` and `core.worktree=<synthetic path>`). Linked worktrees inherit that common config, so a repair that made one workspace look healthy poisoned future sibling workspaces. Separately, leaving the base repo `HEAD` on trunk makes checkout-aware tools such as Graphite interpret the bare cache as the real trunk checkout.

## Implementation

The base repo invariant is now explicit: `.mux-base.git` is bare by layout, has no shared checkout config, and does not advertise a user branch in `git worktree list --porcelain`. Empty base repos point `HEAD` at an unborn internal branch so Git porcelain stays branch-shaped; once Mux selects a real worktree base commit, it detaches the base repo `HEAD` at that commit. This keeps Graphite from seeing trunk checked out at the base repo while preserving normal Git behavior for `worktree add -B`, `worktree add -b`, `fsck`, and `branch -D`.

Both the slow sync path and the fused warm fast-path normalize poisoned base repos before checkout. The warm path treats cleanup failures as cache misses so the slow path can run its retry/error handling instead of materializing from still-poisoned shared config. Branch cleanup after workspace deletion keeps Git's own active-worktree guard by neutralizing base `HEAD` and using `git branch -D` instead of raw ref deletion; the `HEAD` rewrite is scoped to common git dirs named `.mux-base.git` so deleting a worktree of an unmanaged real checkout never strands that checkout on the internal branch. Regression coverage exercises pre-existing poisoned `core.bare` / `core.worktree`, neutral base `HEAD`, warm fast-path healing, missing-object repair, and deleted-worktree branch cleanup.

## Validation

Focused SSH integration coverage was run for the poisoned-config, neutral-HEAD, warm-path healing, missing-object repair, and worktree-delete cases. Local static validation also passed with `MUX_ESLINT_CONCURRENCY=1 make static-check`.

## Risks

This touches remote SSH workspace materialization and deletion. The riskiest area is old/corrupted shared base repos, but the change is intentionally defensive: normalization is idempotent, warm-path cleanup failures fall back to the slow path, missing-object repair remains in the checkout path, and branch cleanup is best-effort.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$47.88`_

### DIFF
--- a/src/node/runtime/SSHRuntime.ts
+++ b/src/node/runtime/SSHRuntime.ts
@@ -3371,9 +3371,16 @@ export class SSHRuntime extends RemoteRuntime {
         // Skip protected trunk branch names to avoid accidental deletion.
         const PROTECTED_BRANCHES = ["main", "master", "trunk", "develop", "default"];
         if (branchToDelete && !PROTECTED_BRANCHES.includes(branchToDelete)) {
+          const branchRefArg = shescape.quote(`refs/heads/${branchToDelete}`);
           await execBuffered(
             this,
-            `${nhp}git --git-dir=${baseRepoPathArg} update-ref -d ${shescape.quote(`refs/heads/${branchToDelete}`)} 2>/dev/null || true`,
+            [
+              `branch_ref=${branchRefArg}`,
+              `branch_oid=$(git --git-dir=${baseRepoPathArg} rev-parse --verify --quiet "$branch_ref" 2>/dev/null || true)`,
+              '[ -n "$branch_oid" ] || exit 0',
+              `if git -C ${baseRepoPathArg} worktree list --porcelain | grep -Fqx "branch $branch_ref"; then exit 0; fi`,
+              `${nhp}git --git-dir=${baseRepoPathArg} update-ref -d "$branch_ref" "$branch_oid" 2>/dev/null || true`,
+            ].join("\n"),
             { cwd: "/tmp", timeout: 10 }
           ).catch(() => undefined);
         }
--- a/src/node/runtime/SSHRuntime.ts
+++ b/src/node/runtime/SSHRuntime.ts
@@ -3373,7 +3373,7 @@ export class SSHRuntime extends RemoteRuntime {
         if (branchToDelete && !PROTECTED_BRANCHES.includes(branchToDelete)) {
           await execBuffered(
             this,
-            `${nhp}git -C ${baseRepoPathArg} branch -D ${shescape.quote(branchToDelete)} 2>/dev/null || true`,
+            `${nhp}git --git-dir=${baseRepoPathArg} update-ref -d ${shescape.quote(`refs/heads/${branchToDelete}`)} 2>/dev/null || true`,
             { cwd: "/tmp", timeout: 10 }
           ).catch(() => undefined);
         }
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Follow-up pushed for the CI Git/fsck behavior: base repo connectivity probes now ignore only the expected HEAD-only badHeadTarget complaint for refs/mux-internal/base-head, while preserving real missing-object failures.

ethanndickson:
@codex review

Addressed the branch deletion finding by replacing `git branch -D` with `git --git-dir ... update-ref -d refs/heads/<workspace>`, and added coverage that deleteWorkspace cleans the branch ref while the base repo HEAD points at the internal sentinel.

ethanndickson:
@codex review

Addressed the active-branch guard concern: the update-ref cleanup now first verifies no worktree still owns the branch and deletes with the expected old OID.

ethanndickson:
@codex review

I pushed an elegant-redesign pass that keeps the base repo HEAD Git-native: unborn internal branch while empty, detached at the selected base commit once materializing a workspace. This lets the PR keep normal Git porcelain (`worktree add -B`, `worktree add -b`, `fsck`, `branch -D`) while still ensuring `.mux-base.git` never advertises a user branch checked out at the base repo.

ethanndickson:
Resolved the remaining Codex thread from the prior `update-ref -d` approach: the redesign removes raw ref deletion entirely and returns branch cleanup to `git branch -D` after neutralizing the base repo HEAD, so Git's active-worktree guard is preserved by porcelain instead of reimplemented.

@codex review

ethanndickson:
Addressed the warm-path cleanup feedback: `core.bare` / `core.worktree` cleanup now only continues on success or "key absent"; any other cleanup failure becomes `WARM_MISS:base-config-normalization-failed`, so the slow path handles retry/error semantics before any worktree materialization.

@codex review

ethanndickson:
@codex review

ethanndickson:
Addressed the P1 (`PRRT_kwDOPxxmWM6I_oAn`) in cab409839: `deleteWorkspace` now gates the HEAD neutralization on the resolved common git dir actually being a managed base repo (`basename === ".mux-base.git"`, which covers both the canonical and legacy hashed layouts). For unmanaged common dirs (e.g. a worktree of a real checkout), the `symbolic-ref HEAD` rewrite is skipped entirely; the best-effort `branch -D` still runs and remains protected by Git's own checked-out-branch guard. Added an integration regression test ("deleteWorkspace leaves an unmanaged source checkout's HEAD untouched") that creates the workspace as a worktree of a normal checkout and asserts the checkout's HEAD ref and commit survive deletion — verified it fails against the previous code and passes with the fix.

ethanndickson:
@codex review

ethanndickson:
@codex review

PR #3498, opened 2026-06-09
🤖 feat: withhold review pane tools from sub-agents

ammar-agent:
## Summary

Withhold the Review pane tools (`review_pane_update` / `review_pane_get`) from sub-agents so child task workspaces can no longer pin code to the user-facing parent Review pane.

## Background

The Review pane is a parent-workspace, user-facing concept. Sub-agents run in isolated child task workspaces with their own session dir, so anything they "pin" lands in a pane the user never sees — pure noise and wasted effort. The pinning tool (`review_pane_update`) and its read counterpart (`review_pane_get`) should simply not be offered to sub-agents.

## Implementation

- `getAvailableTools()` gains an `enableReviewPane` option (defaults to `true`). When `false`, both `review_pane_update` and `review_pane_get` are dropped from the tool allowlist.
- `getToolAvailabilityOptions()` — the single source of truth for these capability flags — now returns `enableReviewPane: !parentWorkspaceId`, so every consumer (system prompt tool instructions, tokenizer estimates, cost debug) disables the pane for sub-agents automatically.
- `getToolsForModel()` mirrors the gate when filtering the concrete toolset to the canonical allowlist, using `config.enableAgentReport` (the existing canonical "is sub-agent" signal, set true iff `parentWorkspaceId` is present) so the toolset and system prompt stay in sync.

`review_pane_get` is gated alongside `review_pane_update` because it only ever reads a sub-agent's own (now always-empty) list, making it useless without the ability to pin.

## Validation

- New behavioral tests: `getToolAvailabilityOptions` flag derivation, `getAvailableTools` gating, and `getToolsForModel` excluding the pane tools for sub-agents (`enableAgentReport=true`).
- `make static-check` and the touched tool test suites pass.

## Risks

Low. Top-level workspaces are unaffected (flag defaults to enabled). The change only removes two tools from sub-agent toolsets; no persisted data or existing pins are touched.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh`_

### DIFF
--- a/src/common/utils/tools/toolAvailability.test.ts
+++ b/src/common/utils/tools/toolAvailability.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from "bun:test";
 
-import { getGoalToolAvailability } from "./toolAvailability";
+import { getGoalToolAvailability, getToolAvailabilityOptions } from "./toolAvailability";
 import type { GoalStatus } from "@/common/types/goal";
 
 const execAgent = {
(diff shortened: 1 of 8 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the P2: `getToolAvailabilityOptions()` now always returns `enableReviewPane`, and `TokenizerService.calculateStats()` forwards that object to `calculateTokenStats`. Updated the exact `toHaveBeenCalledWith` / `toHaveBeenNthCalledWith` expectations in `src/node/services/tokenizerService.test.ts` to include `enableReviewPane: true` (top-level) and `enableReviewPane: false` (sub-agent). Resolved the thread.

ammario (approved):

PR #3508, opened 2026-06-11
🤖 feat: show fallback badge in transcript when a model refused

ThomasK33:
## Summary

When a model refuses and the configured model-fallback chain (#3505) answers instead, the transcript header now shows an amber `⇄ FALLBACK` badge next to the answering model. Hovering (or keyboard-focusing) the badge explains the swap: which model was requested, which models refused, and which model answered.

## Background

Model fallback currently swaps silently: the message header shows only the _answering_ model, so there is no transcript-visible trace that your requested model refused. The metadata (`modelFallback: { requestedModel, refusedModels[] }`) was already persisted by the backend but never reached the frontend — the oRPC zod schemas strip unknown metadata keys at the IPC boundary.

## Implementation

- `MuxMessageSchema` / `StreamEndEventSchema` gain a `modelFallback` passthrough so the record survives the IPC boundary (history replay and live stream-end).
  - The history read path uses `.catch(undefined)` (like `agentId`/`compactionEpoch`): the badge is decorative, so a shape-corrupt persisted record degrades to "no badge" instead of failing whole-chat loading.
- `displayedMessageBuilder` maps the metadata onto displayed assistant rows; `AssistantMessage` renders the new `ModelFallbackBadge` between the model display and the compacted badge.
- Tooltip copy is built by an exported pure helper (`buildModelFallbackTooltipLines`) because Radix portals are unobservable under happy-dom; lines cover single-hop, multi-hop ("Fallback X also refused."), and the answering model.
- Extracted `formatModelStringForDisplay()` (models.ts) — the badge would have been the third copy of the `formatModelDisplayName(getModelName(normalizeToCanonical(...)))` chain; ModelDisplay and ModelSelector now share it.

A deep-review pass found no P0/P1; its P2 (self-healing schema), P3 (schema roundtrip tests), and P4 findings (duplicate React keys on colliding display names, empty-string effective model, keyboard-unreachable tooltip, DRY) are addressed in the second commit. One P4 is intentionally deferred as a follow-up: during a _live_ stream the header shows the refused model until stream-end delivers the corrected metadata (and an interrupt right after a swap only shows the badge after reload) — pre-existing behavior of abort/live metadata paths, not introduced here.

## Validation

- Dogfooded in a `dev-server-sandbox` with seeded `chat.jsonl` fixtures (single-hop, multi-hop, and control messages) via agent-browser: badge renders only on fallback messages, tooltip shows the full chain.
- New behavioral tests: tooltip line building (multi-hop order, missing/empty effective model, malformed empty `refusedModels`), aggregator propagation to displayed rows, schema roundtrip + malformed-record tolerance, and a `StreamEndEventSchema.parse` assertion inside the existing streamManager fallback test so reverting either schema line fails CI.

## Risks

Low. The badge renders only when `metadata.modelFallback` is present (fallback users only); schema additions are optional fields, and the only shared-code change is the display-name helper consolidation (`normalizeToCanonical` is idempotent, so ModelDisplay/ModelSelector behavior is unchanged).

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$68.92`_

### DIFF
--- a/src/browser/components/ModelSelector/ModelSelector.tsx
+++ b/src/browser/components/ModelSelector/ModelSelector.tsx
@@ -26,6 +26,7 @@ import { stopKeyboardPropagation } from "@/browser/utils/events";
 import { formatModelDisplayName } from "@/common/utils/ai/modelDisplay";
 import { formatProviderDisplayName } from "@/common/utils/providers/customProviders";
 import {
+  formatModelStringForDisplay,
   getExplicitGatewayPrefix,
   getModelName,
   getModelProvider,
@@ -277,7 +278,7 @@ export const ModelSelector = forwardRef<ModelSelectorRef, ModelSelectorProps>(
     const canonicalValue = hasValue ? normalizeToCanonical(value) : "";
     const selectedProvider = hasValue ? getModelProvider(canonicalValue) : "";
     const displayValue = hasValue
-      ? formatModelDisplayName(getModelName(canonicalValue))
+      ? formatModelStringForDisplay(canonicalValue)
       : (emptyLabel ?? "");
 
     // Explicit gateway selections short-circuit route resolution: the user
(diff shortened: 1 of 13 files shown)


ThomasK33:
@codex review

PR #3509, opened 2026-06-11
🤖 refactor: upgrade @agentclientprotocol/sdk to 0.25.0

ThomasK33:
## Summary

Upgrades `@agentclientprotocol/sdk` from `^0.14.1` to `^0.25.0`, adopting the stabilized `listSessions` / `resumeSession` agent method names (formerly `unstable_*`), guarding `session/set_config_option` against the newly-schema-legal boolean config values mux doesn't support, and adding a wire-level dispatch regression test.

## Background

The ACP TypeScript SDK moved from 0.14 to 0.25 with several protocol stabilizations:

- 0.16 stabilized `unstable_listSessions` → `listSessions`
- 0.20 stabilized `unstable_resumeSession` → `resumeSession`
- The schema widened `SetSessionConfigOptionRequest.value` to a discriminated union (`string | boolean` via `type: "select" | "boolean"`)

Wire compatibility is preserved: the JSON-RPC method names (`session/list`, `session/resume`, `session/fork`, `session/set_config_option`), `PROTOCOL_VERSION = 1`, capability shapes, and the `Usage`/`ndJsonStream` surfaces are identical between 0.14.1 and 0.25.0 (verified by diffing the installed package against a downloaded 0.14.1 tarball). `unstable_forkSession` remains experimental upstream and is unchanged.

## Implementation

- `src/node/acp/agent.ts`: renamed the two stabilized methods (and their assert-message prefixes). Added a guard in `setSessionConfigOption` rejecting non-string values with a spec-correct JSON-RPC `-32602 InvalidParams` error — pre-0.25 SDK zod schemas rejected boolean values at the schema layer with the same code, so this restores prior wire semantics. Mux only exposes select-type config options.
- `tests/ipc/acp.sessionMethods.test.ts`: renamed call sites; extracted `createMockServer` from the harness and added `createWireHarness` — two cross-wired `ndJsonStream` pipes connecting a real `ClientSideConnection` to the real `AgentSideConnection`/`MuxAgent`. A new wire-level test pins JSON-RPC routing for `session/list`, `session/resume`, and `session/set_config_option`, because Agent-interface methods are _optional_ in the SDK: a missed stabilization rename typechecks cleanly and passes direct-call tests while the wire silently answers `METHOD_NOT_FOUND`. This matters again when `unstable_forkSession` eventually stabilizes.
- New direct-call test covering the boolean config value rejection.

### Deep-review findings addressed

The built-in `deep-review-workflow` was run on the upgrade commit; verdict "sound and safe to land" with five low-severity findings, all fixed here:

| Finding                                                                                         | Severity | Fix                            |
| ----------------------------------------------------------------------------------------------- | -------- | ------------------------------ |
| Boolean-rejection branch (the upgrade's only behavioral change) had zero coverage               | P3       | New direct-call test           |
| Optional Agent-interface methods make missed renames invisible to typecheck + direct-call tests | P3       | New wire-level round-trip test |
| Assert-based rejection surfaced as `-32603 Internal error` instead of `-32602 Invalid params`   | P4       | `RequestError.invalidParams`   |
| Test labels still said "unstable" for stabilized methods                                        | P4       | Renamed describes/test names   |
| Stale `// ACP SDK v0.14.x ...` comment                                                          | P4       | Version pin dropped            |

## Validation

- `make static-check` green; frozen-lockfile install passes under the pinned Bun 1.3.5.
- All 8 ACP jest suites pass (`tests/ipc/acp.*.test.ts`), including the two new tests.
- Live end-to-end verification against the built CLI (`dist/cli/index.js acp`) in an isolated `MUX_ROOT` sandbox:
  - **acpx 0.10.0** (itself on SDK ^0.22.1): `sessions new`, `sessions list` (exercises renamed `listSessions`), `status`, `set agentMode plan`, and a full real prompt turn (initialize → session/new → `file_read` tool call → correct reply → `end_turn`).
  - Raw wire script using SDK 0.25's `ClientSideConnection` over stdio: initialize, session/list, session/resume, session/load, session/fork, set_config_option (string accepted, boolean rejected; connection survives the rejection).
- Known non-regression: `tests/ipc/acp.integration.test.ts` "prompt completes with usage" fails locally with `Forbidden` — reproduced identically on 0.14.1 before the upgrade (local proxy credential issue); expected to pass in CI with its own `ANTHROPIC_API_KEY`.

## Risks

Low. The renames are wire-transparent (same JSON-RPC method strings), and clients that don't send boolean config values see no behavioral change. The one behavioral delta — boolean config values now rejected by an explicit agent-side guard instead of the SDK's zod schema — returns the same `-32602` error code as before. Affected area is limited to the ACP server surface (`mux acp`), used by external editor integrations.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$59.57`_

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -6,7 +6,7 @@
       "name": "mux",
       "dependencies": {
         "@1password/sdk": "^0.4.0",
-        "@agentclientprotocol/sdk": "^0.14.1",
+        "@agentclientprotocol/sdk": "^0.25.0",
         "@ai-sdk/amazon-bedrock": "^4.0.101",
         "@ai-sdk/anthropic": "^3.0.75",
         "@ai-sdk/deepseek": "^2.0.33",
@@ -81,8 +81,8 @@
         "quickjs-emscripten-core": "^0.31.0",
         "react": "18.3.1",
         "react-colorful": "^5.6.1",
-        "react-resizable-panels": "^3.0.6",
         "react-dom": "18.3.1",
+        "react-resizable-panels": "^3.0.6",
         "react-router-dom": "^7.11.0",
         "recharts": "^2.15.3",
         "rehype-harden": "^1.1.5",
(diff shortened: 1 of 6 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

PR #3510, opened 2026-06-11
🤖 feat: seed default Fable 5 → Opus 4.8 refusal fallback (one-time, user-respecting)

ThomasK33:
## Summary

Ships a sensible out-of-the-box refusal fallback: when **Claude Fable 5** refuses with zero output, the turn retries on **Opus 4.8**. The default is seeded into `~/.mux/config.json` exactly once (guarded by `migrations.defaultModelFallbacksSeeded`), so users who edit or delete the chain are never overridden by later app updates.

## Background

#3505 added opt-in per-model refusal fallback chains but shipped no defaults. Fable 5 runs with safeguards that refuse some requests Opus happily serves (Anthropic itself falls back to Opus 4.8 server-side for flagged requests), so Fable → Opus 4.8 is the natural default. The requirement: a one-time seed that respects user intent on every subsequent load/update.

## Implementation

- `DEFAULT_MODEL_FALLBACKS` (`src/common/utils/ai/modelFallbacks.ts`): `anthropic:claude-fable-5 → anthropic:claude-opus-4-8`, built from `KNOWN_MODELS` so IDs cannot drift. Deep-frozen — entries are spread by reference into live configs, so accidental in-place mutation crashes fast.
- **Seed-once on load** (`src/node/config.ts`): if `migrations.defaultModelFallbacksSeeded` is unset, missing defaults are merged into the **raw** on-disk map (gap-checked by canonicalized key) and the flag is persisted via the existing atomic, crash-safe write-back. Existing user entries — including gateway-prefixed keys and hand-edited disabled tombstones whose chains sanitize away — block the seed and survive byte-identical on disk.
- **Fresh installs**: the default config includes the chain and the flag; the first save locks seed-once semantics.
- **Downgrade hardening** (from deep-review finding FB-1): `normalizeConfigMigrations` now passes through *unknown* true-valued migration flags and `AppConfigMigrationsSchema` gained a boolean catchall, so flags introduced by newer versions survive a downgrade-to-this-version + save instead of being stripped (which would re-run their one-time migrations after re-upgrade). Downgrades to versions predating this change remain unfixable retroactively; the constant's doc comment scopes the guarantee accordingly.
- The `saveConfig` migrations gate now keys on presence of any flag instead of a hand-enumerated OR-list.

## Validation

- A deep-review workflow pass (4 lanes + adversarial verification) found 6 issues (2×P3, 4×P4); all are fixed in the second commit, including new regression tests for the merge path, gateway-prefixed keys, tombstones, unknown-flag durability, the `sanitizeModelFallbacks(DEFAULT_MODEL_FALLBACKS)` round-trip invariant, and the frozen constant.
- Dogfooded the full lifecycle against the real `Config` class: pre-update config → chain active + flag persisted; delete chain → restart → stays deleted; fresh install → chain active immediately; newer-version unknown flag survives an unrelated save on this version; hand-edited tombstone survives the seed.
- `make static-check`, `make typecheck`, full `src/node` + config/fallback/schema suites green (10 pre-existing flaky UI failures reproduced identically on `origin/main`).

## Risks

Low. The seed runs inside the existing parse-success branch of `loadConfigOrDefault` with try/catch-wrapped persistence, so startup cannot crash on write failure (idempotent re-seed next launch). Worst verified outcome is the known, documented limitation: a downgrade to a pre-flag version + save + re-upgrade re-seeds the (benign, re-deletable) default chain.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$3.16`_

### DIFF
--- a/src/common/config/schemas/appConfigOnDisk.ts
+++ b/src/common/config/schemas/appConfigOnDisk.ts
@@ -73,6 +73,8 @@ export const ModelFallbacksSchema = z.record(z.string(), ModelFallbackEntrySchem
 export const AppConfigMigrationsSchema = z.object({
   execSubagentDefaultsSplit: z.boolean().optional(),
   userPreferencesInitialized: z.boolean().optional(),
+  /** One-time seed of DEFAULT_MODEL_FALLBACKS; never re-applied once true. */
+  defaultModelFallbacksSeeded: z.boolean().optional(),
 });
 
 export const FeatureFlagOverrideSchema = z.enum(["default", "on", "off"]);
--- a/src/common/config/schemas/appConfigOnDisk.ts
+++ b/src/common/config/schemas/appConfigOnDisk.ts
@@ -70,12 +70,17 @@ export const ModelFallbackEntrySchema = z.object({
 /** Per-model fallback chains, keyed by canonical source model. */
 export const ModelFallbacksSchema = z.record(z.string(), ModelFallbackEntrySchema);
 
-export const AppConfigMigrationsSchema = z.object({
-  execSubagentDefaultsSplit: z.boolean().optional(),
-  userPreferencesInitialized: z.boolean().optional(),
-  /** One-time seed of DEFAULT_MODEL_FALLBACKS; never re-applied once true. */
-  defaultModelFallbacksSeeded: z.boolean().optional(),
-});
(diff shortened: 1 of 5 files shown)


ThomasK33:
@codex review

PR #3511, opened 2026-06-11
🤖 fix: archive selects same-project chat above instead of jumping projects

ethanndickson:
## Summary

Archiving the currently-selected chat now lands on the chat **above** it in the sidebar (instead of below), and prioritises chats in the **same project** before falling back to other projects.

## Background

Archiving a chat picked the next selection via a pure DOM-order walk (`findAdjacentWorkspaceId`), preferring the row below with no awareness of project boundaries. Archiving the last chat in a project would jump the user into a different project's chat, which is disorienting. The delete path already preferred same-project siblings, so archive was the odd one out.

## Implementation

`findAdjacentWorkspaceId` still reads the rendered sidebar DOM (the canonical visual order shared with Ctrl+J/K), but now takes required project context and applies a single preference chain: same-project row above → same-project row below → any row above → any row below → `null`. When the archived row isn't rendered (collapsed project/section), all other visible rows are treated as "below" so the same chain applies. The archive handler in `WorkspaceContext` passes the archived workspace's `projectPath` and resolves rows to projects via the in-memory metadata map; when nothing is selectable it falls back to the project page as before. Ctrl+J/K navigation uses `getVisibleWorkspaceIds` directly and is unaffected.

## Risks

Low. The change is scoped to the post-archive selection path; covered by new happy-dom unit tests for same-project priority, above/below ordering, and the collapsed-row fallback.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$4.60`_

### DIFF
--- a/src/browser/contexts/WorkspaceContext.tsx
+++ b/src/browser/contexts/WorkspaceContext.tsx
@@ -1191,12 +1191,16 @@ export function WorkspaceProvider(props: WorkspaceProviderProps) {
           // removing it from the active metadata map. Otherwise we can briefly render the
           // root-route empty shell while still on `/workspace/:id`.
           //
-          // Prefer the next workspace in sidebar DOM order (like Ctrl+J) so the user
-          // stays in flow; fall back to the project page when no siblings remain.
+          // Prefer nearby chats in the same project first so archive does not jump the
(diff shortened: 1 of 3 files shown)


ethanndickson:
@codex review

PR #3513, opened 2026-06-11
🤖 feat: workflow CLI worktree trust, skip warnings, and `mux trust` command

ThomasK33:
## Summary

Fixes `mux workflow` silently hiding project `.mux/workflows` definitions (worktree trust mismatch + zero feedback) and adds a top-level `mux trust` command so headless environments can grant project trust without the desktop app or server.

## Background

Authoring a valid `.mux/workflows/echo.js` and running `mux workflow -e dynamic-workflows list` from a mux workspace checkout showed only built-in workflows. Two distinct problems:

1. **Worktree trust mismatch.** Trust is keyed in `~/.mux/config.json` by registered project path. The CLI resolved the project dir via `git rev-parse --show-toplevel`, which returns the *worktree* path for mux workspace checkouts (`~/.mux/src/<project>/<branch>`) — never a config key — so trust resolved `false`. Inconsistent with the desktop app, which scans the workspace checkout but checks trust against `metadata.projectPath` (`resolveWorkflowContext` in `src/node/orpc/router.ts`).
2. **Silent filtering.** `list`/`show` omitted untrusted project workflows with no output at any log level, indistinguishable from a discovery bug. (`run` already threw an actionable trust error.)

And once warned, there was no headless remedy: trust could only be granted via desktop Settings → Security, which doesn't help on machines without the desktop app/server or for projects never added to mux.

## Implementation

- `resolveProjectTrusted` accepts trust granted to either the checkout path or, for linked worktrees, the main repository path (parent of `git rev-parse --git-common-dir`), reusing the canonical `isProjectTrusted` helper. Main checkouts/bare repos/submodules resolve to `null` (no false trust); explicit non-worktree `--dir` paths are trusted exactly as given so sub-project trust stays possible.
- `list`/`show` print a stderr warning when skipped project-local workflow candidates exist (stdout stays machine-parseable); the warning and `run`'s trust error give the copy-pasteable `mux trust` remedy.
- New `mux trust [--revoke] [--dir <path>] [--json]` (`src/cli/trust.ts`): writes trust via `Config.editConfig`, mirroring the desktop `projects.setTrust` handler incl. creating a minimal project entry for unregistered projects; a running desktop instance picks the change up through its config watcher. Run from a linked worktree, trust lands on the main repository path instead of accumulating ephemeral worktree entries. Top-level (not under `workflow`) because trust also gates hooks and `.mux` config; no experiment gate for the same reason; no confirmation prompt — an explicit trust command is deliberate consent (precedent: `direnv allow`, `mise trust`) and the output states what trust enables. Idempotent; `--revoke` sets `trusted: false` (matching desktop semantics, preserving the explicitly-declined state).
- Shared helpers (`resolveProjectDir`, `findGitRoot`, `findMainRepoDir`, `resolveProjectTrusted`) live in `trust.ts` with `workflow.ts` importing them; `exitAfterStdoutFlush` moved to dependency-free `processExit.ts` so `index.ts` can import it statically without defeating the CLI's lazy-loading of heavy subcommand modules. Routed like other headless commands (CLI-only, not bundled in Electron).

## Validation

- Reproduced the original bug in an isolated `MUX_ROOT` (untrusted, trusted, and worktree-of-trusted scenarios) before fixing.
- Dogfooded the full headless journey end-to-end: warning → `mux trust` → list/run work → idempotent re-trust → `--revoke` → run blocked with actionable error; plus worktree resolution against a real config (`Resolved linked git worktree … to its main repository.`).
- New tests: trust grant/revoke cycle on an unregistered project (routed through `index.ts` dispatch), worktree trust recorded for the main repo path, untrusted skip warning, and worktree trust inheritance in workflow discovery.

## Risks

Trust gating is a security boundary; changes are deliberately narrow. Worktree fallback only *widens* trust from a path the user already explicitly trusted to its linked worktrees (matching existing desktop behavior). `mux trust` writes the same config shape as the desktop handler. Regression severity: low–moderate, scoped to CLI workflow discovery/trust; the desktop trust path is untouched.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$27.01`_


ThomasK33:
@codex review

ThomasK33:
Addressed the P2 (src/cli/trust.ts:142): `runTrust` now re-reads the config from disk after `editConfig` and throws `Failed to persist trust change …` (exit 1) when the persisted state doesn't match — covering unwritable `MUX_ROOT`, failed disk writes, etc. Verified `loadConfigOrDefault` has no in-memory caching, so the re-read reflects actual disk state in both grant and revoke directions. Added a regression test (`fails loudly when the trust change cannot be persisted`) using `MUX_ROOT` pointed at a regular file.

ThomasK33:
@codex review

Please take another look — addressed the persistence-verification finding.

ThomasK33:
Addressed the P1 (src/cli/trust.ts:135): revoke now also clears direct `trusted` entries for the checkout paths themselves (worktree path / git toplevel) when they differ from the canonical main-repo path, without creating entries where none existed (absence already means untrusted). Additionally, both the idempotency comparison and the post-write verification now use the checkout's *effective* trust (`resolveProjectTrusted(projectDir)` — direct entry or main-repo fallback) instead of only the canonical entry, so the command can no longer report a revocation that leaves the checkout trusted. Added a regression test seeding both a main-repo and a direct worktree `trusted: true` entry and asserting both are `false` after `--revoke` from the worktree.

ThomasK33:
@codex review

Please take another look — revoke now clears direct worktree entries and verifies effective trust.

PR #3514, opened 2026-06-11
🤖 feat: imply dynamic-workflows experiment for mux workflow, add wf alias

ThomasK33:
## Summary

`mux workflow` no longer requires passing `-e dynamic-workflows` on every invocation: choosing the subcommand is already an explicit opt-in, so the experiment is now implied per-invocation (never persisted). The help text carries an experimental disclaimer instead, and `mux wf` is a new shorthand alias.

## Background

Previously every command failed with `mux workflow requires the dynamic-workflows experiment. Re-run with -e dynamic-workflows or enable it in settings.` — forcing a redundant flag on a command the user already deliberately invoked. This was confusing (and the intuitive `mux -e dynamic-workflows workflow list` spelling didn't even parse).

## Implementation

- **`src/cli/workflow.ts`**: removed the `enforceWorkflowExperiment` gate and the now-dead `workflowExperimentEnabled` helper. The experiments object passed to workflow-owned child agents already hardcoded `dynamicWorkflows: true`; the gate was the only consumer of the flag. The program description now prints the experimental disclaimer, and the `-e` option help clarifies it enables *additional* experiments for workflow-owned child agents. `-e dynamic-workflows` remains accepted as a no-op for backward compatibility (locked in by a regression test).
- **`src/cli/index.ts`**: `wf` routes to the same workflow entry point; the top-level help stub registers `.alias("wf")` and is marked `(experimental)`.
- **Dead-code cleanup** (deep-review finding): the orphaned `featureFlagOverrides` config plumbing (`getFeatureFlagOverride`/`setFeatureFlagOverride`, schema field, load/save paths) is removed — its only caller was the deleted gate. Old `config.json` files still parse via schema `passthrough()`.
- **Docs**: `docs/reference/cli.mdx` note/examples/options updated (+ regenerated built-in skill snapshot).

## Validation

- Ran the built-in `deep-review-workflow` against this branch; fixed all 4 verified findings (orphaned config plumbing, missing `-e` backward-compat coverage, `INDEX_ENTRY` version.ts prerequisite comment, `-e` help-string/docs alignment).
- Dogfooded every scenario from the original report: `mux workflow` help shows the disclaimer, `mux workflow list` works without flags, `mux wf list`/`show`/`run` work end-to-end against a trusted project workflow, `-e dynamic-workflows` still accepted.
- `make static-check`, `bun test src/cli/workflow.test.ts src/node/config.test.ts src/common/config/schemas/appConfigOnDisk.test.ts` (127 tests) all green.

## Risks

Low. The CLI gate removal only widens access to an experimental, already-opt-in command. One behavioral nuance: a persisted `featureFlagOverrides: {"dynamic-workflows": "off"}` no longer suppresses the CLI — acceptable since invoking the command is itself the opt-in, and that config mechanism had no UI and is removed here.

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$5.39`_


ThomasK33:
@codex review

PR #3515, opened 2026-06-11
🤖 fix: workflow list/action-list rows on narrow viewports

ThomasK33:
## Summary

`workflow_list` / `workflow_action_list` tool cards broke at narrow container widths (~375px): scope badges stretched to full-card width in the definitions list, and the action list's nowrap description sat in an `auto` grid column that inflated to the text's full width — starving sibling columns, pushing action names off-screen, and clipping effect badges past the card edge.

## Implementation

- Both row layouts now collapse to a stacked form under `@container(max-width:640px)`: name + badges on row 1, truncating description spanning row 2. Badges share one flex cell so they keep natural width, and the description lives in `minmax(0,1fr)` tracks — nowrap text never sits in an `auto` column.
- Definitions list: the trailing `blocked` badge moved inline next to the scope badge (desktop grid 4→3 columns).
- New pinned-mobile stories (`WorkflowListNarrow`, `WorkflowActionListNarrow`) assert the stacked geometry via play functions. The Chromatic mode pins `mobile1` + `hasTouch` and stays inline so the snapshot budget estimator counts the modes. A 375px `maxWidth` decorator engages the container query in the Storybook test-runner too, which applies neither `globals.viewport` nor Chromatic modes and runs plays at desktop window size — that caveat is now documented in AGENTS.md alongside a new "verify UI at mobile widths" styling rule (regenerating `builtInSkillContent.generated.ts`).
- Snapshot budget cap 259→262: main's retained set was already at 260 (the cap lagged one story behind), plus the 2 new pinned-mobile stories.

## Validation

- Full Storybook interaction suite green locally against a CI-style static build (`make storybook-build` + `http-server`): 242 passed, 1 skipped, 0 failed. The narrow plays fail without the width decorator and pass with it, so the assertions are behavioral in both directions.
- Visual checks from the static build: at 375px both lists stack cleanly with natural-width badges and truncated descriptions; at 1000px the wide single-line grid is unchanged.

## Risks

Low. Scoped to the two workflow tool-call renderers; the desktop grid templates changed shape (definitions 4→3 columns, actions 5→4), verified visually at wide widths.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$69.18`_


ThomasK33:
@codex review

PR #3516, opened 2026-06-11
🤖 fix: mask regex literals in workflow action static export detection

ThomasK33:
## Summary

Fixes the built-in `security.hashFiles` workflow action showing as **blocked** with *"Workflow action must export an execute function"*. The action source was fine — the static validator's source masker misparsed regex literals. The fix masks regex literal bodies and adds the standard tokenizer heuristic for `/` division-vs-regex disambiguation, hardened through five Codex review rounds into full token-class coverage.

## Background

`WorkflowActionRunner.describe()` validates actions without executing them by masking comments/strings (`maskStaticJavaScriptSource`) and then scanning the masked source for top-level `execute`/`reconcile` exports. The masker had no concept of regex literals: in `security.hashFiles`, the regex `/\/\*[\s\S]*?\*\//g` ends with `\*\//g`, whose two consecutive `/` characters were misread as a line-comment start. The rest of the line — including the closing `)` of the enclosing `.replace(` call — was masked away, permanently unbalancing the paren-depth counter used by `isTopLevelStaticMatch`. Every subsequent top-level match (including `module.exports.execute = ...`) was then rejected as "not top-level", surfacing in the UI as a blocked built-in action. User-authored actions with regexes like `/^https?:\/\//` hit the same bug.

## Implementation

`maskStaticJavaScriptSource` now masks regex literal bodies (keeping the `/` delimiters), driven by `isRegExpLiteralStart` — the standard lexer heuristic for `/` disambiguation, where division follows a value token and a regex literal follows operators/punctuation/keywords. Every token class to the left of a `/` is classified:

- **identifiers/numbers**: division, unless an expression keyword (`return`, `typeof`, ...);
- **string/template literals**: quote delimiters are kept in the masked output (contents still blanked) so `"10" / 2` stays division;
- **regex literals**: delimiters kept, so `/x/ / 2` stays division while `a / /re/.source` still masks the right-operand regex;
- **postfix `++`/`--`**: division (requiring exactly two — `a+++/x/` lexes as `a++ + /x/`);
- **`}`**: resolved via `isObjectLiteralEnd`, which walks back to the matching `{` (reliable in masked text) and classifies by the preceding token — block contexts (`)`, `;`, `{`, `}`, `=>`, `do`/`else`/class names, file start) keep regex position, any operator/expression context means object literal → division;
- **`)`**: resolved via `isControlHeaderEnd` — `if`/`while`/`for`/`switch`/`with` headers put the following `/` in regex position (`if (x) /re/.test(s)`), call/grouping parens mean division.

As a safety net, a `/` in regex-start position that never closes on its line (regex literals cannot span lines) is treated as division instead of masking the rest of the line — bounding the damage of any residual misclassification (e.g. label/`case`-block edges).

## Validation

- Every fix was verified red→green (failing test first, then the fix).
- New regression tests:
  - regex literals containing `//` mixed with real division;
  - one-line/minified division after string, template, postfix `++`, object literal (incl. ternary/`||`-introduced, and followed by later regexes on the same line), and regex-literal values;
  - block-`}` and control-header-`)` counter-cases where regexes with depth-corrupting `(`/`[` must stay masked;
  - **every built-in workflow action must pass `describe()`** through the real registry→runner path — the coverage gap that let the broken built-in ship (previously only listing was tested).
- Dogfooded the exact failing path from the bug report: resolving + describing `security.hashFiles` (`/__mux_builtin_workflow_actions__/security/hashFiles.js`) returns its metadata cleanly.
- Full workflow suite (225+ tests) and `make static-check` green on the rebased branch.

## Risks

Contained to static action validation (`describe`), not runtime execution. The heuristic now mirrors standard tokenizer disambiguation rules; remaining ambiguities (labels, `case`-block edges) are not reachable from valid code that previously described successfully, and the unclosed-candidate fallback bounds any misclassification to leaving a single `/` unmasked. All built-in actions are describe-tested, so a regression here fails CI rather than shipping a blocked action.

## Pains

The masker is a heuristic lexer, and Codex review correctly identified successive token classes whose value-ness was erased by masking (strings, postfix operators, object literals, ternary operators, regex literals). Each round was fixed red→green; the final design (keep literal delimiters + enumerate block contexts + same-line closing invariant) converges on the standard Esprima-style tokenizer heuristic rather than an enumeration of special cases.

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$4.61`_

### DIFF
--- a/src/node/services/workflows/WorkflowActionRunner.test.ts
+++ b/src/node/services/workflows/WorkflowActionRunner.test.ts
@@ -361,6 +361,55 @@ describe("WorkflowActionRunner", () => {
     expect(description.hasReconcile).toBe(false);
   });
 
+  test("detects exports in sources containing regex literals", async () => {
+    using tmp = new DisposableTempDir("workflow-action-regex-mask");
+    const sourcePath = path.join(tmp.path, "regex.js");
+    // Regression: "//" inside regex literals (URL matchers, comment strippers) was
+    // misread as a line comment, swallowing the rest of the line. That unbalanced the
+    // masked source and hid the execute export ("must export an execute function").
+    const source = `
+      module.exports.metadata = { version: 1, description: "Regex", effect: "read" };
+      function stripBlockComments(text) {
+        return String(text).replace(/\\/\\*[\\s\\S]*?\\*\\//g, "");
+      }
+      function isHttpUrl(text) {
+        return /^https?:\\/\\//.test(text);
+      }
+      const half = 10 / 2;
+      module.exports.execute = async function (input) {
+        return { url: isHttpUrl(String(input)), ratio: stripBlockComments(String(input)).length / half };
+      };
+    `;
+    await fs.writeFile(sourcePath, source, "utf-8");
+    const runner = new WorkflowActionRunner();
+
+    const description = await runner.describe(createAction(sourcePath, source));
+
+    expect(description.metadata.description).toBe("Regex");
+    expect(description.hasReconcile).toBe(false);
+  });
+
+  test("describes every built-in workflow action", async () => {
+    // Built-in sources must always pass static describe validation; a failure here
+    // surfaces in the UI as a "blocked" action (e.g. security.hashFiles, whose regex
+    // literals previously broke the static export detection).
+    using tmp = new DisposableTempDir("workflow-action-built-in-describe");
+    const registry = new WorkflowActionRegistry({
+      projectRoot: path.join(tmp.path, "project-actions"),
+      globalRoot: path.join(tmp.path, "global-actions"),
+    });
+    const runner = new WorkflowActionRunner();
+
+    const actions = await registry.listActions({ projectTrusted: false });
+
+    expect(actions.length).toBeGreaterThan(0);
+    for (const action of actions) {
+      const resolved = await registry.resolveAction(action.name, { projectTrusted: false });
+      const description = await runner.describe(resolved);
+      expect(description.metadata.description).toBeTruthy();
+    }
+  });
+
   test("does not rewrite export syntax inside action strings", async () => {
     using tmp = new DisposableTempDir("workflow-action-export-template");
     const sourcePath = path.join(tmp.path, "template.js");
--- a/src/node/services/workflows/WorkflowActionRunner.test.ts
+++ b/src/node/services/workflows/WorkflowActionRunner.test.ts
@@ -400,6 +400,13 @@ describe("WorkflowActionRunner", () => {
       `module.exports.metadata = { version: 1, description: "Division", effect: "read" }; const n = \`10\` / 2; module.exports.execute = async () => ({ n });`,
       `module.exports.metadata = { version: 1, description: "Division", effect: "read" }; let count = 4; const n = count++ / 2; module.exports.execute = async () => ({ n });`,
       `module.exports.metadata = { version: 1, description: "Division", effect: "read" }; const n = { valueOf() { return 10; } } / 2; module.exports.execute = async () => ({ n });`,
+      // Object-literal division followed by a real regex on the same line: the later
+      // "/" must not be mistaken for the closing delimiter of a regex starting at the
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed the P2 (division after masked literals): `maskStaticJavaScriptSource` now keeps the quote delimiters of string/template literals (contents still masked), and `isRegExpLiteralStart` treats a trailing `"`, `'`, or `` ` `` as a value token, so `const n = "10" / 2; module.exports.execute = ...` is lexed as division. Added a red→green regression test with single-`/` one-line sources for both string and template literals (0f090176b).

ThomasK33:
@codex review

Addressed the P2 (postfix/object value endings): instead of trying to enumerate every value-ending token, the masker now exploits the lexical invariant that regex literals cannot span lines — a `/` in regex-start position with no closing `/` on the same line is left unmasked as division (`findRegExpLiteralEnd`). This covers `count++ / total`, `{ valueOf() { return 10; } } / 2`, and any future unrecognized value ending with a single `/` on the line. Postfix `++`/`--` is also recognized explicitly (requiring exactly two, since `a+++/x/` lexes as `a++ + /x/`) so divisions after them are safe even when another `/` appears later on the line. Both examples from the review are red→green regression tests (1ba8bc5aa).

ThomasK33:
@codex review

Addressed the P2 (`}` before `/` with a later regex on the line): `isRegExpLiteralStart` now resolves the `}` ambiguity the way tokenizer heuristics do — `isObjectLiteralEnd` walks back to the matching `{` in the masked prefix (reliable since strings/comments/regex bodies are already blanked) and classifies it by the preceding token: expression contexts (`=`, `(`, `[`, `,`, `:`, or expression keywords like `return`) open object literals → `/` is division; statement contexts (`)`, `;`, `=>`, `do`/`else`, file start) keep block → regex. Your exact example plus a block counter-case (`if (x) {} /^[(]/.test(...)` where failing to mask the regex would corrupt paren depth) are both red→green regression tests (67c5ee6a1).

ThomasK33:
@codex review

Addressed the P2 (object literals after ternary/logical operators) by flipping `isObjectLiteralEnd`'s default: block contexts are now the enumerated cases (`)`, `;`, `{`, `}`, `=>`, non-expression identifiers like `do`/`else`/class names, file start), and **any other** preceding operator or punctuation (`?`, `:`, `||`, `&&`, arithmetic, `=`, `(`, `[`, `,`) leaves `{` in expression position → object literal → division. Your ternary example plus an `||` variant are regression tests.

Also pre-emptively closed the remaining standard disambiguation rule: `)` is now classified by matching back to its `(` — control-statement headers (`if`/`while`/`for`/`switch`/`with`) put the following `/` in regex position (`if (x) /^[(]/.test(s)` now masks correctly), while call/grouping parens remain division. Both directions are covered by new tests (e80e39f6f).

Remaining known limits are label/`case`-block edges (`foo: {} /re/`), where expression position is assumed — these are not reachable from valid division or regex code that previously described successfully, so no regression class remains there.

ThomasK33:
@codex review

Addressed the P2 (value token after unflagged regex literals) the same way strings were fixed: the masker now keeps the regex's `/` delimiters and masks only the body. `isRegExpLiteralStart` classifies a trailing `/` by walking back across the space-masked body — landing on the opening `/` means the slash closed a regex value (so `/x/ / 2` is division); landing on a value operand means the slash was a division operator (so `a / /re/.source` still opens — and masks — a regex). Your exact example plus that counter-direction (regex as the right operand of division, with depth-corrupting `(`/`[` in its body) are red→green regression tests (3235d6211).

With this, every token class on the left of `/` is now classified (identifiers/keywords, numbers, `)` call vs control header, `]`, `}` object vs block, string/template/regex delimiters, postfix `++`/`--`), matching the standard tokenizer heuristic.

PR #3517, opened 2026-06-11
🤖 feat: surface completed task/workflow reports across compaction

ThomasK33:
## Summary

Makes completed sub-agent/workflow reports recoverable after context compaction: a capped index of re-fetchable report handles is injected via the post-compaction attachment machinery, and completed `task_await`/`workflow_run` results now carry a note that the report is durable and re-fetchable by ID.

## Background

Workflow results and sub-agent reports are already durably persisted (session-dir `workflows/<runId>/run.json` and `subagent-reports/<taskId>/report.json`) and `task_await` can already re-fetch them by ID. But after compaction the model loses the runId/taskId **handles**, so expensive, long-running results become effectively unreachable and the work gets re-run. This change surfaces the existing durable results instead of duplicating them into new files.

## Implementation

1. **New `completed_reports_index` post-compaction attachment** — built in `AttachmentService.generateCompletedReportsAttachment()` from the two existing durable stores (`subagent-reports.json` index + `WorkflowRunStore.listRuns()`). Entries are handles only (ID, kind, title, completed-at, ~token estimate) — never report content. Filters:
   - direct children only (grandchild reports are synthesized into the child's report)
   - workflow-owned sub-agent reports excluded (consumed through the workflow run's report)
   - completed **before** the compaction cutoff (immediate injection: now; periodic re-injection: latest boundary timestamp) so reports still visible in the active epoch aren't duplicated
   - completed workflow runs only; capped at 20, newest first
   - **self-healing**: persisted index rows are validated per-field (null/non-object rows, malformed ids/timestamps, non-array ownership markers are skipped/degraded) so one corrupt row can never break post-compaction injection
2. **Budget-aware rendering** — the renderer emits one line per report plus a `task_await(task_ids: ["<id>"], timeout_secs: 0)` re-fetch instruction, prioritized ahead of bulky skills/diff blocks. Titles (model-provided, unbounded) are capped at 80 chars, and entries pack greedily under the 80k injection budget with an "omitted N handles" note — so a single oversized title can never drop every handle.
3. **`COMPLETED_REPORT_REFETCH_NOTE`** appended to completed `task_await` results (agent tasks + workflow runs) and foreground-completed `workflow_run` results (schema gained an optional `note`), and the `task_await` tool description now documents re-fetch-by-ID across compaction.

## Validation

- New tests: index generation against real artifact writers + a real `WorkflowRunStore` (cutoff/scope filtering, cap+ordering, malformed-row skipping, empty→null) and renderer tests (handle rendering, title capping, partial packing + truncation note, budget priority over file diffs).
- `bun test src/node/services/tools/ src/browser/utils/messages/` — 1,070 tests green; compaction handler/monitor and agentSession post-compaction suites green.

## Risks

Low. The attachment only adds a small injected block post-compaction (existing 80k-char budget applies); generation is read-only over session-dir artifacts, validates each persisted row defensively, and returns null when nothing qualifies. The `note` field is additive on tool results — existing exact-shape test expectations were updated accordingly. Worst-case regression is a missing or slightly stale index entry, which degrades to today's behavior.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$23.53`_

### DIFF
--- a/src/browser/utils/messages/attachmentRenderer.test.ts
+++ b/src/browser/utils/messages/attachmentRenderer.test.ts
@@ -8,6 +8,7 @@ import type {
   PlanFileReferenceAttachment,
   LoadedSkillsSnapshotAttachment,
   EditedFilesReferenceAttachment,
+  CompletedReportsIndexAttachment,
 } from "@/common/types/attachment";
 
 describe("attachmentRenderer", () => {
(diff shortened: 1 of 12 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed the P2: `generateCompletedReportsAttachment` now validates each persisted index entry (`childTaskId` is a non-empty string, `updatedAtMs` passes `Number.isFinite`) and skips corrupt rows before they can reach the cutoff comparison or the renderer's `toISOString()`. Added a regression test that corrupts `subagent-reports.json` with missing/NaN timestamps and a missing id, asserting only the valid entry is surfaced (b7d357201).

ThomasK33:
@codex review

Addressed round 2: rows are now shape-checked (`entry == null || typeof entry !== "object"`) before any field access, corrupt `workflowOwnedAncestorWorkspaceIds` values degrade to "not workflow-owned" via an `Array.isArray` guard instead of throwing, and `title`/`reportTokenEstimate` reads are type-validated too. Regression test extended with null/string rows and a non-array ownership marker (308dbd66f).

ThomasK33:
@codex review

Addressed round 3: titles are now capped at 80 chars per entry, and the index renders via greedy per-entry packing (newest first, header/footer reserved) with an "(omitted N completed report handles)" truncation note — mirroring the skills/diffs budget renderers — instead of all-or-nothing. Added tests for title capping and partial packing under a tight budget (5a25f3b70).

PR #3518, opened 2026-06-11
🤖 feat: maxParallel sliding-window option for workflow parallelAgents

ThomasK33:
## Summary

`parallelAgents(specs, options?)` now accepts `{ maxParallel: number }`, capping how many sub-agent tasks run at once with a sliding window: queued specs start the moment any running one finishes. The deep-research verification phase uses it (smart mode: 12 live verifiers) instead of sequential fixed-size batches, and deep-review/security-scan verify/fix waves are now capped at 12 live agents.

## Background

Deep-research verification previously sliced verifier specs into sequential `parallelAgents` batches (`runParallelAgentBatches`), so each batch stalled on its slowest verifier — execution time was dominated by per-batch stragglers. Meanwhile the other built-in workflows had the opposite problem: single unbounded waves. security-scan's verification fan-out scales with `--max-findings`, clamped only at 1000, so a large budget could launch hundreds of concurrent `exec` verifiers in one wave.

## Implementation

- **Host runner** (`WorkflowRunner.runAgentStepsInParallel`): an optional second sandbox argument is parsed/validated (`maxParallel` must be a positive integer). When `maxParallel < wave size`, a new `AsyncSemaphore` (FIFO, styled after `AsyncMutex`) gates each step: acquire slot → run → release in `finally`.
- **Lazy task creation**: under a window, the bulk `createAgentTasks` fast path is skipped — otherwise all child tasks would be created/started up front, defeating the cap.
- **Failure semantics preserved**: a sibling failure dooms the batch (as before); `batchFailed` is set *before* the slot is released so window-queued steps deterministically observe it and never spawn children after the wave is doomed. Foreground-backgrounding and upstream aborts behave as before.
- **Replay-safe**: step IDs and input hashes are untouched; scheduling options are not part of replay identity, so completed steps from older runs replay cleanly.
- **deep-research**: `verificationBatchSize` → `maxParallelVerifiers` (quick: 4, smart: 12, previously 6 sequential-batch size); `runParallelAgentBatches` deleted.
- **deep-review / security-scan**: verify + fix call sites pass `{ maxParallel: 12 }` (lane fan-outs are ≤ 6/9 by fixed vocabulary and stay uncapped). Behavior is unchanged for waves ≤ 12.
- **Docs**: workflow-authoring skill documents the option and recommends it over manual batch slicing; generated content regenerated.

## Validation

- New deterministic runner test proves window (not batch) semantics: with `maxParallel: 2` and specs a/b/c where a+b block, releasing only **a** admits **c** while **b** still runs; also asserts the bulk-creation path is not used (throwing `createAgentTasks`) and `maxActive` never exceeds 2. Runs through the real QuickJS sandbox, covering second-argument marshaling end-to-end.
- Existing deep-research fan-out cap test (48 verify tasks) now exercises the host-side limiter (`maxActiveVerifyTasks ≤ 12`).
- Full workflows + concurrency suites green locally after rebase; `make static-check` green.

## Risks

- The sliding window only constructs when `maxParallel` is provided and smaller than the wave; all existing `parallelAgents` call sites without options take the exact previous code path. Main regression surface is the parallel-wave failure path (interrupt/abort ordering), which is covered by existing sibling-failure, backgrounding, and resume tests plus the new window tests.
- Built-in workflow behavior change is limited to concurrency scheduling: results, step IDs, vote aggregation, and outputs are unchanged.
---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$13.46`_


ThomasK33:
@codex review

PR #3522, opened 2026-06-11
🤖 feat: coalesce workflow patch events into a single row

ThomasK33:
## Summary

The workflow event list rendered two rows per patch step (`started`, then `applied`/`conflict`/`failed`). Patch events with the same `stepId + sourceTaskId` now coalesce into a single row whose latest status wins, mirroring existing task/action row coalescing.

## Background

A patch step emits a `started` event and later a terminal event for the same step, producing redundant adjacent rows like `apply-p0-p1 / 2a7c53e7f7 / started` followed by `apply-p0-p1 / 2a7c53e7f7 / applied`. Task and action events already collapse into one row each; patch events were the remaining exception.

## Implementation

- `getWorkflowDisplayRows` gains a `patch` branch keyed by `stepId:sourceTaskId` (same pattern as task rows; retries merge into the same row, latest status wins).
- Coalesced patch rows render through the same merged-detail path as action rows (`getWorkflowActionRowDetail` renamed to `getWorkflowMergedRowDetail` and widened), so expanding a row shows `{ started: {...}, applied: {...} }` with no detail loss. Tooltip still references the first raw event sequence.
- An in-flight patch (only `started` seen) renders unchanged.

## Validation

- New UI test: started+applied coalesce into one row with combined expandable details, while a lone `started` patch stays a separate row.
- All 33 `WorkflowRunToolCall` tests pass; `make static-check` green.

## Risks

Low; display-only change in the workflow event list. Worst case is a mis-coalesced row in workflow run history rendering.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$3.83`_

### DIFF
--- a/src/browser/features/Tools/WorkflowRunToolCall.test.tsx
+++ b/src/browser/features/Tools/WorkflowRunToolCall.test.tsx
@@ -626,6 +626,83 @@ describe("WorkflowRunToolCall", () => {
     expect(view.container.textContent).toContain("durationMs");
   });
 
+  test("coalesces patch start and applied events into one row with combined details", () => {
+    const view = render(
+      <ThemeProvider forcedTheme="dark">
+        <TooltipProvider>
+          <WorkflowRunToolCall
+            args={{ name: "implementation", args: {}, run_in_background: true }}
+            status="completed"
+            result={{
+              status: "completed",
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3525, opened 2026-06-11
🤖 fix: wait for pending patch generation in task_apply_git_patch

ThomasK33:
## Summary

`task_apply_git_patch` raced background patch generation: a child task's report is delivered before its `git format-patch` artifact finishes generating, so applying right after task completion failed with "Patch generation is still in progress" — and workflow `applyPatch` steps treated that timing gap as an apply failure needing conflict resolution. The tool now waits for pending artifacts to settle before deciding success/failure.

## Background

Observed in workflow run `wfr_3df02c0fe3d4542c`: a dry-run apply for a freshly completed task returned

```json
{ "success": false, "error": "This task has no ready project patch artifacts.",
  "projectResults": [{ "status": "skipped", "error": "Patch generation is still in progress for this project." }] }
```

`TaskService` kicks off `GitPatchArtifactService.maybeStartGeneration` (writes a `pending` marker, generates in the background) *before* delivering the report to the parent, so the artifact always exists but may still be `pending` when the parent applies. `task_await` already documented this race ("the artifact may still be pending at read time — task_apply_git_patch does a fresh read"), but the fresh read didn't tolerate pending.

## Implementation

- `applyTaskGitPatchArtifact` polls the artifact (500 ms interval, 120 s cap, abort-aware) while any **relevant** (project_path-filtered) project artifact is `pending`, then proceeds with the settled state. Covers both the agent-facing tool and workflow `applyPatch` steps (both route through this function).
- If generation outlives the wait, the failure is explicit and atomic: *"Patch generation has not finished for this task yet. This is not an apply conflict; retry task_apply_git_patch shortly."* No partial apply of ready siblings (which would make workflows checkpoint the step as `applied` while silently dropping the pending project's commits), and no `conflictPaths`/`failedPatchSubject`, so workflow normalization classifies it `failed`, never `conflict`.
- A cancelled call never falls through into a destructive `git am`: the abort signal is re-checked after the wait before any repo mutation.
- Local sleep helper named `sleepResolvingOnAbort` (the shared `sleepWithAbort` in `@/node/utils/abort` rejects on abort; this one must resolve so the tool returns a structured result).
- Test seams: `pendingGenerationWaitMs` / `pendingGenerationPollIntervalMs` / `pendingGenerationOnPoll`.

## Validation

- Ran the `deep-review-workflow` (adversarial multi-agent review) on the initial fix; it verified 8 findings (3×P3, 5×P4). All are addressed in the second commit except WAIT-CHECKPOINT (P4 follow-up: workflow steps durably checkpoint retryable failures — requires replay-semantics design changes out of scope for this surgical fix).
- 6 behavioral tests in `task_apply_git_patch.test.ts` (16 total pass): pending→ready wait (deterministically exercised via the poll seam), still-pending timeout (non-conflict), abort during wait (prompt return, no repo mutation), mixed ready+pending atomic failure, project_path-scoped apply not blocked by a pending sibling, plus existing suite.
- `make static-check` green; `WorkflowRunner`/`WorkflowTaskServiceAdapter`/`gitPatchArtifactService`/`subagentGitPatchArtifacts`/`task_await` suites pass.

## Risks

Medium-low. The wait adds up to 120 s latency per apply when generation genuinely hangs (normal generation settles in <5 s; startup recovery re-kicks orphaned pending artifacts). The mixed ready+pending case changes behavior from partial-apply-success to atomic retryable failure — intentional, since retry converges and the workflow path self-heals via `allowAlreadyApplied`.

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$8.23`_

### DIFF
--- a/src/node/services/tools/task_apply_git_patch.test.ts
+++ b/src/node/services/tools/task_apply_git_patch.test.ts
@@ -987,28 +987,20 @@ describe("task_apply_git_patch tool", () => {
       ],
     });
 
-    // Flip the artifact to ready while the apply call is polling.
-    const markReady = (async () => {
-      await new Promise((resolve) => setTimeout(resolve, 100));
-      await writePatchArtifact({
-        sessionDir,
-        workspaceId,
-        childTaskId,
-        projectArtifacts: [
-          await buildReadyProjectArtifact({
-            sessionDir,
-            childTaskId,
-            storageKey: "target",
-            projectPath: targetRepo,
-            projectName: "target",
-            childRepo,
-            baseSha,
-            headSha,
-          }),
-        ],
-      });
-    })();
+    const readyProjectArtifact = await buildReadyProjectArtifact({
+      sessionDir,
+      childTaskId,
+      storageKey: "target",
+      projectPath: targetRepo,
+      projectName: "target",
+      childRepo,
+      baseSha,
+      headSha,
+    });
 
+    // Flip the artifact to ready only once the wait loop is observably polling,
+    // so the test deterministically exercises the wait path (never vacuous).
+    const markReadyCalls: Array<Promise<void>> = [];
     const result = await applyTaskGitPatchArtifact(
       {
         ...getTestDeps(),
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3528, opened 2026-06-12
🤖 feat: route /orchestrate to durable workflows with a gate/fixup verification loop

ThomasK33:
## Summary

Teach the built-in `/orchestrate` skill to route long-horizon orchestration to durable workflows, codify a gate/fixup verification loop with an independent verifier, and point research-heavy subtasks at existing workflows like the built-in `deep-research`. Also drops a historical "old Orchestrator pattern" reference from the `workflow-authoring` skill that means nothing to a fresh agent.

## Background

The `/orchestrate` playbook drives implement → integrate → verify loops turn-by-turn from the transcript. For long-horizon work (many phases, known dependency DAG, repeated implement → gate → fixup → re-gate cycles) that loop state is fragile: it dies to context compaction, restarts, and interruptions. Durable workflows already encode exactly this shape (`agent`, `applyPatch` with host-side dry-run + structured conflict results, replayable runs via `workflow_resume`), and the `workflow-authoring` skill even references the orchestration pattern — but the orchestrate skill never pointed back. This closes that gap.

## Implementation

- **`orchestrate.md` — "Long-horizon work: prefer a durable workflow"**: routing section that self-gates on workflow tool availability (workflows are behind the `dynamic-workflows` experiment, and built-in skill content is static markdown, so the conditionality is prose: "if `workflow_run`/`workflow_list` are unavailable, skip this section"). Recommends reusing existing workflows first (`workflow_list` → `workflow_run`, e.g. built-in `deep-research` for deep multi-source investigation instead of hand-rolled `explore` fan-out), then authoring a scratch workflow per the `workflow-authoring` skill. Keeps the interactive task loop as the default for exploratory, user-steered, or small-batch work.
- **`orchestrate.md` — "Gate loop (verification)"**: discover gates once via `explore` → dedicated verify-only verifier (never the implementer, so verdicts stay honest) → route structured failures to a fixup `exec` → bounded repeat with user escalation. Implementation agents may *suggest* gates but only additively; a self-reported "tests pass" is evidence, not a gate result. One line maps the loop to workflow mode (`agent({ outputSchema, onRefusal: "fail" })` + bounded `while`). Step 5 of the patch integration loop now references this section instead of duplicating verification guidance.
- **`workflow-authoring.md`**: replaced "should follow the old Orchestrator pattern: …" with the same concrete resolver recipe minus the historical name.
- **`builtInOrchestrateSkill.test.ts`**: one substance spot-check (`/workflow-authoring/`) so gutting the routing section fails the existing contract test, matching the test's existing style of asserting load-bearing directives rather than exact prose.
- Regenerated `builtInSkillContent.generated.ts`.

## Validation

- `bun scripts/gen_builtin_skills.ts` regenerated embedded content; `make static-check` confirms generated file is up to date.
- `bun test src/node/services/agentSkills/` — 38/38 pass.
- `make static-check` green (typecheck, fmt, ESLint, docs sync).

## Risks

Prose-only changes to skill guidance plus regenerated embeds; no runtime logic touched. Worst case is suboptimal agent guidance, trivially revertable.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `high` • Cost: `$5.26`_


ThomasK33:
@codex review

PR #3531, opened 2026-06-12
🤖 fix: memory editor fills the settings pane and gets a bordered panel

ThomasK33:
## Summary

Settings → Memory previously rendered the memory file editor at a fixed `min-h-48` height floating on the page background, which looked odd on large screens and made the editable area hard to distinguish. The editor (and file list) now fill the available height down to the bottom of the settings pane, sit inside a bordered panel, and the redundant collapsible "GLOBAL" scope card is dropped on single-scope surfaces.

## Background

`MemoryFileEditor` already used `flex-1`/`h-full`, but in Settings nothing above it had a definite height (scroll container → auto-height width wrapper → auto-height section), so `h-full` resolved to nothing and the textarea collapsed to its min-height. The sidebar Memory tab was unaffected because its tabpanel gets an explicit `h-full`.

## Implementation

- `SettingsPage`: the bounded-width content wrapper becomes a `min-h-full` flex column, so sections can opt into filling the remaining height with `flex-1`. Content-sized sections (all use `space-y-*`/`gap-*` roots) keep their natural height and long forms still scroll.
- `MemorySection`: opts in via `min-h-0 flex-1`, and wraps `MemoryBrowser` in a bordered rounded panel so the editable area is visually distinct from the page background.
- `MemoryBrowser`: with a single configured scope (Settings → Memory), the file tree renders directly instead of inside the collapsible `ScopeSection` card — the page heading already names the scope, and the card doubled the chrome inside the new bordered panel. The multi-scope sidebar tab keeps scope cards.
- `MemoryBrowser`/`MemoryFileEditor` roots gain `flex-1` alongside the existing `h-full`; each is inert in the other consumer's context.

## Validation

Dogfooded in a dev-server sandbox with seeded global memory files:

- Settings @ 1900×900: textarea bottom = viewport − page padding (fills exactly); list view flat inside one bordered panel.
- Sidebar Memory tab: textarea bottom matches tabpanel bottom exactly (no regression); scope cards still render with usage stats.
- Mobile @ 375×812: fills to bottom, no horizontal overflow.

## Risks

Low. The only shared-surface change is the settings content wrapper becoming a flex column; verified every other settings section root uses content-sized layouts (`space-y-*`/`gap-*`), which are unaffected by the parent's flex direction.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$17.89`_

### DIFF
--- a/src/browser/features/Memory/MemoryBrowser.tsx
+++ b/src/browser/features/Memory/MemoryBrowser.tsx
@@ -183,7 +183,9 @@ export function MemoryBrowser(props: MemoryBrowserProps) {
   const visibleScopes = scopes.filter((scope) => files?.some((file) => file.scope === scope));
 
   return (
-    <div className="flex h-full min-h-0 flex-col">
+    // h-full fills block parents (sidebar tabpanel); flex-1 fills flex parents
+    // (Settings → Memory). Each is inert in the other context.
+    <div className="flex h-full min-h-0 flex-1 flex-col">
       {error !== null && (
         <div className="text-error px-3 py-2 text-xs" role="alert">
           {error}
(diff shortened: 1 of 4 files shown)


ThomasK33:
@codex review

PR #3540, opened 2026-06-12
🤖 perf: skip composer height measurement for empty drafts

ammar-agent:
## Summary

Eliminates the forced-reflow storm that made chat/workspace switching slow: when the composer draft is empty (the common case), `useAutoResizeTextarea` and `focusMessageInput` no longer measure `scrollHeight` inside the commit phase / pre-paint rAF, letting CSS size the empty textarea instead.

## Background

A CDP CPU profile of a live session (30s, repeated chat switches, ~49k-node transcript) showed switches blocking the renderer main thread for 0.6–3.4s each. The hottest application frame by far (2.18s self time, 14% of busy time; up to 991ms inside a single switch) was `useAutoResizeTextarea`'s layout effect, called from `commitLayoutEffects` during the sync render triggered by the switch click.

The cost is not JS: the effect writes `height = "auto"` then reads `scrollHeight`, which forces a synchronous layout of the entire freshly-mounted (fully dirty) document before first paint. `focusMessageInput` in `ChatInput` repeats the same dance in a pre-paint rAF on every switch.

## Implementation

- `useAutoResizeTextarea`: fast path for `value === ""` — clear any stale inline height (matching the existing send-path convention of `style.height = ""` + "let CSS min-height take over"), record state, and skip measurement entirely. No `scrollHeight` read, no reflow. The first keystroke afterwards takes the existing grow-only path (measures once on a clean tree, no `auto` reset).
- `focusMessageInput` (ChatInput): skip the `auto`/`scrollHeight` resize dance when the composer is empty; it fires right after a workspace switch while the new transcript is still dirty.
- `VimTextArea` / `AskUserQuestionToolCall` textareas: `rows={1}` so the natural (un-measured) empty height stays one line — textareas default to `rows=2`. With the consumers' existing `min-h-*` clamps, the rendered empty height is unchanged (verified live: empty composer renders at exactly its CSS `min-height`, same as before).

Non-empty drafts keep the existing measure-on-mount behavior.

## Validation

- Captured the profile against the running dev instance via CDP; per-switch blocking spans attributed 29–45% of their time to this exact reflow path.
- Verified via HMR on the same live instance: empty composer has no inline height, `rows=1`, and `offsetHeight` exactly equals its CSS `min-height` (pixel-identical to the previous measured state).
- Hook unit tests updated/extended: empty mount performs zero `scrollHeight` reads and zero style writes; clearing to empty resets the inline height; first keystroke after the unmeasured empty state grows without the `auto` reset.

## Risks

Low-moderate, scoped to composer sizing. The empty state is now CSS-governed; if a consumer styled the empty composer expecting an inline px height (none found — both consumers have `min-h` clamps that already dominated), its empty height could shift by a couple px. Typing, draft-restore, deletion/shrink, and max-height capping paths are unchanged and covered by existing tests.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `high` • Cost: `$2.11`_

### DIFF
--- a/src/browser/components/VimTextArea/VimTextArea.tsx
+++ b/src/browser/components/VimTextArea/VimTextArea.tsx
@@ -278,6 +278,9 @@ export const VimTextArea = React.forwardRef<HTMLTextAreaElement, VimTextAreaProp
               onChange(e.target.value, e.target.selectionStart ?? e.target.value.length)
             }
             onKeyDown={handleKeyDownInternal}
+            // Keep the empty natural height at one line (textareas default to rows=2) so
+            // useAutoResizeTextarea can leave empty drafts to CSS sizing without measuring.
+            rows={1}
             spellCheck={false}
             autoCorrect="off"
             autoCapitalize="none"
--- a/src/browser/features/ChatInput/index.tsx
+++ b/src/browser/features/ChatInput/index.tsx
@@ -1178,8 +1178,15 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
       const cursor = element.value.length;
       element.selectionStart = cursor;
       element.selectionEnd = cursor;
-      element.style.height = "auto";
-      element.style.height = Math.min(element.scrollHeight, window.innerHeight * 0.5) + "px";
+      // Skip the resize dance when empty: reading scrollHeight after a height write
+      // forces a synchronous reflow, and this rAF fires right after a workspace
+      // switch while the freshly mounted transcript is still dirty — laying out the
+      // whole document before first paint. Empty composers are sized by CSS alone
+      // (rows=1 + min-height; see useAutoResizeTextarea).
+      if (element.value !== "") {
+        element.style.height = "auto";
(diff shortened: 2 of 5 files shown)


ammar-agent:
@codex review

ammar-agent:
**Chromatic note for baseline acceptance** (`UI Tests`: 1 change)

I rebuilt the Storybook for `main` and this branch locally and DOM-diffed textarea geometry across all 110 affected stories:

- 109 stories: inline height `62px` → none, **rendered height identical** (CSS `min-height` clamp dominated before and after) — pixel-invisible.
- 1 real visual change: `app-phoneviewports--i-phone-16-e-sidebar-with-sections`. On `main`, the empty draft composer was stuck at **360px tall (exactly 50vh)** — a bogus pre-layout `scrollHeight` measurement persisted as inline height, pushing the model selector/MCP section below the fold. On this branch it renders at its intended **112px `min-h-28` clamp**.

The flagged change is a strict improvement (fixes a pre-existing empty-composer inflation bug), safe to accept as baseline.

PR #3541, opened 2026-06-12
🤖 perf: rotate sealed chat history out of chat.jsonl at compaction boundaries

ammar-agent:
## Summary

Make chat history size a non-issue for users: when a durable context boundary lands (compaction or `/clear --soft`), HistoryService now rotates everything before it out of `chat.jsonl` into an append-only `chat-archive.jsonl`. Hot-path reads and the per-turn stream-end rewrite become O(active epoch) instead of O(lifetime history), and the "chat.jsonl exceeds 5MB" warning is removed.

## Background

Users were seeing `chat.jsonl exceeds 5MB — full read may be slow, consider compaction` on long-lived workspaces. Compaction appends a boundary summary but never reclaims pre-boundary bytes, so:

- `updateHistory` (runs on **every stream end**), `deleteMessage`, and edit/fork truncation read + atomically rewrite the entire file — cost scales with lifetime history, not the active epoch.
- Uncompacted/boundaryless reads fall back to full-file reads on every request assembly.
- `getMaxHistorySequence` full-scans once per process per workspace.

We don't want users to think about chat size; this is Phase 1 of the fix (Phase 2 — offloading large inline blobs like base64 images to a content-addressed store — is follow-up work).

## Implementation

On-disk layout per session dir (invariant: full history = `chat-archive.jsonl` ++ `chat.jsonl`, archive sequences strictly older):

- `chat.jsonl` — active epoch only (latest durable boundary onward).
- `chat-archive.jsonl` — sealed pre-boundary history, append-only, oldest→newest.

Key pieces:

- **Rotation** (`rotateSealedHistoryUnlocked`): finds the latest durable boundary byte offset, appends the sealed prefix to the archive (fsync first), then atomically rewrites `chat.jsonl` with the tail. Crash-safe: a crash between the two steps duplicates rows, and the next rotation dedupes by skipping prefix rows whose `historySequence` is already covered by the archive.
- **Triggers**: eagerly inside `appendToHistory`/`updateHistory` when the persisted row is a durable boundary (covers compaction summaries written either way, and reset markers); lazily once per process on first `getHistoryFromLatestBoundary` read to migrate legacy files.
- **Archive-aware reads**: `getHistoryFromLatestBoundary` (including `skip>0` spanning the seam), `getHistoryBoundaryWindow` (Load More pages into the archive), `hasHistoryBeforeSequence`, `iterateFullHistory` (usage rebuilds/exports), `getLastMessages` (tail top-up across the seam), `hasHistory`, and sequence-counter init (archive tail consulted so counters never regress if `chat.jsonl` vanishes).
- **Structural ops**: `deleteMessage` falls back to rewriting the archive; `truncateAfterMessage`/percentage `truncateHistory` targeting sealed rows collapse the remainder back into `chat.jsonl` and drop the archive (rare recovery/edit paths, re-sealed by the next boundary); `clearHistory` removes both files; `migrateWorkspaceId` rewrites both.
- **workspaceService**: forks copy `chat-archive.jsonl` alongside `chat.jsonl`; archived sub-agent transcripts concatenate archive + active file into a single snapshot `chat.jsonl` so the transcript reader needs no changes.

Older Mux builds reading a rotated session dir still see a valid `chat.jsonl` (boundary + active epoch) — they only lose Load More access to pre-boundary epochs until this build re-runs.

## Validation

- 10 new tests covering: rotation on boundary append, lazy legacy migration on read, cross-seam `skip`/Load More paging, sequence-counter init from archive, crash-replay dedupe, cross-seam `getLastMessages`, archived-target truncation/delete, `clearHistory`, and archive-only `hasHistory`.
- `make static-check` green; `make test` shows the same 11 pre-existing flaky UI failures on the clean base commit (verified by running the full suite on a stashed tree); all 11 pass standalone with this diff.

## Risks

History persistence is core state management, so regression severity is high if rotation misbehaves (worst case: messages invisible in older epochs or sequence-counter regressions confusing the renderer). Mitigations: rotation is non-fatal/best-effort (reads stay correct on unrotated files), archive-before-rewrite ordering plus sequence-based dedupe makes it crash-safe, the full-history invariant is exercised by both the new tests and all 58 pre-existing HistoryService tests, and per-epoch read semantics for the renderer (`getHistoryFromLatestBoundary(skip=0)`) are unchanged.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `high` • Cost: `$6.21`_

### DIFF
--- a/src/common/constants/paths.ts
+++ b/src/common/constants/paths.ts
@@ -6,6 +6,20 @@ const LEGACY_MUX_DIR_NAME = ".cmux";
 const MUX_DIR_NAME = ".mux";
 
 /**
+ * Session-dir file holding the active chat history epoch (latest compaction
+ * boundary onward). Example: ~/.mux/sessions/<workspace>/chat.jsonl
+ */
+export const CHAT_FILE_NAME = "chat.jsonl";
+
+/**
+ * Session-dir file holding sealed pre-boundary chat history. HistoryService
+ * rotates everything before the latest durable context boundary out of
+ * chat.jsonl into this append-only archive so per-turn reads/rewrites stay
+ * O(active epoch) instead of O(lifetime history).
+ */
+export const CHAT_ARCHIVE_FILE_NAME = "chat-archive.jsonl";
+
+/**
  * Migrate from the legacy ~/.cmux directory into ~/.mux for rebranded installs.
  * Called on startup to preserve data created by earlier releases.
  *
--- a/src/node/orpc/router.ts
+++ b/src/node/orpc/router.ts
@@ -106,6 +106,7 @@ import {
   type SubagentTranscriptArtifactIndexEntry,
 } from "@/node/services/subagentTranscriptArtifacts";
 import { getErrorMessage } from "@/common/utils/errors";
+import { CHAT_FILE_NAME, CHAT_ARCHIVE_FILE_NAME } from "@/common/constants/paths";
 import { WorkflowActionRegistry } from "@/node/services/workflows/WorkflowActionRegistry";
 import {
   shouldDisableHostWorkflowActions,
@@ -4332,6 +4333,8 @@ export const router = (authToken?: string) => {
           const readTranscriptFromPaths = async (params: {
             workspaceId: string;
             chatPath?: string;
+            /** Sealed pre-boundary history (chat-archive.jsonl) for live sessions. */
+            chatArchivePath?: string;
             partialPath?: string;
             logLabel: string;
           }): Promise<MuxMessage[]> => {
@@ -4341,12 +4344,26 @@ export const router = (authToken?: string) => {
             if (params.chatPath && !isPathInsideDir(workspaceSessionDir, params.chatPath)) {
               throw new Error("Refusing to read transcript outside workspace session dir");
             }
+            if (
+              params.chatArchivePath &&
+              !isPathInsideDir(workspaceSessionDir, params.chatArchivePath)
+            ) {
+              throw new Error("Refusing to read transcript archive outside workspace session dir");
+            }
             if (params.partialPath && !isPathInsideDir(workspaceSessionDir, params.partialPath)) {
               throw new Error("Refusing to read partial outside workspace session dir");
             }
 
             const partial = params.partialPath
               ? await readPartialJsonBestEffort(params.partialPath)
+              : null;
+            // Live sessions may have rotated sealed history into chat-archive.jsonl
+            // (older rows), which precedes chat.jsonl (active epoch).
+            const archivedMessages = params.chatArchivePath
+              ? await readChatJsonlAllowMissing({
+                  chatPath: params.chatArchivePath,
+                  logLabel: `${params.logLabel} (archive)`,
+                })
               : null;
             const messages = params.chatPath
               ? await readChatJsonlAllowMissing({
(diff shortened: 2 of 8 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:

1. **Analytics ingestion archive-aware** — `ingestWorkspace`/`parseWorkspaceFromDisk` now read `chat-archive.jsonl` + `chat.jsonl` via shared `statSessionChatHistory`/`readSessionChatHistoryContents` helpers (watermark mtime is the max across both files). Covered by a new ETL test that asserts events from both files are ingested.

2. **Archive sequences preserved after active truncation** — `truncateAfterMessage`'s active-epoch branch now floors the sequence counter with the archive tail max (`getArchiveTailMaxSequence`), so emptying the active file can never reuse archived `historySequence` values. Covered by a new test truncating at the boundary row with `keepTargetMessage: false`.

@codex review

ammar-agent:
Fixed: `truncateHistory` now returns `Ok([])` before any rewrite when `removeCount` is 0, so a no-op percentage truncation can no longer collapse the archive back into `chat.jsonl`. Added a test asserting both files are byte-identical after `truncateHistory(0)`.

@codex review

ammar-agent:
Addressed both findings:

1. **deleteMessage counter floor** — the active-file delete branch now floors the cached counter with `getArchiveTailMaxSequence`, mirroring `truncateAfterMessage`, so deleting the last sequenced active row in a fresh process can never reuse archived sequences. Covered by a new restart-process test.

2. **Live subagent transcript fallback** — `readTranscriptFromPaths` in `router.ts` accepts an optional `chatArchivePath` (with the same path-traversal guard) and prepends sealed rows; the live session-dir fallback now passes the child's `chat-archive.jsonl`.

@codex review

ammar-agent:
Fixed: `migrateWorkspaceId`'s empty-active-file branch now floors the transferred counter with `getArchiveTailMaxSequence(newWorkspaceId) + 1`, matching the delete/truncate paths. Added a fresh-process rename test over an archive-only session asserting the next append continues past the archived sequences.

@codex review

ammar-agent:
Fixed: `statSessionChatHistory` now returns non-null when either file exists, and `readSessionChatHistoryContents` tolerates a missing active file — so archive-only sessions keep their analytics state and are ingested from the archive. Added an `ingestWorkspace` test for the archive-only layout.

@codex review

ammar-agent:
Fixed: the `ingestWorkspace` skip condition now only short-circuits when the combined mtime is *unchanged* (`===` instead of `<=`), so an mtime regression — chat.jsonl deleted leaving an older archive — re-runs ingestion and the existing rebuild path (max-sequence rewind) drops the removed epoch's stale rows. Added a regression test covering exactly that sequence.

@codex review

ammar-agent:
Fixed: the analytics watermark now stores a presence-aware `changeSignal` (sum of each history file's mtime + size, with explicit markers for missing files) instead of the max mtime. Deleting `chat.jsonl` changes the signal even when the surviving archive's mtime equals the stored value, so ingestion re-runs and the rebuild path drops the removed epoch's rows. `mtimeMs` is kept separately for the rebuild dedup-winner recency comparison. Added a same-tick deletion regression test.

@codex review

PR #3542, opened 2026-06-12
🤖 fix: stop counting UI memory reads as memory usage

ThomasK33:
## Summary

Opening a memory file from the Memory tab or Settings no longer counts as a "use" in the memory usage stats. Usage stats now reflect agent reads only, so human browsing can't inflate hot-set ranking.

## Background

The Memory sidecar (`memory-meta.json`) tracks `accessCount`/`lastAccessedAt` per memory file, and that signal drives hot-set preloading (which memories get injected into agent context). The UI read path (`memory.read` → `readFileWithSha`, used by the Memory tab and Settings → Memory) recorded a use on every open — so a human simply browsing their memories made those files look "hot" to the ranking, displacing genuinely agent-used memories.

## Implementation

- `MemoryService.readFileWithSha` no longer calls `recordUsage`; a comment documents the boundary (agent reads via the `view` memory command still count).
- UI **saves** still count as uses: an explicit edit is a deliberate "this file matters" signal, consistent with pinning counting as a use.
- Updated the two tests asserting the old behavior into behavioral assertions of the new rule (save + UI read ⇒ `accessCount === 1`), and fixed the stale "chokepoint" comment.

## Risks

Low. The only behavior change is that UI reads stop bumping usage stats; agent command paths, rename/delete metadata handling, and pinning are untouched. Worst case is slightly different hot-set ranking for users who relied on UI browsing to keep files hot — which is exactly the bias being removed.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$2.95`_

### DIFF
--- a/src/node/orpc/router.memory.test.ts
+++ b/src/node/orpc/router.memory.test.ts
@@ -180,7 +180,7 @@ describe("router memory routes", () => {
     expect(await memoryMetaService.getPinnedKeys()).toEqual(new Set(["global:prefs.md"]));
   });
 
-  test("list exposes usage stats accumulated through reads and writes", async () => {
+  test("list exposes usage stats; UI reads do not count as uses", async () => {
     const client = createClient({ enabled: true });
     await client.memory.save({
       workspaceId: "ws-mem",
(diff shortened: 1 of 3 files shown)


ThomasK33:
@codex review

PR #3544, opened 2026-06-12
🤖 fix: show Dream agent in Settings only when Memory Consolidation is enabled

ThomasK33:
## Summary

The Dream (memory consolidation) agent was missing from the Settings → Agents page in fallback mode, and conversely showed up as dead weight when the Memory Consolidation experiment was off. This PR fixes the fallback inventory and gates the Dream card behind the experiment, following the existing Desktop/Portable Desktop precedent.

## Background

A user enabled the memory dreams experiment but could not find the Dream agent on Settings → Agents. Root cause: `FALLBACK_AGENTS` (used when agent discovery is unavailable, e.g. Settings opened without a workspace selected) was never updated when the dream agent shipped, so the card vanished and any saved override would be mislabeled under "Unknown agents".

While dogfooding the fix, a second UX issue surfaced: the Dream card rendered even with the experiments off — an internal agent the user cannot select or run, which never executes unless both Agent Memory and Memory Consolidation are enabled (see the runtime gate in `memoryConsolidationService`).

## Implementation

- **Commit 1** adds `dream` to `FALLBACK_AGENTS`, mirroring its frontmatter in `src/node/builtinAgents/dream.md`, and replaces the spot-check test with an invariant test pinning `FALLBACK_AGENTS` to `getBuiltInAgentDefinitions()` so future built-ins cannot regress this. (The invariant lives node-side because browser/ cannot value-import from node/.)
- **Commit 2** turns `shouldShowAgentInTasksSettings` into a per-agent rule table: `desktop` → Portable Desktop, `dream` → Memory Consolidation (ANDed with Agent Memory, mirroring the runtime gate). Hidden agents stay in `knownAgentIds`, so saved overrides are not relabeled as unknown when the experiment is toggled off.

## Validation

Dogfooded both states in isolated dev-server sandboxes via agent-browser:

- Experiments off: Internal group shows Compact → Name Workspace, no Dream card.
- Agent Memory + Memory Consolidation on: Dream card appears under Internal with Enabled toggle and Model/Reasoning overrides, live without reload.
- Fallback path (Settings opened from Home, no workspace): Dream listed correctly.

## Risks

Low. Pure Settings-page visibility logic; no change to when consolidation actually runs. Worst case is a Settings card showing/hiding incorrectly; override persistence is covered by tests.
---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$12.77`_

### DIFF
--- a/src/browser/features/Settings/Sections/TasksSection.agents.ts
+++ b/src/browser/features/Settings/Sections/TasksSection.agents.ts
@@ -102,17 +102,24 @@ function compareAgentsByName(a: AgentDefinitionDescriptor, b: AgentDefinitionDes
   return a.name.localeCompare(b.name);
 }
 
+// Experiment-gated agents are dead weight in Settings while their experiment
+// is off (they never run), so hide their cards; their overrides stay "known"
+// via knownAgentIds below.
 function shouldShowAgentInTasksSettings(
   agent: AgentDefinitionDescriptor,
-  portableDesktopEnabled: boolean
+  params: { portableDesktopEnabled: boolean; memoryConsolidationEnabled: boolean }
 ): boolean {
-  return portableDesktopEnabled || agent.id !== "desktop";
+  if (agent.id === "desktop") return params.portableDesktopEnabled;
+  if (agent.id === "dream") return params.memoryConsolidationEnabled;
+  return true;
 }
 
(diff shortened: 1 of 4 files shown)


ThomasK33:
@codex review

PR #3545, opened 2026-06-12
🤖 feat: nest memory sub-experiments under the Agent Memory toggle

ThomasK33:
## Summary

Tucks the **Memory Hot Set** and **Memory Consolidation** experiment toggles into the grey inset settings panel under the **Agent Memory** row (the same pattern used by Advisor Tool and Workspace Heartbeats), shown only while Agent Memory is enabled.

## Background

Both sub-experiments are no-ops while Agent Memory is off — call sites gate on the parent flag — so listing them as top-level experiments invited toggling switches that do nothing. Nesting them makes the dependency obvious and shortens the flat list.

## Implementation

- `ExperimentsSection.tsx`: a `MEMORY_SUB_EXPERIMENT_IDS` group is filtered out of the flat experiments list and rendered via the existing `ExperimentRow` inside an `ExperimentSettingsPanel` under the Agent Memory row, gated on `useExperimentValue(MEMORY)`. Toggle/telemetry behavior is unchanged.
- `experiments.ts`: dropped the now-redundant "(requires Agent Memory)" description suffixes; the flags themselves stay flat and backend gating is untouched.
- Stories: new `MemorySettingsEnabled` story plus play assertions covering both gating branches. The default story seeds an explicit `false` override for Agent Memory because experiment localStorage keys leak between stories in the same browser session.

Sub-flag values are intentionally preserved (not cleared) when the parent is toggled off — same as Advisor Tool — so re-enabling memory restores prior choices.

## Validation

- Dogfooded live in Storybook at a ~420px viewport: toggling Agent Memory on reveals the nested panel with both sub-toggles; toggling off collapses it entirely.
- Storybook test-runner passes on the section including the two new play assertions.

## Risks

Low: Settings-only UI change; experiment flag semantics and backend gating are unchanged.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$8.13`_


ThomasK33:
@codex review

ThomasK33:
@codex review

Removed the accidental `--full-page` screenshot artifact (a stray `agent-browser screenshot` output swept in by `git add -A`) in ac02b6467.

PR #3548, opened 2026-06-13
🤖 fix: harden hot-memory token budgeting

ThomasK33:
## Summary

Hardens memory hot-set preloading by moving selection from byte-only caps to model-aware token budgeting with item and selection-attempt limits. The memory index remains available even when hot-set tokenization fails, and fallback model prompts now resolve hot memories with the fallback model's tokenizer/context only when the hot-set sub-experiment is enabled.

## Background

Hot memories are injected directly into the system prompt to avoid extra tool calls for pinned or frequently used memory files. Byte-only budgeting was an imprecise proxy for actual context usage, and eager hot-set resolution could spend tokenization work before final tool policy and experiment gates determined whether preloaded memory content would actually be sent.

## Implementation

- Adds token, item, and candidate-attempt limits for hot-memory selection while keeping byte caps as an I/O safety guard.
- Counts the exact rendered `<hot_memories>` block, including wrapper guidance, before accepting candidates.
- Caches memory session context per model string and lazily upgrades index-only context to include hot memories only after final tool policy keeps the memory tool and `memory-hot-set` is enabled.
- Rebuilds fallback prompt context with the fallback model's memory context instead of reusing the source model's cached context when hot preloading is active.
- Preserves the memory index when hot-memory tokenization or selection fails, degrading only the preloaded hot block.

## Validation

- `bun test src/node/services/aiService.test.ts --test-name-pattern "does not upgrade memory context when the hot-set sub-experiment is disabled|rebuilds the stream system context without memory availability|prepares fallback system context"`
- `make static-check`
- Targeted regression coverage for per-model caching, index-to-hot upgrade, fallback model memory context, tokenizer failure fallback, exact block token budgeting, bounded selection attempts, and avoiding hot-set upgrades when the sub-experiment is disabled.

## Risks

Medium risk because this touches prompt construction, model fallback preparation, and memory context caching. The changes are scoped to the memory experiment/hot-set path and include fallback behavior so memory access degrades to the index instead of blocking streams.

## Pains

Deep review and Codex review found subtle interactions between token-budgeted memory blocks, fallback model compilation, final tool policy, and the hot-set sub-experiment gate. The implementation now separates cheap index resolution from expensive hot-set selection to keep stream startup predictable.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$59.61`_

### DIFF
--- a/src/common/constants/memory.ts
+++ b/src/common/constants/memory.ts
@@ -64,10 +64,12 @@ export const MEMORY_VIEW_MAX_DEPTH = 2;
 export const MEMORY_HOT_SET_MAX_ITEM_BYTES = 16 * 1024;
 /** Maximum total bytes of selected hot-memory file content. */
 export const MEMORY_HOT_SET_MAX_TOTAL_BYTES = 48 * 1024;
-/** Maximum tokens consumed by rendered hot-memory file blocks. */
+/** Maximum tokens consumed by the rendered <hot_memories> block. */
 export const MEMORY_HOT_SET_MAX_TOTAL_TOKENS = 12_000;
 /** Maximum number of memory files preloaded into the system prompt. */
 export const MEMORY_HOT_SET_MAX_ITEMS = 8;
+/** Maximum ranked candidates to read/tokenize while filling the hot set. */
+export const MEMORY_HOT_SET_MAX_SELECTION_ATTEMPTS = 64;
 /** Half-life of the recency decay applied to access counts when ranking auto-hot files. */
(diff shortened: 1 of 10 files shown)


ThomasK33:
@codex review

Please review the memory hot-set token-budgeting hardening changes.

ThomasK33:
@codex review

Addressed PRRT_kwDOPxxmWM6JUci5 by gating post-policy and fallback hot-memory upgrades on the `memory-hot-set` experiment flag, so memory-only users keep the index-only context without duplicate scans/rebuilds. Added a regression test for the disabled-hot-set path and reran targeted AIService tests plus `make static-check`.

PR #3549, opened 2026-06-13
🤖 fix: clean up completed workflow task workspaces

ThomasK33:
Summary

Allow completed workflow-owned task workspaces to be pruned after their reports/structured outputs have been persisted, instead of retaining those transient worktrees for the full workflow run or behind the user-level preserve-subagents-until-archive setting.

Background

Workflow tasks already persist their report payloads into ancestor report artifacts and the workflow run journal. Keeping the backing task workspaces indefinitely after completion leaves completed workflow lanes in the sidebar even when their structured output is already available from the workflow card.

Implementation

- Removed the in-memory active workflow-run cleanup hold and the corresponding run-start lifecycle callback.
- Kept the run-end lifecycle callback as a best-effort cleanup recheck for workflow children that may have been deferred by other gates.
- Added workflow-owned lineage detection so completed workflow task trees bypass preserve-subagents-until-archive while normal subagents keep the existing behavior.
- Updated workflow lifecycle tests and added regression coverage for workflow-owned cleanup with preserve-subagents-until-archive enabled.

Validation

- `bun test src/node/services/taskService.test.ts -t "preserve subagents until archive"`
- `bun test src/node/services/workflows/WorkflowRunner.test.ts src/node/services/workflows/WorkflowTaskServiceAdapter.test.ts`
- `make typecheck`
- `make lint`
- `make fmt-check`
- `git diff --check`
- `make static-check`

Risks

Medium-low. This touches task cleanup behavior, but the change is scoped to completed workflow-owned task trees and preserves existing cleanup gates for active descendants, pending patch artifacts, and non-workflow subagent preservation.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$16.15`_

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -11160,6 +11160,7 @@ describe("TaskService", () => {
       agentType: string;
       taskStatus?: "reported" | "interrupted";
       reportedAt?: string;
+      workflowTask?: WorkspaceConfigEntry["workflowTask"];
     }
 
     type TaskCleanupEligibility =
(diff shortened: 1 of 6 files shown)


ThomasK33:
@codex review

PR #3553, opened 2026-06-15
🤖 fix: prevent duplicate consolidation tooltip

ThomasK33:
## Summary

Fixes the memory consolidation footer so its summary hover uses the shared portaled tooltip instead of a native `title` tooltip. This leaves only one tooltip visible and lets the app tooltip stack above clipped sidebar containers.

## Background

The consolidation summary hover was using a native `title` attribute, which produced both the app-styled clipped tooltip and the browser/OS native tooltip. Native tooltips cannot participate in the app z-index stack.

## Implementation

- Replaced the footer `title` attribute with `TooltipIfPresent` around the consolidation status lines.
- Preserved multi-line summary formatting with `whitespace-pre-line`.
- Added a regression test that asserts the consolidation status block does not expose a native `title` attribute.
- Recorded the tooltip guidance in `AGENTS.md` and regenerated built-in skill content.

## Validation

- `bun test src/browser/features/RightSidebar/Memory/MemoryTab.test.tsx`
- `make typecheck`
- `make static-check`
- Dogfooded in Storybook at a narrow `390x640` viewport with `agent-browser`; hovered the consolidation footer and verified the DOM had no native title on the consolidation block, one tooltip surface, and a portaled tooltip wrapper at `z-index: 9999`.

## Risks

Low. This is isolated to the memory consolidation footer hover surface and uses the existing shared tooltip primitive already used across the app.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$6.15`_


ThomasK33:
@codex review

PR #3554, opened 2026-06-15
🤖 tests: stabilize deep-research fanout test

ThomasK33:
## Summary

Stabilizes the deep-research capped fanout unit test by letting that heavy fanout fixture use the production WorkflowRunStore lease timing instead of the 100ms stale-lease setting used by retry-focused fixtures.

## Background

The known flaky test `built-in deep-research workflow › caps model-produced deep-research fan-out` failed in merge-queue Unit runs with `Timed out acquiring workflow mutation lock: .../events.jsonl.lock`. I reproduced the same failure locally by pinning the targeted test to one CPU while running a same-core CPU hog; the short 100ms fixture lease renews every 50ms, which adds lock churn while the test is intentionally persisting many parallel workflow steps.

## Implementation

The capped fanout test now constructs `WorkflowRunStore` with its default production lease settings, because the test is asserting source/claim fanout caps rather than stale-lease retry behavior. Its per-test timeout is raised to 30s to leave room for CPU-contention repro runs without changing normal runtime.

## Validation

- Reproduced the pre-fix failure with `taskset -c 0` plus one same-core CPU hog: `Timed out acquiring workflow mutation lock: .../events.jsonl.lock`.
- After the fix, the same targeted test passed under two same-core CPU hogs (~11.5s) and three same-core CPU hogs (~18.0s).
- `bun test src/node/services/workflows/builtInWorkflowDefinitions.test.ts -t "caps model-produced deep-research fan-out"`
- `bun test src/node/services/workflows/builtInWorkflowDefinitions.test.ts`
- `make static-check`

## Risks

Low product risk: this is test-only and avoids applying retry-fixture lease timing to a high-concurrency fanout assertion. The broader workflow runtime code is unchanged.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$11.63`_

### DIFF
--- a/src/node/services/workflows/builtInWorkflowDefinitions.test.ts
+++ b/src/node/services/workflows/builtInWorkflowDefinitions.test.ts
@@ -11,8 +11,7 @@ import { WorkflowActionRegistry } from "./WorkflowActionRegistry";
 import { WorkflowRunStore } from "./WorkflowRunStore";
 import { WorkflowRunner, type WorkflowAgentResult, type WorkflowAgentSpec } from "./WorkflowRunner";
 
-// Keep this above tiny lock timeouts: these fixtures exercise parallel workflow callbacks
-// that legitimately contend for the run-store mutation lock under CI coverage load.
+// Most fixtures use short leases so stale-run retry behavior stays fast.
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

PR #3556, opened 2026-06-15
🤖 tests: stabilize foreground bash migration test

ThomasK33:
## Summary

Stabilizes the foreground-to-background bash migration tests by waiting for actual foreground bash output before clicking the simulated background action, instead of relying on a fixed sleep that is too optimistic on Windows CI.

## Background

The known Windows flake in `tests/ipc/runtime/backgroundBashDirect.test.ts` intermittently returned `Output so far (0 lines)` from `should migrate foreground bash to background and continue running`. A deterministic delayed-output repro showed the same assertion failure when the test backgrounds the process after registration but before the first stdout chunk has reached the foreground output collector.

## Implementation

- Added a small foreground bash-output collector using the existing UI-only `bash-output` event path.
- Updated the foreground migration tests to wait for the first marker text before calling `sendToBackground`.
- Applied the same readiness check to the abort-after-background regression test so it does not depend on Git Bash startup timing either.

## Validation

- Reproduced the missing-marker failure with a temporary delayed-output foreground migration test.
- `bun test ./tests/ipc/runtime/backgroundBashDirect.test.ts`
- `make static-check` (first combined run saw ESLint killed with exit 137; immediate retry passed)

## Risks

Low. This is test-only and replaces fixed sleeps with an observable readiness signal from the same bash output stream the UI uses.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$20.62`_

### DIFF
--- a/tests/ipc/runtime/backgroundBashDirect.test.ts
+++ b/tests/ipc/runtime/backgroundBashDirect.test.ts
@@ -41,7 +41,46 @@ function getBackgroundProcessManager(env: TestEnvironment): BackgroundProcessMan
 }
 
 // Foreground bash startup can be slower on Windows CI (Git Bash init + IO flush).
-const FOREGROUND_MIGRATION_READY_MS = process.platform === "win32" ? 900 : 300;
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

PR #3557, opened 2026-06-15
🤖 fix: remove workflow sidebar progress fraction

ThomasK33:
## Summary

Remove the compact completed/total fraction from workflow task-group rows in the left sidebar while preserving that fraction for best-of and variants task groups.

## Background

Workflow rows already summarize progress in the secondary status line (for example, `1 running` or `1 completed`), so the extra right-aligned `0/1` indicator adds visual noise and truncates the workflow title sooner.

## Implementation

- Gate the compact completed/total fraction behind non-workflow task groups in `TaskGroupListItem`.
- Let workflow headers use a single title column so the workflow name gets the reclaimed space.
- Add regression coverage that workflow groups omit `0/1` while best-of groups still render their fraction.

## Validation

- `bun test src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx`
- `make static-check`
- Storybook dogfood of `Components/ProjectSidebar / Workflow Run Groups` at desktop and narrow/mobile widths; verified workflow row text no longer contains `0/1` while the variants row still shows `0/2`.

## Risks

Low. This is a sidebar presentation-only change scoped to task-group headers. Best-of and variants rows keep their existing progress fraction, and workflow status text remains visible below the header.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$5.75`_

### DIFF
--- a/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
+++ b/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
@@ -942,6 +942,7 @@ describe("ProjectSidebar multi-project completed-subagent toggles", () => {
     expect(view.getByTestId(agentItemTestId("parent"))).toBeTruthy();
     const groupRow = view.getByTestId("task-group-best-of-demo");
     expect(groupRow.textContent).toContain("Best of 3");
+    expect(groupRow.textContent).toContain("0/3");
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3564, opened 2026-06-15
🤖 fix: dedupe memory consolidation summaries

ThomasK33:
## Summary

Deduplicates identical memory-consolidation summaries in the Memory tab tooltip while preserving the separate Workspace, Project, and Global status lines. Also fixes the latest-main compaction epoch reader regression that surfaced when the PR entered the merge queue.

## Background

A single consolidation pass can cover workspace, project, and global memory scopes. When those status records share the same summary, the footer tooltip repeated the same text once per scope, which made the UI look broken.

After rebasing onto the latest `main`, merge-queue Unit runs exposed that `HistoryService.getMessagesForCompactionEpoch` was passing a workspace id into a file-path iterator. That made compaction harvest tests fail on main/merge-group runs because the summary row was never found.

## Implementation

- Added a small helper that trims and deduplicates consolidation summary lines before building the tooltip body.
- Kept the visible per-scope status metadata unchanged so users can still see which scopes were covered.
- Updated the compaction epoch reader to iterate full workspace history, including sealed archives and active chat history.
- Added a MemoryTab regression test for identical workspace/project/global summaries.

## Validation

- `bun test src/browser/features/RightSidebar/Memory/MemoryTab.test.tsx`
- `bun test src/node/services/historyService.test.ts --test-name-pattern "getMessagesForCompactionEpoch"`
- `bun test src/node/services/memoryConsolidationService.test.ts --test-name-pattern "harvest|post-harvest|re-harvest|retryable"`
- `make typecheck`
- `make fmt-check`

## Risks

Low to medium. The tooltip change only affects memory consolidation status rendering. The HistoryService fix restores the intended full-history iteration path used by compaction harvesting; regression coverage already existed and now passes again.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$3.54`_

### DIFF
--- a/src/browser/features/Memory/MemoryBrowser.tsx
+++ b/src/browser/features/Memory/MemoryBrowser.tsx
@@ -485,6 +485,20 @@ function MemoryFileRow(props: MemoryFileRowProps) {
   );
 }
 
+function uniqueConsolidationSummaryLines(
+  summaries: ReadonlyArray<string | null | undefined>
+): string[] {
+  const seen = new Set<string>();
+  const lines: string[] = [];
+  for (const summary of summaries) {
+    const line = summary?.trim();
+    if (!line || seen.has(line)) continue;
+    seen.add(line);
+    lines.push(line);
+  }
+  return lines;
+}
+
 function formatConsolidationRecord(record: MemoryConsolidationRecordPayload | null): string {
   if (record === null) return "never";
   const appliedCount = record.ops.filter((op) => op.applied).length;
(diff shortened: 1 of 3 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Pushed a CI-only test-selector fix after the Unit job exposed that full-suite Radix tooltip rendering can be inline rather than wrapped by the popper content attribute.

ThomasK33:
@codex review

Rebased onto latest main and added a targeted HistoryService fix for the merge-queue Unit failures: compaction epoch reads now iterate full workspace history instead of passing a workspace id to the file-path iterator.

PR #3567, opened 2026-06-15
🤖 fix: stop idle auto-compaction loop on repeated/model-not-found failures

ammar-agent:
## Summary

Stops the hourly idle auto-compaction loop from re-attempting a workspace that keeps failing: it now gives up immediately when an attempt fails with a **model-not-found** error, and otherwise after **two consecutive failures** of any kind. Previously a persistently failing workspace (e.g. a misconfigured/typo'd compaction model) was re-queued every hour forever, since a failed compaction neither marks the workspace `compacted` nor refreshes recency, so it stayed eligible indefinitely.

## Background

`IdleCompactionService` re-checks eligibility every hour and re-queues any idle workspace. There was no failure tracking, so a workspace whose compaction always fails would be retried on every cycle with no cap. The most common cause is a compaction model the provider rejects.

A key subtlety: the failure can surface in two different places.

- **Pre-stream** failures (invalid model string, model unavailable, provider/runtime issues) come back synchronously as the `sendMessage` `Result` that `executeIdleCompaction` already inspects.
- **Mid-stream** failures (the common `model_not_found` case for a well-formed but non-existent Anthropic/OpenAI model) are **not** returned by `sendMessage` — they arrive later as an async `"error"` stream event and are only persisted transiently to `partial.json`, so they can't be recovered from chat history on the next check.

## Implementation

- `IdleCompactionService` now tracks per-workspace consecutive failures and a sticky `suppressedWorkspaceIds` set. `checkEligibility` short-circuits suppressed workspaces (`reason: "suppressed_after_failures"`). A new `recordOutcome(workspaceId, outcome)` applies the policy: success resets the streak; `model_not_found` suppresses immediately; any other failure suppresses on the second consecutive occurrence.
- `WorkspaceService` reports terminal idle-compaction outcomes through a listener:
  - pre-stream failure → reported from `executeIdleCompaction` (model-not-found classified via `formatSendMessageError`),
  - mid-stream failure → reported from the existing `aiService.on("error")` handler when the workspace is mid idle-compaction (read before `stopStreamingStatus` clears the marker),
  - success → reported from `handleStreamCompletion`.
  - `error` / `stream-end` / `stream-abort` are mutually exclusive per turn, so a failed turn never also reports success. Aborts are intentionally neutral (not a failure).
- `ServiceContainer` wires the listener to `IdleCompactionService.recordOutcome`.

Suppression is in-memory and sticky for the service lifetime; restarting the app (e.g. after fixing the configured compaction model) clears it.

## Validation

- New unit tests in `idleCompactionService.test.ts`: suppress after two consecutive failures, suppress immediately on `model_not_found`, success resets the streak, and `checkAllWorkspaces` no longer queues a suppressed workspace.
- New `workspaceService.test.ts` tests: `executeIdleCompaction` reports `modelNotFound: true` for an invalid compaction model and `false` for a generic pre-stream failure.
- `make static-check` and targeted tests pass.

## Risks

Low. Behavior is additive and scoped to the background idle-compaction loop. The outcome listener is optional (no-op until wired), and normal interactive sends are untouched. Worst case of a wrong suppression is that a workspace stops auto-compacting on idle until restart, which a user can already trigger manually.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$1.48`_

### DIFF
--- a/src/node/services/agentSession.ts
+++ b/src/node/services/agentSession.ts
@@ -317,6 +317,8 @@ interface AgentSessionOptions {
   keepBackgroundProcesses?: boolean;
   /** Called when compaction completes (e.g., to clear idle compaction pending state) */
   onCompactionComplete?: (metadata: CompactionCompletionMetadata) => void;
+  /** Called with the terminal outcome of an idle compaction (persisted success / post-stream failure) */
+  onIdleCompactionOutcome?: (success: boolean) => void;
   /** Called when post-compaction context state may have changed (plan/file edits) */
   onPostCompactionStateChange?: () => void;
 }
@@ -515,6 +517,7 @@ export class AgentSession {
       workspaceGoalService,
       keepBackgroundProcesses,
       onCompactionComplete,
+      onIdleCompactionOutcome,
       onPostCompactionStateChange,
     } = options;
 
@@ -539,6 +542,7 @@ export class AgentSession {
       telemetryService,
       emitter: this.emitter,
       onCompactionComplete,
+      onIdleCompactionOutcome,
     });
 
     this.compactionMonitor = new CompactionMonitor(
--- a/src/node/services/compactionHandler.test.ts
+++ b/src/node/services/compactionHandler.test.ts
@@ -236,6 +236,74 @@ describe("CompactionHandler", () => {
       expect(metadata?.previousBoundaryHistorySequence).toBe(1);
     });
 
+    describe("onIdleCompactionOutcome", () => {
+      const createIdleCompactionRequest = (id = "idle-req"): MuxMessage =>
+        createMuxMessage(id, "user", "Please summarize the conversation", {
+          muxMetadata: {
+            type: "compaction-request",
+            rawCommand: "/compact",
+            parsed: {},
+            source: "idle-compaction",
+          },
+        });
+
+      it("reports success only after the summary is persisted", async () => {
+        const onIdleCompactionOutcome = mock((_success: boolean) => undefined);
+        handler = new CompactionHandler({
+          workspaceId,
+          historyService,
+          sessionDir,
+          telemetryService,
+          emitter: mockEmitter,
+          onIdleCompactionOutcome,
+        });
+        await seedHistory(createIdleCompactionRequest());
+
+        const handled = await handler.handleCompletion(createStreamEndEvent("Summary"));
+
+        expect(handled).toBe(true);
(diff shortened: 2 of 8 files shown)


ammar-agent:
@codex review

ammar-agent:
Addressed in 083e71d46.

The idle-compaction outcome is no longer reported from the provider `stream-end`. Instead, `CompactionHandler.handleCompletion()` now reports the terminal outcome via a new `onIdleCompactionOutcome(success)` callback:

- **success** is reported only after the compacted summary is actually persisted (right after `onCompactionComplete`),
- **failure** is reported on the post-stream persistence paths (`performCompaction` returns `Err`, empty summary, raw-JSON summary).

`WorkspaceService.createSession` forwards that to `reportIdleCompactionOutcome` (a persistence failure → `{ success: false, modelNotFound: false }`), so these post-stream failures now count toward the suppression threshold. The unconditional success report in `handleStreamCompletion` was removed. Mid-stream provider errors (no `stream-end`/no persistence) are still reported from the `"error"` handler, and the three terminal events are mutually exclusive per turn, so each idle attempt reports exactly one outcome.

Added `CompactionHandler` tests: success → `[true]`, empty summary → `[false]`, manual (non-idle) compaction → not called.

ammar-agent:
@codex review

Please take another look — addressed the post-stream persistence concern.

ammar-agent:
Addressed in c1a1b76bc.

The `requireIdle` busy-skip (`"Workspace is busy; idle-only send was skipped."`) is now treated as neutral — `executeIdleCompaction` no longer calls `recordOutcome` for it, so normal user-interaction races can't accumulate toward suppression. I factored the message into a shared `IDLE_ONLY_BUSY_SKIP_MESSAGE` constant (used by the `requireIdle` return, the early busy throw, and the skip check) to keep the contract DRY. Extended the busy-skip test to assert no outcome is reported.

ammar-agent:
@codex review

Please take another look — the requireIdle busy-skip is now treated as neutral.

ammar-agent:
Addressed in 866a615c9.

`recordOutcome` now also clears `suppressedWorkspaceIds` on success, so a later in-flight retry that actually persists a compaction lifts the suppression and the workspace becomes eligible again (self-healing). Added a test: two failures suppress → a subsequent success makes the workspace eligible again.

ammar-agent:
@codex review

Please take another look — suppression is now lifted on a later success.

PR #3573, opened 2026-06-16
🤖 tests: stabilize WorkflowService background interrupts

ThomasK33:
## Summary

Stabilizes WorkflowService unit coverage by removing a race in the foreground run-created callback test and treating user-interrupted background workflow exits as expected cancellations rather than background failures.

## Background

Recent unit CI runs showed `WorkflowService > notifies foreground run creation before the first workflow step blocks` intermittently receiving both `run-created` and `agent-started` before the polling assertion ran. The service already awaited `onRunCreated`; the test was racing the runner after a synchronous callback returned.

The same CI window also showed noisy expected background-interrupt stack traces immediately before workflow tests continued. User interruptions should not be logged as background failures, but terminal callbacks that explicitly request interrupted notifications should still be delivered.

## Implementation

- Gates the `onRunCreated` callback in the flaky test so the assertion checks the intended service ordering deterministically.
- Tracks in-progress interrupt status writes so background run cleanup waits for the durable `interrupted` status before deciding whether to log a failure.
- Adds coverage that interrupted background runs can still notify terminal observers without emitting `console.error` noise.

## Validation

- `bun test --max-concurrency=1 --coverage --coverage-reporter=lcov src/node/services/workflows/WorkflowService.test.ts`
- Stress-ran the foreground run-created test 100 times locally before final validation.

## Risks

Low. The production change is scoped to WorkflowService background failure reporting after an explicit interrupt. Non-interrupted background failures still log and notify as before.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$32.88`_

### DIFF
--- a/src/node/services/workflows/WorkflowService.test.ts
+++ b/src/node/services/workflows/WorkflowService.test.ts
@@ -620,6 +620,7 @@ export default function workflow({ action }) {
     );
 
     let releaseAgent: ((value: { taskId: string; reportMarkdown: string }) => void) | undefined;
+    const runCreatedGate = Promise.withResolvers<void>();
     const lifecycleEvents: string[] = [];
     const runStore = new WorkflowRunStore({ sessionDir: tmp.path });
     const service = new WorkflowService({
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3574, opened 2026-06-16
🤖 tests: isolate renderer test globals

ThomasK33:
Summary
- Stabilizes local renderer unit-test runs by keeping test-only module mocks and DOM globals scoped to the suites that need them.

Background
- `bun test src` runs many renderer test files in one process locally, so file-scope `mock.module()` registrations and direct `globalThis.window/document` overwrites can leak into later suites.
- The repro surfaced as `useChatTranscriptFullWidth` not calling its real API context after `useModelsFromSettings`, `events.test` seeing `document` as undefined, `chatCommands` inheriting a remote `window.location`, and `ImmersiveReviewView` rendering a stubbed diff renderer from `ProposePlanToolCall`.

Implementation
- Moves `useModelsFromSettings` and `ProposePlanToolCall` module mocks behind setup/teardown helpers that restore the real module exports after each test.
- Uses `installDom()` or explicit previous-global restoration so renderer tests leave the global DOM state as they found it.
- Makes `events.test` install its own DOM per test instead of relying on a one-time side-effect import.

Validation
- `bun test src/browser/hooks/useModelsFromSettings.test.ts src/browser/hooks/useChatTranscriptFullWidth.test.tsx src/browser/utils/events.test.ts`
- `bun test src/browser/hooks/useRouting.test.ts src/browser/utils/chatCommands.test.ts`
- `bun test src/browser/features/Tools/ProposePlanToolCall.test.tsx src/browser/features/RightSidebar/CodeReview/ImmersiveReviewView.test.tsx`
- `bun test src/browser/hooks src/browser/utils`
- `make static-check`

Risks
- Low product risk: this is test-only isolation. The main risk is that moving mocks into hooks could reveal tests that were unintentionally depending on cross-file pollution; the targeted cross-suite commands cover the known leak chains.

Pains
- `make test` now gets past the renderer failures but still exits on the pre-existing local Storybook budget guard (`71 > 70` snapshot-enabled story files, `276 > 262` estimated snapshots), which is unrelated to this diff and not part of the CI unit shard.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$46.22`_

### DIFF
--- a/src/browser/features/Tools/ProposePlanToolCall.test.tsx
+++ b/src/browser/features/Tools/ProposePlanToolCall.test.tsx
@@ -1,8 +1,15 @@
 import type { ComponentProps } from "react";
-import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import { afterAll, afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 import { installDom } from "../../../../tests/ui/dom";
 import { cleanup, fireEvent, render, waitFor } from "@testing-library/react";
 
+import * as APIModule from "@/browser/contexts/API";
+import * as WorkspaceContextModule from "@/browser/contexts/WorkspaceContext";
+import * as UseOpenInEditorModule from "@/browser/hooks/useOpenInEditor";
+import * as UseReviewsModule from "@/browser/hooks/useReviews";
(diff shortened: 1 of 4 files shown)


ThomasK33:
@codex review

PR #3577, opened 2026-06-17
🤖 fix: detect MCP OAuth on POST-only servers

ethanndickson:
## Summary
Restore OAuth-required detection for MCP servers that only advertise `WWW-Authenticate` on POST responses, so Settings can offer the OAuth login callout instead of only surfacing the raw 401 error.

## Background
`@ai-sdk/mcp` collapses failed HTTP MCP requests into `MCPClientError` without preserving response headers. Mux previously fell back to a GET probe when that happened, but Streamable HTTP servers like Carta commonly reject GET on the MCP endpoint and only emit the Bearer challenge on POST.

## Implementation
- capture `WWW-Authenticate` headers from the actual MCP test transport fetch path before the SDK erases them
- make the shared OAuth hint probe transport-aware so HTTP/auto prefer POST first while SSE still prefers GET
- add regression coverage for POST-only auth challenges in both the manager test path and the standalone probe helper

## Validation
Focused MCP OAuth tests now cover the Carta-style POST-only challenge and the HTTP/auto probe ordering, alongside targeted ESLint and full typecheck coverage.

## Risks
Low. The behavior change is limited to OAuth challenge discovery during MCP connection tests and login hint probing, and still falls back across both remote verbs for mixed deployments.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh` • Cost: `$2.64`_

### DIFF
--- a/src/node/services/mcpOauthService.test.ts
+++ b/src/node/services/mcpOauthService.test.ts
@@ -5,7 +5,11 @@ import * as os from "os";
 import * as path from "path";
 import { Config } from "@/node/config";
 import { MCPConfigService } from "./mcpConfigService";
-import { McpOauthService, parseBearerWwwAuthenticate } from "./mcpOauthService";
+import {
+  McpOauthService,
+  parseBearerWwwAuthenticate,
+  probeServerForBearerChallenge,
+} from "./mcpOauthService";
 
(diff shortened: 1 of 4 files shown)


ethanndickson:
@codex review

PR #3582, opened 2026-06-17
🤖 fix: default slash workflow project path args

ThomasK33:
## Summary

Fix slash-command workflow starts so workflows that declare a `projectPath` argument receive the active workspace checkout path (including sub-project checkout paths) from the server-side workflow context instead of failing before a run is created.

## Background

Running a workflow from a slash command could hit `/workflows/start` with only the typed workflow args. Workflows whose `metadata.argsSchema` required `projectPath` then failed during argument normalization with `Workflow argument projectPath is required`, which surfaced as a 500.

## Implementation

- Added invocation-context defaults to workflow argument normalization, applied only to schema-declared missing fields.
- Passed the resolved active workflow execution checkout path as a slash-command-only `projectPath` default, including sub-project workspaces, without persisting hidden fields into command invocation metadata.
- Added `WorkflowArgsValidationError` so workflow argument/schema validation failures map to oRPC `BAD_REQUEST` instead of generic internal errors.
- Ran the `simplify` workflow and incorporated its small follow-up simplifications and malformed-schema validation coverage.

## Validation

- `bun test src/node/services/workflows/WorkflowService.test.ts --test-name-pattern "applies invocation default args|workflow args"`
- `bun test src/node/orpc/router.test.ts --test-name-pattern "router workflow routes"`
- `bun test src/node/orpc/router.test.ts --test-name-pattern "projectPath defaults|active sub-project|bad request"`
- `make static-check`
- `workflow_run simplify --base origin/main`

## Risks

Low. The default context only applies to workflow args declared in a workflow schema and only when the caller omitted that field. Explicit user args and schema defaults keep their existing precedence, persisted slash-command args remain unchanged, and sub-project workspaces now receive the same active checkout path that workflow execution should operate on.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$19.51`_

### DIFF
--- a/src/node/orpc/router.test.ts
+++ b/src/node/orpc/router.test.ts
@@ -801,11 +801,13 @@ export default function workflow({ args }) { return { reportMarkdown: args.proje
     );
   });
 
-  test("fills slash projectPath defaults from the active sub-project", async () => {
+  test("fills slash projectPath defaults from the active sub-project checkout", async () => {
+    const workspacePath = path.join(tempDir, "workspace-checkout");
     const subProjectPath = path.join(projectPath, "packages", "api");
-    fs.mkdirSync(subProjectPath, { recursive: true });
+    const workspaceSubProjectPath = path.join(workspacePath, "packages", "api");
+    fs.mkdirSync(path.join(workspacePath, ".mux", "workflows"), { recursive: true });
     fs.writeFileSync(
-      path.join(projectPath, ".mux", "workflows", "needs-active-project.js"),
+      path.join(workspacePath, ".mux", "workflows", "needs-active-project.js"),
       `const s = mux.schema;
 export const metadata = {
   description: "Needs active project",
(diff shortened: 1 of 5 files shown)


ThomasK33:
@codex review

ThomasK33:
Addressed the Codex finding about sub-project workspaces.

- The slash workflow default now uses `workflowExecutionProjectPath`, which falls back to the active `metadata.subProjectPath` when no explicit workflow project path is supplied.
- Added regression coverage for slash workflow `projectPath` defaults in an active sub-project workspace.
- Re-ran `make static-check` and the targeted router sub-project/bad-request tests.

ThomasK33:
@codex review

Please take another look after the sub-project path default fix.

ThomasK33:
Addressed the follow-up Codex finding about worktree/SSH checkout paths.

- The slash workflow default now derives `projectPath` from the resolved workspace checkout root and appends the active sub-project relative path when applicable.
- Updated the sub-project regression test to use a separate workspace checkout and assert the checkout sub-project path, not the original source project identity.
- Re-ran `make static-check` and the targeted router sub-project/bad-request tests.

ThomasK33:
@codex review

Please take another look after the checkout-path default fix.

PR #3585, opened 2026-06-17
🤖 fix: highlight scheduled workflow trigger names

ThomasK33:
## Summary

Render scheduled workflow trigger messages so the `Automation:` label stays in the normal user-message color while the workflow name receives the blue workflow badge styling.

## Background

Scheduled workflow invocations display as synthetic user messages like `Automation: github-issue-triage`. The UI previously highlighted `Automation:` because it was treated as the command prefix, but the visually important workflow identity is the workflow name.

## Implementation

- Added a shared scheduled-workflow trigger label constant used by backend message construction and frontend rendering.
- Special-cased scheduled workflow trigger rendering to keep `Automation:` unbadged and badge/color the workflow name instead.
- Kept existing workflow definition popover behavior attached to the workflow-name badge.
- Added a Storybook dogfood scenario and a regression component test for the scheduled-trigger rendering.

## Validation

- `bun test src/browser/features/Messages/UserMessageContent.inline-skill.test.tsx`
- `make typecheck`
- `make lint`
- `make fmt-check`
- `make static-check`
- Storybook dogfood via `agent-browser` at desktop and mobile widths with screenshots and WebM recording captured locally.

## Risks

Low. The change is scoped to user-message rendering for the scheduled workflow trigger label and reuses the existing workflow badge/popover path for workflow names.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$10.82`_

### DIFF
--- a/src/browser/features/Messages/MessageRenderer.stories.tsx
+++ b/src/browser/features/Messages/MessageRenderer.stories.tsx
@@ -14,6 +14,7 @@ import {
   createUserMessage,
 } from "@/browser/stories/mocks/messages";
 import {
+  SCHEDULED_WORKFLOW_TRIGGER_LABEL,
   WORKFLOW_RESULT_METADATA_TYPE,
   WORKFLOW_RUN_CARD_DISPLAY_METADATA_TYPE,
   WORKFLOW_TRIGGER_DISPLAY_METADATA_TYPE,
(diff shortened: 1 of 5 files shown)


ThomasK33:
@codex review

PR #3568, opened 2026-06-15
🤖 refactor: remove mux.md link support

ammar-agent:
## Summary

Removes all **mux.md link support** — the end-to-end-encrypted share/paste feature — which has regressed and is no longer maintained. This strips the sharing UI, the link-consumption path, the signing service that existed solely to sign mux.md shares, the client library and its dependency, and the related preferences/docs. In the plan-file display, the former mux.md **Share** (copy link) action is replaced with a **Copy file path** action.

## Background

The mux.md feature let users share assistant messages / transcripts to an external encrypted paste service (producing `https://mux.md/<id>#<key>` links), consume such links via `web_fetch`, and cryptographically sign shares. That product area has regressed and will no longer be maintained, so it's being removed wholesale.

## Implementation

- **Sharing UI removed:** `ShareTranscriptDialog`, `ShareMessagePopover`, `ShareSigningBadges`, the `AssistantMessage` "Share" button, the "Share transcript" workspace-menu item + `SHARE_TRANSCRIPT` keybind, and the telemetry-gated `TelemetryEnabledContext`/`useLinkSharingEnabled` (its only purpose was gating the share UI). The `api.telemetry` namespace itself is untouched.
- **Link consumption removed:** `web_fetch` no longer special-cases mux.md URLs (no more client-side decrypt branch).
- **Signing removed:** `signingService` + the orpc `signing` namespace/schemas/context wiring (only consumed by the two share components).
- **Plumbing removed:** `muxMd` client lib, `shareExpiration` helpers, `sharedUrlCache`, `SHARE_EXPIRATION_KEY`/`SHARE_SIGNING_KEY`, the `userPreferences.sharing` block + its storage round-trip, the `muxMdUrlOverride` preload bridge, `__MUX_MD_URL_OVERRIDE__` Vite define, and ambient global types.
- **Dependency:** dropped `@coder/mux-md-client`.
- **Docs/skill:** removed `workspaces/sharing.mdx` (+ image, nav, redirects), trimmed the telemetry doc's Share note, and regenerated the `mux-docs` built-in skill index.
- **Plan-file display:** the `propose_plan` card's `Share` action is replaced with a `Copy file path` action that copies the plan's on-disk path (hidden when no path is available).

## Validation

- `make static-check` passes (lint, typecheck, formatting, docs-sync, code→docs links).
[...]

PR #3586, opened 2026-06-17
🤖 fix: distinguish plan Copy file path button icon

ammar-agent:
## Summary

Differentiate the plan-file display's two copy buttons. Follow-up to #3568, which replaced the removed mux.md "Share" link action with a "Copy file path" action — but both that button and the existing "Copy" (content) button used the same `Clipboard`/`ClipboardCheck` icons, making them visually indistinguishable.

## Background

#3568 removed mux.md link support and, in the plan display, swapped the "Share" link for a "Copy file path" action. Because the new button reused the content button's `Clipboard`/`ClipboardCheck` icons, the two adjacent actions looked identical and only differed by label.

## Implementation

The "Copy file path" button now uses lucide's `Copy`/`Check` icons — the repo's existing copy-a-value idiom (see `RuntimeBadge`, which copies workspace paths) — while the content "Copy" button keeps `Clipboard`/`ClipboardCheck`. The two buttons now read as distinct actions at rest and in their copied state.

## Risks

Minimal — icon-only change in `ProposePlanToolCall.tsx`. No behavior, props, or copy-to-clipboard logic changed.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$25.12`_

### DIFF
--- a/src/browser/features/Tools/ProposePlanToolCall.tsx
+++ b/src/browser/features/Tools/ProposePlanToolCall.tsx
@@ -61,9 +61,11 @@ import type { ReviewActionCallbacks } from "../Shared/InlineReviewNote";
 import { isPlanFilePath, normalizePlanFilePath } from "@/common/types/review";
 import type { ThinkingLevel } from "@/common/types/thinking";
 import {
+  Check,
   Clipboard,
   ClipboardCheck,
   ClipboardList,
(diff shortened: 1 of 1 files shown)


ammar-agent:
@codex review

PR #3588, opened 2026-06-18
🤖 fix: request OAuth scopes from protected resource metadata

ethanndickson:
## Summary

Request OAuth scopes from protected resource metadata when an MCP server omits `scope=` in its Bearer challenge.

## Background

Some remote MCP servers advertise their available scopes in protected resource metadata but do not repeat them in the initial 401 challenge. In that case the SDK still discovers the authorization server correctly, but the authorization request can end up narrower than the server expects for subsequent tool calls.

## Implementation

Keep the existing challenge probe, but when the challenge has no `scope`, fetch the protected resource metadata URL from the same challenge and use its `scopes_supported` as the requested OAuth scope string. If the challenge already includes `scope`, keep that behavior unchanged.

## Validation

Validated locally with `make static-check` and the focused OAuth service test file covering both the challenge-scope path and the protected-resource-metadata fallback path.

## Risks

Low and isolated to MCP OAuth startup for remote servers. If the metadata fetch fails or advertises no scopes, the flow falls back to the previous behavior.

---

_Generated with `mux` • Model: `openai:gpt-5.4` • Thinking: `xhigh`_

### DIFF
--- a/src/node/services/mcpOauthService.test.ts
+++ b/src/node/services/mcpOauthService.test.ts
@@ -9,6 +9,7 @@ import {
   McpOauthService,
   parseBearerWwwAuthenticate,
   probeServerForBearerChallenge,
+  resolveOAuthScope,
 } from "./mcpOauthService";
 
 function getStoreFilePath(muxHome: string): string {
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

PR #3590, opened 2026-06-18
🤖 fix: retry workflow validation failures immediately

ThomasK33:
## Summary

Updates the workflow runner's `parallelAgents` scheduler so structured-output validation failures retry immediately when a slot is available instead of waiting for unrelated sibling tasks to finish.

## Background

The deep-review workflow fans out lane/review tasks in parallel. When one lane returned invalid structured output, the runner recorded the validation failure but queued the retry for a later wave, so users could be left waiting on a slow sibling before the failed lane retried. This PR makes schema-validation retries opportunistic and concurrent with the rest of the fan-out.

## Implementation

- Replaced the fixed-wave `parallelAgents` settle loop with a dynamic queue that launches more work whenever a running task settles.
- Requeues retryable validation failures at the front of the queue, preserving the existing max-attempt limit and retry prompt behavior.
- Preserves terminal behavior for non-validation failures by interrupting active siblings and failing the batch.
- Propagates runtime/upstream aborts before returning from a partially launched queue, including interrupting any bulk-created children that can no longer be waited.
- Keeps bulk task creation for unconstrained first-attempt fan-outs, while retry attempts are launched through the normal single-task path so they can start immediately.

## Validation

- `bun test src/node/services/workflows/WorkflowRunner.test.ts --test-name-pattern "validation|parallelAgents"`
- `make typecheck`
- `workflow_run simplify --target HEAD --fix`
- `bun test src/node/services/workflows/WorkflowRunner.test.ts`

## Risks

This touches workflow runner scheduling for `parallelAgents`, which is central workflow infrastructure. The risk is bounded by focused regression coverage for immediate validation retry, retry priority under `maxParallel`, aborted queued work, existing sibling-interrupt behavior, and the full workflow runner test file.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `311517{MUX_COSTS_USD:-unknown}`_

### DIFF
--- a/src/node/services/workflows/WorkflowRunner.test.ts
+++ b/src/node/services/workflows/WorkflowRunner.test.ts
@@ -1487,6 +1487,61 @@ describe("WorkflowRunner", () => {
     expect(interruptRunCalls).toBe(0);
   });
 
+  test("does not complete parallelAgents when abort prevents queued tasks from launching", async () => {
+    using tmp = new DisposableTempDir("workflow-runner-parallel-abort-queued");
+    const store = new WorkflowRunStore({
+      sessionDir: tmp.path,
+      staleLeaseMs: WORKFLOW_RUNNER_TEST_STALE_LEASE_MS,
+    });
+    await store.createRun({
+      id: "wfr_parallel_abort_queued",
+      workspaceId: "workspace-1",
+      definition,
+      definitionSource: `export default function workflow({ parallelAgents }) {
+        parallelAgents([
+          { id: "source-a", prompt: "Read source A" },
+          { id: "source-b", prompt: "Read source B" },
+        ]);
+        return { reportMarkdown: "must not complete" };
+      }`,
+      args: {},
+      now: "2026-05-29T00:00:00.000Z",
+    });
+    const abortController = new AbortController();
+    let interruptRunCalls = 0;
+    const runner = createRunner(store, {
+      async runAgent() {
+        throw new Error("queued agents should not start after abort");
+      },
+      async createAgentTasks(
+        specs,
+        lifecycle?: { onTaskCreated?: (index: number, taskId: string) => Promise<void> | void }
+      ) {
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

ThomasK33:
Addressed Codex review thread `PRRT_kwDOPxxmWM6KeFcS`:

- Added abort propagation for queued `parallelAgents` work so an aborted batch cannot return a partially filled result array.
- Interrupts any bulk-created child tasks when abort prevents queued tasks from launching.
- Added regression coverage: `does not complete parallelAgents when abort prevents queued tasks from launching`.
- Re-ran `bun test src/node/services/workflows/WorkflowRunner.test.ts` and `MUX_ESLINT_CONCURRENCY=1 make static-check` successfully.

ThomasK33:
@codex review

Please take another look.

ThomasK33:
Addressed Codex review threads from the second round:

- `PRRT_kwDOPxxmWM6KeVma` Preserve original sibling failure: changed the queued-work guard to only synthesize an abort error when the abort signal prevents queued work from launching, so already-settled child failures are consumed and surfaced through the existing failure path. Added regression coverage for a failed windowed sibling with queued specs.
- `PRRT_kwDOPxxmWM6KeVmg` Keep bulk creation for non-windowed fanouts: bulk `createAgentTasks` is now used whenever `maxParallel` is absent or does not actually cap the current fanout (`maxParallel >= queued.length`). Updated the bulk-creation test to pin the `maxParallel === length` case.

Validation after the fixes:
- `bun test src/node/services/workflows/WorkflowRunner.test.ts --test-name-pattern "bulk creates|original parallelAgents child failure|abort prevents queued|validation|parallelAgents"`
- `bun test src/node/services/workflows/WorkflowRunner.test.ts`

ThomasK33:
@codex review

Please take another look.

PR #3593, opened 2026-06-18
🤖 fix: remove deep research quick flag

ThomasK33:
## Summary

Remove the redundant `quick` / `--quick` argument from the built-in `deep-research` workflow so callers use `mode` consistently for quick/smart/fast selection.

## Background

The workflow previously exposed both `{ "quick": true }` and `{ "mode": "quick" }`, which made the structured argument contract redundant and confusing. Keeping only `--mode quick` makes the schema match the actual mode selection API.

## Implementation

- Removed `quick` from the `deep-research` metadata args schema.
- Removed `args.quick` normalization from `normalizeDeepResearchInput`.
- Regenerated the embedded built-in workflow content.
- Updated workflow metadata tests to exercise `mode` instead of `quick`.

## Validation

- `bun test src/node/builtinWorkflows/deep-research.test.ts`
- `bun test src/node/services/tools/workflow_definitions.test.ts`
- `bun test src/node/services/workflows/builtInWorkflowDefinitions.test.ts -t "quick mode applies smaller caps"`
- `make fmt-check`
- `make typecheck`
- `make static-check`
- Dogfood metadata check confirmed `deep-research` exposes `topic`, `input`, `query`, and `mode` only; no `quick` field or `--quick` alias remains.

## Risks

Low. This intentionally removes a workflow argument alias, so callers using `--quick` or `{ "quick": true }` must switch to `--mode quick` or `{ "mode": "quick" }`.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$4.65`_


ThomasK33:
@codex review

PR #3594, opened 2026-06-18
🤖 fix: prevent workflow task recovery after interrupts

ThomasK33:
## Summary

Prevents task-level startup, queue, stream-end, and stream-error recovery from resurrecting workflow-owned child tasks after their owning workflow run is inactive.

## Background

Workflow interrupt/abort is the source of truth for workflow execution. Before this change, TaskService recovery paths could still prompt or restart child agent tasks after the owning workflow was interrupted, which could burn tokens and violate workflow lifecycle boundaries.

## Implementation

- Resolves the nearest workflow-owned task in a task ancestry chain and reads the durable workflow run status from that run's store.
- Interrupts workflow-owned tasks, including nested descendants, when their owning run is inactive or unavailable instead of prompting for recovery.
- Reuses task indexes across recovery and descendant checks to avoid repeated full-workspace indexing.
- Extracts the shared interrupted-task mutation logic so inactive-owner recovery and descendant interruption preserve completed reports and queued prompts consistently.
- Adds regression coverage for startup recovery and stream-end behavior after an owning workflow is interrupted.

## Validation

- Ran simplify workflow `wfr_ed8469895aaa74a1`; it applied cleanup commit `e0ab3551b` and reported targeted validation passed.

## Risks

Moderate TaskService lifecycle risk: this touches recovery paths for queued/running/awaiting-report tasks. The guard is limited to tasks that are workflow-owned directly or through ancestry, and active workflow runs still allow normal recovery.

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$17.65`_

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -5756,6 +5756,92 @@ describe("TaskService", () => {
     );
   });
 
+  test("initialize drains queued tasks after interrupting inactive workflow children", async () => {
+    const config = await createTestConfig(rootDir);
+
+    const projectPath = await createTestProject(rootDir, "repo-queue-after-interrupt");
+    const runtimeConfig = { type: "worktree" as const, srcBaseDir: config.srcDir };
+    const runtime = createRuntime(runtimeConfig, { projectPath });
+    const rootName = "root";
+    await runtime.createWorkspace({
+      projectPath,
+      branchName: rootName,
+      trunkBranch: "main",
+      directoryName: rootName,
+      initLogger: createNullInitLogger(),
+    });
+
+    const rootId = "root-queue-after-interrupt";
+    const runningTaskId = "running-inactive-workflow-occupies-slot";
+    const queuedTaskId = "queued-starts-after-interrupt";
+    const workflowRunId = "wfr_queue_after_interrupt";
+
+    await saveWorkspaces(
+      config,
+      projectPath,
+      [
+        {
+          path: runtime.getWorkspacePath(projectPath, rootName),
+          id: rootId,
+          name: rootName,
+          createdAt: new Date().toISOString(),
+          runtimeConfig,
+        },
+        projectWorkspace(projectPath, "running", runningTaskId, {
+          name: "agent_explore_running",
+          parentWorkspaceId: rootId,
+          agentId: "explore",
+          agentType: "explore",
+          taskStatus: "running",
+          taskModelString: defaultModel,
+          runtimeConfig,
+          workflowTask: { runId: workflowRunId, stepId: "running" },
+        }),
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

ThomasK33:
Addressed Codex thread `PRRT_kwDOPxxmWM6Klrcy`: startup recovery now pre-interrupts inactive workflow-owned awaiting/running tasks, refreshes the config/task index, and then runs parent descendant checks so an awaiting parent is not left stuck behind a child that was just interrupted. Added a regression test for that parent-after-child recovery case.

ThomasK33:
@codex review

Please take another look after the startup recovery refresh fix.

ThomasK33:
Addressed Codex thread `PRRT_kwDOPxxmWM6Kl4iu`: after startup interrupts inactive workflow-owned tasks and finishes recovery prompts, TaskService now awaits another queue drain so freed capacity can launch unrelated queued tasks immediately. Added a regression test where maxParallelAgentTasks=1 is initially occupied by an inactive workflow child and a queued task starts in the same initialize pass after the interrupt.

ThomasK33:
@codex review

Please take another look after the post-interrupt queue drain fix.

ThomasK33:
Addressed Codex thread `PRRT_kwDOPxxmWM6KmEKo`: `maybeStartQueuedTasksFromReservations` now scans queued tasks for inactive workflow owners before capacity gating or slot consumption, so abandoned queued workflow steps get interrupted even when no slots are available. Extended the startup queue regression to cover an inactive queued workflow task alongside an unrelated queued task.

ThomasK33:
@codex review

Please take another look after the queued inactive-workflow prepass fix.

ThomasK33:
Addressed Codex thread `PRRT_kwDOPxxmWM6KmN8Z`: inactive workflow-owner interruptions now schedule queued-task draining by default after a real interruption, and scheduler callsites opt out to avoid recursive queue scans. Extended the stream-end inactive workflow regression to assert that the queue drain is scheduled.

ThomasK33:
@codex review

Please take another look after the inactive-interrupt queue-drain scheduling fix.

ThomasK33:
Addressed Codex thread `PRRT_kwDOPxxmWM6KmZWY`: workflow recovery now checks every workflow owner in the ancestry chain, returning inactive if any owner is inactive. The regression now covers a nested task whose nearest inner workflow run is active while an outer workflow owner is interrupted.

ThomasK33:
@codex review

Please take another look after the ancestry-wide workflow owner check fix.

PR #3596, opened 2026-06-19
🤖 feat: add workspace heartbeat tool

ThomasK33:
## Summary

Adds a model-facing `heartbeat` tool so an agent can read, set, or unset the scheduled heartbeat for its current workspace without editing config files directly.

## Background

Workspace heartbeats already exist as a backend scheduling mechanism, but agents could not adjust their own idle check-in cadence or disable/reset workspace-specific overrides during long-running work. This exposes that capability through the existing tool pipeline while keeping the target scoped to the executing workspace.

## Implementation

- Defines heartbeat tool input/result schemas, bridgeable PTC result typing, generated hook docs, and shared `HeartbeatToolArgs`/`HeartbeatToolResult` types.
- Adds `createHeartbeatTool`, wired into `getToolsForModel` only when a heartbeat service is available, the `workspace-heartbeats` experiment is enabled, and the workspace is not an agent-report child workspace.
- Injects `WorkspaceService` into `AIService` as the heartbeat tool service and propagates the resolved heartbeat experiment flag through preserved send options/workflow task experiment context.
- Extends `WorkspaceService` to resolve heartbeat workspace entries, return canonical settings from sparse updates, default sparse persisted heartbeat intervals on read, unset workspace-specific heartbeat settings, and preserve ORPC compatibility by mapping set results back to `Result<void, string>`.
- Normalizes heartbeat metadata in `Config.getAllWorkspaceMetadata()` so sparse persisted heartbeat settings stay schema-valid before the renderer lists workspaces.
- Maps `workspace-heartbeats` through `mux run --experiment` and `mux workflow --experiment` request-scoped experiment options.
- Updates hook/context tests to use real providers with injected clients instead of process-global API module mocks, preventing leakage into unrelated hook tests during the unit suite.

## Validation

- `bun test src/node/services/tools/heartbeat.test.ts src/node/services/workspaceService.heartbeatSettings.test.ts src/common/utils/tools/tools.test.ts`
- `bun test src/common/utils/tools/toolDefinitions.test.ts src/node/services/ptc/typeGenerator.test.ts`
- `bun test src/node/services/workspaceService.heartbeatSettings.test.ts src/node/services/tools/heartbeat.test.ts`
- `bun test src/node/config.test.ts -t "defaults sparse persisted heartbeat intervals in workspace metadata"`
- `bun test src/cli/workflow.test.ts -t "CLI commands work without the dynamic-workflows experiment"`
- `bun test src/browser/hooks/useWorkspaceHeartbeat.test.tsx src/browser/hooks/useChatTranscriptFullWidth.test.tsx`
- `bun test src/browser/contexts/WorkspaceContext.test.tsx src/browser/hooks/useWorkspaceHeartbeat.test.tsx src/browser/hooks/useChatTranscriptFullWidth.test.tsx`
- `bun test src/node/services/workflows/WorkflowTaskServiceAdapter.test.ts`
- `make typecheck`
- `git diff --check`
- `make static-check`
- `MUX_ESLINT_CONCURRENCY=1 make static-check` after the hook/context test isolation and experiment-gating/sparse-metadata fixes (default local ESLint concurrency was killed by the workspace memory limit)
- Ran the `simplify` workflow and committed its actionable cleanup pass.

## Risks

Low-to-medium. The change touches model tool assembly, shared tool schemas, workspace config persistence/metadata normalization, CLI experiment option mapping, and focused test fixtures. The heartbeat tool is gated by the same experiment used by the UI/keybind/slash-command paths, only exposed for normal workspaces with the heartbeat service injected, and existing ORPC heartbeat settings behavior is preserved.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$33.49`_


ThomasK33:
@codex review

ThomasK33:
@codex review

Re-requesting after the hook test isolation fix for the unit-suite failure.

ThomasK33:
@codex review

Addressed your heartbeat experiment-gating feedback by requiring the resolved `workspace-heartbeats` experiment before registering the model-facing `heartbeat` tool, and by propagating that resolved flag through preserved send/workflow task experiment context.

ThomasK33:
@codex review

Re-requesting after the latest unit-suite fix: `WorkspaceContext.test.tsx` now uses the real APIProvider with an injected client instead of a process-global API module mock, which should stop leaking the mocked APIProvider into the heartbeat/full-width hook tests.

ThomasK33:
@codex review

Addressed the sparse heartbeat interval feedback by normalizing persisted heartbeat settings with the resolved global default interval before returning them from WorkspaceService, and added a regression test for missing persisted `intervalMs`.

ThomasK33:
@codex review

Addressed both latest findings:
- heartbeat metadata is now normalized in `Config.getAllWorkspaceMetadata()` so sparse persisted intervals are defaulted before `workspace.list`/renderer validation
- `workspace-heartbeats` now maps through both `mux run --experiment` and `mux workflow --experiment` request-scoped experiment objects

PR #3605, opened 2026-06-23
🤖 fix: background foreground waits for prequeued messages

ethanndickson:
## Summary

Fix foreground sub-agent waits so they are sent to the background when a tool-end user message was already queued before the foreground wait registered.

## Background

The existing auto-backgrounding path was edge-triggered from message enqueue. If the user queued a message while a foreground task wait was already registered, the wait was rejected and the task continued in the background. If the user queued the message first and the agent later spawned a foreground task in the same step, the enqueue-time signal hit an empty waiter set and the foreground wait blocked until the child completed.

## Implementation

`AgentSession.hasQueuedMessages` now accepts an optional dispatch mode, and both the stream step-end callback and task wait registration use the same tool-end queue predicate. When a backgroundable foreground waiter registers, `TaskService` re-checks the requesting workspace's current tool-end queue state and reuses the existing `backgroundForegroundWaitsForWorkspace` path to mark and reject the wait.

## Risks

Low. The behavior is scoped to foreground waits that already opt into queue-triggered backgrounding, and the new check only fires for existing tool-end queued messages.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$9.72`_

### DIFF
--- a/src/node/services/agentSession.ts
+++ b/src/node/services/agentSession.ts
@@ -3608,7 +3608,7 @@ export class AgentSession {
       workspaceGoalService: this.workspaceGoalService,
       experiments: options?.experiments,
       disableWorkspaceAgents: options?.disableWorkspaceAgents,
-      hasQueuedMessage: () => this.hasQueuedMessages("tool-end"),
+      hasQueuedMessages: this.hasQueuedMessages.bind(this),
       openaiTruncationModeOverride,
     });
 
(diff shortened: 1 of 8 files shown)


ethanndickson:
@codex review

ethanndickson:
Addressed the deferred setup cleanup issue: the prequeued-message background check now runs only after each foreground waiter path has installed its cleanup-managed state (timers, abort listeners, and queued-task start waiters). I also added focused coverage for both agent report waits and workspace-turn waits.

@codex review

ethanndickson:
@codex review

PR #3610, opened 2026-06-23
🤖 fix: remove skill pill from slash suggestions

ThomasK33:
## Summary

Removes the `SKILL` text pill from chat slash command suggestions so skill entries and model one-shot aliases use the same row layout.

## Background

The skill rows already include scope context in their descriptions, and the extra pill made skill suggestions feel visually different from nearby model aliases without adding much value.

## Implementation

Deleted the slash suggestion badge component and its row render call. The underlying suggestion `kind` metadata is unchanged for future UI use.

## Validation

- `bun test src/browser/features/ChatInput/CommandSuggestions.test.tsx`
- `make typecheck`
- `git diff --check`
- Storybook dogfood at desktop and mobile widths confirmed the `SKILL` text no longer appears in the slash suggestion list.

## Risks

Low. This is a presentational removal in the shared command suggestion row; selection behavior and suggestion data are unchanged.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `244294{MUX_COSTS_USD:-0}`_

### DIFF
--- a/src/browser/features/ChatInput/CommandSuggestions.tsx
+++ b/src/browser/features/ChatInput/CommandSuggestions.tsx
@@ -67,29 +67,6 @@ function HighlightedText({
   return <span className={className}>{parts}</span>;
 }
 
-const SUGGESTION_KIND_BADGES = {
-  skill: { label: "Skill", className: "text-medium" },
-} satisfies Partial<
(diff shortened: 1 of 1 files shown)


ThomasK33:
@codex review

PR #3616, opened 2026-06-23
🤖 ci: cache Playwright browsers on macOS

ThomasK33:
## Summary

Cache Playwright browser installs on the correct OS-specific paths and bound the shared setup step so CI fails fast if a browser download stalls.

## Background

The macOS E2E job on PR #3610 got stuck inside `setup-playwright` after Chromium reached 100%. The existing cache path only covered Linux (`~/.cache/ms-playwright`), so macOS runners redownloaded browsers into `~/Library/Caches/ms-playwright` on every cache miss instead of saving/restoring that directory.

## Implementation

- Cache the Linux, macOS, and Windows Playwright browser cache roots in `.github/actions/setup-playwright`.
- Add `runner.arch` to the Playwright cache key so arm64/x64 browser caches stay separate.
- Add a 10-minute timeout to each workflow callsite of the shared setup action.

## Validation

- `make lint-actions`
- `make static-check`
- `git diff --check`

## Risks

Low. This only changes CI cache paths/keys and setup-step timeouts. The main behavior change is that a future stalled browser install fails in 10 minutes rather than blocking the workflow for hours.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `3451720{MUX_COSTS_USD:-unknown}`_

### DIFF
--- a/.github/actions/setup-playwright/action.yml
+++ b/.github/actions/setup-playwright/action.yml
@@ -21,10 +21,15 @@ runs:
       id: cache-playwright
       uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
       with:
-        path: ~/.cache/ms-playwright
-        key: ${{ runner.os }}-playwright-${{ steps.playwright-version.outputs.version }}-${{ inputs.browsers }}
+        # Playwright uses OS-specific browser cache roots; include each default so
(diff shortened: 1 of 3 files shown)


ThomasK33:
@codex review

PR #3618, opened 2026-06-23
🤖 refactor: remove workflow agentType option

ThomasK33:
## Summary
Remove the legacy `agentType` workflow-agent option so workflow authors must use `agentId` consistently.

## Background
The workflow authoring skill still advertised `legacy agentType` as an optional alias. This removes the alias from new workflow runs to avoid two names for the same workflow-agent selector.

## Implementation
- Reject new `agent(..., { agentType: ... })` calls with a clear error directing authors to `agentId`.
- Preserve replay/resume compatibility for persisted workflow run source snapshots that already contain `agentType`, including current-schema runs created before this change.
- Store a new-run marker so runs created after this change reject the alias, while older run records that lack the marker can still replay.
- Keep `agentId`, `model`, and `thinking` normalization unchanged.
- Update workflow-authoring skill content and tests to use `agentId` only.

## Validation
- `bun test src/node/services/workflows/WorkflowRunner.test.ts`
- `bun test src/node/services/workflows/WorkflowRunStore.test.ts`
- `make typecheck`
- `make fmt-check`
- `make static-check`
- `git diff --check`

## Risks
Low. New workflow runs intentionally fail on the legacy `agentType` alias, while existing persisted runs keep enough compatibility to replay or resume their original source snapshots.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$2.84`_

### DIFF
--- a/src/common/orpc/schemas/workflow.ts
+++ b/src/common/orpc/schemas/workflow.ts
@@ -244,6 +244,7 @@ export const WorkflowRunRecordSchema = z.object({
   sourceHash: z.string().min(1),
   args: JsonValueSchema,
   agentOutputSchemaRequired: z.boolean().optional(),
+  agentTypeAliasAllowed: z.boolean().optional(),
   parentWorkflow: WorkflowRunParentSchema.optional(),
   status: WorkflowRunStatusSchema,
   createdAt: IsoDateTimeSchema,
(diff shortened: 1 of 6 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed the replay concern by rejecting `agentType` for new workflow runs while preserving alias normalization for legacy persisted workflow run source snapshots.

ThomasK33:
@codex review

Addressed the current-schema replay concern by adding a new-run marker: fresh runs reject `agentType`, while older persisted run records that lack the marker can still normalize `agentType` while replaying their source snapshot.

PR #3619, opened 2026-06-23
🤖 fix: recover stale workspace turn handles

ThomasK33:
Summary

Fix stale workspace-turn recovery so completed delegated workspace turns do not remain blocked by archived descendant tasks whose persisted status still says running. The change also lets stale running task entries return persisted report artifacts instead of timing out task_await/waitForAgentReport.

Background

Workspace aba21cd8d7 had a workspace-turn handle waiting on child workspace 9a69e6b062 even though the child had already produced final assistant messages. The blocker was archived descendant task metadata that still carried taskStatus: running, causing Mux to treat the workspace turn as active forever and to skip deferred completion history during recovery.

Implementation

- Treat archived, non-streaming task workspaces as inactive for active-descendant and active-task accounting.
- Recover deferred workspace-turn stream-end history once descendant work is no longer active.
- Self-heal workspace-turn handles that older recovery incorrectly marked interrupted after restart.
- Return persisted sub-agent report artifacts after waiter registration even when config still says the task is running.

Validation

Risks

Medium-low risk in task/workspace-turn orchestration. The behavioral change is scoped to archived non-streaming tasks and durable report/turn recovery paths; active streams and queued/preparing turns still count as live.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `488084{MUX_COSTS_USD:-unknown}`_

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -1409,6 +1409,291 @@ describe("TaskService", () => {
     expect(recovered?.reportMarkdown).toBeUndefined();
   });
 
+  test("workspace-turn stale recovery repairs restart-interrupted deferred handles after descendants stop blocking", async () => {
+    const { config, parentId, projectPath, taskService, historyService, workspaceMocks } =
+      await startWorkspaceTurnForTest({ disposable: true });
+    await config.editConfig((cfg) => {
+      const project = Array.from(cfg.projects.values())[0];
+      assert(project, "test project must exist");
+      project.workspaces.push({
+        path: path.join(projectPath, "descendant-task"),
+        id: "descendant-task",
+        name: "descendant-task",
+        createdAt: "2026-06-19T00:00:00.000Z",
+        runtimeConfig: { type: "local" },
+        parentWorkspaceId: "childworkspace",
+        taskStatus: "running",
+      });
+      return cfg;
+    });
+    const muxMetadata = {
+      type: "workspace-turn-task" as const,
+      taskHandleId: "wst_handle",
+      ownerWorkspaceId: parentId,
+      turnId: "turn",
+    };
+    const appendResult = await historyService.appendToHistory(
+      "childworkspace",
+      createMuxMessage("msg_prehandoff", "assistant", "Recovered final text", {
+        model: "anthropic:claude-opus-4-6",
+        agentId: "exec",
+        finishReason: "stop",
+        muxMetadata,
+      })
+    );
+    expect(appendResult.success).toBe(true);
+    const internal = taskService as unknown as {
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

ThomasK33:
Addressed Codex review findings:

- `Recover deferred error turns as terminal errors`: updated stale interrupted workspace-turn self-heal to persist any recovered terminal record, including `error` records reconstructed from non-stop finish reasons. Added regression coverage for a deferred `finishReason: "length"` history message repairing to the real workspace-turn error.
- `Let task_list inspect archived stale tasks`: kept archived/non-streaming filtering out of the general `listDescendantAgentTasks` listing path so `task_list(includeArchived: true)` can still inspect stale archived running tasks; active accounting continues to use the narrower active-task predicate.

ThomasK33:
@codex review

Please take another look.

ThomasK33:
Addressed second Codex review findings:

- `Gate deferred recovery on all background work`: deferred workspace-turn recovery now checks the same categories that caused deferral—active descendant agent tasks, active referenced workflow runs, and active nested workspace-turn handles—before replaying deferred final history. Added regression coverage for an active child workflow blocker.
- `Clean up disposable workspaces after repair`: the interrupted-handle repair path now runs disposable workspace cleanup after persisting the recovered terminal record. Added regression coverage that a repaired disposable workspace turn calls the cleanup path.

ThomasK33:
@codex review

Please take another look.

ThomasK33:
Addressed third Codex review finding:

- `Mark recovered report tasks terminal before resolving`: when `waitForAgentReport` finds a durable report artifact for a stale child still marked running, it now marks the persisted workspace entry `reported`, emits metadata, drains queued work, and runs reported-leaf cleanup before resolving the waiter. The stale-running artifact regression now asserts the persisted task status is repaired.

ThomasK33:
@codex review

Please take another look.

ThomasK33:
Addressed fourth Codex review finding:

- `Drop stale errors on recovered completions`: recovered completed workspace-turn records now clear the old restart-interruption error, while recovered error records still set their real terminal error. The restart-interrupted completion regression now asserts the repaired completed snapshot has no stale error.

ThomasK33:
@codex review

Please take another look.

ThomasK33:
Addressed fifth Codex review finding:

- `Start patch generation before cleaning stale reports`: stale-running report recovery now starts subagent git patch artifact generation after marking the child reported and before queue drain / reported-leaf cleanup. The persisted-artifact regression now spies on patch generation to ensure it runs before the stale task is resolved.

ThomasK33:
@codex review

Please take another look.

PR #3624, opened 2026-06-24
feat: add Workflows right-sidebar tab with live run streaming

ThomasK33:
## Summary

Adds a **Workflows** tab to the right sidebar so a workspace's durable workflow runs get a dedicated, live observation surface — instead of relying on the large in-chat workflow tool call card. Implements the imported "Workflows Tab" design.

The tab is gated on the existing **dynamic-workflows** experiment (it is that feature's observation surface) and shows the focused run live, with run history and an empty-state launcher.

## Backend

- **`workflows.subscribe`** — push stream (initial snapshot + per-write deltas) of a workspace's top-level runs. `WorkflowService` is constructed per oRPC request, so it can't be the singleton event source a subscription needs; instead every run mutation already flows through `WorkflowRunStore.writeRunFile`, which now notifies a module-level `workflowRunStreamHub` that the subscribe handler listens to. This decouples *who runs the workflow* (chat tool, CLI, background, crash recovery) from *who is watching it*.
- **`workflows.listScripts`** — enumerates runnable workflow scripts (built-in + project + global skills exposing a `workflow.js`) for the empty-state launcher, reusing the existing resolver/skill discovery and surfacing each script's declared args.

## Frontend

- A pure, unit-tested **projector** folds the intentionally-lean run record (`events` + `steps`; steps carry no phase/title/usage) into the phase→step view: phase assignment by event sequence, titles from task events, durations from timestamps, final report/error from events.
- **Timeline UI**: run header (interrupt / resume / retry-from-checkpoint / re-run), phases (collapsed by default so a 20+ step fan-out stays scannable, with expandable phase info objects), expandable steps with per-step output + structured output, run history, and an empty-state launcher with a per-arg input form. The final report is collapsible and renders the full **structured output** returned to the agent, not just summary chips.
- Tab registration mirrors the memory/browser experiment-gated tabs (added to the layout via a per-experiment effect), with a live active-run count badge on the label.

## Behavior change

When the Workflows tab is available, the in-chat workflow tool card **auto-collapses by default** (any status). Without it, the prior behavior (auto-collapse completed runs only) is preserved — so behavior and existing tests are unchanged when the experiment is off.

## Deferred (deliberate)

- **Per-step token/cost**: sub-agent usage rolls up into the parent workspace and isn't reliably queryable by task once a step completes. The projector/UI already accept a usage overlay for when per-step capture lands.
- **Background a live run**: needs runner detach support; Interrupt covers the stop case.

## Testing

`make typecheck`, `make lint`, and the workflow suites all green (205 tests across the frontend Workflows + WorkflowRunToolCall + backend workflow + schema tests, including a dedicated projector test).

### DIFF
--- a/src/browser/features/RightSidebar/RightSidebar.tsx
+++ b/src/browser/features/RightSidebar/RightSidebar.tsx
@@ -674,6 +674,7 @@ const RightSidebarComponent: React.FC<RightSidebarProps> = ({
   const desktopExperimentEnabled = useExperimentValue(EXPERIMENT_IDS.PORTABLE_DESKTOP);
   const browserExperimentEnabled = useExperimentValue(EXPERIMENT_IDS.AGENT_BROWSER);
   const memoryExperimentEnabled = useExperimentValue(EXPERIMENT_IDS.MEMORY);
+  const workflowsExperimentEnabled = useExperimentValue(EXPERIMENT_IDS.DYNAMIC_WORKFLOWS);
   // Child task workspaces can't run goal actions — backend rejects them
   // via `WorkspaceGoalService.assertParentWorkspace`. We use this flag
   // both to hide the Goal tab below and to gate any inline goal UX.
@@ -979,6 +980,26 @@ const RightSidebarComponent: React.FC<RightSidebarProps> = ({
       return prev;
     });
   }, [memoryExperimentEnabled, initialActiveTab, setLayoutRaw]);
+
+  // Workflows tab follows the dynamic-workflows experiment (same shape as the memory tab):
+  // experimental tabs are added/removed from the persisted layout here, not via tabConfig's
+  // featureFlag (which only filters the Add-Tool picker / command palette).
+  React.useEffect(() => {
+    setLayoutRaw((prevRaw) => {
+      const prev = parseRightSidebarLayoutState(prevRaw, initialActiveTab);
+      const hasWorkflows = collectAllTabs(prev.root).includes("workflows");
+
+      if (workflowsExperimentEnabled && !hasWorkflows) {
+        return addTabToFocusedTabset(prev, "workflows", false);
+      }
+
+      if (!workflowsExperimentEnabled && hasWorkflows) {
+        return removeTabEverywhere(prev, "workflows");
+      }
+
+      return prev;
+    });
+  }, [workflowsExperimentEnabled, initialActiveTab, setLayoutRaw]);
 
   React.useEffect(() => {
     setLayoutRaw((prevRaw) => {
--- a/src/browser/features/RightSidebar/Tabs/TabLabels.tsx
+++ b/src/browser/features/RightSidebar/Tabs/TabLabels.tsx
@@ -16,6 +16,7 @@ import {
   Sparkles,
   Target,
   Terminal as TerminalIcon,
+  Workflow,
   X,
 } from "lucide-react";
 import { Tooltip, TooltipContent, TooltipTrigger } from "@/browser/components/Tooltip/Tooltip";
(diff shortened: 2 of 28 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed the P2: phase assignment now derives each step's first sequence from all step-bearing events (`task`, `patch`, `workflow`, and legacy `action`/`validation`) instead of only `task` events, so apply-patch and nested-workflow steps are grouped under their active phase rather than the ungrouped bucket. Nested-workflow steps also take their title from the child workflow name. Added a regression test (`projectWorkflowRun — non-task step events`).

ThomasK33:
@codex review

Addressed all three P2 findings:
- WorkflowsTab: reset `scripts`/`busyScriptName`/`runError` at the start of the discovery effect so a workspace switch can't show or launch a stale script.
- workflowScriptDiscovery: moved descriptor build + arg summarization inside the per-skill try/catch, so a single workflow with malformed arg metadata is skipped instead of aborting discovery for the whole workspace.
- WorkflowTimeline final report: gate the structured-output block on `!== undefined` (not `!= null`) so a valid `structuredOutput: null` still renders instead of leaving an empty report.

ThomasK33:
@codex review

Addressed all three P2 findings:
- Failed phases now initialize expanded, and the phase header shows a destructive alert icon when the phase has a failed step, so the error isn't hidden behind a collapsed phase.
- The timeline now renders a run-level error banner from `view.errorMessage` for failed runs, so pre-step setup/compile/eval failures show a reason instead of "No steps yet".
- Re-run is only offered when the record stores a resolvable script path (canonical/source/requested); the bare-`name` fallback is removed so legacy records don't surface a Re-run button that would fail.

ThomasK33:
@codex review

Addressed both P2s: phase and step disclosure now use a shared `useDisclosureOpenOnFailure` helper (React "adjust state during render" pattern) that opens not only when the phase/step is failed at first mount but also when it transitions to failed via the live subscription — so failures in an in-flight run are no longer hidden behind a collapsed phase/step. Manual collapse still wins; only a fresh false→true failure transition forces it open.

ThomasK33:
@codex review

Addressed both P2s in the empty-state launcher:
- Untouched fields (including booleans) are now omitted from the submitted args, so the backend applies the script's declared default / omit semantics instead of us forcing `false`. Only fields the user actually edited are sent.
- Integer inputs are only coerced when the whole string is a valid integer (`/^[+-]?\d+$/`); partial input like `1.9` or `10abc` is passed through as a string so the backend validates and reports it rather than parseInt silently truncating it.

ThomasK33:
@codex review

Addressed the P2: the launcher's required-field check now ignores required args that declare a default (`arg.default === undefined` added to the predicate). Such fields no longer block Start — submit omits the untouched field and the backend applies the default before validating required, matching slash/chat launches.

ThomasK33:
@codex review

Addressed the P2: required boolean args with no default now block Start until the user sets the checkbox (the required-field predicate handles booleans via `values[arg.name] === undefined` rather than excluding them), so the launcher no longer offers an enabled Start that fails backend `validateRequired`. Required args with a declared default still don't block (backend applies the default).

ThomasK33:
@codex review

Addressed both P2s in the launcher:
- Row state (`configuring`/busy) is now keyed by the unique `scriptPath` instead of `descriptor.name`, so duplicate-named skill workflows (which normalize to "workflow") no longer expand/disable each other's rows.
- The boolean checkbox now reflects the declared default when untouched (`arg.default === true`), so the visible state matches the value that will actually run (submit still omits untouched fields and the backend applies the default).

ThomasK33:
@codex review

Test-only change: `WorkflowRunToolCall.test.tsx` now mocks `useExperimentValue → false` so the suite is hermetic. The component reads the dynamic-workflows experiment to gate auto-collapse; in the full test run, cross-file experiment state (shared happy-dom localStorage) was flipping it on and collapsing the card, hiding the expanded content these tests assert. Forcing the experiment off restores the intended experiment-off behavior under test. No production changes.

ThomasK33:
@codex review

Addressed the P3: required boolean args with no default are now seeded to `false` when the form mounts, so their unchecked state is immediately submittable (Start enabled) and launching with `false` no longer requires a check→uncheck dance. Optional booleans and booleans with a declared default remain untouched (omitted on submit; backend applies the default).

ThomasK33:
@codex review

Addressed both:
- P2: the in-chat workflow card no longer auto-collapses when it carries a tool error (`errorResult != null`). Pre-run failures (invalid args / unresolved script path) create no durable run, so the Workflows tab has nothing to show and the error must stay visible in chat. Non-error cards still auto-collapse when the tab is enabled (and completed runs without it).
- P3: the focused-run header/timeline is now keyed by `focusedRun.id`, so switching runs from history gives fresh UI state (no stale action error, and `useDisclosureOpenOnFailure` correctly auto-opens a new failed phase even if a prior same-named phase was collapsed).

PR #3628, opened 2026-06-24
🤖 fix: apply default budget for model goals

ThomasK33:
## Summary

Model-created `set_goal` calls now inherit the effective positive default goal budget even when user-authored goal forms allow omitted budgets to mean “no budget”. This prevents an omitted model budget/turn cap from unexpectedly resolving to an unbounded goal and failing before the workspace/global defaults can apply.

## Background

The `set_goal` tool description says omitted or null budget/turn fields use the effective workspace goal defaults. In practice, the model-specific resolver still respected `alwaysRequireExplicitBudget: false` like user-facing forms do, so a model call with omitted bounds could resolve to no budget and no turn cap and trip the safety guard.

## Implementation

- Updated `resolveModelGoalSetIntent` so omitted/null model budgets always use the positive effective default budget when one exists.
- Kept the safety guard for configurations whose effective defaults truly have no positive budget and no turn cap.
- Clarified the guard error message and added coverage for the model-tool behavior.

## Validation

- `bun test src/common/utils/goals/resolveGoalSetIntent.test.ts src/node/services/tools/goal.test.ts`
- `make typecheck`

## Risks

Low. This only changes model-authored goal default resolution; user-authored goal creation still uses the existing tri-state semantics where blank budget fields can mean no budget when that setting is configured.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$3.93`_

### DIFF
--- a/src/common/utils/goals/resolveGoalSetIntent.test.ts
+++ b/src/common/utils/goals/resolveGoalSetIntent.test.ts
@@ -40,10 +40,25 @@ describe("resolveModelGoalSetIntent", () => {
     expect(intent.turnCap).toBe(8);
   });
 
-  test("honors no-budget defaults for model tools without treating null as explicit clear", () => {
+  test("applies the positive budget default even when user-authored goals may omit budgets", () => {
     const intent = resolveModelGoalSetIntent(
       { objective: "ship", budgetCents: null, turnCap: null },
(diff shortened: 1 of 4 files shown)


ThomasK33:
@codex review

PR #3630, opened 2026-06-25
🤖 fix: workflow result chip labels

ThomasK33:
## Summary

Render workflow final-report stat chips as **key** value instead of value key, with the key emphasized as the field label.

## Background

The Workflows tab's structured-output chips were displaying the value first and the key second, which made labels like outcome, target, and issue read backwards.

## Implementation

- Swap the final-report stat chip ordering to key before value.
- Move the bold emphasis from the value to the key.
- Add a focused UI test covering both ordering and emphasis.

## Validation

- `bun test src/browser/features/RightSidebar/Workflows/WorkflowTimeline.test.tsx src/browser/features/RightSidebar/Workflows/workflowDisplay.test.ts`
- `make static-check`

## Risks

Low. This only changes presentation for primitive structured-output stat chips in the Workflows tab final report.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$3.49`_

### DIFF
--- a/src/browser/features/RightSidebar/Workflows/WorkflowTimeline.test.tsx
+++ b/src/browser/features/RightSidebar/Workflows/WorkflowTimeline.test.tsx
@@ -1,0 +1,78 @@
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import { cleanup, render } from "@testing-library/react";
+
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3631, opened 2026-06-25
feat: add HeartbeatToolCall transcript card

ThomasK33:
## Summary

Render the `heartbeat` tool — the agent's recurring, idle-gated self check-in — with a dedicated transcript card (`HeartbeatToolCall`) instead of the generic fallback. The header is a glanceable status pill (live cadence / paused / cleared); expanding reveals a scanning ECG pulse trace, the State/Cadence/Context/Trigger stats, and the check-in prompt.

## Background

The `heartbeat` tool already exists end-to-end (`HeartbeatToolArgsSchema` / `HeartbeatToolResultSchema`, the config modal, the slash command), but agent `heartbeat` tool calls in the transcript fell back to `GenericToolCall` — just raw args/result JSON. This implements the proposed `Heartbeat Tool Call.html` design as a real card so the schedule reads at a glance.

## Implementation

- `HeartbeatToolCall.tsx` mirrors `GetGoalToolCall`: shared `ToolPrimitives`, `useToolExpansion`, and `StatusIndicator`/`getStatusDisplay`. The result is validated against `HeartbeatToolResultSchema` before use, so a malformed persisted transcript degrades to "no settings" rather than throwing (self-healing).
- Registered `heartbeat` in `getToolComponent`, and added the `Activity` (ECG) icon to `TOOL_NAME_TO_ICON` to match the card's pulse-trace motif (the config menu keeps `HeartPulse`).
- Added reduced-motion-gated `heartbeat-dot-pulse` / `heartbeat-trace-scan` keyframes to `globals.css`, following the existing keyframe + utility-class convention.
- Design-fidelity note: the mockup's `text-foreground-secondary` is an undefined token in the app (renders nothing), so muted text maps to the real `text-secondary` / `text-muted`.

## Validation

- Registry wiring test: valid args → `HeartbeatToolCall`; out-of-range `intervalMs` → `GenericToolCall`.
- Interval formatter unit tests (hour/minute boundaries, singular/plural).
- 8 Storybook stories cover the design states. Chromatic snapshots are **disabled** on the file because the repo-wide snapshot budget (`tests/ui/storybook/budget.test.ts`) is already at its ceiling; the stories still render under local Storybook and the CI Storybook test-runner. Flip to a single mode once the budget is raised.

## Risks

Low. Additive — a new entry in the tool-component registry and an icon-map key; no existing tool rendering changes. Worst case for the new card is a degraded/empty detail section on a malformed result, which is handled defensively.

### DIFF
--- a/src/browser/features/Tools/HeartbeatToolCall.stories.tsx
+++ b/src/browser/features/Tools/HeartbeatToolCall.stories.tsx
@@ -63,7 +63,10 @@ export const ScheduledEnabled: Story = {
   },
 };
 
-/** set · long cadence that compacts context before each check-in. */
+/**
+ * set · long cadence that compacts context first, with no custom message — exercises
+ * the default-prompt fallback (the common case, since `message` is only stored when set).
+ */
 export const LongCadenceCompact: Story = {
   args: {
     args: { action: "set", enabled: true, intervalMs: 2 * 3_600_000, contextMode: "compact" },
@@ -73,12 +76,7 @@ export const LongCadenceCompact: Story = {
       success: true,
       action: "set",
       configured: true,
-      settings: {
-        enabled: true,
-        intervalMs: 2 * 3_600_000,
-        contextMode: "compact",
-        message: HEARTBEAT_DEFAULT_MESSAGE_BODY,
-      },
+      settings: { enabled: true, intervalMs: 2 * 3_600_000, contextMode: "compact" },
       summary: "Heartbeat is enabled for this workspace at 2 hours.",
     },
   },
--- a/src/browser/features/Tools/HeartbeatToolCall.stories.tsx
+++ b/src/browser/features/Tools/HeartbeatToolCall.stories.tsx
@@ -221,12 +221,38 @@ export const Executing: Story = {
   },
 };
 
-/** Error · interval outside the supported range. */
-export const ErrorOutOfRange: Story = {
+/**
+ * Error · a reachable failure result. Out-of-range intervals are rejected by the tool
+ * schema before this card renders (they route to GenericToolCall), so the card's ErrorBox
+ * is exercised here with valid args and a server-side failure.
+ */
+export const ErrorResult: Story = {
   args: {
-    args: { action: "set", intervalMs: 30_000 },
+    args: { action: "set", enabled: true, intervalMs: 30 * 60_000 },
     status: "failed",
     defaultExpanded: true,
-    result: { success: false, error: "intervalMs must be between 5 minutes and 24 hours." },
+    result: {
+      success: false,
+      error: "Failed to update heartbeat settings: workspace configuration is unavailable.",
+    },
+  },
+};
+
+/**
+ * Interrupted before the result arrived (no settings/error/executing state). The expanded
+ * body falls back to the requested args instead of going blank — the generic renderer used
+ * to surface the args here.
+ */
+export const Interrupted: Story = {
+  args: {
+    args: {
+      action: "set",
+      enabled: true,
+      intervalMs: 2 * 3_600_000,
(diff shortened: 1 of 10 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed both findings in 576ee052e:

1. **Default prompt now shown.** When no custom `message` is stored (the common `set`/`get` case, since WorkspaceService only persists `message` when explicitly provided), the card renders the built-in `HEARTBEAT_DEFAULT_MESSAGE_BODY` marked "· default" instead of hiding the Check-in prompt section. An explicitly-cleared empty string is treated as "no custom prompt" too. Added story coverage (`LongCadenceCompact` omits `message`).
2. **No signature destructuring.** `HeartbeatToolCall` and its helper components (`HeartbeatBadge`, `PulseTrace`, `HeartbeatStat`) now access `props.field` per the repo component rule.

ThomasK33:
@codex review

Addressed both round-2 findings:

1. **Global default prompt.** The card no longer hardcodes `HEARTBEAT_DEFAULT_MESSAGE_BODY`. When there's no custom `settings.message`, it shows a neutral "Uses the default check-in prompt." note instead of a body it can't know (the effective default may be `config.heartbeatDefaultPrompt`, which isn't in the tool result). Custom messages still render in full.
2. **Premature unset/badge confirmation.** "Recurring check-ins removed", the "No heartbeat is configured" note, and the header badge now gate on a successful result (`success?.action`) rather than the requested action, so an executing/interrupted unset no longer confirms removal before completion.

ThomasK33:
@codex review

Addressed: the custom check-in prompt body now uses `whitespace-pre-wrap break-words` with a `max-h-[200px]` scroll cap (matching `DetailContent`), so multiline messages keep their newlines and long URLs/paths wrap instead of overflowing the card on narrow/mobile widths. Added a `CustomMessageWrapping` story covering a multiline message with a long unbroken token.

ThomasK33:
@codex review

Addressed: `CustomMessageWrapping` is now pinned to a fixed 375px container (the Storybook test-runner renders at desktop width and ignores viewport/Chromatic modes, so the narrow case is forced with a wrapper rather than a mode) and has a `play` that fails if the prompt's long URL/path overflows the container — mirroring the `TodoToolCall` overflow regression test. I also dropped the prompt body's inner `max-h/overflow-y-auto` so horizontal overflow propagates to the asserted container; `break-words` + `whitespace-pre-wrap` remain.

ThomasK33:
@codex review

A max-effort self-review pass surfaced a few more issues; addressed in this push:

1. **No more blank expanded body.** Interrupted/in-flight calls and a degraded `success:true`/`settings:null` result hit no detail branch before — added a `RequestedArgs` fallback that shows what the agent requested (restoring the visibility GenericToolCall gave).
2. **Formatter consolidated + rounding fixed.** Moved the interval formatter to `src/constants/heartbeat.ts` (now shared by the backend summary and the card, killing the duplicate), and it rounds in-range non-whole-minute intervals to the nearest minute instead of emitting a raw `"<ms> ms"` string.
3. **Animation no longer stutters.** Trace offset is now `-243` (a multiple of the 9px dash period) so it scrolls seamlessly across the loop.
4. **Reachable error story.** Out-of-range intervals route to GenericToolCall before this card renders, so the ErrorBox story now uses valid args + a server-side failure; added an `Interrupted` story for the fallback.

ThomasK33:
@codex review

Addressed: `RequestedArgs` now renders while executing too (the executing guard previously suppressed it), so an in-flight heartbeat card shows the requested interval/context/message alongside the "Updating heartbeat settings…" note instead of hiding them.

PR #3634, opened 2026-06-25
🤖 feat: continue interrupted stream by clicking the interrupted splitter

ethanndickson:
## Summary

Clicking the "interrupted" splitter now continues the stream from where it stopped, identical to the existing RetryBarrier "Retry" button and the backend auto-retry path. There is also a keyboard shortcut (Ctrl/Cmd+Shift+Enter) for the same action. Previously the splitter was purely decorative, which left user-initiated (Esc) interrupts with no continue affordance at all — the warning RetryBarrier is intentionally suppressed for `reason: "user"` aborts, so only the subtle "interrupted" line showed and there was no way to resume short of typing a new message.

## Background

When a stream is interrupted, the backend keeps the partial assistant turn in history. Both manual Retry and auto-retry converge on `resumeStream`, which does no history shaping; the model simply continues its partial turn (plus an ephemeral `[CONTINUE]` sentinel injected at transform time, never persisted). The continue behavior already existed — it just wasn't reachable after a user-initiated Esc, because the RetryBarrier only mounts for system/error interrupts.

## Implementation

- Adds `useResumeStream(workspaceId, resetKey?)`, a small one-shot "continue from where it stopped" hook for the divider and its keybind. A user-aborted Esc means "continue once", so the hook never touches the auto-retry preference — auto-retry-enabled users still get backend recovery because the backend consults the persisted preference on failure. `RetryBarrier` is unchanged and keeps its own temporary-auto-retry recovery for system/error interrupts.
- Resume ownership for the divider lives in `ChatPane` (a single `useResumeStream` instance), so the click and keybind paths share one source of truth and one error value, and unmounting a transient divider can't cancel an in-flight retry. `InterruptedBarrier` is a presentational component receiving `resumable`/`onResume`/`error`. The hook's `resetKey` is the resume target message id, so a stale error/spinner resets when the interrupted turn changes (and on workspace switch, since `ChatPane` stays mounted), with a late-resolve guard so a slow resume can't write onto a different turn.
- Only the divider on the current resume target (history tail) is clickable, and only when the workspace is writable (`!transcriptOnly`) and the warning RetryBarrier is suppressed (`!showRetryBarrierUI`). `resumeStream` always continues the tail, so older partial dividers and archived transcripts stay decorative; when RetryBarrier is visible its button owns resume, so the divider can't bypass that recovery.
- Handles the pre-token case: when a user interrupts before the first token, the transcript tail is the user message with no assistant row, so `shouldShowInterruptedBarrier` never marks it. `isPreTokenInterruptedUserTurn(tail, lastAbortReason)` detects this (user/startup abort on a trailing user message) and `ChatPane` marks that message in the divider id set — under the same hydration/auto-retry/streaming suppression — so the stopped request is still continuable.
- `BaseBarrier` gains an optional `onClick`. When provided, only the centered label becomes a `<button>` with a hover underline and pointer cursor; the gradient lines and every other visual stay exactly as before. Barriers without `onClick` (e.g. `EditCutoffBarrier`) are unchanged.
- Adds the `RESUME_STREAM` keybind (Ctrl/Cmd+Shift+Enter, the inverse of the interrupt keybind), wired through `useAIViewKeybinds`. It fires only when a resumable interrupted turn is shown, and a keybind-triggered failure surfaces inline on the divider. The shortcut is listed in Settings > Keybinds; there is no inline hint on the divider, preserving its no-visual-change contract.

## Visual contract

No visual change except: hovering the "interrupted" label underlines it and shows a pointer cursor; a resume failure renders a small inline message below the divider. The gradient lines are inert.

## Risks

Low. The divider is presentational and the keybind only acts on a resumable interrupted turn. `RetryBarrier`'s resume/auto-retry recovery is untouched (no diff against the base). Resume + gating are covered by `useResumeStream`, `useAIViewKeybinds`, `InterruptedBarrier`, and `retryEligibility` tests. The new surface in `BaseBarrier` is the conditional `<button>`, scoped to callers that pass `onClick`.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$6.82`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -1260,7 +1260,13 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
         )}
         {isAtCutoff && <EditCutoffBarrier />}
         {interruptedBarrierMessageIds.has(message.id) && (
-          <InterruptedBarrier workspaceId={workspaceId} />
+          // Only the divider on the resume target (history tail) is clickable;
+          // resumeStream always continues the tail, so older partial dividers
+          // must stay decorative to avoid resuming the wrong turn.
+          <InterruptedBarrier
+            workspaceId={workspaceId}
+            resumable={message.id === lastRetryCandidateMessage?.id}
+          />
         )}
       </React.Fragment>
     );
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -1065,12 +1065,17 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
   // divider on the writable resume target is interactive.
   const { resume: resumeInterruptedStreamAsync, error: resumeInterruptedError } = useResumeStream(
     workspaceId,
-    { autoRetryOnFailure: false }
+    { autoRetryOnFailure: false, resetKey: lastRetryCandidateMessage?.id }
   );
   // Adapt the async resume to the void-returning shape the click/keybind props expect.
   const resumeInterruptedStream = () => void resumeInterruptedStreamAsync();
+  // The divider is the resume affordance only when the warning RetryBarrier is
+  // NOT shown (the user-aborted / suppressed case). When RetryBarrier is visible,
+  // its Retry button owns resume — with temporary auto-retry-on-failure recovery —
+  // so the divider stays decorative to avoid bypassing that recovery.
   const interruptedTailResumable =
     !transcriptOnly &&
+    !showRetryBarrierUI &&
     lastRetryCandidateMessage != null &&
     interruptedBarrierMessageIds.has(lastRetryCandidateMessage.id);
   const transcriptTailItems: TranscriptTailStackItem[] = [];
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -1058,21 +1058,15 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
     }
   }
 
-  // Resume is owned here (not in the divider) so the click and keybind paths
-  // share one instance/error, and so unmounting a transient divider can't cancel
-  // an in-flight retry. autoRetryOnFailure:false keeps the user-abort "continue
-  // once" semantic. resumeStream always continues the history tail, so only the
-  // divider on the writable resume target is interactive.
+  // Owned here so the click and keybind paths share one resume/error. resetKey is
+  // the resume target, so error/spinner reset when the interrupted turn changes.
   const { resume: resumeInterruptedStreamAsync, error: resumeInterruptedError } = useResumeStream(
     workspaceId,
-    { autoRetryOnFailure: false, resetKey: lastRetryCandidateMessage?.id }
+    lastRetryCandidateMessage?.id
   );
-  // Adapt the async resume to the void-returning shape the click/keybind props expect.
   const resumeInterruptedStream = () => void resumeInterruptedStreamAsync();
-  // The divider is the resume affordance only when the warning RetryBarrier is
-  // NOT shown (the user-aborted / suppressed case). When RetryBarrier is visible,
-  // its Retry button owns resume — with temporary auto-retry-on-failure recovery —
-  // so the divider stays decorative to avoid bypassing that recovery.
+  // Resumable only on the writable tail and only when RetryBarrier is suppressed
+  // (user-aborted case). When RetryBarrier is visible, its button owns resume.
   const interruptedTailResumable =
     !transcriptOnly &&
     !showRetryBarrierUI &&
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -1259,7 +1259,9 @@ const ChatPaneContent: React.FC<ChatPaneContentProps> = (props) => {
           />
         )}
         {isAtCutoff && <EditCutoffBarrier />}
-        {interruptedBarrierMessageIds.has(message.id) && <InterruptedBarrier />}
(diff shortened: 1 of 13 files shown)


ethanndickson:
Addressed both Codex P2 findings in d89eba2017.

**P2-1 — Keep resume rollback outside the transient divider** (`InterruptedBarrier.tsx`): the divider now resumes via `useResumeStream(workspaceId, { autoRetryOnFailure: false })`. It no longer enables auto-retry or owns rollback refs, so the unmount that happens once `isAutoRetryActive` becomes true can't cancel an in-flight scheduled retry. A user-initiated Esc means "continue once"; auto-retry-enabled users still get backend recovery because the backend consults the persisted preference on failure regardless.

**P2-2 — Limit clicks to the resumable interrupted turn** (`ChatPane.tsx`): the divider is only clickable when it sits on the current resume target (`message.id === lastRetryCandidateMessage?.id`). Since `resumeStream` always continues the history tail, older partial dividers now stay decorative instead of resuming the wrong turn.

ethanndickson:
@codex review

Addressed both P2 findings (auto-retry rollback ownership + clickable-only-on-resume-target). Please take another look.

ethanndickson:
Addressed in 9364f85358.

**P2 — Surface interrupted-resume failures** (`InterruptedBarrier.tsx`): the resumable divider now renders the hook's `error` inline below it ("Couldn't continue: …"), mirroring `RetryBarrier`'s error treatment. A resume failure before any stream-error (disconnected API, `SendMessageError` from `resumeStream`) is now visible instead of looking like a no-op. The idle/hover appearance is unchanged, and the label stays clickable to try again.

ethanndickson:
@codex review

Addressed the resume-failure feedback finding (errors now surface inline on the resumable divider). Please take another look.

ethanndickson:
Addressed in c4d3d054fb.

**P2 — Disable resume in transcript-only chats** (`ChatPane.tsx`): the divider's `resumable` flag is now gated on `!transcriptOnly`. Archived/no-checkout workspaces (where the composer is already replaced with a read-only notice) render a decorative divider instead of a live continue button, so they stay non-mutating.

ethanndickson:
@codex review

Gated the divider's resume on !transcriptOnly so archived transcripts stay read-only. Please take another look.

ethanndickson:
Addressed in 928bc363b3.

**P2 — Add a shortcut for continuing interrupted responses**: added the `RESUME_STREAM` keybind (Ctrl/Cmd+Shift+Enter, the inverse of the interrupt keybind), wired through `useAIViewKeybinds`. It fires only when a resumable interrupted turn is shown, and a keybind-triggered failure surfaces inline on the divider. It's listed in Settings > Keybinds. There's no inline hint on the divider itself, keeping the no-visual-change contract (so nothing to hide on mobile). Resume ownership moved to `ChatPane` (one `useResumeStream` instance) so the click and keybind paths share one source of truth and error.

ethanndickson:
@codex review

Added a keyboard shortcut (Ctrl/Cmd+Shift+Enter) for continuing interrupted streams and centralized resume ownership. Please take another look.

ethanndickson:
Addressed both findings in 9299d1db22.

**P2 — Reset resume state when workspace changes** (`useResumeStream.ts`): since `ChatPane` owns the hook and stays mounted across workspace switches, `error`/`isResuming` could bleed into the next workspace. The hook now resets them during render on `workspaceId` change (React's "adjust state during render" pattern — no effect), and guards every post-await state update with a `latestWorkspaceIdRef` check so a resume that resolves *after* a switch can't surface its error/spinner on a different workspace.

**P3 — Add the resume shortcut to a visible keybind group** (`KeybindsSection.tsx`): `RESUME_STREAM` is now listed in the `Chat` group next to the interrupt keybinds, so it renders in the Settings → Keybinds reference (it previously only had a label and was filtered out).

ethanndickson:
@codex review

Scoped resume state per-workspace and listed the resume shortcut in Settings > Keybinds. Please take another look.

ethanndickson:
Addressed both findings in 1fcb5b4ddb.

**P2 — Restrict divider resume to suppressed-Retry turns** (`ChatPane.tsx`): the divider's `resumable` flag is now gated on `!showRetryBarrierUI`. When a system/error abort leaves a retryable partial turn, the warning RetryBarrier is shown and its Retry button owns resume (with the temporary auto-retry-on-failure recovery); the divider stays decorative there, so a click can't bypass that recovery. The divider is the resume affordance only in the user-aborted / suppressed case where RetryBarrier is hidden.

**P3 — Clear stale divider resume errors on target change** (`useResumeStream.ts` + `ChatPane.tsx`): the hook now takes a `resetKey` (the resume target message id) and resets transient state when `workspaceId` *or* `resetKey` changes. A failed continue on one interrupted turn no longer shows its "Couldn't continue" error on a later interrupted turn in the same workspace. The late-resolve guard keys on the composite identity too, so a resume that fails after the target changes can't write onto the new one.

ethanndickson:
@codex review

Gated the divider resume to the suppressed-Retry case and scoped its error to the specific interrupted turn. Please take another look.

ethanndickson:
@codex review

ethanndickson:
> Support pre-token interrupted turns

Fixed in `483bd8facc`. Added `isPreTokenInterruptedUserTurn(tail, lastAbortReason)` to `retryEligibility.ts` and, in `ChatPane`, mark the trailing user message in `interruptedBarrierMessageIds` when the turn was interrupted before its first token and RetryBarrier is suppressed (user/startup abort). Both `interruptedTailResumable` and the per-message render key off that set, so the divider button and Ctrl/Cmd+Shift+Enter now continue the stopped request. The set-add is gated by the same hydration/auto-retry/streaming suppression as the partial path, and on `!showRetryBarrierUI` so system aborts still let RetryBarrier own recovery. Covered by new unit tests in `retryEligibility.test.ts`.

ethanndickson:
@codex review

ethanndickson:
@codex review

PR #3637, opened 2026-06-25
🤖 fix: clarify workflow slash command hints

ThomasK33:
Summary

Clarifies the Workflows sidebar empty-state copy and `/workflow` slash-command ghost hint so both describe explicit workflow script-path invocation without implying a special workspace workflow directory.

Background

The previous sidebar text pointed users at `.mux/workflows/`, but workspace-file workflows are not discovered from that folder. They are loaded by explicit `.js` path, while the Workflows tab only lists workflow-bearing skills with a conventional `workflow.js` entry. The slash-command ghost hint also repeated `/workflow` after the user had already typed it.

Implementation

- Added a canonical workflow slash-command argument hint: `<script_path> [args]`.
- Reused that hint for the `/workflow` command usage, ghost text, and Workflows tab helper copy.
- Added a regression test that verifies the ghost hint shows only arguments after `/workflow `.

Validation

- `bun test src/browser/utils/slashCommands/ghostHint.test.ts`
- `bun test src/browser/utils/slashCommands/parser.test.ts`
- `bun x prettier --check src/common/constants/slashCommandHints.ts src/browser/utils/slashCommands/registry.ts src/browser/utils/slashCommands/ghostHint.test.ts src/browser/features/RightSidebar/Workflows/WorkflowEmptyState.tsx`
- `make typecheck`
- `env MUX_ESLINT_CONCURRENCY=1 make lint`
- `env MUX_ESLINT_CONCURRENCY=1 make static-check`
- Dogfooded in an isolated `make dev-server-sandbox` session with Agent Browser; captured screenshot and WebM evidence for the updated Workflows tab copy and slash-command hint.

Risks

Low. This changes display copy and shared command hint constants only; workflow resolution and execution behavior are unchanged.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$11.27`_

### DIFF
--- a/src/browser/features/RightSidebar/Workflows/WorkflowEmptyState.tsx
+++ b/src/browser/features/RightSidebar/Workflows/WorkflowEmptyState.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { BookOpen, Play, Workflow } from "lucide-react";
 
+import { SLASH_COMMAND_HINTS } from "@/common/constants/slashCommandHints";
 import type { AvailableWorkflow, WorkflowArgSummary } from "@/common/types/workflow";
 
 import { WorkflowScopeBadge } from "./WorkflowBadges";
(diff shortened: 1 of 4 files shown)


ThomasK33:
@codex review

ThomasK33:
Addressed Codex finding PRRT_kwDOPxxmWM6MPYUs:

- Updated the workflow ghost-hint regression test to assert the user-visible suffix directly (`<script_path> [args]`).
- Added an explicit guard that the returned hint does not include `/workflow`, preventing the duplicated command regression from returning.
- Re-ran `bun test src/browser/utils/slashCommands/ghostHint.test.ts` and `env MUX_ESLINT_CONCURRENCY=1 make static-check` successfully.

ThomasK33:
@codex review

Please take another look.

PR #3639, opened 2026-06-26
🤖 feat: update skills for background wakeups

ThomasK33:
## Summary

Updates built-in skill guidance so agents can use the new notify-on-terminal background task/workflow behavior without forced awaits when the current decision does not depend on the result.

## Background

PR #3632 changed `run_in_background: true` semantics for tasks, workspace turns, and workflows so background work can finish asynchronously and wake the owning workspace at terminal state. Several built-in skills still emphasized waiting or did not spell out when to rely on that terminal wake-up.

## Implementation

- Clarified `loop` guidance to record background IDs and stop polling when terminal wake-up is the right next event.
- Updated `orchestrate` so background implementation fan-out can stop after recording task IDs when no parent-side setup remains.
- Documented notify-later background behavior for `deep-research`, `/spawn`, and workflow runs while keeping foreground as the default when the result is immediately needed.
- Regenerated the built-in skill bundle.

## Validation

- `bun scripts/gen_builtin_skills.ts check`
- `bun test src/node/services/agentSkills/builtInLoopSkill.test.ts src/node/services/agentSkills/builtInOrchestrateSkill.test.ts src/node/services/agentSkills/builtInSpawnSkill.test.ts src/node/services/agentSkills/builtInWorkflowSkillDescriptions.test.ts src/node/services/agentSkills/agentSkillsService.test.ts`
- `git diff --check`
- `make fmt-check`

## Risks

Low. This is guidance-only and preserves foreground/default behavior for cases where the current answer depends on the result. The main risk is overusing background dispatch; the text explicitly keeps foreground as default for result-dependent work.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$5.77`_


ThomasK33:
@codex review

PR #3649, opened 2026-06-27
🤖 fix: keep heartbeat message editable when disabled

ThomasK33:
## Summary

Keep the workspace heartbeat message editor visible and editable even when heartbeat scheduling is disabled.

## Background

The heartbeat configuration dialog previously hid the custom message textarea when the workspace heartbeat switch was off. That made it hard to inspect or revise the stored heartbeat prompt before re-enabling scheduled follow-ups.

## Implementation

- Always render the message textarea alongside the heartbeat scheduling controls.
- Preserve the roomy two-column layout on desktop and stacked layout on narrow viewports.
- Update modal tests so the disabled state verifies the textarea remains editable and survives enable/disable toggles.
- Add disabled long-message Storybook stories for desktop and mobile viewport coverage.

## Validation

- `bun test src/browser/components/WorkspaceHeartbeatModal/WorkspaceHeartbeatModal.test.tsx`
- Dogfooded the disabled heartbeat dialog in Storybook at 1280×800 and 375×667, including toggling the switch and editing the message while disabled.

## Risks

Low. The change only removes conditional rendering around an existing textarea and updates tests/stories for the new always-visible behavior.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$6.27`_

### DIFF
--- a/src/browser/components/WorkspaceHeartbeatModal/WorkspaceHeartbeatModal.stories.tsx
+++ b/src/browser/components/WorkspaceHeartbeatModal/WorkspaceHeartbeatModal.stories.tsx
@@ -50,16 +50,24 @@ function createWorkspaceContextValue(): WorkspaceContextValue {
   } as unknown as WorkspaceContextValue;
 }
 
-function WorkspaceHeartbeatModalStoryShell(props: { children: ReactNode }) {
-  const settings: HeartbeatSettings = {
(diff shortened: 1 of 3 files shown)


ThomasK33:
@codex review

PR #3652, opened 2026-06-29
feat: add agent_skill_list tool-call card

ThomasK33:
## Summary

Adds a dedicated transcript card for the `agent_skill_list` tool, ported from the "Agent Skill List Tool Call" mockup in the Mux Design System. Collapsed, it reads as a glanceable count ("Listed skills · N skills"); expanded, it groups the discovered skills by scope (project / global / built-in) and lets you open any row to read the full description and how to invoke it.

## Background

`agent_skill_list` previously fell through to the `GenericToolCall` renderer (raw JSON). This brings it in line with the other skill-family cards (`agent_skill_read` / `agent_skill_read_file`) and the session-level `SkillIndicator`, giving the skill-discovery call a first-class, scannable presentation.

## Implementation

- New `AgentSkillListToolCall` built on the shared tool primitives, modeled on the newest card template (`HeartbeatToolCall`).
- **Bound to the real backend shape** (`AgentSkillListToolResult = { success: true; skills } | { success: false; error }`), not the mockup's invented shape. Two exported pure helpers carry the logic:
  - `groupSkillsByScope` — stable scope ordering, drops empty groups.
  - `toSkillListView` — unwraps the SDK JSON container, detects both the thrown `{ success:false, error }` and the nested `{ error }` (code_execution/PTC) failure shapes, and validates the success payload, filtering any malformed skill rows (self-healing).
- Scope colors use existing theme tokens (`--color-task-mode` teal / `--color-ask-mode` indigo / `--color-muted-foreground` neutral); no hardcoded hex.
- Per-row disclosure is a real `<button>` (keyboard-reachable) with 2-line clamped descriptions; unadvertised skills get an `EyeOff` chip and a "call it explicitly" note. Invocation is surfaced with the `$name` inline syntax; scopes are described as project-local / global / built-in.
- Registered in `getToolComponent` with the tool's own schema (bad args still degrade to `GenericToolCall`); added a `LayoutGrid` header icon and the `AgentSkillListToolArgs` type.

### Notable deviation from the mockup

The mockup sketched a "N skills failed to load" / issues panel. The real `agent_skill_list` result carries **no** load diagnostics (the backend logs-and-skips invalid skills, and that surface already lives in `SkillIndicator`), so wiring it here would be dead UI — it is intentionally omitted.

## Validation

- Behavioral unit tests for `groupSkillsByScope` and `toSkillListView` (ordering, malformed-row filtering, JSON-container unwrap, both failure shapes), plus rendered tests for scope grouping, the per-row disclosure interaction, the error branch, and the scope-aware empty state.
- Registration test in `getToolComponent.test.ts`.
- Storybook stories for all states incl. a pinned 375px story with an overflow-guard play; verified each state visually via Storybook + agent-browser screenshots.

> Note: `Test / E2E (macos)` is currently red due to a repo-wide CI infra bug (Node 24 / extract-zip breaking Playwright + Electron installs), unrelated to this card. Fixed separately; this PR will be rebased once that lands.

## Risks

Low. Additive — a new card plus a one-line registry entry, an additive icon-map entry, and a derived args type. Unrecognized/old result shapes self-heal to the generic renderer. No backend or shared-runtime logic changed.

### DIFF
--- a/src/browser/features/Tools/AgentSkillListToolCall.stories.tsx
+++ b/src/browser/features/Tools/AgentSkillListToolCall.stories.tsx
@@ -1,0 +1,191 @@
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import { AgentSkillListToolCall } from "@/browser/features/Tools/AgentSkillListToolCall";
+import { CHROMATIC_DISABLED, lightweightMeta, StoryUiShell } from "@/browser/stories/meta.js";
+import type { AgentSkillDescriptor } from "@/common/types/agentSkill";
+
+const meta = {
+  ...lightweightMeta,
+  title: "App/Chat/Tools/AgentSkillList",
+  component: AgentSkillListToolCall,
+  parameters: {
+    ...lightweightMeta.parameters,
+    // The repo-wide Chromatic snapshot budget (tests/ui/storybook/budget.test.ts) is
+    // already at its ceiling, so these states stay out of paid visual snapshots. They
+    // still render under local Storybook and the CI Storybook test-runner smoke pass.
+    chromatic: CHROMATIC_DISABLED,
+  },
+  decorators: [
+    (Story) => (
+      <StoryUiShell>
+        <div className="bg-background p-6">
+          <div className="w-full max-w-2xl">
+            <Story />
+          </div>
+        </div>
+      </StoryUiShell>
+    ),
+  ],
+} satisfies Meta<typeof AgentSkillListToolCall>;
+
+export default meta;
+
+type Story = StoryObj<typeof meta>;
+
+const BROWSER_DESC =
+  "Browser automation CLI for AI agents. Use when the user needs to interact with websites — " +
+  "navigating pages, filling forms, clicking buttons, taking screenshots, extracting data, " +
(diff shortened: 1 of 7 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Pushed a follow-up commit bumping Playwright 1.56 → 1.61.1 to fix the deterministic `Test / E2E (macos)` install hang (Node 24 + extract-zip, microsoft/playwright#40998). Please re-review the full diff.

ThomasK33:
@codex review

Rebased onto main, which now includes the Node-22 CI fix (#3656). The card diff is unchanged from your prior approval; the rebase just lets `Test / E2E (macos)` run under Node 22. Please re-confirm.

ThomasK33:
Resolved the stale P1 on `bun.lock`: it flagged the Playwright 1.56→1.61.1 bump causing a `@storybook/test-runner` version skew. That bump was **reverted** from this PR — the macOS E2E fix landed separately as #3656 (a Node 22 pin in `setup-mux`, no Playwright change), so there's no longer any `bun.lock`/Playwright delta or test-runner skew here. This PR is now just the `agent_skill_list` card.

PR #3657, opened 2026-06-29
🤖 fix: recover workflow_run after post-create abort

ThomasK33:
## Summary

Recover foreground `workflow_run` calls that abort after durable run creation by returning a model-visible run result instead of a bare error. This gives agents the durable `runId`, preserves active-run rediscovery, and documents the safe recovery path to avoid duplicate workflow starts.

## Background

A foreground workflow can be aborted after its durable run already exists. Previously the tool only surfaced the thrown error, so recovery could miss the still-running workflow and launch another copy. The new behavior makes the durable run the source of truth once `onRunCreated` has fired.

## Implementation

- Track the created workflow run id from `onRunCreated`.
- On post-create errors, fetch and validate the durable run record before returning a structured `workflow_run` result.
- Return the latest durable result if the run already completed; otherwise return `null` and recovery guidance.
- Record active recovered runs as background workflow references so `task_list` can rediscover them.
- Update tool descriptions and generated docs to warn against terminal-only recovery queries.

## Validation

- `bun test src/node/services/tools/workflow_run.test.ts`
- `make typecheck`
- `make lint`
- `make static-check`
- Dogfooded with a disposable harness that simulated a post-create abort, confirmed the returned `runId`/note, verified `task_list` rediscovered the same run, and confirmed `startWorkflow` was invoked exactly once.

## Risks

Low-to-moderate risk in workflow tool error handling. The recovery path only changes behavior after durable run creation is observed; pre-create failures still reject normally. The durable run record is schema-validated before its status/result are trusted.

---

<details>

# Implementation Plan: Prevent Duplicate Workflow Runs After Aborted `workflow_run`

## Context

In workspace `d4987a2655`, two `issue-implementation-loop` / nested `implementation-loop` workflows ran for the same GitHub issue because the first foreground `workflow_run` was aborted after the durable run existed, but the model only saw `Execution aborted`. During recovery it called `task_list` with `statuses: ["failed", "interrupted", "completed"]`, which hid the still-`running` original workflow. It then launched a second copy.

Verified repo facts from inspection:

- `src/node/services/tools/workflow_run.ts` calls `workflowService.startWorkflow(...)` / `startWorkflowInBackground(...)`; `onRunCreated` emits a workflow attached event, but errors from the start call currently propagate without surfacing the already-created `runId`.
- `src/node/services/workflows/WorkflowService.ts` persists a run before foreground execution and supports backgrounding/resume. Foreground abort handling can leave durable state behind, and recovery should use that state rather than a fresh run.
- `src/node/services/tools/task_list.ts` defaults to active statuses (`queued`, `starting`, `running`, `awaiting_report`, `pending`, `backgrounded`) and lists top-level workflow runs, but explicit terminal-only filters hide active workflow runs.
- `src/common/utils/tools/toolDefinitions.ts` documents `workflow_run`, `workflow_resume`, and `task_list`, but currently does not strongly warn against starting a fresh workflow after an abort without first rediscovering active runs.
- `src/node/services/tools/workflow_run.test.ts` already covers foreground starts, background starts, workflow run attachment events, and missing service behavior.

## Goals

1. Make a foreground `workflow_run` abort/error after durable run creation recoverable from the model-visible tool result.
2. Prevent or strongly hinder agents from launching a second identical workflow when a previous matching workflow is still active or resumable.
3. Improve agent-facing guidance in tool descriptions and the relevant skill so the correct recovery path is `task_await` / `workflow_resume`, not `workflow_run` retry.
4. Keep the change minimal and surgical: backend/tool behavior first, prompt/skill guidance second, no broad workflow runtime refactor.

## Non-goals

- Do not redesign workflow execution, nested workflow replay, or lease semantics.
- Do not add a full global duplicate-run registry in the first pass unless review shows it is necessary.
- Do not change nested workflow deterministic IDs; those already dedupe by parent run + step + input hash.
- Do not change GitHub issue watcher parent/orchestrator behavior except for recovery guidance.

## Recommended approach A — tool-level recovery result (primary)

**Net product LoC estimate:** +35 to +65 LoC in `src/node/services/tools/workflow_run.ts` and shared tool definitions. Tests are additional and not counted as product LoC.

### Design

When `workflow_run` creates durable workflow state and then the underlying start call throws, return a structured workflow tool result instead of a bare thrown error whenever possible.

Implementation shape:

1. In `createWorkflowRunTool`, introduce local state:
   - `createdRunId: string | null`
   - `createdRun: unknown | null` if useful for fallback
2. In `onRunCreated`, set `createdRunId = event.runId` before or alongside `emitWorkflowRunAttachedEvent(...)`.
3. Wrap the `startWorkflow` / `startWorkflowInBackground` call in `try/catch`.
4. In the catch path:
   - if `createdRunId == null`, rethrow the original error unchanged;
   - otherwise fetch the durable run with `workflowService.getRun?.({ workspaceId, runId: createdRunId })`;
   - explicitly parse the fetched run with `WorkflowRunRecordSchema.safeParse` or an equivalent existing parser before reading `run.status`;
   - if the run can be fetched and parsed, return `parseToolResult(WorkflowRunToolResultSchema, ...)` with:
     - `status: run.status`
     - `runId: createdRunId`
     - `result`: the latest durable workflow result from `run.events` when `run.status === "completed"`; otherwise `null`
     - `run`
     - `note`: concise recovery guidance that includes the original error message, for example:
       - ``workflow_run errored after creating durable run `wfr_...`: Execution aborted. Do not start another copy; use task_await if active, workflow_resume default mode if interrupted, or workflow_resume with mode='retry_from_checkpoint' for eligible failed runs.``
   - if fetching or parsing the run fails, rethrow/augment the original error with the created `runId`; do not fabricate a status.
5. If the catch path returns an active/detached status (`pending`, `running`, or `backgrounded`), record a background workflow reference when feasible so the run remains rediscoverable and attention-worthy after the tool call stops blocking.
6. Preserve existing successful behavior, including:
   - background run reference recording when status is `running` / `backgrounded`;
   - completed-run note and completed durable result refetch behavior;
   - workflow attached event emission.

### Error/status handling details

- Prefer the durable run record as source of truth for `status` because abort/interrupt races can produce `running`, `backgrounded`, or `interrupted`.
- Do not mark the run terminal or mutate runtime state in the tool catch handler. It should only report already-durable state.
- Do not swallow errors that happen before durable run creation, such as path resolution failures or workflow service absence.
- If `workflowService.getRun` is absent, rethrow with the `createdRunId` in the error message. In the normal app path `getRun` is available; fabricating a status in fallback code would be worse than a clear recoverable error containing the run id.

### Acceptance criteria

- A foreground `workflow_run` whose `startWorkflow` calls `onRunCreated` and then throws returns a model-visible result containing the created `runId`.
- The returned result includes a note with the original error message and tells the agent not to start another copy; active statuses go to `task_await`, interrupted runs go to default `workflow_resume`, and failed runs use `workflow_resume({ mode: "retry_from_checkpoint" })` only when safe/eligible.
- If the fetched durable run is already `completed`, the returned `result` is the latest durable workflow result rather than `null`.
- Errors before durable creation still reject normally.
- Existing background and completed foreground behavior is unchanged.

## Recommended approach B — active duplicate guard (optional second phase)

**Net product LoC estimate:** +70 to +140 LoC if implemented now; +0 LoC if deferred.

### Recommendation

Defer this unless approach A plus guidance is considered insufficient. It is stronger, but it changes semantics for intentional parallel identical workflow runs.

### Possible design if needed

Before creating a new top-level workflow run, scan existing top-level workflow runs in the same workspace and look for a non-terminal run with the same effective identity:

- canonical script path or source hash;
- normalized args;
- same workspace;
- no `parentWorkflow`.

If a matching active run exists, return that run instead of creating a duplicate, with a note explaining that an active matching workflow already exists. If intentional duplicates are a required use case, add an explicit `allow_duplicate` argument to `WorkflowRunToolArgsSchema` and default it to false.

### Risks

- Workflows may intentionally run identical args in parallel for benchmarking or best-of-N experiments.
- Normalizing args for comparison must be stable and must not reintroduce the `input` tokenization pitfalls already documented in repo memory.
- Dedupe based on source hash + args could surprise users who expect a fresh run after a stale but still-active run.

### Acceptance criteria if implemented

- Starting a second active top-level workflow with the same identity returns the existing run instead of creating another run.
- Users/agents have a documented explicit override if intentional duplicates are supported.
- Nested workflows remain unaffected and continue using deterministic child IDs.

## Recommended approach C — agent-facing guidance updates (secondary, do with A)

**Net product LoC estimate:** +10 to +25 LoC in tool descriptions; skill update is outside product code and not counted.

### Tool descriptions

Update `src/common/utils/tools/toolDefinitions.ts`:

1. `workflow_run.description`
   - Add: after a previous `workflow_run` error, abort, timeout, or uncertain result, do not start a fresh run until existing workflow runs have been rediscovered.
   - Direct the model to rediscover existing runs with either:
     - one explicit broad query including `pending`, `running`, `backgrounded`, `interrupted`, `failed`, and `completed`; or
     - two queries: default active `task_list` first, then `interrupted`/`failed`/`completed`.
   - Direct active runs to `task_await`, interrupted runs to default `workflow_resume`, failed runs to eligibility-aware `workflow_resume({ mode: "retry_from_checkpoint" })`, and completed runs to result refetch/inspection rather than a fresh run.
2. `task_list.description` and/or `TaskListToolArgsSchema.statuses.describe(...)`
   - Clarify that omitting `statuses` is the safe recovery default because it includes active workflow runs.
   - Warn that passing only terminal/resumable statuses can hide still-running workflows.
3. `workflow_resume.description`
   - Optionally clarify that `running`/`backgrounded` workflows do not need resume; await them.

### Skill guidance

Update the global `issue-implementation-loop` skill (and any repo-local copy only if intentionally still used) with a “Workflow recovery before retry” rule in child/issue mode:

1. If the adjacent workflow errors, aborts, times out, or the agent is unsure whether it is still running, first rediscover existing workflow runs.
2. Use one broad `task_list` query with `pending`, `running`, `backgrounded`, `interrupted`, `failed`, and `completed`, or run default active discovery first and then a terminal/resumable discovery query.
3. If a matching issue workflow is `pending`, `running`, or `backgrounded`, call `task_await` on its run id.
4. If a matching issue workflow is `interrupted`, call default `workflow_resume`; if it is `failed`, inspect eligibility and use `workflow_resume({ mode: "retry_from_checkpoint" })` only when safe/eligible.
5. If a matching issue workflow is `completed`, inspect/refetch its report rather than rerunning it.
6. Start a fresh `workflow_run` only after confirming no matching active/resumable/completed run exists, or after intentionally terminating/abandoning the old run with user-visible rationale.

Because global skills are not part of the tracked repo, keep PR correctness independent of this skill edit. Update the global skill only if explicitly allowed in the implementation turn, or include the proposed skill patch in final notes as an operational follow-up.

### Acceptance criteria

- Tool descriptions contain explicit recovery guidance for aborted/uncertain workflow starts.
- The issue implementation skill contains an explicit no-fresh-retry-until-rediscovery checklist.
- The guidance specifically prevents the exact bad query pattern: `task_list({ statuses: ["failed", "interrupted", "completed"] })` as the only recovery check.

### Unit tests

1. Add a test in `src/node/services/tools/workflow_run.test.ts`:
   - create a mock `startWorkflow` that invokes `onRunCreated({ runId: "wfr_created_then_aborted", ... })`, then throws `new Error("Execution aborted")`;
   - mock `getRun` to return a `WorkflowRunRecord` with `status: "running"`;
   - assert `tool.execute(...)` resolves, not rejects;
   - assert the result includes `status: "running"`, `runId: "wfr_created_then_aborted"`, `result: null`, `run`, and recovery `note`;
   - assert `readAgentWorkflowRunReferences(...)` / `agent-workflow-runs.json` contains the active run so it remains rediscoverable after the aborted foreground call detaches.
2. Add a test that errors before `onRunCreated` still reject normally.
3. Add a cheap completed-run variant if implementation cost stays low: post-create throw + fetched `completed` run returns the latest durable result instead of `null`.
4. If adding duplicate guard, add tests for:
   - existing active matching run is returned;
   - terminal matching run does not block a new run;
   - `allow_duplicate: true` (if added) creates a fresh run.

### Type/lint validation

Run targeted validation first, then broader validation if touched code is small:

```bash
bun test src/node/services/tools/workflow_run.test.ts
make typecheck
make lint
```

If resources are tight, use the repo memory guidance:

```bash
MUX_ESLINT_CONCURRENCY=1 make lint
```

Before claiming completion, run the highest practical aggregate check:

```bash
make static-check
```

If `make static-check` is unavailable or blocked by unrelated environment issues, record the exact blocker and include successful targeted checks.

### Backend/tool behavior dogfood

Create a disposable workflow or test harness that starts a foreground workflow, ensures the durable run is created, then simulates an abort/error. Verify the model-visible result includes the run id and recovery note.

Evidence to collect:

- terminal transcript or asciinema cast showing:
  - the workflow start;
  - the aborted/error result;
  - the returned `runId`;
  - `task_await` or `task_list` rediscovering the same run instead of starting a fresh copy;
- attach the cast or converted recording if the environment supports it.

### UI/workflow sidebar smoke check

If the implementation changes visible workflow run attachment behavior or result rendering, run a Mux dev server/sandbox and inspect the workflow run card:

1. Start the appropriate dev sandbox using existing repo skills/scripts.
2. Trigger the simulated aborted-after-create workflow.
3. Confirm the workflow appears once in the Workflows/sidebar UI with the same run id.
4. Capture screenshots and a short video recording for reviewer evidence.
5. Check narrow/mobile viewport if any UI layout changes are touched; otherwise note that no UI code changed.

If only backend/tool code and descriptions are touched, UI screenshots are optional, but a terminal recording remains the preferred dogfood artifact.

## Rollout order

1. Add the `workflow_run` created-run recovery path.
2. Add unit tests for post-create throw and pre-create throw behavior.
3. Update tool descriptions to steer recovery correctly.
4. Optionally update the global `issue-implementation-loop` skill once product-code behavior is validated.
5. Run targeted tests and static validation.
6. Dogfood the recovery path with a disposable workflow/harness and capture evidence.
7. Only consider the stronger duplicate guard if tests/dogfood/advisor review indicate approach A is not enough.

## Risks and mitigations

- **Risk:** Returning structured results for thrown foreground starts could mask a real failure.
  - **Mitigation:** Only do this after durable run creation; include an explicit note that the start errored and the run must be awaited/resumed based on durable status.
- **Risk:** Durable run status may be stale during an abort race.
  - **Mitigation:** Use `workflowService.getRun` after the catch; do not infer status from the caught error.
- **Risk:** Agents ignore guidance and still launch duplicates.
  - **Mitigation:** Approach A gives the model the actual `runId`; guidance tells it exactly what to do. Add optional duplicate guard if this remains insufficient.
- **Risk:** Global skill edits are not tracked with the product PR.
  - **Mitigation:** Treat skill edits as operational guardrails; product correctness lives in tool behavior and tests.

## Recommended final scope for first PR

Implement approaches A and C only:

- `workflow_run` returns recoverable structured output after durable creation + thrown start.
- Tool descriptions explicitly document aborted-workflow recovery.
- Unit tests cover the failure mode, including active-run background reference recording.
- Skill update can be done separately or included as a local/global operational change if requested.

This scope is minimal, directly addresses the observed incident, and avoids changing semantics for intentional duplicate workflows.

</details>

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$23.74`_


ThomasK33:
@codex review

ThomasK33:
Addressed Codex feedback `PRRT_kwDOPxxmWM6M-eii`: recovered `pending` workflow runs are now treated as resumable via `workflow_resume` rather than awaitable active work, and they are not recorded as background workflow references. Added test coverage for the pending recovery path.

ThomasK33:
@codex review

Please take another look.

PR #3658, opened 2026-06-29
🤖 fix: show workspace lifecycle display names

ThomasK33:
## Summary

Show the workspace display name in `task_workspace_lifecycle` transcript rows, while keeping the workspace ID and originating `wst_...` task ID visible as secondary identifiers.

## Background

Archived workspaces disappear from the active sidebar, but the lifecycle tool UI previously surfaced only stable IDs. That made it hard to map an archive result back to the workspace title the user saw in the sidebar.

## Implementation

- Add optional `displayName` support to workspace lifecycle tool result rows.
- Populate `displayName` from workspace metadata (`title` first, then `name`, then the task-turn title fallback).
- Render the display name as the primary row label and preserve copyable IDs on the secondary line.
- Extend UI and service tests plus Storybook fixture data for the new display shape.

## Validation

- `bun test src/browser/features/Tools/WorkspaceLifecycleToolCall.ui.test.tsx`
- `bun test src/node/services/tools/task_workspace_lifecycle.test.ts`
- `bun test src/node/services/taskService.test.ts --test-name-pattern "workspace lifecycle"`
- `make typecheck`
- `make fmt-check`
- Dogfooded with an `agent-browser` screenshot/video capture against a temporary Vite component harness at a narrow viewport.

## Risks

Low. Existing transcripts without `displayName` continue to render the prior ID-first fallback, and the schema field is optional.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$21.03`_

### DIFF
--- a/src/browser/features/Tools/WorkspaceLifecycleToolCall.stories.tsx
+++ b/src/browser/features/Tools/WorkspaceLifecycleToolCall.stories.tsx
@@ -46,9 +46,24 @@ export const ArchiveMultiple: Story = {
     defaultExpanded: true,
     result: {
       results: [
-        { status: "archived", action: "archive", workspaceId: "24e33167af" },
-        { status: "archived", action: "archive", workspaceId: "4a92f76fbf" },
-        { status: "archived", action: "archive", workspaceId: "0b71c40e21" },
+        {
+          status: "archived",
+          action: "archive",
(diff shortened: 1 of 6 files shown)


ThomasK33:
@codex review

PR #3659, opened 2026-06-29
🤖 fix: simplify nested workflow sidebar

ThomasK33:
## Summary

Simplifies the right-sidebar nested workflow presentation so a nested-only workflow step no longer renders an empty details card or horizontal divider. The nested workflow still keeps a single bordered panel for grouping.

## Background

Nested workflow steps do not expose the same parent-level controls as regular step details, so the extra details wrapper added visual noise without adding functionality.

## Implementation

- Detect whether a workflow step has actual detail content before rendering the generic step-details card.
- Render nested-only workflow content directly under the parent row.
- Keep one subtle border around the nested workflow panel for visual grouping.

## Validation

- `bun test src/browser/features/RightSidebar/Workflows/WorkflowTimeline.test.tsx`
- `make typecheck`
- `make fmt-check`
- Dogfooded the Storybook nested workflow story with `agent-browser`, expanded the nested step, and captured screenshot/video evidence.

## Risks

Low risk: this is a scoped presentation change in the workflows timeline renderer. Existing nested workflow polling and disclosure behavior is unchanged.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$7.05`_

### DIFF
--- a/src/browser/features/RightSidebar/Workflows/WorkflowTimeline.test.tsx
+++ b/src/browser/features/RightSidebar/Workflows/WorkflowTimeline.test.tsx
@@ -389,6 +389,46 @@ describe("WorkflowTimeline", () => {
     expect(view.getByText("running · 0/1 steps · child-phase")).toBeDefined();
   });
 
+  test("keeps completed nested workflow duration out of the generic details card", async () => {
+    const client = {
+      workflows: {
+        getRun: () => Promise.resolve(makeChildWorkflowRun()),
+      },
+    };
+    const completedParentView = makeNestedWorkflowParentView();
+    completedParentView.phases[0].done = 1;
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

ThomasK33:
Addressed the Codex finding by excluding nested workflow rows from the metadata-only details-card path, and added a regression test for completed nested workflow duration.

@codex review

ThomasK33:
Addressed the second Codex finding by ignoring stored nested workflow result payloads for the parent step details-card decision. The nested child timeline remains the source of visible result/report details, and the regression test now includes the structured result that production records.

@codex review

PR #3661, opened 2026-06-29
🤖 fix: recover task_list workspace-turn statuses

ThomasK33:
## Summary

Fixes stale `task_list` output for workspace-turn tasks by making list results use the same restart-recovery normalization as `task_await` before applying status filters.

## Background

Workspace-turn handles can be persisted as `interrupted` during restart recovery even when their child workspace history later proves the turn completed. `task_await` already self-healed those handles, but `task_list` read the raw handle rows and could keep showing recovered workspaces as interrupted, encouraging agents to queue unnecessary continuation messages.

## Implementation

- Added a shared workspace-turn record normalization path in `TaskService`.
- Routed both `getWorkspaceTurnSnapshot` and `listWorkspaceTurnTasks` through that path.
- Changed `listWorkspaceTurnTasks` to repair handles before applying the caller's status filter, so a repaired `completed` handle no longer appears in `interrupted`-only lists.
- Added a regression test for restart-interrupted deferred handles listed through `listWorkspaceTurnTasks`.

## Validation

- `bun test src/node/services/taskService.test.ts -t "listWorkspaceTurnTasks repairs restart-interrupted deferred handles before filtering"`
- `bun test src/node/services/taskService.test.ts -t "workspace-turn stale recovery repairs restart-interrupted deferred handles after descendants stop blocking"`
- `bun test src/node/services/taskService.test.ts -t "listWorkspaceTurnTasks"`
- `bun test src/node/services/tools/task_list.test.ts`
- `make typecheck`
- `MUX_ESLINT_CONCURRENCY=1 make static-check` after rebasing onto current `origin/main`
- `git diff --check`

## Dogfooding

Backend/tooling-only change; no visual UI path changed. Dogfooded via the targeted task-list and workspace-turn recovery tests above, which exercise the same backend listing path used by the `task_list` tool.

## Risks

Low-to-moderate risk in workspace-turn task accounting: listing now reads all owner handles before filtering so stale rows can self-heal. The behavior is intentional, but it means listing may do recovery work even for terminal-status filters. The touched path is limited to workspace-turn task handles.

---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `$8.42`_

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -3181,6 +3181,57 @@ describe("TaskService", () => {
     expect(workspaceMocks.remove).toHaveBeenCalledWith("childworkspace", true);
   });
 
+  test("listWorkspaceTurnTasks repairs restart-interrupted deferred handles before filtering", async () => {
+    const { config, parentId, taskService, historyService } = await startWorkspaceTurnForTest();
+    const muxMetadata = {
+      type: "workspace-turn-task" as const,
+      taskHandleId: "wst_handle",
+      ownerWorkspaceId: parentId,
+      turnId: "turn",
+    };
+    const appendResult = await historyService.appendToHistory(
+      "childworkspace",
+      createMuxMessage("msg_recovered_list", "assistant", "Recovered list text", {
+        model: "anthropic:claude-opus-4-6",
+        agentId: "exec",
+        finishReason: "stop",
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3663, opened 2026-06-30
🤖 fix: stop background monitor from re-waking on already-shown output

ethanndickson:
## Summary

A background bash `monitor` could wake the agent about output that had already been delivered inline, producing a redundant synthetic wake immediately after the read returned (e.g. a `task_await` on a bash task returns the matching FAIL line, then a wake about that same line fires the instant the await resolves).

## Background

`monitor` matches and the agent's own reads (`task_await` / `bash_output`) run on independent paths through `BackgroundProcessManager`. The monitor scans output and emits `monitor:match`; agent reads advance a read cursor via `getOutput()`. Nothing connected the two, so when a match flush was deferred (cooldown window) and the agent read the same bytes inline in the meantime, the deferred flush still fired and double-reported output the agent had already seen.

## Implementation

Suppression is decided at the single emit chokepoint, `emitMonitorMatch`, by comparing two absolute file byte offsets:

- `monitor.matchedThroughOffset` — the end offset of the last complete line that produced a match. Tracked per matched line (computed by walking the chunk's newlines from its start offset), so a match is never inflated by later or unmatched output in the same poll.
- `proc.shownThroughOffset` — the end offset of the last complete line actually delivered to the agent by an **unfiltered** `getOutput` read. It advances in exactly one place and only when no `filter`/`filter_exclude` is in play, so a filtered read (which may drop matched lines) never counts as having shown them.

If `shownThroughOffset >= matchedThroughOffset`, the matched output was already delivered inline and the pending match is dropped instead of emitted. Because both sides are absolute file offsets, the comparison is order-independent — there is no race between the reader advancing its cursor and the monitor recording its match.

This replaces an earlier, fragile approach that reused the raw read cursor (`outputBytesRead`) as the "agent has seen this" signal. That cursor advances past content the agent is not actually shown — buffered unterminated fragments, and lines filtered out of a read — which is what made the naive comparison leak wakes in several edge cases.

The change is upstream of `WorkspaceService`: once a wake is emitted, all existing queue/drain/tool-end delivery behavior is unchanged. Genuinely-new matches and fire-and-forget monitors (whose shown offset stays at 0) wake exactly as before.

## Risks

Low. The new state is monitor-local and the suppression only narrows emission in the already-shown case. The main behaviors to preserve are covered by tests: drop after an unfiltered read past the match; still wake for a line only buffered unterminated (matched on exit); still wake when later/unmatched output shares the poll; and still wake when a filtered read advanced the cursor without showing the match.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$15.86`_

### DIFF
--- a/src/node/services/backgroundProcessManager.test.ts
+++ b/src/node/services/backgroundProcessManager.test.ts
@@ -451,6 +451,45 @@ describe("BackgroundProcessManager", () => {
       manager.off("monitor:match", handler);
     });
 
+    it("anchors matchedThroughOffset to the matched line, not later output in the same chunk", async () => {
+      // A matched line followed by a non-matching complete line in one poll: matchedThroughOffset
+      // must point at the end of the matched line (byte 8 of "ERR foo\n"), not the end of the whole
+      // complete region (byte 16 of "...nomatch\n"). Otherwise an agent that read only the matched
+      // line stays below the inflated offset and gets a redundant wake.
+      const result = await manager.spawn(
+        runtime,
+        testWorkspaceId,
+        "printf 'ERR foo\\nnomatch\\n'; sleep 5",
+        {
+          cwd: process.cwd(),
+          displayName: "monitor-matched-line-anchor",
+          monitor: {
+            filter: "ERR",
+            pattern: /ERR/,
+            exclude: false,
+            cooldownMs: 10_000,
+          },
+        }
+      );
+      expect(result.success).toBe(true);
+      if (!result.success) return;
+
+      // Wait until both lines have been scanned (cursor at 16, no fragment buffered).
+      let proc = await manager.getProcess(result.processId);
+      for (let attempt = 0; attempt < 60; attempt++) {
+        proc = await manager.getProcess(result.processId);
+        if (
+          (proc?.monitor?.lastReadOffset ?? 0) >= 16 &&
+          (proc?.monitor?.pendingLines.length ?? 0) > 0
+        )
+          break;
+        await new Promise((resolve) => setTimeout(resolve, 25));
+      }
+      expect(proc?.monitor?.pendingLines ?? []).toEqual(["ERR foo"]);
+      expect(proc?.monitor?.matchedThroughOffset).toBe(8);
+      expect(proc?.monitor?.lastReadOffset).toBe(16);
+    });
+
     it("strips ANSI before matching and emitting matched lines", async () => {
       const eventPromise = waitForMonitorMatch(manager);
       const result = await manager.spawn(
--- a/src/node/services/backgroundProcessManager.test.ts
+++ b/src/node/services/backgroundProcessManager.test.ts
@@ -395,6 +395,62 @@ describe("BackgroundProcessManager", () => {
       expect(event.payload.lines).toEqual(["ERR boom"]);
     });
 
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

Addressed the P2 (buffered unterminated match). The suppression now compares the monitor scan offset against the *shown* offset — `proc.outputBytesRead - byteLength(proc.incompleteLineBuffer)` — instead of the raw read cursor. `getOutput` advances `outputBytesRead` past an unterminated trailing line it only buffers, so subtracting that still-buffered fragment keeps the buffered-but-unshown case below the scan offset and its on-exit match still wakes. Added a regression test (`still wakes when the matched line was only buffered (unterminated), not shown`).

ethanndickson:
@codex review

Addressed the second P2 (complete matched line followed by a trailing fragment). The monitor now tracks `matchedThroughOffset` — the byte offset through the end of the last complete line that produced a match, computed as `lastReadOffset - byteLength(rawTrailingFragment)` at record time — and suppression compares the agent's shown offset against that instead of the raw scan cursor. Both sides now use end-of-complete-line semantics, so the comparison is exact: a matched line the agent has been shown drops even when an unterminated fragment trails it, while a line still buffered unterminated (matched only on exit) stays below the shown offset and still wakes. Added a regression test (`drops the wake for a matched line even when a trailing fragment follows it`); confirmed it fails against the previous `lastReadOffset` comparison.

ethanndickson:
@codex review

Addressed the third P2 (per-line vs per-chunk offset). `processMonitorContent` now computes each complete line's absolute end offset by walking `content`'s newlines from the chunk's start offset, and `recordMonitorMatch` is given that specific line's end rather than the chunk's complete-region end. So a matched line followed by later complete non-matching output in the same poll gets `matchedThroughOffset` anchored at the matched line — an agent that read only that line is suppressed exactly, with no redundant wake. The chunk-start offset is passed explicitly from `monitorTailLoop` (the pre-read `lastReadOffset`), avoiding any reliance on `byteLength(content)`. Added a regression test (`anchors matchedThroughOffset to the matched line, not later output in the same chunk`); confirmed it fails against the previous per-chunk offset.

ethanndickson:
@codex review

Addressed the filtered-read P2 — and stepped back to fix the root cause rather than patch another offset edge.

The fragility came from overloading `outputBytesRead` as the "agent has seen this" signal: `getOutput` advances it past content it doesn't actually show (buffered fragments, and lines dropped by a `filter`/`filter_exclude` read). That's why each round surfaced a new edge.

New design: a dedicated `proc.shownThroughOffset` = the file offset through the end of the last complete line delivered by an **unfiltered** read. It advances in exactly one place (`getOutput`'s normal return, guarded by `!filter`), so a filtered read never counts as showing output. `emitMonitorMatch` compares it against the matched line's end offset (`matchedThroughOffset`); both are absolute file offsets, so suppression is order-independent — no race between the reader and the monitor, and the buffered/fragment/per-chunk/filtered cases all collapse out of the comparison instead of each needing bespoke accounting.

Added a regression test (`still wakes when a filtered read advanced past but did not show the matched line`): a deferred `ERR` match, a `filter="DONE"` read, then a terminate-triggered flush still wakes; confirmed it fails if the `!filter` guard is removed.

PR #3668, opened 2026-07-01
🤖 fix: allow heartbeat turns to set goals

ThomasK33:
Summary
- Allow heartbeat-triggered agent turns to opt into model-created goals so heartbeat prompts can instruct agents to call `set_goal`.

Background
- Heartbeats are backend-generated synthetic turns. They were missing `allowAgentSetGoal`, so `AIService` treated them like one-shot hosts and withheld `set_goal` even when the heartbeat prompt asked the agent to seed a bounded goal.

Implementation
- Set `allowAgentSetGoal: true` in heartbeat send options.
- Covered normal heartbeat dispatch plus compact/reset heartbeat follow-up preservation.
- Existing goal-tool availability still gates `set_goal` to top-level exec-like agents, and heartbeat eligibility still excludes child workspaces.

Validation
- `bun test src/node/services/heartbeatService.test.ts`
- `make static-check`

Risks
- Low. The change is limited to heartbeat-initiated sends and still flows through existing goal-tool availability checks.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `1168530{MUX_COSTS_USD:-unknown}`_

### DIFF
--- a/src/node/services/heartbeatService.test.ts
+++ b/src/node/services/heartbeatService.test.ts
@@ -620,6 +620,7 @@ describe("HeartbeatService", () => {
         message?: string;
       }
       interface HeartbeatSendOptions {
+        allowAgentSetGoal?: boolean;
         muxMetadata?: {
           type?: string;
           source?: string;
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3669, opened 2026-07-01
🤖 fix: prevent bash monitor wake redelivery

ThomasK33:
## Summary

Fix bash monitor wake delivery so accepted synthetic wake messages are marked delivered as soon as they are persisted to chat history. This prevents pre-stream provider startup failures from repeatedly appending the same background-monitor wake to the workspace transcript.

## Background

Workspace `f5d85266cf` showed one persisted bash monitor wake with a handful of matched lines, but hundreds of duplicate synthetic user wake messages. The wake record stayed pending after the synthetic message was accepted because delivery waited for provider `stream-start`; later error/drain events retried the same pending wake and appended duplicates.

## Implementation

`WorkspaceService.drainBashMonitorWakes()` now treats chat-history acceptance as the delivery boundary. If startup fails after acceptance, AgentSession's existing startup retry resumes the accepted turn instead of relying on bash-monitor wake re-enqueue. Rejected sends before acceptance still leave the wake pending for retry.

## Validation

- `make static-check`
- `make typecheck`
- `bun test src/node/services/workspaceService.test.ts --test-name-pattern "WorkspaceService bash monitor wakes"`
- `make fmt-check`

## Risks

Low-to-medium risk, limited to bash monitor wake auto-delivery. The main semantic change is that an accepted synthetic wake is considered delivered before provider streaming starts; this relies on existing accepted-turn startup retry, which is already set before the acceptance callback runs.

---

_Generated with `mux` • Model: `openai:gpt-5.5` • Thinking: `xhigh` • Cost: `1731989{MUX_COSTS_USD:-unknown}`_

### DIFF
--- a/src/node/services/workspaceService.test.ts
+++ b/src/node/services/workspaceService.test.ts
@@ -278,6 +278,67 @@ describe("WorkspaceService bash monitor wakes", () => {
     }
   });
 
+  test("marks an accepted wake delivered when stream startup fails before provider start", async () => {
+    const { config, cleanup } = await createTestHistoryService();
+    try {
+      const workspaceId = "bash-monitor-accepted-startup-failure";
+      const projectPath = path.join(config.rootDir, "project");
+      await config.addWorkspace(projectPath, {
+        id: workspaceId,
+        name: workspaceId,
+        projectName: "project",
+        projectPath,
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3671, opened 2026-07-02
🤖 feat: encourage codifying prose skills as one-off inline workflows

ThomasK33:
## Summary

Encourages the model to codify prose-described processes (skills, instruction blocks, plans that read like workflows) into one-off inline `script_source` workflows instead of executing every phase in-context, when the dynamic-workflows experiment is enabled.

## Background

Several skills are effectively workflow descriptions — ordered phases, loops, parallel lanes, verification gates — but ship no packaged `workflow.js`. Executing them fully in-context tends to drift from the documented process as context grows. Existing guidance actively biased against inline workflows ("use script_source **only** for **small** one-off conductors"; "Choose a workflow **only** when…"), which also meant the `script_source` path saw little organic use while the experiment is trying to gather signal.

## Implementation

Three surfaces, no system-prompt change (the `workflow_run` tool description is only visible when the dynamic-workflows experiment is enabled, so the nudge is experiment-gated for free):

- **`workflow_run` tool description + `script_source` field** (`toolDefinitions.ts`): states the trigger — prose-described multi-phase/looping/multi-agent process with no packaged workflow script → prefer codifying it as a one-off `script_source` workflow (durable checkpoints, resume, fresh delegated context per phase).
- **`loop` built-in skill**: new mechanism-chooser row routing "skill/instructions describing a multi-phase process" to a one-off `script_source` workflow; workflow route softened from "only when" to a positive trigger signal. The "do not create a workflow just because the prompt says loop" guardrail is retained.
- **`workflow-authoring` built-in skill**: new "Codifying prose-described processes" section with a fit check (every phase must be expressible as sub-agent delegation + patch integration; processes needing the parent's working tree/uncommitted state stay in-context) and a promotion path (inline conductor → `./workflows/<name>.js` → skill-packaged `workflow.js` once it proves reusable).

`builtInSkillContent.generated.ts` and `docs/hooks/tools.mdx` are regenerated output.

## Risks

Low. Prompt/guidance-only change; no runtime behavior touched. Main failure mode would be over-triggering (wrapping one-shot tasks in workflow scaffolding), mitigated by keeping the trigger specific to prose-described multi-phase processes, retaining the existing anti-overkill guardrails, and adding the delegation fit check.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$4.87`_


ThomasK33:
@codex review

PR #3672, opened 2026-07-03
🤖 fix: keep kickoff-window goals active through synthetic bash-monitor wake turns

ThomasK33:
## Summary

Background bash-monitor wake turns could permanently pause a freshly set workspace goal (Goal panel stuck at "ACTIVE (PAUSED)", kickoff never fires). This fixes the goal kickoff-window reconciliation so synthetic wake turns — and any concurrent `getGoal()` read — can no longer strand a goal.

## Background

When a background bash `monitor` matches output, Mux wakes the workspace by sending a synthetic user message. That delivery path was already correct (`synthetic: true`, chat-tail-neutral, same as heartbeats and sub-agent report wakes). The actual bug was a two-step failure in `WorkspaceGoalService` during the kickoff window:

1. **Kickoff-window flip.** Activating a goal (model `set_goal` or user Resume) arms a `kickoff` continuation candidate, but the first `goal_continuation` history row has not landed yet — the chat tail still ends at the pre-goal manual user row. Any `getGoal()` during that window (the wake turn's tool build guarantees one) ran chat-tail reconciliation and flipped the goal `active → paused`.
2. **Candidate deletion.** At the wake turn's stream end, `requestContinuationAfterStreamEnd`'s kickoff protection required persisted status `"active"`; seeing `"paused"` it deleted the armed kickoff candidate, permanently stranding the goal.

## Implementation

- `readChatTailGoalMode` now tags *why* the tail resolved to paused: explicit `goal-pause-boundary` row vs ordinary manual user row.
- `applyChatTailGoalMode` skips only the **implicit** `active → paused` downgrade (`pausedBy: "manual_user"`) while an armed kickoff candidate matches `goal.goalId`. Explicit pauses are unaffected: every explicit pause path deletes the candidate first and appends a pause-boundary row, which is deliberately not suppressed.
- `requestContinuationAfterStreamEnd`'s kickoff protection now also accepts a `paused` kickoff goal instead of deleting the candidate — mirroring existing semantics where eligibility/dispatch tolerate paused-kickoff and `recordContinuationFired` flips the goal back to active.

## Validation

- 3 new regression tests simulate a bash-monitor wake turn during the kickoff window (synthetic wake row + `getGoal` + wake-turn stream end). All 3 **fail without the fix** (verified via `git stash`) and pass with it.
- Full goal suites (`workspaceGoalService`, `agentSession.goalAutoPause`, `workspaceService.goalDefaults` — 160 tests), `bashMonitorWakeStore`, and `workspaceService` monitor-wake tests: green.

## Risks

Low-moderate; touches goal status reconciliation. The suppression is narrowly scoped: only `active → paused`, only for the implicit manual-user-tail case, and only while a kickoff candidate with a matching `goalId` is armed. Kickoff candidates cannot keep a goal artificially active: explicit pause, manual user messages, and user stops all delete the candidate before/at reconciliation, and dispatch/stream-end paths clear or replace it. Worst regression case would be a goal staying active slightly longer during the kickoff window — previous behavior (silent permanent pause) was strictly worse.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$16.04`_

### DIFF
--- a/src/node/services/workspaceGoalService.test.ts
+++ b/src/node/services/workspaceGoalService.test.ts
@@ -774,6 +774,150 @@ describe("WorkspaceGoalService", () => {
     });
   });
 
+  // Regression: bash-monitor wake turns used to disable freshly set goals.
+  // The wake turn's synthetic user row lands before the kickoff continuation
+  // row, so chat-tail reconciliation saw the pre-goal manual user row and
+  // flipped the goal active→paused mid-window.
+  test("getGoal keeps a kickoff-window goal active while a synthetic wake turn runs", async () => {
+    await appendUserHistoryMessage(historyService, workspaceId, "Set yourself a goal and continue");
+    const busy = true;
+    const dispatcher = new IdleDispatcher();
+    const execute = mock(() => Promise.resolve(true));
+    service.registerGoalContinuationConsumer(dispatcher, {
+      ...continuationBridge(execute),
+      getRuntimeState: () => ({ isRuntimeCompatible: true, isBusy: busy }),
+      getKickoffSendOptions: () => ({ model: "openai:gpt-4o", agentId: "exec" }),
+    });
+
+    const goal = await setGoalOk(service, {
+      workspaceId,
+      objective: "Survive a wake turn",
+      status: "active",
+      initiator: "model",
+    });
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3673, opened 2026-07-06
🤖 fix: set Linux app name so windows group under mux.desktop

ethanndickson:
## Summary

Give Mux windows a stable Linux desktop identity so they group under the pinned `mux.desktop` launcher instead of spawning a separate, generic "Electron" taskbar entry — and contain that identity so apps launched *from* Mux (terminals, custom editors) don't inherit it.

## Background

On Linux, Electron derives the X11/XWayland `WM_CLASS` and native-Wayland `app_id` from the app name, normally read from `package.json`. Launch modes that don't expose our `package.json` (notably the Nix package, which runs `electron dist/cli/index.js`) fall back to the built-in `Electron` default, so the window doesn't match `mux.desktop` (`StartupWMClass=mux`) and KDE/GNOME open a second taskbar entry with the wrong icon. Verified live on KDE Plasma 6: the Mux window reported `resourceName=electron` / `resourceClass=Electron` and userData resolved to `~/.config/Electron`.

## Implementation

- `src/desktop/main.ts`: before the ready event on Linux, `app.setName("mux")` (sets `WM_CLASS`, which XWayland/X11 groups on) and `CHROME_DESKTOP=mux.desktop` (sets the native-Wayland `app_id`; same env var `app.setDesktopName` writes).
- `package.json`: `"desktopName": "mux.desktop"` — the upstream-recommended declarative identity for packaged builds (same approach as Signal), and guards against a future `productName` case change breaking the default-derived name.
- Containment (Codex findings): `CHROME_DESKTOP` is stripped in `sanitizeMuxChildEnv` (covers bash/pty children), and the two Linux paths that hand control to user GUI apps — native terminal launch and custom editor launch — now spawn with a sanitized env. Without this, a Chromium/Electron app launched from a Mux terminal would group under Mux's launcher; packaged builds had this leak already since Electron always sets `CHROME_DESKTOP` when it can read `package.json`.
- macOS (`open`/`osascript`) and Windows paths are unaffected: the identity block is Linux-gated.

## Risks

Low, Linux-only. Setting the app name also moves `userData`/crash dumps from `~/.config/Electron` to `~/.config/mux` on builds that previously fell back to the default name (dev and Nix), aligning them with packaged builds; one-time renderer `localStorage` reset on those installs. Packaged builds already resolving the name to `mux` are unaffected.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$6.19`_

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "mux",
   "version": "0.27.0",
+  "desktopName": "mux.desktop",
   "description": "mux - coder multiplexer",
   "author": "Coder",
   "main": "dist/cli/index.js",
--- a/src/desktop/main.ts
+++ b/src/desktop/main.ts
@@ -63,8 +63,10 @@ import {
 // generic "Electron" entry. Launch modes that don't expose our package.json
 // (e.g. the Nix package's `electron dist/cli/index.js`) otherwise fall back to
 // the "Electron" default. setName sets WM_CLASS (XWayland matches
-// StartupWMClass=mux); CHROME_DESKTOP sets the native-Wayland app_id. Must run
-// before the ready event.
+// StartupWMClass=mux); CHROME_DESKTOP sets the native-Wayland app_id (same var
+// Electron sets from package.json desktopName when it can read it). Must run
+// before the ready event. sanitizeMuxChildEnv strips CHROME_DESKTOP from child
+// processes so apps launched from mux terminals don't inherit our identity.
 if (process.platform === "linux") {
   app.setName("mux");
   process.env.CHROME_DESKTOP = "mux.desktop";
--- a/src/desktop/main.ts
+++ b/src/desktop/main.ts
@@ -59,6 +59,17 @@ import {
   shell,
 } from "electron";
 
+// Pin the Linux app name so the window groups under mux.desktop instead of a
+// generic "Electron" entry. Launch modes that don't expose our package.json
+// (e.g. the Nix package's `electron dist/cli/index.js`) otherwise fall back to
+// the "Electron" default. setName sets WM_CLASS (XWayland matches
+// StartupWMClass=mux); CHROME_DESKTOP sets the native-Wayland app_id. Must run
+// before the ready event.
+if (process.platform === "linux") {
+  app.setName("mux");
+  process.env.CHROME_DESKTOP = "mux.desktop";
+}
+
 // Enable local crash dump collection so renderer SIGSEGV produces a minidump
 // at ~/.config/mux/Crashpad/completed/*.dmp — no data leaves the machine.
 // Must be called as early as possible, before app.whenReady().
--- a/src/node/runtime/childProcessEnv.ts
+++ b/src/node/runtime/childProcessEnv.ts
@@ -1,10 +1,15 @@
 import * as path from "node:path";
 import { getMuxHome } from "@/common/constants/paths";
 
-const BROWSER_ENV_KEYS_TO_STRIP = [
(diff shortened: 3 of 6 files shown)


ethanndickson:
@codex review

Addressed the P2 in 54f52f9cf: `CHROME_DESKTOP` is now stripped in `sanitizeMuxChildEnv` (the shared choke point both `LocalBaseRuntime` and `spawnPtyProcess` already funnel through), with test coverage. Note the leak predated this PR for packaged builds — Electron's init script always calls `setDesktopName` (default `${app.name}.desktop`) when it finds a package.json — so the strip also fixes existing packaged installs. Also added `desktopName: "mux.desktop"` to package.json as the declarative identity for packaged builds.

ethanndickson:
@codex review

Addressed: `openNativeTerminalLinux` now spawns with `env: sanitizeMuxChildEnv({ ...process.env })`. That was the one remaining path handing control to user apps with an unsanitized env — the macOS branches go through `open`/`osascript` (children get launchd's env, and `CHROME_DESKTOP` is Linux-only), Windows is unaffected, and the remaining spawn sites (git/coder CLIs) don't launch user GUI apps. Keeping the var in `process.env` itself is required: Chromium reads `CHROME_DESKTOP` at window creation (including later terminal pop-out windows), and packaged Electron always sets it anyway — so sanitizing the user-app launch paths is the correct containment.

ethanndickson:
@codex review

Addressed: the custom editor spawn in `editorService.ts` now uses `env: sanitizeMuxChildEnv({ ...process.env })`. I also audited all remaining `detached: true` spawn sites: the other three in `terminalService.ts` are the macOS (`open`/`osascript`) and Windows (`cmd`) branches, where `CHROME_DESKTOP` is never set (Linux-only), and `shell.openExternal` is Electron-internal with no env override (identical behavior to every packaged Electron app). All Linux user-app launch paths now sanitize.

ethanndickson:
@codex review

No change needed here — both premises of this finding don't hold against Electron 40 source (v40.10.0 tag):

1. **`app.setName()` does set the Linux WM_CLASS.** `shell/browser/native_window_views.cc` builds it at window creation from `Browser::Get()->GetName()`:
   ```cpp
   std::string name = Browser::Get()->GetName();
   params.wm_class_name = base::ToLowerASCII(name);
   params.wm_class_class = name;
   ```
   and `GetApplicationName()` (`shell/common/application_info_linux.cc`) resolves "attempt #1: the string set in app.setName()". The docs note that `setName` "does not affect the name that the OS uses" refers to the OS-displayed application name, not WM_CLASS. This was also verified empirically on a live KDE Wayland session: with `setName`, the window reports `resourceClass=mux` and groups under the pinned launcher.

2. **`app.setDesktopName("mux.desktop")` is byte-for-byte equivalent to what we already do.** `App::SetDesktopName` (`shell/browser/api/electron_api_app.cc:964`) is a three-line wrapper: `env->SetVar("CHROME_DESKTOP", desktop_name)`. Our `process.env.CHROME_DESKTOP = "mux.desktop"` sets the same variable at the same pre-ready time. Additionally, `setDesktopName` is not present in Electron 40.9.3's public `electron.d.ts`, so calling it would require a type cast that this repo's TypeScript discipline forbids.

So in the Nix/dev launch path: `setName("mux")` covers WM_CLASS (X11/XWayland grouping) and the `CHROME_DESKTOP` assignment covers the native-Wayland app_id — the two mechanisms this comment asks for are both already in place.

PR #3678, opened 2026-07-06
🤖 feat: add setting to disable sidebar age grouping

ibetitsmike:
## Summary

Adds a "Group sidebar workspaces by age" toggle in Settings → General → Appearance (default on). When switched off, the sidebar skips the collapsible "Older than 1/7/30 days" tiers and renders every workspace as one flat recency-sorted list.

## Background

The age tiers keep long-lived projects tidy, but some users prefer seeing all workspaces at once without expanding tiers. This makes the grouping opt-out while preserving current behavior as the default.

## Implementation

- New `SIDEBAR_AGE_GROUPING_KEY` persisted UI pref (`src/common/constants/storage.ts`), read in `ProjectSidebar` with `{ listener: true }` so toggling in Settings applies live.
- When disabled, `renderAgeTiers` bypasses `partitionWorkspacesByAge` and routes all workspaces through the existing recent-row path with full-length empty buckets, preserving task-group coalescing, connector geometry, and tier-index assumptions. `expandedOldWorkspaces` state is untouched, so re-enabling restores prior expand/collapse state.
- Settings UI: a `Switch` row next to the other appearance toggles in `GeneralSection`.

## Validation

- New Storybook story `FlatListWhenAgeGroupingDisabled` asserts old workspaces render inline and no "Older than" toggle exists when the pref is off; the two existing tier-dependent stories now pin the pref on so state cannot leak between runs. Storybook test runner: 22/22 LeftSidebar stories pass.
- Manually dogfooded against a sandboxed dev server with workspaces backdated 2/9/45 days: verified default tiers, live flatten on toggle-off, live restore on toggle-on, persistence across reload, tier expansion state surviving an off/on cycle, and 375px mobile rendering.

## Risks

Low. The change is a display-only branch in the sidebar render path; grouping-on behavior is byte-identical to before, and the pure `partitionWorkspacesByAge` util is unchanged.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$45.00`_

### DIFF
--- a/src/browser/components/LeftSidebar/LeftSidebar.stories.tsx
+++ b/src/browser/components/LeftSidebar/LeftSidebar.stories.tsx
@@ -109,6 +109,9 @@ function resetStorybookPersistedStateForStory(): void {
   if (typeof localStorage !== "undefined") {
     localStorage.removeItem(SELECTED_WORKSPACE_KEY);
     localStorage.setItem(UI_THEME_KEY, JSON.stringify("dark"));
+    // FlatListWhenAgeGroupingDisabled writes this key; clear it so later
+    // stories are not affected by story execution order.
+    localStorage.removeItem(SIDEBAR_AGE_GROUPING_KEY);
   }
 }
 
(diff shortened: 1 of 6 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

PR #3680, opened 2026-07-06
🤖 feat: add Claude Mythos 5 as a first-class known model

ammar-agent:
## Summary

Registers Claude Mythos 5 (`anthropic:claude-mythos-5`) as a first-class known model, making it selectable in the model picker and reachable via the `mythos` alias.

## Background

The Fable 5 work already landed all Mythos-class plumbing for `claude-mythos-5`: thinking policy (thinking can't be disabled), native 1M context classification, display-name formatting ("Mythos 5"), and token/pricing metadata in `models-extra.ts` ($10/$50 per MTok, 1M context, 128K output). But the model was never added to the curated registry, so users with Project Glasswing / trusted access could not select it without typing a custom model string.

Mythos 5 shares Fable 5's specs and pricing; it ships with safeguards lifted in some areas and is restricted to approved trusted-access customers (no self-serve sign-up).

## Implementation

- Single `MYTHOS` entry in `MODEL_DEFINITIONS` (`src/common/constants/knownModels.ts`), alias `mythos`, same Opus 4.5 tokenizer override as Fable.
- Not `warm`: most users can't access it, and the shared tokenizer override is already warmed via `FABLE`.
- No default refusal-fallback chain (unlike Fable → Opus): Mythos runs with safeguards lifted, and `DEFAULT_MODEL_FALLBACKS` is seeded exactly once so a new entry wouldn't reach existing configs anyway.
- `docs/config/models.mdx` and `builtInSkillContent.generated.ts` regenerated via `make fmt`.

## Validation

- Existing registry-wide tests cover the new entry automatically (token metadata presence, alias/id uniqueness); targeted suites for knownModels, modelDisplay, models, thinking policy, and normalizeModelInput all pass.
- `make static-check` green (including docs sync).

## Risks

Low: additive registry entry; all Mythos-specific behavior paths were already in place and tested from the Fable 5 rollout.

---

### DIFF
--- a/docs/config/models.mdx
+++ b/docs/config/models.mdx
@@ -14,6 +14,7 @@ Mux ships with curated models kept up to date with the frontier. Use any custom 
 | Model                  | ID                            | Aliases                                                      | Default |
 | ---------------------- | ----------------------------- | ------------------------------------------------------------ | ------- |
 | Fable 5                | anthropic:claude-fable-5      | `fable`                                                      |         |
+| Mythos 5               | anthropic:claude-mythos-5     | `mythos`                                                     |         |
 | Opus 4.8               | anthropic:claude-opus-4-8     | `opus`                                                       | ✓       |
 | Sonnet 5               | anthropic:claude-sonnet-5     | `sonnet`                                                     |         |
 | Haiku 4.5              | anthropic:claude-haiku-4-5    | `haiku`                                                      |         |
--- a/src/common/constants/knownModels.ts
+++ b/src/common/constants/knownModels.ts
@@ -36,8 +36,10 @@ const MODEL_DEFINITIONS = {
     providerModelId: "claude-fable-5",
     aliases: ["fable"],
     warm: true,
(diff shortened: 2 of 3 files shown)


ammar-agent:
@codex review

ammar-agent:
Re the tokenizer-undercount finding (P2, `knownModels.ts:54`): a current tokenizer isn't available — `anthropic/claude-opus-4.5` is the newest Anthropic encoding published in `ai-tokenizer`; no Opus 4.7+/Fable-generation tokenizer exists upstream. Reuse-closest is the established convention for every post-4.5 Anthropic entry in this registry (Sonnet 5, Opus 4.8, Fable 5), and Mythos 5 simply mirrors Fable 5 (same underlying model). An adjusted-estimate multiplier would be a new counting mechanism affecting all four models, out of scope for this additive registry entry.

Addressed in bb65d85 by correcting the stale "tokenization is unchanged across the 4.x / Mythos line" claim on FABLE and documenting the ~1.0-1.3x undercount on both FABLE and MYTHOS overrides, so the limitation is explicit at the sink that drives context budgeting.

@codex review

PR #3681, opened 2026-07-06
🤖 feat: compact expandable rendering for bash monitor wake messages

ammar-agent:
## Summary

Bash monitor wake turns ("A background bash monitor matched output.") now render as a compact card in the transcript — a title, one summary line per monitor (`process · /filter/`), and a "Show details" toggle that reveals the full raw prompt — instead of dumping the entire multi-line model prompt into the user bubble.

## Background

Background bash monitors wake the agent with a synthetic user message built by `buildBashMonitorWakePrompt`. That prompt is model-facing plumbing (matched output blockquotes, `task_await` guidance, restart notices) and previously rendered verbatim in chat, taking up a large chunk of the transcript for what is essentially a one-line event.

## Implementation

- New `bash-monitor-wake` member in the `MuxMessageMetadata` union carrying compact per-record summaries (`kind`, `displayName`, `filter`, `filterExclude`). `buildBashMonitorWakeMetadata` builds it next to the prompt builder, and `drainBashMonitorWakes` stamps it on the synthetic send. The full prompt stays in the message text for the model; metadata only drives display.
- `displayedMessageBuilder` surfaces it as `bashMonitorWake` on the user `DisplayedMessage`; `UserMessage` branches to a new `BashMonitorWakeMessageContent` (patterned after `GoalSyntheticMessageContent`) with a collapsed-by-default `aria-expanded` details toggle, plus a dedicated "monitor wake" pill instead of the generic "auto" pill.
- Old wake messages without the metadata keep the previous full-text rendering (no migration needed).

## Validation

- New Storybook story `BashMonitorWakeMessages` (match + monitor-lost cards; play expands the match card so the snapshot covers both states).
- Unit tests: `MessageRenderer.test.tsx` asserts the prompt body is hidden until expanded and re-hidden on collapse; `workspaceService.test.ts` asserts the stamped metadata including the displayName→processId fallback.
- `make static-check`, targeted `bun test` suites pass locally.

## Risks

Low. Display-only branch keyed off new metadata; wake delivery/merge semantics untouched. Worst case is a wake card rendering with the old full-text layout.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `high` • Cost: `$11.77`_

### DIFF
--- a/src/browser/features/Messages/BashMonitorWakeMessageContent.tsx
+++ b/src/browser/features/Messages/BashMonitorWakeMessageContent.tsx
@@ -1,0 +1,72 @@
+import { useState, type ReactElement } from "react";
+import { ChevronRight, Radar } from "lucide-react";
+import { cn } from "@/common/lib/utils";
+import type { BashMonitorWakeDisplayRecord } from "@/common/types/message";
+
+interface BashMonitorWakeMessageContentProps {
+  /** Full wake prompt text (matched lines, task_await guidance) shown when expanded. */
+  content: string;
+  records: BashMonitorWakeDisplayRecord[];
+}
+
+function describeRecord(record: BashMonitorWakeDisplayRecord): string {
+  const inverted = record.filterExclude ? " (inverted)" : "";
+  const lost = record.kind === "monitor-lost" ? " — monitor lost" : "";
+  return `${record.displayName} · /${record.filter}/${inverted}${lost}`;
+}
+
+/**
+ * Compact card for synthetic bash-monitor wake turns. The raw prompt is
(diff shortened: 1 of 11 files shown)


ammar-agent:
@codex review

ammar-agent:
Addressed the P1: `displayedMessageBuilder` now validates the `bash-monitor-wake` records shape (`Array.isArray` + per-record field checks via `getValidBashMonitorWakeRecords`) and falls back to the normal full-text user rendering when metadata is malformed, so corrupted persisted metadata can't crash transcript rendering. Covered by new tests in `displayedMessageBuilder.bashMonitorWake.test.ts`.

ammar-agent:
@codex review

PR #3685, opened 2026-07-07
🤖 fix: let attach_file accept any file type with display-only fallback

ethanndickson:
## Summary

`attach_file` now accepts any file type. Images, SVG, and PDF are still forwarded to the model as real attachments; every other file type (text, source, diffs, logs, video, audio, archives, etc.) is shown to the user in chat for preview/download while the model receives a clear notice that the contents were not forwarded.

This PR also pins the `mintlify` CLI in `check-docs-links` to unbreak `static-check-full` / the merge queue (see Background).

## Background

`attach_file` previously hard-rejected text files it could not turn into a model attachment, throwing `Unsupported attachment type`. This surfaced when an agent tried to hand the user a `.diff` file: the only workaround was renaming it to `.md` to squeeze through the markdown display-only path. The rejection was a deliberate guardrail (steer agents toward `file_read` for text contents) but it blocked the legitimate "give the user a downloadable artifact" use case for any non-markdown text file.

Separately, the merge queue's `Static Checks` job runs `make static-check-full`, which includes `check-docs-links` (regular PR CI runs only `make static-check` and skips it). That step invoked `bun x mintlify` unpinned, resolving to `mintlify@4.2.681`, whose transitive dependency `@mintlify/auth-edge@0.0.36` was unpublished from npm and now 404s. Resolution failed and the merge queue's static checks errored out even though the PR's own checks were green. This is an external breakage affecting any PR reaching `static-check-full`.

## Implementation

- `readAttachmentFromPath.ts`: the `mediaType == null` branch no longer throws. It always falls through to a display-only result after enforcing the size cap. A new `resolveDisplayMediaType` picks the media type by precedence: caller override, then known audio/video/markdown extensions, then byte-sniffing (`text/plain` vs `application/octet-stream`). The now-dead text-extension/media-type reject sets and the `requireBinaryContent` sniff-reject were removed.
- `attach_file.ts`: the display-only notice returned to the model now states plainly that only images/SVG/PDF are sent as attachments, that this file's contents were not forwarded, and to use `file_read` for contents.
- `toolDefinitions.ts`: tool description rewritten to describe the accept-any-file behavior and the two tiers.
- `Makefile`: pin `check-docs-links` to `mintlify@4.2.680` (last known-good; `4.2.681` pulls the unpublished `@mintlify/auth-edge`). Comment notes to bump once a fixed release ships.
- The 10MB size cap remains the only hard rejection. The image/SVG/PDF real-attachment path is unchanged.

## Risks

Low. The attach_file change only affects the previously-rejected branch of a single tool; the real-attachment path and size cap are untouched. Frontend rendering (`DisplayOnlyFile`) already handles arbitrary media types with a download affordance, so no UI change was needed. The Makefile pin is a version constraint on a dev/CI-only docs-links check and does not affect shipped code.

---

_Generated with `mux` • Model: `anthropic:claude-opus-4-8` • Thinking: `xhigh` • Cost: `$3.46`_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -501,7 +501,9 @@ check-docs-links: ## Check documentation for broken links
 	@echo "🔗 Checking documentation links..."
 	# Workaround: katex@0.16.34 ships broken ESM with unreplaced __VERSION__ placeholder.
 	# Remove this NODE_OPTIONS prefix once katex publishes a fixed build.
-	@cd docs && NODE_OPTIONS="$${NODE_OPTIONS:+$$NODE_OPTIONS }--import data:text/javascript,globalThis.__VERSION__=%220.16.34%22" bun x mintlify broken-links
+	# Pin mintlify: 4.2.681 pulls @mintlify/auth-edge@0.0.36, which was unpublished
+	# from npm (404) and breaks resolution. Bump once a fixed release ships.
+	@cd docs && NODE_OPTIONS="$${NODE_OPTIONS:+$$NODE_OPTIONS }--import data:text/javascript,globalThis.__VERSION__=%220.16.34%22" bun x mintlify@4.2.680 broken-links
 
 check-code-docs-links: ## Validate code references to docs paths
 	@./scripts/check-code-docs-links.sh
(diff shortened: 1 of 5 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

PR #3683, opened 2026-07-06
🤖 feat: pin chats to the top of the sidebar

ibetitsmike:
## Summary

Adds chat pinning: workspaces can be pinned to the top of the project sidebar in a stable, user-controlled order that is immune to recency reshuffling.

## Background

Sidebar ordering is recency-based, so long-running or reference chats constantly drift down as other workspaces see activity. Pinning gives important chats a fixed spot at the top of their project.

## Implementation

- `pinnedAt` timestamp on `WorkspaceConfigSchema` / `WorkspaceMetadataSchema`; shared `isWorkspacePinned` / `isWorkspacePinnable` helpers (`src/common/utils/pin.ts`). Only root chats can be pinned; sub-agents and archived workspaces cannot.
- `WorkspaceService.setPinned` with a server-generated monotonic timestamp (`max(now, maxPinInProject + 1ms)`) so pin order is append-only and stable even with client clock skew; exposed via `workspace.setPinned` ORPC route. Archiving clears `pinnedAt`; unarchive does not restore it.
- Sorting: pinned roots sort first by `pinnedAt` ascending (id tie-break) in `buildSortedWorkspacesByProject`; `partitionWorkspacesByAge` treats pinned roots (and their children via tier inheritance) as recent so pins never fall into the aged bucket.
- Frontend applies an optimistic `pinnedAt` on click (mirroring the server's append-only ordering) so rows reorder instantly; the metadata subscription reconciles and failures revert.
- UI: Pin/Unpin chat item in the sidebar row menu and workspace menu bar, pin icon on pinned rows, `Ctrl/Cmd+Alt+P` keybind, and a Pin/Unpin Current Chat palette action.

## Validation

Full UAT against a `dev-server-sandbox` instance driven with agent-browser (16 scenarios): pin/unpin via row menu, menu bar, keybind, and palette; stable append-below-existing-pins order; new workspaces landing below pins; persistence across reload; `pinnedAt` cleared on archive and not restored on unarchive; two-client live sync (pin in client A instantly reorders client B); pin icon rendering at 375px mobile width with no overflow. Rapid-pinning four chats via the row menu reorders the sidebar immediately on each click (verified with no-wait DOM snapshots) and converges with the server's persisted order.

## Risks

[...]

PR #3687, opened 2026-07-07
🤖 fix: sidebar not updating when pin chat doesn't reorder rows

ibetitsmike:
## Summary

Pinning a chat persisted `pinnedAt` on the backend but the sidebar often showed no change: no pin icon, row menu stuck on "Pin chat", and `Ctrl+Alt+P` could not unpin.

## Background

Follow-up to #3683. The bug triggers whenever pinning does not reorder rows, which is common: new pins append at the bottom of the pinned block, so pinning the row directly below it (or the only row in a project) is an order no-op.

Two causes:

1. `getWorkspaceSidebarKey()` omitted `pinnedAt`, so `useStableReference` in `App.tsx` kept the stale `sortedWorkspacesByProject` map whenever pin state changed without changing order.
2. `WorkspaceStore.syncWorkspaces()` never refreshed metadata for existing workspaces, so the `Ctrl+Alt+P` keybind read a stale `pinnedAt` and kept re-sending `pinned=true` instead of toggling.

## Implementation

- Add `pinnedAt` to the sidebar key in `src/browser/utils/workspace.ts`.
- Refresh metadata for existing workspaces in `WorkspaceStore.syncWorkspaces()` so keybind toggles read current pin state.

## Validation

- Unit tests added for both causes (sidebar key changes with `pinnedAt`; `syncWorkspaces` refreshes existing metadata).
- Verified live twice against the sole-workspace repro (zero reorder): once via dev-server sandbox, and once end-to-end over the production remote path (Docker container, HTTP + WS, second observer client with no optimistic update). Pre-fix build reproduces the bug; this build shows the pin icon instantly in both acting and observer clients, and `Ctrl+Alt+P` toggles both ways with `config.json` following each toggle.

## Risks

Low. `syncWorkspaces` now replaces metadata for existing workspaces on every sync; this is the same data the backend just sent, so the only behavior change is dropping stale local copies. Sidebar keys gain one field, which can only invalidate a memoized map that should have been invalidated.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$42.07`_

### DIFF
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -1692,6 +1692,19 @@ describe("WorkspaceStore", () => {
       expect(global.window.localStorage.getItem(autoRetryKey)).toBeNull();
     });
 
+    it("refreshes metadata for existing workspaces", () => {
+      const metadata = makeWorkspaceMetadata("workspace-1");
+      store.syncWorkspaces(new Map([[metadata.id, metadata]]));
+      expect(store.getWorkspaceMetadata("workspace-1")?.pinnedAt).toBeUndefined();
+
+      // Same workspace id, updated field (e.g. pinned after initial load).
+      // Imperative readers like the pin keybind must see the fresh value.
+      const pinned = { ...metadata, pinnedAt: "2026-01-01T00:00:00.000Z" };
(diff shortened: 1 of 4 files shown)


ibetitsmike:
@codex review

PR #3696, opened 2026-07-08
🤖 feat: queue messages behind special sends instead of erroring (FIFO message queue)

ammar-agent:
## Summary

Sending a message while an internal workspace-turn follow-up (or another metadata-bound send) was queued failed with a hard error dialog. This PR restructures `MessageQueue` from a single merged batch into FIFO dispatchable entries, so special sends own their turn and later messages queue behind them without errors. It also keeps backend-initiated queue work non-disruptive: background monitor wakes are hidden from/restored separately from the user's composer, and live foreground bashes are moved to background before a monitor wake dispatches after the current step.

## Background

`MessageQueue` previously merged every queued message into one combined send (joined text, one metadata object, one callback set). That forced hard rejections for sends that must stay attached to exactly one dispatch, including workspace-turn follow-ups, agent-skill invocations, and compaction requests.

The FIFO change made background and user entries able to coexist, exposing two related UX risks that this PR also addresses:

- synthetic/background queue entries could appear as editable queued input or be restored over an unsent user draft;
- a queued background monitor wake used tool-end dispatch, which could soft-stop the stream while a foreground bash was still attached and abort that process instead of preserving its work.

## Implementation

- `MessageQueue` now holds ordered `QueueEntry` items and dispatches one entry at a time with `dequeueNext()`.
- Plain user follow-ups still batch into the newest compatible open entry. Compaction, agent-skill, workspace-turn, callback-carrying, and ownership-changing sends start separate entries so metadata, callbacks, and origin never leak between turns.
- All previous `Cannot queue ...` throw paths are removed.
- User-visible queue projections expose only user-authored entries. Synthetic/background entries remain in the real FIFO for dispatch and completion bookkeeping but do not render an editable queue card or restore into the composer.
- Queue events retain an all-entry `hasQueuedMessages` signal so background follow-up bookkeeping remains correct even when the renderer has no user-visible queued message.
- `restore-to-input` clears/cancels the complete queue but emits only user-authored text/files/reviews. Restore events are workspace-scoped so a background workspace cannot overwrite the currently mounted composer.
- Bash monitor wakes still use tool-end (send-after-step) dispatch, but first background any live foreground bash tool calls. The process keeps running under background tracking instead of being killed by the ensuing stream abort.
- Interrupting a queued workspace turn removes only its matching FIFO entry, preserving unrelated user messages.

## Validation

- Regression coverage for FIFO ordering and metadata/callback isolation.
- Regression coverage for ownership-separated entries and user-visible queue projections.
- Regression coverage proving synthetic background entries do not surface in queue UI or restore over user input.
- Regression coverage proving streaming monitor wakes send foreground bashes to background.
- Targeted suites: 516 tests passed across `messageQueue`, `agentSession.queueDispatch`, `workspaceService`, and `WorkspaceStore`.
- `make typecheck` passed.
- `make static-check` passed.

## Risks

Medium: this changes the hot path for mid-stream sends, queue rendering/restoration, and bash-monitor wake dispatch. Existing batching semantics remain for compatible user messages; special sends and ownership boundaries are isolated into FIFO entries. Synthetic entries are hidden only from user-editable projections, while all-entry queue state is preserved for dispatch and response-completion logic. Monitor wake behavior matches the existing user send-after-step UX by migrating foreground bash work before soft-stopping the stream.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `high` • Cost: `$41.31`_

### DIFF
--- a/src/browser/features/ChatInput/index.tsx
+++ b/src/browser/features/ChatInput/index.tsx
@@ -1741,7 +1741,15 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
         mode?: "append" | "replace";
         fileParts?: FilePart[];
         reviews?: ReviewNoteDataForDisplay[];
+        workspaceId?: string;
       }>;
+
+      if (
+        customEvent.detail.workspaceId != null &&
+        workspaceIdForComposerClear !== customEvent.detail.workspaceId
+      ) {
+        return;
+      }
 
       const { text, mode = "append", fileParts, reviews } = customEvent.detail;
       const restoredIdPrefix = `restored-${Date.now()}`;
@@ -1781,7 +1789,15 @@ const ChatInputInner: React.FC<ChatInputProps> = (props) => {
     window.addEventListener(CUSTOM_EVENTS.UPDATE_CHAT_INPUT, handler as EventListener);
     return () =>
       window.removeEventListener(CUSTOM_EVENTS.UPDATE_CHAT_INPUT, handler as EventListener);
-  }, [appendText, restoreText, restoreDraft, applyDraftFromPending, getDraft, editingMessageForUi]);
+  }, [
+    appendText,
+    restoreText,
+    restoreDraft,
+    applyDraftFromPending,
+    getDraft,
+    editingMessageForUi,
+    workspaceIdForComposerClear,
+  ]);
 
   useEffect(() => {
     const handler = (event: CustomEvent<{ workspaceId: string }>) => {
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -4405,6 +4405,42 @@ describe("WorkspaceStore", () => {
     });
   });
 
+  it("queued message identity changes when a partial FIFO drain changes visible content", async () => {
+    const workspaceId = "queued-message-identity";
+    mockChatStreamFor(workspaceId, async function* () {
+      yield {
+        type: "queued-message-changed",
+        workspaceId,
+        hasQueuedMessages: true,
+        queuedMessages: ["first", "second"],
+        displayText: "first\nsecond",
+      };
+      yield { type: "caught-up", hasOlderHistory: false };
+      await tick(25);
+      yield {
+        type: "queued-message-changed",
+        workspaceId,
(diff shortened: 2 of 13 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the P2: `interruptWorkspaceTurn` now uses a targeted `removeQueuedWorkspaceTurn(workspaceId, handleId, ...)` path that removes only the matching workspace-turn entry (and notifies its cancel callback), leaving unrelated queued user messages pending. Covered by a new `MessageQueue.removeWorkspaceTurn` unit test.

ammar-agent:
@codex review

Updated the queue fix to keep background work non-disruptive:
- synthetic/background entries no longer appear in or restore over the user's composer;
- restore events are scoped to the owning workspace;
- bash monitor wakes background live foreground bashes before tool-end dispatch, matching send-after-step semantics and preserving in-flight work.

Local targeted tests, typecheck, and `make static-check` are green.

ammar-agent:
@codex review

Addressed the send-now priority issue: the queued-message card now promotes the oldest user-authored entry ahead of hidden synthetic/background entries before dispatch. Hidden work stays queued and resumes through the normal drain lifecycle.

Added a regression test for user/background/user ordering; targeted tests, typecheck, and `make static-check` pass.

ammar-agent:
@codex review

Addressed the partial-drain send-now guard issue: queued card identity now derives from the visible queue projection, so dispatching one FIFO entry changes the id and resets the in-flight guard for the remaining card.

Added a WorkspaceStore regression test; targeted tests, typecheck, and `make static-check` pass.

ammar-agent:
@codex review

Addressed the queue-card id allocation issue: identity now includes only lightweight attachment/review metadata (media type, filename, URL length, file/line), never attachment data URLs or review bodies.

WorkspaceStore tests, typecheck, and `make static-check` pass.

ammar-agent:
@codex review

Addressed the attachment-id performance concern: queued card identity now includes only lightweight attachment/review metadata (media type, filename, URL length, file path, line range), never serialized data URLs or review payloads.

WorkspaceStore regression test, typecheck, and `make static-check` pass.

PR #3697, opened 2026-07-08
🤖 fix: distinct waiting-on-bash-monitor indicators in chat and sidebar

ammar-agent:
## Summary

When a turn ends while a background bash monitor is still armed, both the chat pane and the workspace sidebar now show a distinct "waiting on monitor" state instead of looking like active streaming (or going blank), so users don't mistake the wake-waiting state for a running or finished conversation.

## Background

Background bash processes spawned with a `monitor` block wake the workspace when matching output lands. Previously:

- The **chat pane** hid the streaming barrier as soon as the stream ended — the transcript looked idle even though the agent was deliberately parked waiting for a wake.
- The **sidebar** funneled armed monitors into the same `"active"` visual state as live streams, so a waiting workspace showed the identical pulsing green dot as an actively streaming one.

Both read as "still streaming" or "done," causing user confusion.

## Implementation

**Chat pane**

- `StreamingBarrier` gains a `waiting-on-monitor` phase, entered when no stream is active (`!canInterrupt`, not starting) and `WorkspaceState.activeBashMonitorCount > 0` — the count is already plumbed from `BackgroundProcessManager` via workspace activity snapshots, so no backend/IPC changes were needed.
- The phase renders informational copy ("agent wakes on matching output") in place of the Stop control, hides token stats, and is not debounced so the stream-end → waiting handoff is immediate.
- `ChatPane` keeps the barrier mounted while a monitor is armed (`shouldMountStreamingBarrier`); reveal/hydration logic is unchanged.

**Sidebar**

- `StatusDot` gains a `"waiting"` visual state: a static dot in the shared `--color-backgrounded` soft blue (the token background-process badges already use, no `workspace-status-dot-active` pulse), so parked wake-waiting is visually distinct from both the pulsing-green streaming state and the finished/idle states.
- `AgentListItem.getVisualState` maps armed monitors to `"waiting"` via a dedicated `isWaitingOnBashMonitor` flag instead of piggybacking on `hasActiveDelegatedWork`; real work (streaming/starting/delegated/workflow) still wins and keeps the pulsing dot.
- The existing "Watching background bash" caption is unchanged.

Added unit tests (chat: waiting state shown/hidden/precedence; sidebar: static-vs-pulsing dot assertions) and Storybook stories for both surfaces.

## Risks

Low. Active streaming/interrupting/awaiting-input states take precedence over the new states, and with zero armed monitors behavior is byte-for-byte the prior logic. Worst case is a lingering waiting indicator if a monitor count fails to clear, which would equally affect the pre-existing sidebar caption.

---

### DIFF
--- a/src/browser/components/AgentListItem/AgentListItem.stories.tsx
+++ b/src/browser/components/AgentListItem/AgentListItem.stories.tsx
@@ -487,12 +487,47 @@ function renderWorkflowOnlyActivity() {
   );
 }
 
+// Idle workspace parked on an armed background bash monitor: static green
+// "waiting" dot (no streaming pulse) + "Watching background bash" caption.
+function renderBashMonitorWaiting() {
+  const monitorWorkspace = createWorkspace({
+    id: "ws-bash-monitor",
+    name: "bash-monitor",
+    title: "Waiting on background bash monitor",
+    projectName: PROJECT_NAME,
+    projectPath: PROJECT_PATH,
+    createdAt: new Date(NOW - 6_000).toISOString(),
+  });
+
+  return (
+    <StoryScaffold
+      workspaces={[monitorWorkspace]}
+      workspaceActivitySnapshots={{
+        [monitorWorkspace.id]: {
+          recency: NOW,
+          streaming: false,
+          lastModel: null,
+          lastThinkingLevel: null,
+          activeBashMonitorCount: 1,
+        },
+      }}
+    >
+      <WorkspaceRow workspace={monitorWorkspace} />
+    </StoryScaffold>
+  );
+}
+
 // Composite gallery covering the primary single-workspace states. Replaces the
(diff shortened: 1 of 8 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the P2: `computeChatViewReveal` and `showEmptyTranscriptPlaceholder` now gate on `shouldMountStreamingBarrier`, so a cleared transcript with an armed monitor shows the waiting barrier without the "No Messages Yet" placeholder. Retry-barrier/interrupted-marker gates intentionally stay on `shouldShowStreamingBarrier` since an armed monitor shouldn't suppress retry affordances.

ammar-agent:
@codex review

PR #3701, opened 2026-07-09
🤖 fix: re-land waiting-on-bash-monitor indicators with a subtle pulsing blue dot

ammar-agent:
## Summary

Re-lands #3697 (distinct waiting-on-bash-monitor indicators in chat and sidebar), which was reverted on `main` (09a1857fa1) because the sidebar's blue waiting dot rendered far too large. The dot is now the same size/geometry as the green streaming dot and shares its pulse animation, in the `--color-backgrounded` soft blue.

## Background

The original PR gave idle-but-waiting-on-monitor workspaces a blue sidebar dot styled `bg-backgrounded border-backgrounded/25`. The `StatusDot` circle is 12px with a 3.5px border; the green active dot uses a near-background surface tint for that border (`border-surface-green`), so only its small core reads as colored. The translucent blue border made the entire 12px circle read as blue — visually enormous next to its siblings, prompting the revert.

## Implementation

First commit is a clean `git revert` of the revert (restoring #3697 verbatim); the second commit applies the fix on top:

- `StatusDot` waiting state is now `bg-backgrounded border-surface-sky workspace-status-dot-active`: the `--color-surface-sky` border is the same near-background surface treatment the green/question dots use (defined in all four themes), so the visible blue core matches the green core's size, and the shared `workspace-status-dot-pulse` opacity animation now runs on it too.
- Tests updated: since the pulse class is shared between active and waiting, assertions distinguish the states by color token (`bg-content-success` vs `bg-backgrounded`) instead of by animation class.

Everything else from #3697 (chat-pane waiting barrier, empty-transcript gating, stories) is unchanged.

## Risks

Low, same envelope as #3697: zero armed monitors reproduces prior behavior exactly, and streaming/error/question states still outrank the waiting dot. The only new surface is the dot styling itself, now built from the same tokens/animation as the existing active dot.

---

### DIFF
--- a/src/browser/components/AgentListItem/AgentListItem.stories.tsx
+++ b/src/browser/components/AgentListItem/AgentListItem.stories.tsx
@@ -487,9 +487,8 @@ function renderWorkflowOnlyActivity() {
   );
 }
 
-// Idle workspace parked on an armed background bash monitor: static
-// backgrounded-blue "waiting" dot (no streaming pulse) + "Watching background
-// bash" caption.
+// Idle workspace parked on an armed background bash monitor: pulsing
+// backgrounded-blue "waiting" dot + "Watching background bash" caption.
 function renderBashMonitorWaiting() {
   const monitorWorkspace = createWorkspace({
     id: "ws-bash-monitor",
--- a/src/browser/components/AgentListItem/AgentListItem.stories.tsx
+++ b/src/browser/components/AgentListItem/AgentListItem.stories.tsx
@@ -487,12 +487,48 @@ function renderWorkflowOnlyActivity() {
   );
 }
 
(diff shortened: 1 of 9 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the P2 (narrow-pane overflow): the waiting-on-monitor phase no longer renders the reserved `min-w-[14ch]` stats slot, and its informational hint now hides below `@md` via a container query scoped to the barrier's own width (so it tracks narrow chat panes on wide desktops, not just the viewport). Added a fixed-width `WaitingOnBackgroundBashMonitorNarrow` story validating the narrow layout.

PR #3702, opened 2026-07-09
🤖 feat: add GPT-5.6 Sol/Terra/Luna models and move gpt alias to Sol

ammar-agent:
## Summary

Adds first-class support for OpenAI's GPT-5.6 family (Sol / Terra / Luna) and moves the bare `gpt` alias to `gpt-5.6-sol` now that the family is generally available.

## Background

OpenAI previewed GPT-5.6 on June 26, 2026 and released it generally on July 9, 2026. The family introduces a new naming system: the number is the generation, while Sol (flagship), Terra (balanced), and Luna (fast/cheap) are durable capability tiers.

GA availability was verified against the live API with a plain API key: all three ids (`gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`) appear in `/v1/models` and successfully complete minimal Responses API calls (HTTP 200, `status: completed`).

## Implementation

- `knownModels.ts`: three new registry entries with `sol` / `terra` / `luna` aliases. Bare `gpt` alias moves from `gpt-5.5` to `gpt-5.6-sol` (Sol succeeds 5.5 as the flagship at the same list price); `gpt-5.5` stays selectable via its explicit id alias. Sol is warmed.
- `models-extra.ts`: published pricing — Sol $5/$30, Terra $2.50/$15, Luna $1/$6 per 1M tokens; cache reads keep the 90% discount and cache writes bill at 1.25x input. Context window mirrors GPT-5.5 (1.05M in / 128K out) as a documented provisional value — OpenAI hasn't published official context specs yet.
- `thinking/policy.ts`: GPT-5.6 tiers expose all 6 thinking levels including the documented native `max` reasoning effort ("for demanding tasks that need more exploration and verification"). The Responses API path sends `max` through as-is; Chat Completions clamps to `xhigh` (the SDK enum rejects `max`).
- Thinking labels stay distinguishable on GPT-5.6: `max` renders as MAX (native effort) while `xhigh` stays XHIGH; other OpenAI models keep XHIGH for both since they clamp max→xhigh on the wire.
- **Pro mode** (`reasoning.mode: "pro"`) is deliberately **not** in this PR: it belongs as a toggle inside the thinking slider (it composes with effort rather than replacing it). Known gap, tracked in #3704.
- `codexOAuth.ts`: all three tiers added to `CODEX_OAUTH_ALLOWED_MODELS` since the family ships in both Codex and the public API.
- Docs table + built-in skill snapshot regenerated via `bun scripts/gen_docs.ts`.

## Validation

- Live API check: `/v1/models` listing + successful Responses API smoke call per tier.
- Live-verified `reasoning.effort: "max"` completes on `gpt-5.6-sol` and that `gpt-5.5` rejects `max` (confirms per-model gating).
- New policy tests per tier, including gateway-prefixed and version-suffixed forms.
- `make static-check`, typecheck, and 600+ registry/alias-dependent tests pass locally.

## Risks

Low. The riskiest bit is the `gpt` alias move: users relying on `gpt` now get Sol instead of GPT-5.5. Pricing is identical and `gpt-5.5` remains addressable, so blast radius is a model-behavior change only. Context-window metadata is provisional (mirrors GPT-5.5) until OpenAI publishes official specs; worst case is premature compaction, not request failures.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `high` • Cost: `$10.12`_


ammar-agent:
@codex review

ibetitsmike:
1. Max reasoning effort: GPT-5.6 supports max reasoning effort for demanding tasks that need more exploration and verification. If you currently use xhigh, compare both settings on representative workloads.
2. Pro mode: GPT-5.6 can perform more model work to improve reliability on difficult tasks and return a single final answer. Enable it with reasoning.mode: "pro" when quality matters more than latency and token usage. Learn how to [use pro mode](https://developers.openai.com/api/docs/guides/reasoning#reasoning-mode).

ammar-agent:
@codex review

Added GPT-5.6 native `max` reasoning effort (6-level thinking policy, Responses-only passthrough with Chat Completions clamp) and `reasoning.mode: "pro"` support via a Mux-internal header + fetch-wrapper body injection, both live-verified against the API.

ammar-agent:
Addressed both review findings in 75119b702a:

- **Keep GPT-5.6 max distinguishable in thinking labels** (PRRT_kwDOPxxmWM6PrcuR): fixed. `getThinkingDisplayLabel`/`getThinkingOptionLabel` now return MAX/`max` for the max level on GPT-5.6 models (gated on `openaiSupportsNativeMaxEffort`), while xhigh stays XHIGH/`xhigh`. Other OpenAI models keep XHIGH for both levels since they clamp max→xhigh on the wire. Covered by new tests in `thinking.test.ts`.
- **Honor provider-level pro mode outside main chat** (PRRT_kwDOPxxmWM6PrcuT): mooted by removal. Per maintainer decision, pro mode is out of this PR entirely — it belongs as a toggle inside the thinking slider, not backend-only config plumbing. All pro-mode code (schema knobs, header, fetch-wrapper injection) is reverted; the gap you identified (headless `createModel()` callers bypassing `buildRequestHeaders`) is recorded in the tracking issue #3704 so the follow-up implementation handles it from the start.

ammar-agent:
@codex review

ammar-agent:
**Wire OpenAI cache-write tokens into GPT-5.6 pricing** (PRRT_kwDOPxxmWM6PrtRN): verified and triaged as a real but currently-unfixable-in-Mux gap; documented in 30baf4bd6d and tracked in #3705.

Findings:
- Confirmed live that the Responses API reports `input_tokens_details.cache_write_tokens` for GPT-5.6 (3207 write tokens on first call, 3207 `cached_tokens` on repeat).
- However `@ai-sdk/openai` cannot surface it: `convertOpenAIResponsesUsage` hardcodes `cacheWrite: undefined` and its usage Zod schema strips `cache_write_tokens`. Verified on both our installed 3.0.62 and the latest 3.0.83 — an SDK bump does not help, so `createDisplayUsage()` has no OpenAI-side source for `cacheCreateTokens`.
- Folding the 1.25x rate into `input_cost_per_token` would overcharge every non-write request, so we're not doing that.

Resolution in this PR: a `KNOWN COST GAP` comment at the metadata site explains the bounded underreport (0.25x input rate, first-write tokens only; reads are costed correctly) and why the `cache_creation_input_token_cost` values are kept — pricing lights up automatically once the SDK surfaces cache-write usage. #3705 captures the verified evidence and the upstream-first fix plan.

ammar-agent:
@codex review

ammar-agent:
**Gate native max to OpenAI Responses routes** (PRRT_kwDOPxxmWM6Pr3GG): fixed in 6d233d779c. The GPT-5.6 policy now exposes `max` only when the route delivers Mux-built openai providerOptions verbatim — direct `openai:` or a `PROVIDER_DEFINITIONS` passthrough gateway (mux-gateway). Non-passthrough routes like `openrouter:openai/gpt-5.6-sol` or `github-copilot:openai/gpt-5.6-terra` keep the 5-level xhigh-capped policy, so the slider never offers an effort the wire clamps away. Covered by new route-gating tests alongside the existing passthrough-gateway case.

ammar-agent:
@codex review

ammar-agent:
**Gate GPT-5.6 max on the resolved route** (PRRT_kwDOPxxmWM6Pr3GG follow-up, thread PRRT_kwDOPxxmWM6Pr-wp): fixed in fc03cbe059. You're right that the policy gate only sees the model string while routePriority/overrides pick the route at request time. Added `clampThinkingLevelForRoute()`: when a GPT-5.6 model resolves onto a non-passthrough route, `max` is clamped to `xhigh` in `AIService` immediately after route resolution — before provider options, headers, replay transforms, and metadata are built — so everything downstream agrees with what the wire can deliver (mirrors the existing Mythos-class clamp precedent). Applied on both the primary and fallback-model paths; no-op for non-GPT-5.6 models where max/xhigh already share a wire value. Covered by new clamp tests (resolved-route override, model-string fallback, passthrough routes, non-max levels).

ammar-agent:
@codex review

coadler:
image

been working fine for me, only thing that feels odd is the model ordering in the menu. i feel like they should be before 5.5

PR #3713, opened 2026-07-12
🤖 fix: prevent concurrent agent banner wake flicker

ammar-agent:
## Summary

Keep the concurrent local-agent warning stable while another agent cycles through background wake-ups, including the handoff gap after a one-shot monitor or terminal task notification is consumed but before the synthetic turn starts streaming.

## Background

The warning originally only considered an interruptible stream active. Background monitor matches and workflow/task terminal messages transition the owning workspace through waiting, queued/preparing, and streaming states. For inactive workspaces, pre-stream startup was not exposed in activity snapshots, so the banner could still unmount and remount on every wake cycle.

## Implementation

- Add an authoritative `pendingBackgroundWake` workspace activity field.
- Publish it before durable bash-monitor or terminal-attention sources disappear.
- Recompute it from pending wake stores plus queued/preparing/retrying session state, with versioned async refreshes to prevent stale clears.
- Include the signal in activity list bootstrap/reconnect and frontend sidebar state.
- Keep the concurrent warning visible for streaming, startup, pending background wakes, active workflows, and armed monitors.
- Add focused coverage for selector behavior, inactive workspace activity propagation, monitor wakes, and terminal wakes.

## Validation

- Focused warning, workspace store, monitor wake, and terminal wake regression suites
- `make typecheck`
- `make static-check`

## Risks

Moderate and localized to workspace activity signaling. The pending flag is recomputed from durable stores and live session state, and is cleared on stream start, cancellation, supersession, or consumed terminal attention; versioned refreshes prevent older async reads from restoring stale state.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `high` • Cost: `$2.94`_

### DIFF
--- a/src/browser/components/ChatPane/ChatPane.tsx
+++ b/src/browser/components/ChatPane/ChatPane.tsx
@@ -71,7 +71,7 @@ import { CompactionWarning } from "../CompactionWarning/CompactionWarning";
 import { ContextSwitchWarning as ContextSwitchWarningBanner } from "../ContextSwitchWarning/ContextSwitchWarning";
 import {
   ConcurrentLocalWarningDecoration,
-  useConcurrentLocalStreamingWorkspaceName,
+  useConcurrentLocalActiveWorkspaceName,
 } from "../ConcurrentLocalWarning/ConcurrentLocalWarning";
 import { BackgroundProcessesBanner } from "../BackgroundProcessesBanner/BackgroundProcessesBanner";
 import { checkAutoCompaction } from "@/common/utils/compaction/autoCompactionCheck";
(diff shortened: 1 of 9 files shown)


ammar-agent:
@codex review

ammar-agent:
Addressed the inactive-workspace startup gap by adding an authoritative `pendingBackgroundWake` activity signal. Durable bash-monitor and task/workflow terminal notifications now publish the handoff before their source activity disappears, and the signal stays set through queued/preparing startup until streaming begins or the wake is canceled/superseded. Added focused backend/store regression coverage.

@codex review

ammar-agent:
Addressed the activity-list race: `getActivityList()` now computes the current pending wake for its response without mutating or versioning the live cache, so it cannot invalidate a newer refresh or restore stale pending state.

@codex review

ammar-agent:
Seeded `pendingBackgroundWake` for every persisted terminal-attention owner during `TaskService.initialize()` before scheduling its drain, and added restart recovery coverage that asserts the handoff signal is published.

@codex review

ammar-agent:
Seeded `pendingBackgroundWake` for every persisted terminal-attention owner during `TaskService.initialize()` before scheduling drains, including records that already existed on disk and therefore were not recreated by recovery. Added a restart regression assertion.

@codex review

ammar-agent:
Cleared the live pending wake signal after terminal-attention notifications transition to delivered (and after re-pending on pre-stream failure), so normal `requireIdle` delivery cannot leave the concurrent warning stuck. Added a focused assertion around the normal drain path.

@codex review

PR #3715, opened 2026-07-13
🤖 fix: wait for sibling tools before send-after-step dispatch

ThomasK33:
## Summary

Fix **Send after step** so a queued follow-up waits for every known sibling tool call in the current step while still preempting immediately after provider-native tools such as `web_search` and `web_fetch` finish.

## Background

`AgentSession` previously soft-stopped the stream on every individual `tool-call-end`. When a model emitted multiple client-executed tools in the same step, the first completion could cancel or background remaining siblings before their results were committed.

Moving dispatch exclusively to the AI SDK `stopWhen` step boundary fixed client sibling batches, but provider-executed tools run inside a single provider response and do not create the client-tool continuation where `stopWhen` is evaluated. They therefore need a distinct native-tool boundary.

## Implementation

- Remove per-client-tool soft aborts from `AgentSession`; client tools now stop at `StreamManager`'s AI SDK completed-step boundary after all sibling results settle.
- Propagate the AI SDK's `providerExecuted` marker on `tool-call-end` events.
- Track live tool-call IDs in `AgentSession` and reserve soft preemption for a provider-executed result after every currently known sibling has completed.
- Preserve intentional-abort retry suppression, edited-queue behavior, and hard-user-interrupt cancellation for that native-only path.
- Add delayed two-tool mock coverage for the full client sibling step plus focused provider-native result, sibling, metadata, and interrupt-race tests.

## Validation

- `bun test src/node/services/streamManager.modelOnlyNotifications.test.ts`
- `TEST_INTEGRATION=1 bun x jest tests/ui/chat/sendModeDropdown.test.ts --runInBand`
- `git diff --check`

## Dogfooding

Verified the client sibling path in an isolated dev-server sandbox with Claude Opus 4.8. The model planned two sibling foreground Bash calls in one assistant step, each sleeping for 45 seconds, while a follow-up was queued in **Send after step** mode.

Observed UTC ordering:

1. Follow-up queued: `09:12:48.941`
2. Sibling A completed: `09:13:33.923`
3. At the first tool boundary, sibling B was alive and the follow-up was absent from durable history.
4. Sibling B completed: `09:14:18.953`
5. Follow-up entered durable history: `09:14:19.025` — 72 ms after sibling B completed.

Captured four progression screenshots and a continuous 103.4-second VP8 WebM recording at 1280×578 covering queueing, sibling handoff, and final dispatch. The provider-native branch is additionally locked down with deterministic event-level tests because native search/fetch execution depends on external provider behavior.

<details>
<summary>Middle-boundary process evidence</summary>

```text
/tmp/mux-after-step-video-final/markers/sibling-a.done 2026-07-13 09:13:33.923199472 +0000
second_sibling=b pid=1026746 alive=yes at 2026-07-13T09:13:34,031728440+00:00
    PID    PPID                  STARTED     ELAPSED COMMAND
1026746  979179 Mon Jul 13 09:13:33 2026       00:00 bash -c ... sleep 45 ... sibling-b.done
queued_followup_in_history=no
```

</details>

## Risks

This changes queue-dispatch timing for **Send after step** only. Client tool calls remain on the SDK's transactional step boundary; provider-native calls use a narrowly gated soft interrupt only after the emitted result is persisted and no known sibling remains active. Regression tests cover both paths and the hard-interrupt race.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `max` • Cost: `$79.65`_

### DIFF
--- a/src/common/orpc/schemas/stream.ts
+++ b/src/common/orpc/schemas/stream.ts
@@ -450,6 +450,10 @@ export const ToolCallEndEventSchema = z.object({
   toolCallId: z.string(),
   toolName: z.string(),
   result: z.unknown(),
+  providerExecuted: z
+    .boolean()
+    .optional()
+    .meta({ description: "True when the provider executed the tool server-side" }),
   timestamp: z.number().meta({ description: "When tool call completed (Date.now())" }),
   parentToolCallId: z.string().optional().meta({ description: "Set for nested PTC calls" }),
 });
--- a/src/node/services/agentSession.queueDispatch.test.ts
+++ b/src/node/services/agentSession.queueDispatch.test.ts
@@ -99,6 +99,96 @@ describe("AgentSession queued message tool-call dispatch", () => {
     }
   });
 
+  test("soft-stops after a provider-executed tool result and dispatches after abort", async () => {
+    const workspaceId = "queue-dispatch-provider-tool";
+    const { session, cleanup, aiEmitter, aiService } = await createAgentSessionHarness({
+      workspaceId,
+    });
+    const stopStream = spyOn(aiService, "stopStream").mockResolvedValue(Ok(undefined));
+    const sendQueuedMessages = spyOn(session, "sendQueuedMessages").mockImplementation(
+      () => undefined
+    );
+
+    try {
+      aiEmitter.emit("stream-start", streamStartEvent(workspaceId));
+      session.queueMessage("follow up", { model: TEST_MODEL, agentId: "exec" });
+
+      aiEmitter.emit("tool-call-end", {
+        ...toolCallEndEvent(workspaceId),
+        toolName: "web_search",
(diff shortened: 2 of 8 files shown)


ThomasK33:
@codex review

ThomasK33:
Addressed the Codex P2 finding **“Restore send-after-step for native tools”** in `4d9987e9a`:

- propagate `providerExecuted` from AI SDK tool results,
- soft-stop only for provider-native results,
- wait for every currently known sibling tool call before stopping,
- keep client tool batches on the SDK `stopWhen` step boundary,
- preserve the hard-user-interrupt race guard.

Added deterministic native-result, native-sibling, event-metadata, and interrupt-race coverage. Targeted tests, the full send-mode UI test, and `make static-check` pass.

@codex review

PR #3719, opened 2026-07-13
🤖 fix: avoid OpenAI tool search name collision

ibetitsmike:
## Summary

Rename Mux's deferred MCP catalog search tool from `tool_search` to `tool_catalog_search` so OpenAI Responses treats it as a custom function tool rather than its native tool-search primitive.

## Background

AI SDK 7 reserves `tool_search` for OpenAI's native tool search and validates results against `{ tools: [...] }`. Mux returned `{ query, matches, totalDeferred }`, so continuing after a catalog search failed with a type-validation error.

## Implementation

- Register and advertise the live Mux tool as `tool_catalog_search`.
- Normalize completed historical Mux `tool_search` calls in provider-bound messages without changing persisted history.
- Keep legacy transcript rendering and deferred-tool activation compatible with existing workspaces.
- Avoid rewriting unrelated MCP tools that happen to use the legacy name.
- Update tool definitions, UI mappings, generated documentation, and regression coverage.

## Validation

- OpenAI Responses wire regression for completed legacy history
- Tool catalog, legacy activation, tool renderer, tool definition, and stream active-tool tests

## Risks

Low. The live name changes only while the tool-search experiment is enabled. Compatibility handling is request-only and requires the historical result to match Mux's catalog-search shape.

> Mux worked on behalf of Mike.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `xhigh` • Cost: `$27.56`_

### DIFF
--- a/docs/hooks/tools.mdx
+++ b/docs/hooks/tools.mdx
@@ -750,7 +750,7 @@ If a value is too large for the environment, it may be omitted (not set). Mux al
 </details>
 
 <details>
-<summary>tool_search (2)</summary>
+<summary>tool_catalog_search (2)</summary>
 
 | Env var                | JSON path | Type   | Description                                                                                               |
 | ---------------------- | --------- | ------ | --------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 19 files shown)


ibetitsmike:
@codex review

PR #3724, opened 2026-07-14
🤖 fix: cap GPT-5.6 context over Codex OAuth

coadler:
## Summary

Apply the ChatGPT/Codex OAuth context caps to the GPT-5.6 family while retaining the public 1.05M context window for API-key requests. Sol and the bare GPT-5.6 alias use 272K; Terra and Luna use 128K.

## Background

The ChatGPT/Codex OAuth routing layer has lower context limits than the public OpenAI API. GPT-5.6 was initially allowed through OAuth without route-specific context overrides, causing compaction and the token meter to use the API's 1.05M limit.

## Implementation

- Add tier-specific Codex OAuth overrides for the bare GPT-5.6 alias and the Sol, Terra, and Luna tiers.
- Cover every GPT-5.6 OAuth model ID in the context-limit tests.
- Update token-meter coverage to use the effective OAuth limit.

## Risks

Low. The overrides only apply when direct OpenAI requests actually route through Codex OAuth. API-key and gateway-routed requests continue using their existing model metadata.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `medium` • Cost: `$5.63`_

### DIFF
--- a/src/common/constants/codexOAuth.test.ts
+++ b/src/common/constants/codexOAuth.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from "bun:test";
 import {
+  getCodexOauthContextWindowOverride,
   isCodexOauthAllowedModel,
   isCodexOauthAllowedModelId,
   isCodexOauthRequiredModel,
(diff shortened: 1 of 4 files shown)


coadler:
@codex review

coadler:
@codex review

Updated the OAuth overrides to use 272K for Sol/the bare alias and 128K for Terra/Luna, with tier-specific test coverage.

PR #3728, opened 2026-07-15
🤖 feat: skills refresh — invocation control, $ARGUMENTS, dynamic context, .claude compat

ThomasK33:
## Summary

Skills refresh: adopt ecosystem-standard SKILL.md frontmatter used by other agent tools (Claude Code, VS Code Copilot), add Claude Code-compatible `$ARGUMENTS` substitution, and introduce two default-off experiments — `.claude/skills` discovery compat and dynamic context injection.

## Background

Mux already follows the agentskills.io core spec, but the ecosystem has converged on a handful of extension fields that portable skills increasingly rely on. This PR closes the gaps identified in a survey of Claude Code / VS Code / Codex CLI skill features:

1. **`disable-model-invocation: true`** — ecosystem spelling for hiding a skill from the model-facing index (alias of Mux's `advertise: false`; most restrictive wins).
2. **`user-invocable: false`** — the inverse: hide a skill from user-facing invocation surfaces (`/` menu, typed `/name`, `$refs`, command palette, ACP command list) while staying model-visible. Combining both makes a reference-only skill.
3. **`argument-hint`** — shown in invocation UIs; drives ACP `input.hint`.
4. **`when_to_use` / `when-to-use`** — extra model-facing guidance appended to the skill's index entry.
5. **`$ARGUMENTS` and `$1`–`$9`** — placeholder substitution in skill bodies for slash invocations (`/fix-issue 123 high`); no-placeholder bodies keep today's behavior (arguments stay visible in the user message).
6. **`claude-skills-compat` experiment (default off)** — read-only discovery from `.claude/skills` and `~/.claude/skills` at lowest precedence per scope, so repos shipping Claude skills work in Mux without copying files.
7. **`skill-dynamic-context` experiment (default off)** — whole-line `` !`command` `` directives in a skill body execute in the workspace and are replaced with fenced output when a **user** invokes the skill.

## Implementation

- All new frontmatter is normalized at descriptor-construction time via resolver helpers in `src/common/orpc/schemas/agentSkill.ts` (`resolveSkillAdvertise`, `resolveSkillUserInvocable`, `resolveSkillWhenToUse`), so downstream consumers never see raw aliases.
- The user-invoked snapshot pipeline in `agentSession.materializeAgentSkillSnapshots` composes: claude-compat resolution → `$ARGUMENTS` substitution → dynamic context injection → snapshot hash (hash covers the final body, so different args/outputs produce distinct snapshots).
- Argument text travels as a new optional `arguments` field on the shared `agent-skill` message metadata, populated by both producers (ChatInput and ACP).
- **Deliberate ACP behavior change:** `availableCommands` now filters on `user-invocable` instead of `advertise` — unadvertised skills are exactly the ones users should invoke manually, so they now appear in ACP clients' command lists.
- Dynamic context injection is bounded (max 10 directives, 10s timeout, 16KB output cap per command, sequential) and failure-isolated (a failing directive injects a note; an unexpected error degrades to the raw body). Execution happens only on user-initiated invocations; model-side `agent_skill_read` always returns the raw body. Trust model documented via a `SECURITY AUDIT` comment at the exec sink.
- Both experiments follow existing registry conventions and surface automatically in Settings → Experiments.

## Validation

- Full `make static-check` green; generated content (`builtInSkillContent.generated.ts`, `docs/hooks/tools.mdx`) verified up-to-date via check mode.
- Broad sweep: 1698 pass / 0 fail across 121 test files (aiService, agentSession, agentSkills, tools, ACP, browser slash/inline utils, ChatInput).
- Dogfooded all features end-to-end through production code paths (real discovery + real tool factories in a temp project): descriptor normalization, model-index gating + when-to-use text, ACP list semantics + argument hint, `$ARGUMENTS`/positional/`$10` substitution output, directive injection with fence formatting, and `.claude` compat discovery + precedence (`.mux` > `.agents` > `.claude`).

## Risks

- **Skill visibility surfaces**: the highest-risk area is the ACP command-list semantic change (unadvertised skills now listed). Severity low; affects ACP clients only, and the new behavior matches the documented intent of `disable-model-invocation`.
- **Snapshot materialization** is on the message-send path; substitution and injection are wrapped so failures degrade to the raw body rather than blocking sends.
- **Dynamic context injection executes repo-controlled commands**, but only behind a default-off experiment, only on user-initiated invocation, with bounded execution — no change for users who don't opt in.
- Legacy persisted slash invocations without the new `arguments` metadata substitute placeholders as empty strings (documented on the type).

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$83.54`_


ThomasK33:
@codex review

ThomasK33:
@codex review

Both findings addressed:

- **P2 (bounded output)**: `execBuffered` now accepts `maxOutputBytes`; the dynamic-context exec sink passes the 16KB cap (+1 byte so the module's truncation marker still fires). The new `streamToStringCapped` bounds accumulation *while reading* and drains the remainder, so memory stays capped without disturbing exit codes; the 10s runtime timeout bounds duration. Covered by unit tests and a seam test running a real 64KB-output directive.
- **P3 (argument hints)**: slash suggestions (chat menu + command palette share this path) now render `argument-hint` next to the skill name, display-only — matching and replacement still use just the name.

ThomasK33:
@codex review

A final pre-merge review pass surfaced three blockers, now fixed (commit above):

1. **Remote stderr unbounded**: `RemoteRuntime.exec` kept an uncapped stderr side-buffer for error reporting, defeating `maxOutputBytes` on SSH/Docker runtimes. Now capped at 16KB.
2. **Directives could run twice**: skill snapshots were materialized before the on-send compaction decision; a deferring turn discarded the first materialization and re-ran it on the follow-up — executing dynamic-context directives twice. Materialization now happens only for turns that send immediately (regression test added).
3. **Security gate hardening**: the dynamic-context experiment gate now requires an explicit local Settings override (`isExperimentLocallyEnabled`); a remote/cached experiment assignment can never enable shell execution. SECURITY AUDIT comment updated to match.

Also: capped reader drops split code points instead of emitting U+FFFD, the Workflows launcher honors `user-invocable: false`, and docs warn about shell quoting when combining `$1` placeholders with directives.

ThomasK33:
@codex review

Addressed: `skill-dynamic-context` is now `localOverrideOnly` in the experiment registry — `getExperimentValue` excludes remote/cached assignment from its evaluation path entirely (reads `disabled` without an explicit local toggle). Since the Settings UI consumes the backend's experiment values, the toggle and the execution gate can no longer disagree; a remote-enabled bucket shows OFF and stays OFF until the user flips it locally. Regression test asserts both gates agree across seeded-remote/override/cleared states.

ThomasK33:
@codex review

Please take another look. Latest commit `9c4fb50e3` makes the skill-dynamic-context experiment local-override-only in the registry (remote assignment excluded from evaluation), addressing the prior P2.

PR #3731, opened 2026-07-15
🤖 fix: honor agent AI defaults and target workspace settings for workspace turns

ThomasK33:
## Summary

Workspace turns (`task(kind: "workspace")`) now resolve model/thinking with a saner per-field precedence: explicit per-launch override → the target workspace's own persisted settings (`mode: "existing"` follow-ups) → configured `agentAiDefaults` for the agent the turn runs → owner's live runtime settings → owner's persisted settings → app default.

## Background

Two surprising behaviors when an orchestrator workspace delegates via workspace turns:

1. **Creation ignored configured agent defaults.** An orchestrator turned down to `medium` spawned `medium` children even when the user's exec default was `max` — `createWorkspaceTurn` skipped straight from the explicit override to the owner's live settings, unlike sub-agent tasks (`resolveTaskAISettings`), which already consult `subagentAiDefaults`/`agentAiDefaults` first. Agent-created workspaces now match what the user would get creating the same workspace by hand.
2. **Follow-ups dragged children along.** Every `mode: "existing"` message re-inherited the owner's *current* settings and persisted them onto the child (persist-on-send), so bumping the orchestrator to `high` silently flipped all of its children to `high`, clobbering any level the user had deliberately set on a child. Follow-ups now continue the target workspace's own last-used settings, mirroring how a workspace behaves when the user sends a message in it directly.

## Implementation

- All lookups key off a single `workspaceTurnAgentId` constant (`exec` today, including the previously hardcoded `sendMessage` agent id), so if workspace turns ever run other agents, that agent's defaults follow automatically.
- Pro `reasoningMode` follows the same ownership rule: creation still inherits the owner's toggle, but once the target has its own persisted settings, follow-ups respect them outright (absent = standard per `WorkspaceAISettingsSchema`) instead of re-injecting the owner's choice.
- Explicit `model`/`thinking` args still outrank everything and persist onto the child via the existing persist-on-send path; for unchanged follow-ups that persist becomes a no-op writing back identical values.

## Validation

- New test walks the full precedence chain end-to-end: agent defaults beat owner runtime on create; the child's own settings beat both on follow-up; explicit override beats the child's settings.
- New test covers pro-mode ownership: inherited on create, not re-injected on follow-up after the child switched back to standard.
- Full `taskService.test.ts` suite (293) and `tools/task.test.ts` (28) pass; `make static-check` green.

## Risks

Behavioral change for orchestrator workflows: the accidental "turn the orchestrator down to spawn cheap children" trick no longer works when an exec agent default is configured — economizing children now requires an explicit `thinking:` arg on the task call (this is the intended semantics). Blast radius is confined to workspace-turn AI-settings resolution; regular sub-agent tasks and UI-created workspaces are untouched.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$11.66`_

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -1324,6 +1324,207 @@ describe("TaskService", () => {
     expect(sendMessageCall[2]).toMatchObject({ reasoningMode: "pro" });
   });
 
+  test("createWorkspaceTurn resolves AI settings: agent defaults on create, target settings on follow-up, explicit override wins", async () => {
+    const config = await createTestConfig(rootDir);
+    stubStableIds(config, [
+      "firsthandle",
+      "firstturn",
+      "secondhandle",
+      "secondturn",
+      "thirdhandle",
+      "thirdturn",
+    ]);
+    // Owner persisted at opus/high (helper default); configured exec agent
+    // defaults differ from both the owner's persisted and live settings.
+    const { parentId, projectPath } = await saveLocalParentWorkspace(config, rootDir, {
+      agentAiDefaults: { exec: { modelString: "openai:gpt-5.2", thinkingLevel: "xhigh" } },
+    });
+
+    const createWorkspace = mock(
+      async (...args: unknown[]): Promise<Result<{ metadata: WorkspaceMetadata }>> => {
+        const tags = args[7] as Record<string, string> | undefined;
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3732, opened 2026-07-16
🤖 fix: defer bash monitor wakes during task_await

ethanndickson:
## Summary

Defers a bash monitor wake while `task_await` is already waiting on the same process, preventing the wake from interrupting the wait it is meant to supplement without delaying monitor wakes from unrelated processes.

## Background

The settled shown-frontier gate only synchronized with unfiltered output reads. A filtered `task_await` was invisible to that gate, so a monitor match could enqueue a synthetic wake during the active wait. The queued wake interrupted `task_await`, and agents could repeatedly restart the wait and be interrupted by the next monitor match.

## Implementation

- Track unfiltered reads and filtered `task_await` reads as blockers for same-process monitor wake delivery.
- Expose a non-blocking delivery-state snapshot from `BackgroundProcessManager` so the workspace can defer only records for processes with active blocking reads.
- Schedule a fresh drain when that process's read settles while continuing to deliver unrelated monitor matches from the current batch.
- Reuse the shown-frontier gate after settlement: output displayed by an unfiltered await suppresses the redundant wake, while matched output hidden by a filter remains eligible for delivery.
- Add regression coverage for both the filtered-await synchronization and the unrelated-process delivery invariant.

## Risks

A monitor wake for a process can now be delayed for the duration of an active filtered `task_await` on that same process. This is intentional and bounded by the await. Wakes for other processes and filtered reads from other callers are unaffected.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `xhigh` • Cost: `$22.23`_

### DIFF
--- a/src/node/services/backgroundProcessManager.test.ts
+++ b/src/node/services/backgroundProcessManager.test.ts
@@ -847,10 +847,10 @@ describe("BackgroundProcessManager", () => {
       let proc = await manager.getProcess(result.processId);
       for (let attempt = 0; attempt < 200; attempt++) {
         proc = await manager.getProcess(result.processId);
-        if (proc?.unfilteredReadSettled) break;
+        if (proc?.monitorWakeBlockingReadSettled) break;
         await new Promise((resolve) => setTimeout(resolve, 5));
       }
-      expect(proc?.unfilteredReadSettled).toBeDefined();
+      expect(proc?.monitorWakeBlockingReadSettled).toBeDefined();
       expect(proc?.shownThroughOffset ?? -1).toBe(0);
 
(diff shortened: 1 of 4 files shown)


ethanndickson:
@codex review

ethanndickson:
Addressed the Codex P2 finding about unrelated monitor wakes.

The drain now takes a non-blocking delivery-state snapshot per process. It leaves the blocked process's record pending, schedules a retry when that read settles, and continues delivering other processes' records from the same owner batch. Added a regression proving the unrelated wake is delivered before the blocked read resolves.

ethanndickson:
@codex review

PR #3733, opened 2026-07-16
🤖 fix: clean up devtools.jsonl on archive/remove and reap orphaned session dirs

ThomasK33:
## Summary

DevTools debug logs (`devtools.jsonl`) are now deleted when a workspace is archived or removed, and startup sweeps reclaim disk from orphaned session directories and from workspaces archived before this cleanup existed.

## Background

A disk audit found ~57.7 GiB of `devtools.jsonl` files across 442 archived workspaces (mux never pruned these; the worst case after unarchive is an empty DevTools panel), plus 96 orphaned session directories whose workspaces no longer exist in `~/.mux/config.json` at all (left behind by crashed or partially-failed removals).

## Implementation

- `DevToolsService.removeWorkspaceData(workspaceId)`: drops in-memory run/step state and deletes the on-disk `devtools.jsonl`. The deletion serializes behind pending appends via the existing per-workspace write queue, and deleting the in-memory entry makes stale queued appends no-ops via the existing `appendToFile` existence guards. Runs even when debug logging is disabled, so files written while logging was on still get cleaned.
- `WorkspaceService.archive()` calls it best-effort **after** `archivedAt` is durably persisted — archive never fails because of log cleanup.
- `WorkspaceService.remove()` additionally drops in-memory DevTools state (the file already dies with the session directory).
- Startup sweeps in `WorkspaceService.initialize()` (next to the existing orphan-scratch-workdir reaper):
  - **Orphan session dirs**: deletes `~/.mux/sessions/<id>` directories with no config entry. Safety rails: config read failures are never treated as an empty reference set (no mass deletion), legacy `<project>-<workspace>` session IDs are preserved for un-migrated entries, and a 24h mtime grace window plus a fresh `findWorkspace()` re-check guard against racing workspace creation.
  - **Retroactive devtools sweep**: deletes `devtools.jsonl` for currently-archived workspaces, healing archives that predate archive-time cleanup and any archive-time cleanup failures.
- Wired via a narrow `setDevToolsService()` setter in `coreServices` (same pattern as `setTaskService`). Chat history, plans, and archive snapshots are untouched.

## Validation

End-to-end dogfood with real `Config` + `DevToolsService` + `WorkspaceService` on a temp root: `initialize()` shrank an archived workspace 8.1M→4K (chat kept), removed a 4.1M stale orphan, spared a fresh orphan; `archive()` then dropped the live workspace's `devtools.jsonl` leaving only `chat.jsonl`.

## Risks

- Deletion sweeps are the main risk surface. Mitigations: orphan reaping requires a strict config load (`throwOnError`), a 24h grace window, and a per-directory fresh config re-check immediately before `rm`; devtools cleanup only ever touches `devtools.jsonl`.
- Severity if wrong: loss of debug logs (low; regenerable) or, guarded against as above, session data of a brand-new workspace.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$13.32`_

### DIFF
--- a/src/node/services/__tests__/devToolsService.test.ts
+++ b/src/node/services/__tests__/devToolsService.test.ts
@@ -402,6 +402,30 @@ describe("DevToolsService", () => {
       expect(await service.getRunWithSteps("ws-1", "run-1")).toBeNull();
       expect(await fs.readFile(getDevtoolsLogPath(sessionsDir, "ws-1"), "utf-8")).toBe("");
     });
+
+    it("removeWorkspaceData deletes the log file and in-memory state", async () => {
+      const service = new DevToolsService(createTestConfig({ sessionsDir, enabled: true }));
+      await service.createRun("ws-1", makeRun("run-1"));
+      await service.createStep("ws-1", makeStep({ id: "step-1", runId: "run-1" }));
+
+      await service.removeWorkspaceData("ws-1");
+
+      expect(await pathExists(getDevtoolsLogPath(sessionsDir, "ws-1"))).toBe(false);
+      expect(await service.getRuns("ws-1")).toEqual([]);
+      expect(await service.getRunWithSteps("ws-1", "run-1")).toBeNull();
+    });
+
+    it("removeWorkspaceData deletes stale log files even when logging is disabled", async () => {
+      // Simulate a file written while logging was enabled, then the user disabling it.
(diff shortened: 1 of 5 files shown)


ThomasK33:
@codex review

PR #3739, opened 2026-07-20
🤖 feat: show per-model cost breakdown in workspace Costs tab

ibetitsmike:
## Summary

Adds a per-model cost breakdown table to the workspace Costs tab (Stats → Cost, Session view), so multi-model sessions show where tokens and spend went per model.

## Background

Session cost was only shown as component totals (input/output/cache), with no visibility into which model consumed them. Sessions that mix models (tool-invoked models, sub-agent roll-ups, mid-session model switches) had no per-model attribution. The per-model data already existed in `session-usage.json` (`byModel`); it was summed into `sessionTotal` and discarded.

## Implementation

- `WorkspaceStore`: `WorkspaceUsageState` now exposes `sessionByModel`, a fresh shallow copy of the per-canonical-model usage record already loaded from `session-usage.json` (copied per recompute because the underlying record is mutated in place on stream-end).
- `CostsTab`: new Model / Tokens / Cost table below the component table, Session view only (Last Request is single-model by definition). Live streaming usage is merged into the active model's bucket the same way the displayed session total already merges it, so model rows always sum to the header total. Rows sort by cost, then tokens, descending; names render via `formatModelStringForDisplay`.
- Storybook: `CostsTabMultiModel` story with a play assertion that both model rows render.

## Validation

- UAT against a sandboxed dev server with real API calls across two models (Haiku 4.5 + GPT-5.4 Nano): verified Session breakdown rows sum to the header total, the table hides in Last Request view, rows stay consistent mid-stream while usage grows live, and reload renders identical values from pure disk state matching `session-usage.json`.

## Risks

Low. Read-only UI addition fed by existing usage data; no changes to usage accounting or persistence. Worst case is a display inconsistency in the new table itself.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$20.48`_

### DIFF
--- a/src/browser/features/RightSidebar/CostsTab.tsx
+++ b/src/browser/features/RightSidebar/CostsTab.tsx
@@ -3,8 +3,10 @@ import { useWorkspaceUsage, useWorkspaceConsumers } from "@/browser/stores/Works
 import {
   sumUsageHistory,
   formatCostWithDollar,
+  getTotalCost,
   type ChatUsageDisplay,
 } from "@/common/utils/tokens/usageAggregator";
+import { normalizeToCanonical, formatModelStringForDisplay } from "@/common/utils/ai/models";
 import { usePersistedState } from "@/browser/hooks/usePersistedState";
 import { AGENT_AI_DEFAULTS_KEY } from "@/common/constants/storage";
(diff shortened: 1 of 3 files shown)


ibetitsmike:
@codex review

PR #3740, opened 2026-07-20
🤖 fix: keep analytics sidebar controllable on mobile

ammar-agent:
## Summary

Keeps the left sidebar controllable from the Analytics route at narrow widths. Analytics previously hid its sidebar opener below the `md` breakpoint even though the collapsed sidebar is fully off-canvas there, leaving no pointer-accessible way to reopen it.

## Implementation

- Opt the Analytics sidebar opener into the shared `mobile-menu-btn` responsive behavior used by other main-pane routes.
- Add a full-app UI regression test that collapses and reopens the sidebar from Analytics.
- Add a phone-viewport Storybook story with a pinned Pixel viewport that exercises the control at the responsive breakpoint.
- Make the existing review auto-refresh integration test deterministic by emitting the same WorkspaceStore file-tool completion event directly instead of relying on a live model to choose and call `bash`.

## Validation

- `TEST_INTEGRATION=1 bun x jest tests/ui/layout/analyticsHeader.test.ts --runInBand --silent`
- `make static-check`
- 3 consecutive targeted runs of the review auto-refresh integration test
- `make storybook-build`
- `STORYBOOK_PORT=6108 make storybook-run CMD='bun x test-storybook --url http://127.0.0.1:6108 --testTimeout 30000 --maxWorkers 1 src/browser/stories/App.phoneViewports.stories.tsx'`

## Risks

Low. The production change only adds the existing responsive utility class to one route-specific button. Desktop behavior is unchanged. The test-only flake fix preserves the same WorkspaceStore event path while removing live-provider tool-selection variance.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `high` • Cost: `$23.02`_

PR #3741, opened 2026-07-20
🤖 fix: archive all sidebar variants

coadler:
## Summary

Adds a context menu action to variant group rows in the sidebar so all variants can be archived together without expanding the group and archiving each chat individually.

## Implementation

- adds an **Archive all variants** action on right-click and supports the existing archive keyboard shortcut when the group row is focused
- preflights every variant before changing the group to avoid leaving it partially archived
- combines streaming and untracked-file warnings into one group confirmation
- archives each member through the existing workspace archive path
- adds UI coverage for opening the group context menu and archiving every member

## Risks

Low and limited to sidebar variant-group archive behavior. Individual workspace archive behavior continues to use the existing preflight, confirmation, and error handling paths.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `medium` • Cost: `$13.98`_

### DIFF
--- a/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
+++ b/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
@@ -1226,7 +1226,7 @@ describe("ProjectSidebar multi-project completed-subagent toggles", () => {
     expect(parentRow.getAttribute("aria-describedby")).toBe("workspace-status-description-parent");
   });
 
-  test("renders variants groups with a shared row and labeled members when expanded", async () => {
+  test("archives a variants group from its context menu and renders labeled members when expanded", async () => {
     window.localStorage.setItem(EXPANDED_PROJECTS_KEY, JSON.stringify(["/projects/demo-project"]));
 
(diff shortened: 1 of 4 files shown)


coadler:
@codex review

coadler:
@codex review

Addressed both findings:
- the group archive keybind now stops native propagation so the selected-workspace handler cannot also fire
- bulk archive now stops immediately when any member archive fails, with coverage for the failure path

coadler:
@codex review

Addressed the portal keyboard finding by limiting row shortcuts to events targeting the group row itself, with coverage for Enter on the nested menu action.

coadler:
@codex review

Addressed the nested-menu archive shortcut finding: the shared archive keybind is handled and propagation is stopped before row-target filtering, while plain Enter/Space on menu controls remains isolated from row expansion. Added coverage for both paths.

coadler:
@codex review

Rebased onto current main, which includes #3740's deterministic replacement for the flaky live-LLM review refresh test. Local targeted tests and static checks pass after the rebase.

PR #3746, opened 2026-07-22
🤖 feat: stage arbitrary pasted/dropped files into the workspace from chat

ibetitsmike:
## Summary

Pasting, dropping, or picking any file type in the chat input now works. Images and PDFs keep the existing native provider attachment behavior. Every other file (.md, .txt, .csv, binaries, unknown extensions) is staged into the workspace at `.mux/user-attachments/<uuid>/<name>` and the outgoing message carries the existing generated `<attached-files>` notice pointing the agent at the path.

## Background

The chat input previously discarded all non-image/PDF/ZIP files during clipboard/drop extraction, so text files like `.md` or `.csv` could not be attached at all. The ZIP staged-attachment pipeline already provided exactly the needed mechanism (workspace write through the Runtime abstraction, git-excluded dir, composer chips, draft persistence, hidden notice, fork copying, edit/failure restoration), so this change generalizes it from ZIP-only to arbitrary files.

## Implementation

- Classification (`supportedAttachmentMediaTypes.ts`): provider check (image/PDF) is unchanged and evaluated first; everything else resolves to a staged media type via browser MIME, a small text extension map, or `application/octet-stream`. Staged MIME strings are validated against an RFC-style pattern (and length-capped) before landing in the notice, since the notice is backtick-delimited and reparsed.
- Browser: extractors accept all files; `processAttachmentFiles` routes provider-first with staged fallback; paste/drop/picker share one per-file handler so a single oversized or unreadable file toasts an error without rejecting the rest of the batch.
- Node (`stageWorkspaceAttachment.ts`): ZIP-only gates removed; the sanitizer preserves the original extension; staged-path normalization, history-text path extraction (fork copying), and directory listing accept any file under `.mux/user-attachments/` while keeping all traversal checks; download MIME is inferred from the extension. Legacy `.zip` paths still parse.
- File picker: `accept` filter removed.

Kept as-is: the 10 MiB cap for all staged files, the staged directory, the notice format, and the ORPC schema (already generic).

## Validation

- Dogfood UAT in a sandboxed dev-server with a live model: dropped/pasted/picked `.md`, `.txt`, `.csv`, extension-less binary, PNG, PDF, ZIP, a mixed PNG+md drop, an 11 MiB oversize batch, text-only paste fall-through, failed-send composer restoration, and the no-workspace creation view. The agent read a staged file by its notice path and echoed the content. Long-filename chip verified at 375px with no overflow.

## Risks

Medium-touch area: attachment classification and the staged path validation feed both send and fork-copy flows. Regressions would surface as wrongly-staged images/PDFs (guarded by the provider-first routing test), broken fork copying of staged files (covered by non-zip + legacy-zip path tests), or rejected uploads. Traversal checks on staged paths are unchanged in strictness.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$36.68`_

### DIFF
--- a/src/browser/features/ChatInput/AttachFileButton.tsx
+++ b/src/browser/features/ChatInput/AttachFileButton.tsx
@@ -1,15 +1,12 @@
 /**
- * Attach file button - floats inside the chat input textarea next to the voice input button.
- * Opens a native file picker for images, SVGs, PDFs, and workspace-staged ZIPs.
+ * Attach file button that opens a native picker for any file type.
+ * Images and PDFs attach natively. Other files are saved into the workspace.
  */
 
 import React, { useRef } from "react";
 import { Paperclip } from "lucide-react";
 import { Tooltip, TooltipTrigger, TooltipContent } from "@/browser/components/Tooltip/Tooltip";
 import { cn } from "@/common/lib/utils";
-
-/** Accept filter for the file picker: provider attachments plus workspace-staged ZIPs. */
-const FILE_ACCEPT = "image/*,.svg,.pdf,.zip,application/zip,application/x-zip-compressed";
 
 interface AttachFileButtonProps {
   onFiles: (files: File[]) => void;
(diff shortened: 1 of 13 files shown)


ibetitsmike:
@codex review

ibetitsmike:
Addressed: the picker is now gated by a `canStageFiles` prop. Creation/scratch composers keep a provider-only accept filter (images, SVGs, PDFs) and matching tooltip; the unfiltered any-file picker only appears in workspace composers where staging works.

ibetitsmike:
@codex review
Please take another look.

PR #3749, opened 2026-07-23
🤖 chore: bump @coder/pixel-storybook to 0.2.2

ibetitsmike:
## Summary

Bump `@coder/pixel-storybook` from 0.2.1 to 0.2.2.

## Background

0.2.2 ships two changes:

- coder/pixel-storybook#55: snapshot display names now include the story name, so multiple stories from one CSF file no longer collapse into a single dashboard group. This re-keys all snapshot groups, so the first Pixel build on this PR reports everything as new snapshots.
- coder/pixel-storybook#56: screenshots are held in memory instead of round-tripping through disk. Internal change, no config impact (we only set env vars in `pixel.yml`).

Also updates the Nix offline dep cache hash in `flake.nix` for the lockfile change.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$1.76`_

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -111,7 +111,7 @@
         "@babel/preset-env": "^7.28.5",
         "@babel/preset-react": "^7.28.5",
         "@babel/preset-typescript": "^7.28.5",
-        "@coder/pixel-storybook": "0.2.1",
+        "@coder/pixel-storybook": "0.2.2",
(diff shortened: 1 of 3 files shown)


ibetitsmike:
@codex review

PR #3758, opened 2026-07-29
🤖 fix: drop the creation headline and align docked surfaces with the transcript column

ibetitsmike:
## Summary

Removes the "Let's get building." headline from the creation page, and lines every surface in the composer dock up with the transcript column so the composer stops floating in the middle of a full-width transcript.

## Background

Both reported from the chrome redesign in #3753. Rebased onto `main` now that #3757 has merged, so this diff is only the two commits below.

The headline gave a static string the largest type on the page. It was already `sr-only` below 768px, so this removes it outright.

The alignment bug is structural. The composer dock lives *inside* the transcript scrollport and cancels its `px-[15px]` gutter with `mx-[-15px]` so it can paint full-bleed. Every surface inside the dock therefore has to re-apply that gutter and follow the transcript's width mode. They all hardcoded `px-4` (16px) plus `mx-auto max-w-4xl` instead.

Centered mode hid this completely: the transcript and the docked surfaces shared both a center line and the same `4xl` cap, so their edges coincided by accident. Turn on `chatTranscriptFullWidth` and the transcript uncaps while everything docked stays 896px wide.

## Implementation

`CHAT_DOCK_GUTTER_CLASS` (in `constants/layout`) plus `ChatDockColumnProvider` / `useChatDockColumnWidthClass` / `ChatDockSurface` (in `components/ChatPane/chatDockColumn`) now carry the contract. The dock provides the width mode; surfaces inside consume it.

Applied to the composer section, `ChatInputDecoration` (todos, background bashes, reviews, queued messages, chat instructions), `ConcurrentLocalWarningDecoration`, and `TranscriptOnlyNoticePane`. The second commit sweeps three decoration-lane entries that never followed the column in *either* mode: the compaction warning and context-switch banner carried their own `mx-4`, and the pre-stream task card had no gutter at all. All three go through `ChatDockSurface`, and the two `mx-4` values are removed so the wrapper is the single source of the inset.

The width mode resolves through context rather than calling `useChatTranscriptFullWidth` per surface, because that hook also issues a `config.getConfig` fetch and an `onConfigChanged` subscription per instance, and the dock can hold several decorations at once.

## Validation

Measured at 1500px with the left sidebar collapsed: in full-width mode the transcript rows, composer box, background-bash decoration, and auto-compact label all span 35..1465; centered, all cap at 896.

Two play tests (`FullWidthTranscriptAlignment`, `CenteredTranscriptAlignment`) assert every docked surface shares the transcript column's left and right edges, including each `ChatDockSurface`-wrapped entry. Four red-green toggles:

| toggle | failure |
| --- | --- |
| width class reverted to the hardcoded cap | `Docked composer is inset ... by 157px left and 157px right` |
| gutter reverted to `px-4` | `... by 1px left and 1px right` |
| compaction warning unwrapped | `No ChatDockSurface-wrapped entry rendered` |
| transcript pinned to centered | `Transcript is 896px wide, so this story is not in full-width mode` |

The centered story separately asserts the scrollport has room to exceed the cap, so it proves the cap is what limits the column rather than the available width.

One trap worth recording: `updatePersistedState(CHAT_TRANSCRIPT_FULL_WIDTH_KEY, true)` does **not** enable full width in a story, because `useChatTranscriptFullWidth` syncs from `config.getConfig` and overwrites it. The story would silently run centered and pass with the bug present. `setupSimpleChatStory` now takes `chatTranscriptFullWidth` and threads it to the mock config.

## Risks

Expect Pixel diffs on any story showing the compaction warning, context-switch banner, or pre-stream task card: those move from a 16px full-bleed inset onto the composer column, which is the intended correction.

Not fixed here, pre-existing and unrelated to width mode: in centered mode the composer's toast overlay is positioned against the outer section rather than the inner column, so it spans nearly the whole dock. Moving its containing block would also shift it vertically, so it needs its own change.

---

_Generated with `mux` • Model: `anthropic:claude-opus-5` • Thinking: `xhigh`_


ibetitsmike:
@codex review

ibetitsmike:
@codex review

PR #3760, opened 2026-07-29
🤖 fix: align composer pickers and size local workers by memory

ibetitsmike:
## Summary

Aligns the model selector with the adjacent agent picker and makes local/default Jest and ESLint worker counts respect shared cgroup memory pressure.

## Background

The model selector reserved a checkmark column on every row, which shifted its leading icons away from the agent picker alignment. During validation, local lint and test runs also exhausted a 32 GiB shared cgroup because the visible 96-core host and leaf cgroup did not expose the tighter ancestor memory cap.

## Implementation

- Shares panel chrome and option-row styling between the model and agent pickers.
- Distinguishes hover with a neutral row background and selection with an accent provider icon and label, while preserving `aria-selected` and the separate default-model star.
- Walks cgroup v1 and v2 ancestors to find the effective memory constraint, accounts for broader equal-cap usage, discounts reclaimable file cache, and reserves headroom before choosing local worker counts.
- Lints and formats the new runtime helper, with a one-worker ESLint fallback if budget detection fails.

CI integration jobs still pass an explicit `--maxWorkers=100%` on fixed runners. This PR changes the default worker budget used by local commands and other invocations that do not override it.

## Validation

- `make static-check`
- `bun x jest tests/workerBudget.test.ts --runInBand --silent` (13 tests)
- `bun test ./src/browser/components/AgentModePicker/AgentModePicker.test.tsx` (5 tests)
- Storybook visual comparison of normal, hovered, selected, and selected-plus-hover model rows

## Risks

The worker profiles use measured repository workloads and may need retuning if tool memory behavior changes. CI's explicit worker override is unchanged.

> Mux prepared this PR on behalf of Mike.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `max`_

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -737,14 +737,15 @@ export default defineConfig([
     },
   },
   {
-    // Workflow/action/runtime sources are plain JS evaluated outside the TS
-    // program (QuickJS, skill assets, or generated child-process wrappers), so
-    // type-aware rules cannot apply. Lint them with core untyped rules so typos
-    // and dead helpers fail loudly instead of becoming silent sandbox globals.
+    // Workflow/action/runtime and script helper sources are plain JS evaluated outside the TS
+    // program (QuickJS, skill assets, generated child-process wrappers, or local tooling), so
+    // type-aware rules cannot apply. Lint them with core untyped rules so typos and dead helpers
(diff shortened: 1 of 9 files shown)


ibetitsmike:
@codex review

PR #3774, opened 2026-07-31
🤖 fix: harmonize concurrent local agent warning

ammar-agent:
## Summary

Restyles the concurrent local-agent warning as a standard composer decoration row and keeps it mounted across brief activity-state handoffs, reducing visual mismatch and flicker.

## Implementation

- aligns icon, height, spacing, typography, color tokens, and transcript-column width with neighboring composer decorations
- truncates the warning cleanly at phone widths instead of wrapping or overflowing
- reuses the existing streaming-status transition hold so transient idle snapshots do not blink the warning
- adds a behavioral regression test and a Pixel-pinned phone story

## Validation

- `bun test src/browser/components/ConcurrentLocalWarning/ConcurrentLocalWarning.test.tsx`
- `make typecheck`
- `make static-check`
- visually inspected Storybook at 1200px and 390px widths

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `high` • Cost: `$7.99`_


ammar-agent:
@codex review

PR #3776, opened 2026-07-31
🤖 fix: cancel stale background monitor wakes

ammar-agent:
## Summary

Prevents stale background bash monitor output from creating a synthetic follow-up turn after an agent has intentionally canceled the monitor, and adds explicit steering to clean up monitors before a final response when future wake-ups are no longer useful.

## Background

Monitored bash processes are intentionally workspace-lifetime, so they can wake an agent after its current response. That is useful for deliberate long-running watchers, but it also meant irrelevant monitors left behind at turn end could clutter the transcript. Worse, explicit `task_terminate` flushed coalesced monitor matches during teardown and did not retract an already-persisted, queued, or dequeued-but-still-preparing synthetic wake, so attempting cleanup could itself trigger the late turn.

## Implementation

- Treat explicit background process termination as monitor cancellation: discard in-memory pending matches and emit a cancellation disposition.
- Preserve pending match delivery for natural process exit, configured timeout, and shutdown recovery.
- Tombstone canceled monitors in `WorkspaceService`, supersede persisted wake records, and selectively remove queued synthetic wake entries without clearing unrelated queued work.
- Carry an internal cancellation signal through `MessageQueue` into `AgentSession`, so cancellation also aborts a wake already dequeued into `PREPARING`.
- Atomically delete only the recorded rows for the not-yet-accepted wake, preserving later rows from concurrent history writers such as goal pause boundaries. If deletion reports an error, verify durable history to distinguish a committed rewrite from rows that remain.
- Treat goal synchronization as the cancellation point of no return: once goal state can observe the durable wake row, a concurrent cancellation lets that wake finish acceptance. If goal sync then fails, finalize the wake as accepted so startup recovery can resume the existing row without redelivering it.
- Finalize durable wakes before disposed-session early returns. Accepted chat rows persist their wake snapshot identity, so later drains and restarts suppress duplicate turns while retrying transient wake-store delivery failures.
- Serialize match, cancellation, and re-arm persistence by process key so cleanup from an old process generation cannot supersede a new monitor reusing the same display-derived ID.
- Keep mixed wake batches safe by re-draining records for unrelated monitors when one monitor in the batch is canceled.
- Update the bash tool contract, background-monitor skill, and completion discipline to make post-response wake behavior and cleanup expectations explicit.

## Validation

- `bun test src/node/services/historyService.test.ts`
- `bun test src/node/services/backgroundProcessManager.test.ts`
- `bun test src/node/services/workspaceService.test.ts --test-name-pattern 'bash monitor wakes'`
- `bun test src/node/services/tools/task.bash.test.ts src/node/services/tools/bash_background_terminate.test.ts`
- `bun run typecheck`
- `make fmt`
- `make static-check` (repeated after review fixes and after rebasing onto the latest `main`)

## Risks

Moderate but localized to monitored background bash teardown and internal queued-send cancellation. Intentional monitors still survive turn end and wake normally; only explicit termination changes from flushing to canceling undelivered monitor output. Natural exit, timeout, and shutdown retain their previous delivery semantics. Cancellation signals are internal-only and applied to sealed synthetic queue entries, so user-authored queue behavior remains unchanged.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `high` • Cost: `$225.23`_


ammar-agent:
@codex review

Please review the explicit monitor cancellation semantics, queued wake retraction, and preservation of natural exit/timeout wake delivery.

ammar-agent:
Addressed both Codex findings:

- **P1 (dequeued/PREPARING wake):** monitor wake batches now carry an internal `AbortSignal` through `MessageQueue` into `AgentSession`. Explicit cancellation aborts even after dequeue; `AgentSession` rolls back any not-yet-accepted history rows, suppresses transcript emission/stream startup, releases `PREPARING`, and continues later queue entries.
- **P2 (reused process ID):** monitor match, cancellation retirement, and re-arm persistence now share the per-process registry lock. The re-arm clears the cancellation tombstone only after old-generation supersession completes, and new match persistence queues behind that re-arm.

Added regression tests for both races and reran the full targeted queue/monitor suites plus `make static-check` after rebasing onto current `main`.

ammar-agent:
@codex review

Please re-review the fixes for dequeued PREPARING cancellation and reused process-ID serialization.

ammar-agent:
Addressed the latest Codex findings:

- **Goal reconciliation:** preparing-wake cancellation now atomically truncates the temporary turn rows and then reruns `syncGoalModeWithChatTail()` before cancellation is finalized, restoring goal mode from the post-deletion chat tail.
- **Rollback failure:** cancellation no longer supersedes the durable wake up front when a queued/preparing dispatch owns it. If atomic truncation fails, the cancellation is disabled for that dispatch, `onCanceled` is not called, and the persisted wake continues normal acceptance so no hidden synthetic row is left for a later request.

Added focused tests for both paths, reran the complete queue/monitor test set, and passed `make static-check`.

ammar-agent:
@codex review

Please re-review the goal reconciliation and rollback-failure handling for canceled PREPARING monitor wakes.

ammar-agent:
Addressed both new Codex findings by tightening the acceptance boundary:

- **Post-write truncation errors:** after `truncateAfterMessage()` returns an error, the send now reads durable history and checks every row from the preparing turn. If those rows are absent, cancellation finalizes; if any remain (or verification fails), cancellation is disabled and the persisted wake continues normal acceptance.
- **Goal-sync failures:** cancellation is now allowed only before goal synchronization begins. Goal sync is the explicit point of no return; a stop arriving during/after it lets the durable wake finish acceptance, so no history rollback can race goal mutation and no canceled registration is stranded on reconciliation errors.

Added focused tests for committed post-write failures and cancellation during goal sync, then reran the full queued-dispatch suite and `make static-check`.

ammar-agent:
@codex review

Please re-review the durable rollback verification and explicit goal-sync point of no return.

ammar-agent:
Addressed the latest Codex findings:

- **Concurrent history writers:** added `HistoryService.deleteMessages()` for a single atomic rewrite that deletes exactly the preparing wake's recorded IDs. Later rows (including goal pause boundaries) are preserved, and missing targets fail before any rewrite.
- **Goal-sync failure after point of no return:** if cancellation arrived but goal synchronization throws after the durable row crossed the boundary, `onAccepted` is invoked before propagating the startup failure. The wake record is finalized and startup recovery can resume the existing row without a duplicate synthetic wake.

Also retained durable-history verification for post-write delete errors and added focused HistoryService/AgentSession regression tests. Full HistoryService, queued-dispatch, monitor-wake suites and `make static-check` pass locally.

ammar-agent:
@codex review

Please re-review the targeted atomic history deletion and accepted-wake finalization on goal-sync failure.

ammar-agent:
Addressed the remaining Codex finding: every cancelable monitor wake is now finalized via `onAccepted` on any goal-sync failure after the durable row crosses the cancellation point of no return, regardless of whether the abort signal fired concurrently. The focused test now covers a sync failure with an un-aborted cancellation signal, and `make static-check` passes.

ammar-agent:
@codex review

Please re-review the unconditional post-boundary wake finalization.

ammar-agent:
Addressed both findings:

- **Disposed-session success path:** after the goal-sync point of no return, a disposed session now invokes `onAccepted` before its early return, finalizing the durable wake without emitting or starting a stream.
- **Wake-store transition failures:** `delivered` is now set only after `markDeliveredAfterAccepted` succeeds. Failed attempts leave retryable state, and the normal post-send/accepted-failure paths retry instead of becoming permanent no-ops.

Added focused tests for disposed-session finalization and a first-attempt wake-store failure followed by successful retry. Typecheck, tests, formatting, and `make static-check` pass.

ammar-agent:
@codex review

Please re-review disposed-session finalization and retryable wake-store delivery.

ammar-agent:
Addressed the delivery-retry lifetime finding with durable accepted-history recovery:

- Accepted wake metadata now includes the process ID plus the exact persisted wake `updatedAt` snapshot.
- Every drain (including startup recovery) scans history for matching accepted snapshots before sending. Matching pending records are delivery-only reconciliation candidates: Mux retries `markDelivered`, but suppresses duplicate synthetic turns even if the wake store remains unavailable.
- Snapshot identity prevents suppressing newer matches merged for the same process after the accepted turn.

Added regression coverage for persistent wake-store failures across recovery drains; typecheck, wake metadata tests, monitor wake tests, formatting, and `make static-check` pass.

ammar-agent:
@codex review

Please re-review durable accepted-history suppression and delivery-only reconciliation.

ammar-agent:
Addressed all three recovery findings:

- Accepted recovery now uses `markDeliveredSnapshot(record)`; if newer matches merged, only the accepted snapshot is consumed and the remainder is scheduled for a separate drain.
- Historical wake metadata is runtime-validated (`records` must be an array; entries and identity fields must be objects/strings), so malformed rows are skipped without disabling recovery.
- Full chat clears now retire all pending monitor wakes before deleting history. If retirement fails, the clear is rejected, preserving the acceptance proof and preventing stale output from reappearing in an empty transcript.

Rebased onto current `main` (including #3779), reran targeted recovery tests and `make static-check` after the rebase.

ammar-agent:
@codex review

Please re-review snapshot-aware accepted recovery, malformed metadata handling, and full-clear retirement.

ammar-agent:
Addressed both edge cases:

- Accepted-history scanning now rejects empty/whitespace-only `processId` and `wakeUpdatedAt` fields before snapshot-key construction, in addition to the existing runtime shape checks.
- Pending-wake retirement is centralized and invoked for every destructive clear: percentage truncation uses `>= 1.0`, and destructive `replaceHistory()` retires wakes before calling `clearHistory()`. A retirement failure blocks the destructive clear so acceptance proof cannot be lost.

Added coverage for empty identifiers, percentages above 1, and destructive history replacement. Rebased onto current `main` and reran `make static-check` afterward.

ammar-agent:
@codex review

Please re-review non-empty recovery identifiers and centralized destructive-clear retirement.

ammar-agent:
Addressed both clear-boundary findings:

- Added a per-workspace monitor/history mutex. Monitor match persistence and wake drains use the same lock as destructive history retirement+deletion, eliminating the enqueue/accept gap between separate awaits.
- Added an AgentSession hard-restart clear coordinator supplied by WorkspaceService, so `execSubagentHardRestart()` routes through the same locked retirement boundary instead of calling `HistoryService.clearHistory()` directly.

Typecheck, monitor-wake/clear tests, formatting, and `make static-check` pass.

ammar-agent:
@codex review

Please re-review serialized destructive clears and exec hard-restart routing.

ammar-agent:
Addressed failed-clear restoration:

- Wake retirement now returns the staged pending snapshots.
- The serialized clear coordinator restores those snapshots when the history clear returns an error or throws.
- `supersedeAllPending()` also rolls back partial staging if a later wake transition fails.

Added a failed full-clear regression test proving the pending wake is restored; typecheck, tests, formatting, and `make static-check` pass.

ammar-agent:
@codex review

Please re-review reversible wake retirement on failed destructive clears.

ammar-agent:
Addressed both clear-outcome findings:

- Armed/stopped lifecycle store mutations now share the per-workspace history lock, so an explicit cancellation cannot be overwritten by failed-clear restoration.
- Failed clears compare accepted snapshot proof before and after the operation. Unaccepted wakes are restored; accepted wakes are restored only if their proof still exists. Partially committed clears therefore keep affected accepted wakes retired.

Added a partial-clear regression test; typecheck, monitor/clear tests, formatting, and `make static-check` pass.

ammar-agent:
@codex review

Please re-review failed/partial clear restoration and serialized lifecycle mutations.

ammar-agent:
Addressed all four findings:

- **Lock re-entry:** monitor-originated streams record that the workspace monitor/history lock is already held. Exec hard restart uses an already-locked clear path for that call chain and the normal locked path otherwise.
- **Startup recovery:** stale registry conversion now uses the same per-workspace history lock plus per-process registry lock as live lifecycle mutations and destructive clears.
- **Pre-clear scan failures:** accepted-wake history scanning now happens before any retirement; a failed scan blocks the clear without changing wake state.
- **Partial truncation deleting all rows:** every positive-percentage truncation is coordinated. Unaccepted wakes are restored after successful partial truncation, while accepted wakes remain retired only when their acceptance marker was actually deleted.

Added coverage for scan failures and sub-100% truncation deleting the entire transcript. Full targeted suites and `make static-check` pass.

ammar-agent:
@codex review

Please re-review lock re-entry handling, startup recovery serialization, pre-clear verification, and partial all-row truncation.

ammar-agent:
Addressed both findings:

- **Queued lock ownership:** `monitorHistoryLockHeld` is now an explicit internal dispatch field. Immediate monitor sends set it true; queued wakes preserve false through MessageQueue, so hard restart acquires the coordinator lock after deferred dispatch.
- **Startup recovery discovery:** WorkspaceService retains the constructor recovery promise, and destructive clears await the whole recovery operation before acquiring the history mutex. A clear cannot overtake owner discovery and later receive a stale recovered wake.

Added focused MessageQueue and destructive-clear gating tests; typecheck, tests, formatting, and `make static-check` pass.

ammar-agent:
@codex review

Please re-review explicit queued lock ownership and startup recovery gating.

ammar-agent:
Addressed both findings:

- **Actual dispatch ownership:** monitor wakes carry a mutable lock-state object. The real WorkspaceService queue path flips it to `held=false` immediately before enqueue, including races during async send preparation. Immediate sends retain true; deferred MessageQueue dispatches therefore reacquire the mutex during hard restart.
- **Recovery gate for lock-held restart:** all monitor drains now await the complete constructor recovery promise before acquiring the workspace history mutex. A stream can only claim lock ownership after startup owner discovery/conversion has finished.

Focused dispatch-state/recovery tests and `make static-check` pass.

ammar-agent:
@codex review

Please re-review actual queue-derived lock state and recovery-gated drains.

PR #3783, opened 2026-08-02
🤖 fix: avoid duplicate subagent completion responses

ammar-agent:
## Summary

Prevents background sub-agent completion from triggering a second, redundant assistant ending when the parent already responded to that child’s latest incremental progress update.

## Background

`agent_report` progress updates are themselves synthetic wake turns. After the parent responds, the child’s terminal report is injected into history and the terminal-attention notifier currently sends another generic prompt beginning “Background sub-agent task(s) have completed… Write the final response now.” That forces another model response even though the chat already reached a natural conclusion, creating confusing double endings.

## Implementation

- Inspect the active context epoch before delivering completed agent-task attention.
- Suppress only successful completion wakes where the parent has an assistant response after that child’s latest `in_progress` report.
- Keep the terminal report visible by persisting it with `uiVisible: true` and emitting the live report card event.
- Preserve the generic terminal wake when the latest progress update has not been answered, when no progress update exists, and for failures, workspace turns, and workflow results.
- Track the latest progress report rather than any historical update, so a newer unprocessed progress message still receives a terminal handoff.

## Validation

- Full `src/node/services/taskService.test.ts` suite: 327 passing tests.
- Targeted terminal-attention, progress-report, and report-card tests.
- `make static-check`.

## Risks

Low-to-moderate and scoped to successful background sub-agent terminal attention. The suppression decision fails open: history read failures, compaction that moves the progress turn out of the active context, or an unanswered latest update retain the existing completion wake.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `high` • Cost: `$272.47`_

### DIFF
--- a/src/node/services/agentSession.startupAutoRetry.test.ts
+++ b/src/node/services/agentSession.startupAutoRetry.test.ts
@@ -14,6 +14,7 @@ import { DEFAULT_RUNTIME_CONFIG } from "@/common/constants/workspace";
 import type { WorkspaceMetadata } from "@/common/types/workspace";
 import { Ok } from "@/common/types/result";
 import { GOAL_CONTINUATION_KIND } from "@/constants/goals";
+import { formatSubagentReportEnvelope } from "@/common/utils/subagentReportEnvelope";
 import { WORKSPACE_DEFAULTS } from "@/constants/workspaceDefaults";
 
 interface AutoRetryResumeRequest {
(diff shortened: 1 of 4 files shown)


ammar-agent:
@codex review

Please review the progress-aware completion-wake suppression and visible terminal report fallback.

ammar-agent:
Addressed both findings:

- Completion suppression now also requires a matching durable, `uiVisible` completed report card in the active context. Early-return delivery paths or append failures retain the generic terminal wake.
- Completed sub-agent report envelopes are explicitly non-retryable during startup recovery, even though their cards are UI-visible.

Added missing-card and startup auto-retry regression tests; targeted suites and `make static-check` pass.

ammar-agent:
@codex review

Please re-review durable terminal-card verification and startup retry exclusion.

ammar-agent:
Addressed both startup recovery gaps:

- Only `uiVisible` completed report cards are excluded from startup retry. Hidden completed reports keep the existing fallback if terminal attention was not persisted.
- The visible-card early return now runs only when there is no recoverable assistant partial; an interrupted parent stream still schedules startup recovery.

Added regression tests for hidden terminal reports and visible cards with assistant partials; targeted suites and `make static-check` pass.

ammar-agent:
@codex review

Please re-review startup recovery behavior for hidden reports and assistant partials.

PR #3791, opened 2026-08-03
🤖 fix: avoid subagent report interruption flash

ammar-agent:
## Summary

Prevents a completed subagent report card from briefly putting the parent workspace into a phantom response lifecycle and flashing the **Stream interrupted** barrier.

## Background

When a parent had already answered an incremental subagent progress update, the terminal report was appended as a visible synthetic user message without starting another parent response. The frontend treated that display-only card as a new main-thread request, both setting pending-stream state and selecting it as an unanswered retry candidate even though no stream would follow.

## Implementation

- detect visible synthetic completed subagent report envelopes in the streaming message aggregator
- keep rendering the report card while preserving the parent workspace's existing idle stream lifecycle
- exclude display-only completed report cards from main-turn retry candidate selection
- add a regression test proving the card remains visible without pending-stream or interrupted-retry state

## Validation

- `bun test src/browser/utils/messages/StreamingMessageAggregator.test.ts src/common/utils/messages/retryEligibility.test.ts`
- `make static-check`

## Risks

Low. The exception is limited to completed subagent report envelopes explicitly marked synthetic and UI-visible; regular user messages and hidden report prompts retain the existing stream-start behavior.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `high`_

### DIFF
--- a/src/browser/utils/messages/StreamingMessageAggregator.test.ts
+++ b/src/browser/utils/messages/StreamingMessageAggregator.test.ts
@@ -3,6 +3,7 @@ import { CONTEXT_BOUNDARY_KINDS } from "@/common/constants/contextBoundary";
 import { createMuxMessage, type DisplayedMessage } from "@/common/types/message";
 import { formatSubagentReportEnvelope } from "@/common/utils/subagentReportEnvelope";
 import { buildWorkflowRunCardMessage } from "@/common/utils/workflowRunMessages";
+import { getInterruptionContext } from "@/common/utils/messages/retryEligibility";
 import { shouldNotifyOnResponseComplete } from "./responseCompletionMetadata";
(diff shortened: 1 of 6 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Addressed the retry-candidate gap: visible completed report cards now preserve idle pending-stream state and are excluded from main-turn retry selection. The regression test asserts both invariants.

ammar-agent:
@codex review

Please review the latest force-pushed revision; it includes the retry-candidate exclusion and regression assertion.

ammar-agent:
@codex review

Addressed the visibility distinction: displayed user rows now retain the UI-visible marker, so only normal-transcript completed report cards are excluded from retry selection. Hidden synthetic reports remain retry candidates, with regression coverage for debug display mode.

PR #3793, opened 2026-08-04
🤖 feat: show task kind and spawn intent in single-task task_await summary

ibetitsmike:
## Summary

For a single-task `task_await`, the collapsed transcript row now shows what actually finished: the task kind (`bash`, `workflow`, `workspace`, or the sub-agent type) plus the spawn intent or title, e.g. `1 task completed · bash · Resolving pagination feedback and requesting another review`.

## Background

The collapsed row previously said only `1 task completed`, which conveys nothing about what finished, forcing the user to expand every await row to see which background work landed.

## Implementation

- Kind is derived from the task handle (`bash:` → bash, `wfr_` → workflow, `wst_` → workspace) or from the spawning `task` call's `agentId`/`subagent_type`.
- `task_await` results don't carry `model_intent`, so the existing render-time `computeTaskReportLinking` pass (which already links spawn titles) now also indexes background `bash` spawns by taskId and captures their `model_intent` + script.
- The intent is sanitized via a shared `sanitizeDisplayableModelIntent` helper (extracted from `buildBashCollapsedSummary`) so an intent that merely restates the command is rejected, falling back to the result title (bash `display_name` or agent report title).
- Multi-task summaries stay count-only to avoid noise; the detail renders in the existing truncating muted span so long text clips gracefully at narrow widths.

Scope note: only the completed state is enriched; the `Waiting for N tasks` state could get the same suffix via the same helper if desired.

## Validation

- 5 new tests in `TaskToolCall.test.tsx` (bash intent, command-echo intent fallback, title fallback, agent type, multi-task stays count-only), red-green verified against the summary wiring and the echo-intent fallback.
- Sandbox UAT with seeded real sessions: verified bash and explore awaits at desktop and 375px mobile widths (no overflow), plus expanded-row rendering.

---

### DIFF
--- a/src/browser/features/Tools/TaskToolCall.test.tsx
+++ b/src/browser/features/Tools/TaskToolCall.test.tsx
@@ -4,7 +4,9 @@ import { GlobalWindow } from "happy-dom";
 
 import { TooltipProvider } from "@/browser/components/Tooltip/Tooltip";
 
+import type { DisplayedMessage } from "@/common/types/message";
 import type { FrontendWorkspaceMetadata } from "@/common/types/workspace";
+import { computeTaskReportLinking } from "@/browser/utils/messages/taskReportLinking";
 
 let workspaceContextMock: {
   workspaceMetadata: Map<string, FrontendWorkspaceMetadata>;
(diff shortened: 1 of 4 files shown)


ibetitsmike:
@codex review

ibetitsmike:
> Mux is acting on Mike's behalf.

ibetitsmike:
@codex review

PR #3796, opened 2026-08-05
🤖 feat: show per-task kinds and intents in multi-task task_await summaries

ibetitsmike:
## Summary

Multi-task `task_await` completion rows previously said only `N tasks completed`, hiding what actually finished. The collapsed row now lists each completed task's kind and spawn intent/title: a single detail stays inline in the header (unchanged from #3793), while multiple details render as stacked truncated lines below it.

## Background

#3793 added the `kind · intent/title` detail for single-task awaits but deliberately kept multi-task summaries count-only. In practice that forces expanding every multi-task await row to see which background work landed. A comma-joined inline list was considered first but truncates quickly and reads poorly because `,` and `·` separators compete, so multiple details stack as one line per task.

## Implementation

- The single-task detail computation now runs for every completed result, reusing the same kind derivation (`bash:`/`wfr_`/`wst_` handles or spawn `agentId`) and `sanitizeDisplayableModelIntent` fallback chain per task.
- With exactly one detail it remains the inline header suffix; with more, a stacked list renders below the header, gated to collapsed + success state since the expanded view already shows per-task result rows.
- Tasks with no resolvable kind/intent/title are skipped rather than rendering empty lines.

## Validation

- Storybook screenshots (built Storybook, dark theme) comparing inline comma-joined vs stacked layouts for 1/2/3-task completions; stacked layout verified to align with header text and truncate per line.

---

### DIFF
--- a/src/browser/features/Tools/TaskToolCall.test.tsx
+++ b/src/browser/features/Tools/TaskToolCall.test.tsx
@@ -542,20 +542,61 @@ describe("TaskAwaitToolCall", () => {
     expect(view.queryByText(/Investigation Complete/)).toBeNull();
   });
 
-  test("keeps multi-task completion summaries count-only", () => {
+  test("shows each task's kind and intent for multi-task completions", () => {
+    const bashSpawn = createToolMessage({
+      toolName: "bash",
+      args: {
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #3800, opened 2026-08-05
🤖 fix: collapse subagent reports by default

ammar-agent:
## Summary

Collapse subagent report details by default so transcript timelines remain scannable, while keeping the report body and structured output available through accessible disclosure controls.

## Implementation

- turn the report header into an accessible expand/collapse control
- hide report markdown until the user expands the report
- preserve the existing nested structured-output disclosure
- cover collapsed, expanded, legacy, and structured-output behavior in message renderer tests

## Validation

- `bun test src/browser/features/Messages/MessageRenderer.test.tsx`
- pinned 390×844 Storybook verification: both reports defaulted to `aria-expanded=false` with no horizontal overflow
- `make static-check`

## Risks

Low and limited to subagent report presentation in the transcript. Copy behavior and report envelope parsing are unchanged.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `high` • Cost: `$0.00`_

### DIFF
--- a/src/browser/features/Messages/MessageRenderer.test.tsx
+++ b/src/browser/features/Messages/MessageRenderer.test.tsx
@@ -250,7 +250,7 @@ describe("MessageRenderer subagent report rows", () => {
     };
   }
 
-  test("presents incremental synthetic reports without exposing the protocol envelope", () => {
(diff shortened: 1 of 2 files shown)


ammar-agent:
@codex review

PR #3803, opened 2026-08-05
🤖 fix: repair transcript tail ordering without migrations

ammar-agent:
## Summary

Repair transcript-tail ordering for both existing idle histories and new live reports, without requiring a migration or newly persisted metadata. Completed sub-agent cards are reconstructed from transcript structure, exact anchors remain optional precision, and unrelated tail accumulators no longer pile up.

## Background

Completed sub-agent cards are durable synthetic user messages appended after the parent assistant row, but they do not start a new turn. Sorting only by `historySequence` therefore leaves them after the parent’s final response. Earlier anchor-only fixes helped only reports carrying new metadata and missed old, stale, invalid, or tail anchors.

The same audit found two related tail defects: splitting an assistant around an inserted card could duplicate terminal error/refusal rows, and repeated `/plan show` commands created unique maximum-sequence previews that accumulated at the bottom.

## Implementation

- derive historical card placement from the matching progress turn and its first assistant response
- repair trailing legacy cards against the immediately preceding assistant within the current context epoch
- use valid anchors only as optional within-response precision; settled tail anchors fall back to a stable message boundary
- prevent crossing context boundaries or reparenting into later unrelated turns
- emit assistant terminal decorations only once when an inserted card splits the response
- preserve `historySequence` on live `stream-end` events for reconnect reconstruction
- make the ephemeral `/plan show` preview a replace-in-place singleton

## Validation

- `make static-check` after rebasing onto current `main`
- `StreamingMessageAggregator.test.ts`: 111 passing tests
- `chatCommands.test.ts`: 81 passing tests
- full `taskService.test.ts`: 342 passing tests
- focused `streamManager.test.ts` reconnect metadata coverage
- regression matrix covers old unanchored cards, missing/stale anchors, reasoning/tool-only responses, later unrelated turns, context boundaries, active vs settled tail anchors, and duplicate terminal rows

## Risks

Moderate and localized to transcript projection. Historical exact mid-stream offsets cannot be reconstructed when no anchor exists, so those cards use a deterministic message boundary. The renderer refuses to cross context boundaries or unrelated user turns; ambiguous cases remain chronological rather than guessing.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `max` • Cost: `$152.55`_

### DIFF
--- a/src/browser/utils/chatCommands.test.ts
+++ b/src/browser/utils/chatCommands.test.ts
@@ -13,6 +13,8 @@ import {
 import { parseCommand } from "./slashCommands/parser";
 import type { CommandHandlerContext, SlashCommandContext } from "./chatCommands";
 import type { ReviewNoteData } from "@/common/types/review";
+import { useWorkspaceStoreRaw, workspaceStore } from "@/browser/stores/WorkspaceStore";
+import type { FrontendWorkspaceMetadata } from "@/common/types/workspace";
 import { HEARTBEAT_DEFAULT_INTERVAL_MS } from "@/constants/heartbeat";
 
 // Simple mock for localStorage to satisfy resolveCompactionModel and experiment gating.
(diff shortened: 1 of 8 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

The PR was substantially revised after reproducing the remaining failures on `5887c541c`: correctness now repairs existing idle histories at render time and no longer depends on newly persisted anchor state. It is rebased on current `main`; please review the full diff.

ammar-agent:
Addressed `PRRT_kwDOPxxmWM6Wz3XU`: the forward progress-response scan now skips display-only terminal cards, so another child completion cannot abort causal placement. Added a regression with two completed reports before the assistant response; full transcript tests and `make static-check` pass.

@codex review

ammar-agent:
Rebased onto the latest `main` after #3804 merged; post-rebase transcript suites and `make static-check` pass.

@codex review

ammar-agent:
Addressed `PRRT_kwDOPxxmWM6Wz-wt`: no-progress fallback is now disabled whenever a context boundary precedes the completion card, preserving chronological order for old tasks completing in a new epoch. Added a regression with a reset followed by an unrelated new user/assistant turn; 113 transcript tests and `make static-check` pass.

@codex review

ammar-agent:
Addressed `PRRT_kwDOPxxmWM6W0DAu`: a valid same-epoch anchor now takes precedence over the earlier response inferred from the task's progress turn. Added a regression where completion occurs during a later active parent response; 114 transcript tests and `make static-check` pass.

@codex review

PR #3798, opened 2026-08-05
🤖 feat: add Grok 4.5 priority processing

ammar-agent:
## Summary

Adds Grok 4.5 as the sole curated xAI model and integrates xAI Priority Processing with Mux's shared thinking and Fast mode controls, exact billed-cost tracking, and native web/X search.

## Background

xAI exposes `grok-4.5` with configurable low/medium/high reasoning effort. Earlier built-in Grok entries were retired on May 15, 2026 and now redirect to newer models, so keeping them in Mux's curated index is misleading. xAI Priority Processing uses `service_tier: "priority"` and is billed at 2× standard token rates when granted.

## Implementation

- replaces the curated Grok 4.1 Fast and Grok Code entries with `xai:grok-4.5`
- adds Grok 4.5 token limits, tiered pricing, capabilities, and a low/medium/high reasoning policy
- plugs Grok into the shared `ThinkingSelector`; the configured minimum-thinking floor controls whether Low is shown, and Grok never exposes Off or Pro mode
- generalizes the shared Fast mode selector, keyboard shortcut, command palette action, and closed-state indicator to the active direct provider
- keeps OpenAI and xAI Fast mode state independent: OpenAI uses its existing restore-tier behavior, while Grok toggles xAI between default and priority
- hides xAI Fast mode on gateway routes and for legacy/non-Grok-4.5 xAI model strings
- adds an xAI provider setting for standard or fast processing with serialized, failure-aware persistence
- injects xAI's `service_tier` request field through the provider fetch wrapper
- uses the xAI Responses path for Grok 4.5 and reconciles usage to exact `costInUsdTicks` billing metadata across multi-step turns
- updates and patches the xAI SDK for current Responses limits, then translates existing xAI Live Search configuration into Responses-native web and X search tools, including source filters, disabled mode, and required first-step search for `mode: "on"` on direct xAI routes
- keeps older custom Grok strings on Chat Completions for legacy compatibility
- updates compaction coverage and regenerates model documentation and embedded skill content

## Validation

- `make static-check`
- targeted unit suite for Fast mode provider resolution, service-tier persistence, command availability, provider config, provider service, and settings regressions: 154 tests passed
- full-app Jest UI integration covering shared thinking controls and Fast mode for OpenAI and Grok 4.5
- targeted Grok/provider regression suite covering aliases, pricing metadata, exact billing, thinking policy, native search tools, provider options, request shaping, and routing
- real xAI Grok 4.5 priority inference test in the CI integration harness
[...]

PR #3804, opened 2026-08-05
🤖 fix: honor mapped Grok 4.5 aliases

ammar-agent:
## Summary

Honor configured `mappedToModel` aliases that target Grok 4.5, and make Grok 4.5 usage meters/pricing resolve known context limits for family aliases.

## Background

#3798 added Grok 4.5 Responses routing and native search tools, but some checks used the raw runtime model id. A provider entry like `{ id: "team-grok", mappedToModel: "xai:grok-4.5" }` still took the Grok 4.5 provider-options path while model creation stayed on Chat Completions, so those requests lost both legacy Live Search and Responses-native `web_search`/`x_search`.

Separately, context usage could show "Unknown model limits" for Grok 4.5 when the selected/runtime id used a rolling alias (`-latest`) or mixed case, because stats lookup only hit the exact `xai/grok-4.5` key.

## Implementation

- resolve mapped Grok 4.5 capability through `resolveModelForMetadata` before choosing Responses vs Chat Completions
- pass the capability model into tool construction and forced first-step search so native xAI tools and prompt allowlists stay aligned
- let Fast mode treat mapped Grok aliases as xAI priority-capable on direct routes
- harden model-stats lookup for `-latest` and case variants, and publish Grok 4.5 `max_output_tokens`
- add regression coverage for mapped alias routing, native tools, Fast mode availability, and context-limit resolution

## Validation

- `bun test src/browser/utils/fastModeServiceTier.test.ts src/common/utils/tools/tools.test.ts src/node/services/providerModelFactory.test.ts src/common/utils/tokens/modelStats.test.ts src/common/utils/compaction/contextLimit.test.ts`
- `make static-check`

## Risks

Low. Behavior only expands known Grok 4.5 metadata resolution and mapped-alias capability inheritance; unmapped legacy Grok strings remain on Chat Completions.

---

_Generated with `mux` • Model: `xai:grok-4.5` • Thinking: `high` • Cost: `$159.82`_

### DIFF
--- a/src/browser/utils/fastModeServiceTier.test.ts
+++ b/src/browser/utils/fastModeServiceTier.test.ts
@@ -27,6 +27,19 @@ describe("fast mode service tier", () => {
     expect(getFastModeProvider("xai:grok-4.5", { resolvedRouteProvider: "direct" })).toBe("xai");
     expect(getFastModeProvider("xai:grok-4.5", { resolvedRouteProvider: "openrouter" })).toBeNull();
     expect(
+      getFastModeProvider("xai:team-grok", {
+        resolvedRouteProvider: "direct",
+        providersConfig: {
+          xai: {
+            apiKeySet: true,
+            isEnabled: true,
+            isConfigured: true,
+            models: [{ id: "team-grok", mappedToModel: "xai:grok-4.5" }],
+          },
+        },
+      })
+    ).toBe("xai");
+    expect(
       getFastModeProvider("xai:grok-code-fast-1", { resolvedRouteProvider: "direct" })
     ).toBeNull();
     expect(getFastModeProvider("anthropic:claude-sonnet-4-5")).toBeNull();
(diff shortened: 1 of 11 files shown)


ammar-agent:
@codex review

Please review the mapped Grok 4.5 alias follow-up: Responses routing, native search tools, forced search mode, and Fast mode now all use the shared capability resolver.

ammar-agent:
@codex review

Also fixed Grok 4.5 usage meters: family aliases/case variants now resolve the 500K context window, and max_output_tokens is published so the UI no longer reports unknown model limits.

ammar-agent:
Addressed Codex finding **“Preserve mixed-case model catalog lookups”**:

- exact catalog casing is probed first
- lowercase keys are only fallbacks for family/case variants
- added a regression covering `deepinfra:Qwen/Qwen3-14B`

@codex review

ammar-agent:
Addressed Codex finding **“Do not use the context window as the output cap”**:

- removed the invented `max_output_tokens: 500000` from Grok 4.5 metadata
- left output uncapped unless independently documented
- context-limit resolution for Grok aliases remains covered

@codex review

PR #3814, opened 2026-08-06
feat: add a timeline_event transcript card

ThomasK33:
## Summary

Adds a dedicated transcript card for the `timeline_event` tool, which currently falls back to the generic renderer. Collapsed, the agent's one-sentence note is the summary, tagged with an ask-mode category chip. Expanded, the card previews the row exactly as it lands in the Timeline tab (ask-mode ink, Sparkles ring, category badge) under its day header, and explains where to find it.

## Background

The durable workspace timeline landed in #3755/#3756 with the `timeline_event` tool, but calls render as generic JSON dumps in the transcript. The card design comes from the Mux Design System mockup (`Timeline Event Tool Call.html`), bound here to the real backend shapes rather than the mockup's idealized ones.

## Implementation

- `TimelineEventToolResultSchema` (+ derived `TimelineEventToolArgs`/`TimelineEventToolResult` types) — the real result is `{ success: true, recorded: boolean }`, where `recorded: false` means TimelineService throttled the note (duplicate or rate-limited) and nothing was written. The card surfaces that as an amber "Not recorded" chip plus an explanation instead of a preview that would overstate what happened.
- The feed-row preview is a static replica of `TimelineEventRow`'s agent-authored variant. To keep it provably in sync with the tab, the row icon comes from the shared `getTimelinePresentation("agent.event")`, and the day/time formatters (`getTimelineDayLabel`, `formatTimelineTime`) are extracted from `TimelinePanel.tsx` into `timelinePresentation.ts` and used by both. The preview's day header and time cell derive from `toolCallTimestamp` and are omitted when unavailable — nothing is fabricated.
- Result handling follows the current card conventions: `unwrapResult` for the SDK JSON container, `safeParse` for self-healing on malformed persisted results, and both persisted error shapes (`{ success: false, error }` and the nested bare `{ error }`).
- Registered in `getToolComponent.ts`; header icon (Sparkles, matching the feed's `agent.event` rows) in `TOOL_NAME_TO_ICON`.

## Validation

- New happy-dom suite covers the behavioral branches: recorded/throttled gating, badge derivation (`picked_up` → "picked up", missing category → "Agent"), both error shapes, JSON-container unwrap, and timestamp gating.
- Stories mirror the design gallery (all five categories, fallback, throttled, executing, error) plus a 375px pinned-width play test asserting no horizontal overflow. Snapshots are `pixel: PIXEL_DISABLED` since the repo-wide budget is at its ceiling.
- Full `bun test src` matches the clean-main baseline exactly (the same 30 pre-existing env-dependent failures, none in touched areas); TimelinePanel's 16 jest tests and a static Storybook build pass.

## Risks

Low. The card and registry entry are new code paths gated to `timeline_event`. The only shared-code change is moving the day/time formatters out of `TimelinePanel.tsx` (behavior-preserving, covered by the existing timeline jest suite).

### DIFF
--- a/src/browser/features/RightSidebar/Timeline/TimelinePanel.tsx
+++ b/src/browser/features/RightSidebar/Timeline/TimelinePanel.tsx
@@ -29,6 +29,8 @@ import type {
 
 import {
   TIMELINE_CATEGORIES,
+  formatTimelineTime,
+  getTimelineDayLabel,
   getTimelineEventCategories,
   getTimelineEventKind,
   getTimelineEventTitle,
@@ -89,17 +91,6 @@ const FILTERS: Array<{ value: TimelineFilter; label: string }> = [
   })),
 ];
 
-const timeFormatter = new Intl.DateTimeFormat(undefined, {
-  hour: "numeric",
-  minute: "2-digit",
-});
-
-const dayFormatter = new Intl.DateTimeFormat(undefined, {
-  month: "short",
-  day: "numeric",
-  year: "numeric",
-});
-
 function isTimelineFilter(value: string): value is TimelineFilter {
   return value === "all" || TIMELINE_CATEGORIES.some((category) => category === value);
 }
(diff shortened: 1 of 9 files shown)


ThomasK33:
@codex review

PR #3816, opened 2026-08-06
🤖 fix: resume parents directly from sub-agent reports

ammar-agent:
## Summary

Removes the redundant synthetic “Background sub-agent task(s) have completed…” handoff turn. Completed sub-agent reports and failures now resume the parent directly from the durable report rows already in conversation history.

The refactor also removes 1,001 net lines by collapsing terminal-attention branching and deleting tests coupled to the old prompt/output state machine.

## Background

Sub-agent completion previously used a two-message handoff:

1. persist the real report/failure into parent history
2. start another visible synthetic user turn that only told the model the report was already present

That second turn surfaced as an `AUTO` chat bubble and forced a generic instruction into the transcript. Recent suppression logic only handled the narrow case where the parent had answered an incremental progress update, so normal background completions still showed the prompt.

## Implementation

- resume parent streams directly from existing terminal sub-agent report/failure rows instead of sending a second handoff message
- make terminal report and failure rows visible, ordered append-only messages
- reconstruct a missing terminal row from durable success/failure artifacts before resuming, preserving self-healing after failed history appends
- correlate assistant responses with the highest history sequence included in their provider request, so a concurrently appended report is not incorrectly treated as consumed
- supersede orphaned agent notifications without blocking valid workspace-turn or workflow notifications for the same owner
- reduce active terminal-attention behavior to source identity plus delivery state while retaining the legacy output/outcome fields on disk for downgrade compatibility
- retain explicit prompts only for sources whose output is not already in history (workspace-turn handles and workflow results)

## Validation

- `bun test src/node/services/taskService.test.ts src/node/services/terminalAttentionStore.test.ts --timeout 30000` — 337 passing
- `bun test src/node/services/agentSession.startupAutoRetry.test.ts` — 30 passing
- `bun test src/node/services/aiService.test.ts -t 'uses the latest durable boundary slice'`
- `make typecheck`
- `make static-check`
- verified production no longer contains or sends the completed/failed sub-agent handoff prompt; legacy prompt openings remain only for persisted timeline classification compatibility
- diff: 665 insertions, 1,666 deletions (1,001 net lines removed)

## Risks

Moderate, scoped to terminal background-work delivery. Workspace-turn and workflow wake prompts retain their existing behavior. Sub-agent completion now depends on `resumeStream` over durable history, with provider-request correlation, artifact-backed repair, orphan isolation, and upgrade/downgrade-compatible outbox records covering concurrency and restart cases.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `xhigh` • Cost: `$91.81`_

### DIFF
--- a/src/common/types/message.ts
+++ b/src/common/types/message.ts
@@ -532,6 +532,8 @@ export interface ModelFallbackRecord {
 
 // Our custom metadata type
 export interface MuxMetadata {
+  /** Highest persisted history sequence included in the provider request that produced this assistant. */
+  requestHistorySequence?: number;
   historySequence?: number; // Assigned by backend for global message ordering (required when writing to history)
   duration?: number;
   ttftMs?: number; // Time-to-first-token measured from stream start; omitted when unavailable
--- a/src/node/services/aiService.test.ts
+++ b/src/node/services/aiService.test.ts
@@ -2228,7 +2228,7 @@ describe("AIService.streamMessage compaction boundary slicing", () => {
         compactionEpoch: 2,
         model: "openai:gpt-5.2",
       }),
-      createMuxMessage("latest-user", "user", "continue"),
+      createMuxMessage("latest-user", "user", "continue", { historySequence: 42 }),
     ];
 
     const result = await harness.service.streamMessage({
@@ -2251,6 +2251,7 @@ describe("AIService.streamMessage compaction boundary slicing", () => {
 
     const startStreamMessageIds = messageIdsFromUnknownArray(startStreamCall[1]);
     expect(startStreamMessageIds).toEqual(["boundary-2", "latest-user"]);
+    expect(initialMetadataFromStartStreamCall(startStreamCall).requestHistorySequence).toBe(42);
 
     const openaiOptions = openAIOptionsFromStartStreamCall(startStreamCall);
     expect(openaiOptions.previousResponseId).toBeUndefined();
--- a/src/node/services/aiService.ts
+++ b/src/node/services/aiService.ts
@@ -2386,7 +2386,12 @@ export class AIService extends EventEmitter {
         return Ok(undefined);
       }
 
+      const requestHistorySequence = providerRequestMessages.reduce(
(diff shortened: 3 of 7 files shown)


ammar-agent:
@codex review

Please review the prompt-free terminal sub-agent resume refactor, especially crash recovery and mixed background-work delivery.

ammar-agent:
Addressed all three Codex findings:

- `PRRT_kwDOPxxmWM6XHxVN` (P1): assistant messages now persist the highest history sequence actually included in their provider request. Terminal attention is consumed only when that request sequence includes the report/failure row, so a concurrently appended report remains pending and triggers a follow-up resume.
- `PRRT_kwDOPxxmWM6XHxVR` (P2): orphaned agent notifications are superseded independently. Valid workspace-turn/workflow notifications in the same drain continue normally, while transient report-row repair failures remain pending without blocking unrelated work.
- `PRRT_kwDOPxxmWM6XHxVT` (P2): new outbox records continue writing the legacy `outputDelivery` and `terminalOutcome` fields. Current logic ignores those derived fields, but older strict readers can still process pending notifications after a downgrade.

Added regression coverage for each case and pushed the fixes in `47b14bcff`.

ammar-agent:
@codex review

Please take another look at the provider-request correlation, orphan isolation, and downgrade-compatible outbox fixes in `47b14bcff`.

ammar-agent:
@codex review

The latest commit only retriggers CI after GitHub's Actions webhook outage; please confirm the reviewed code remains approved at the current head.

ammar-agent:
Addressed `PRRT_kwDOPxxmWM6XINr_` (P1) in `f183c00a1`:

- persistent `resumeStream` errors now leave terminal attention pending without scheduling an immediate idle retry
- only the successful `started: false` busy/no-op case waits for the owner to become idle and retries
- added regression coverage proving a persistent budget/model-style error performs one resume attempt, does not register an idle waiter, and keeps the notification pending

Local task tests, typecheck, and `make static-check` pass.

ammar-agent:
@codex review

Please review the final persistent-resume-error fix in `f183c00a1`.

PR #3824, opened 2026-08-09
🤖 feat: copy/download context menu and richer previews for tool result images

ammar-agent:
## Summary

Renews the image experience for `attach_file` (and all other tool results that render images: `desktop_screenshot`, MCP tools) — adds a right-click context menu with **Copy image**, **Download image**, and **View full size** (plus keyboard shortcuts), filename captions on thumbnails, a copy/download action bar **and the same context menu** in the expanded lightbox, and download cards for PDF/SVG attachments that previously rendered no visual at all.

## Background

Tool result images were view-only: a thumbnail that opened a bare lightbox. There was no way to copy an image to the clipboard, no way to save it, no way to tell which file a thumbnail came from, and PDF/SVG attachments (which are real model attachments but excluded from the raster-image allowlist) were completely invisible in the tool card.

## Implementation

- **`src/browser/utils/imageActions.ts`** (new): shared image actions for base64 data URLs.
  - `copyImageDataUrlToClipboard` — Chromium's async Clipboard API only accepts `image/png` payloads, so non-PNG sources (JPEG/WebP/GIF/…) are re-encoded through an offscreen canvas first.
  - `downloadDataUrl` + `getImageDownloadFilename` — download with a sensible filename (attachment filename, else extension derived from media type, with `jpeg→jpg`, `svg+xml→svg` mapping).
  - `handleImageActionKeyDown` — shared handler for the new centralized keybinds `KEYBINDS.IMAGE_COPY` (mod+C) and `KEYBINDS.IMAGE_DOWNLOAD` (mod+S); defers mod+C to native copy when text is selected, suppresses the browser save dialog on mod+S, and stops propagation so global handlers (vim-mode Ctrl+C stream interrupt) don't fire.
- **`src/browser/components/ImageActionsMenu.tsx`** (new): shared right-click menu (Copy image / Download image + shortcut handling) used by both the thumbnail grid and the expanded lightbox, so the two surfaces stay consistent by construction.
- **`ToolResultImages`** (shared by attach_file / desktop_screenshot / MCP / generic tool cards):
  - Right-click (or 500ms long-press on touch) opens the shared menu with a thumbnail-only "View full size" item prepended, reusing the existing `useContextMenuPosition` infrastructure. Menu items show shortcut hints (hidden on mobile); the shortcuts work while the menu is open via a new `PositionedMenu.onKeyDown` prop.
  - Thumbnails now show a truncating filename caption when the media part carries one.
  - `extractImagesFromToolResult` normalizes `filename` (non-string/empty → `undefined`) without dropping the image.
- **`ImageLightbox`**: adds a footer bar with the filename plus Copy (with copied-feedback via the existing `useCopyToClipboard` hook) and Download actions. The full-size image offers the same right-click/long-press context menu as the thumbnails (menu copies route through the feedback hook so the Copy button flashes "Copied"); the same keyboard shortcuts work while the lightbox is open. Escape layering is preserved: Escape closes the menu first, then the lightbox.
- **`AttachFileToolCall`**: PDF and SVG media attachments — which are deliberately never rendered inline (SVG can contain scripts; see `sanitizeImageData`) — now surface as a download card instead of being invisible.

Security note: the existing raster-only allowlist and base64 validation in `sanitizeImageData` are unchanged; SVG remains never rendered in the DOM and is only offered as a download.

## Validation

- `make static-check` passes; 374 tests across `src/browser/features/Tools/` + new `imageActions` tests pass; full Storybook test-runner passes locally (343 tests).
- New unit tests: data-URL → Blob decoding, download filename derivation, filename normalization in `extractImagesFromToolResult`, image caption rendering, the PDF download-card branch, and keyboard-shortcut behavior (copy/download dispatch, selection deferral, unrelated-key passthrough).
- New Storybook coverage: Gallery gains a PDF download-card section; `ImageContextMenu` opens the right-click menu via a play function so Pixel snapshots capture it; `ImageLongPressMenu` exercises the touch-only 500ms long-press path and asserts the follow-up click is suppressed; `LightboxContextMenu` opens the menu on the expanded image and asserts Escape closes only the menu while the lightbox stays open.

## Risks

Low. The context menu wraps the existing thumbnail button without changing click-to-lightbox behavior, and the transcript-wide right-click menu is unaffected (buttons are already excluded via its interactive-target selector). Clipboard copy requires `ClipboardItem` support (available in Electron/Chromium); it fails gracefully with a console error elsewhere. New keybinds are scoped to focused image surfaces only (menu content / lightbox dialog), so they cannot shadow global shortcuts elsewhere.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$0.00`_

### DIFF
--- a/src/browser/components/ImageActionsMenu.tsx
+++ b/src/browser/components/ImageActionsMenu.tsx
@@ -1,0 +1,87 @@
+import type React from "react";
+import { Copy, Download } from "lucide-react";
+import type { UseContextMenuPositionReturn } from "@/browser/hooks/useContextMenuPosition";
+import {
+  PositionedMenu,
+  PositionedMenuItem,
+} from "@/browser/components/PositionedMenu/PositionedMenu";
+import {
+  copyImageDataUrlToClipboard,
+  downloadDataUrl,
+  handleImageActionKeyDown,
+} from "@/browser/utils/imageActions";
+import { KEYBINDS, formatKeybind } from "@/browser/utils/ui/keybinds";
+
+export interface ImageActionsMenuImage {
+  dataUrl: string;
+  downloadFilename: string;
+}
+
+/** Copy an image data URL, logging (not throwing) on failure. */
+export function copyImageWithErrorLogging(dataUrl: string): void {
+  copyImageDataUrlToClipboard(dataUrl).catch((err: unknown) => {
+    console.error("Failed to copy image:", err);
+  });
+}
+
+interface ImageActionsMenuProps {
+  contextMenu: UseContextMenuPositionReturn;
+  /** The image the menu targets; null renders an empty (closed) menu */
+  image: ImageActionsMenuImage | null;
+  /** Extra leading menu items (e.g. "View full size" on thumbnails) */
+  children?: React.ReactNode;
+  /** Override the copy action (e.g. to surface copied-feedback in the lightbox) */
+  onCopy?: (dataUrl: string) => void;
+}
+
+/**
+ * Shared right-click menu for image surfaces (tool result thumbnails and the
+ * expanded lightbox) so both offer the same Copy/Download actions and
+ * keyboard shortcuts.
+ */
+export function ImageActionsMenu(props: ImageActionsMenuProps) {
+  const image = props.image;
+  const copy = props.onCopy ?? copyImageWithErrorLogging;
+
+  return (
+    <PositionedMenu
+      open={props.contextMenu.isOpen}
+      onOpenChange={props.contextMenu.onOpenChange}
+      position={props.contextMenu.position}
+      onKeyDown={(e) => {
+        if (!image) return;
+        const consumed = handleImageActionKeyDown(e, {
+          copy: () => copy(image.dataUrl),
+          download: () => downloadDataUrl(image.dataUrl, image.downloadFilename),
+        });
+        if (consumed) {
+          props.contextMenu.close();
+        }
+      }}
+    >
+      {image && (
+        <>
+          {props.children}
+          <PositionedMenuItem
+            icon={<Copy />}
+            label="Copy image"
+            shortcut={formatKeybind(KEYBINDS.IMAGE_COPY)}
+            onClick={() => {
+              copy(image.dataUrl);
+              props.contextMenu.close();
+            }}
+          />
+          <PositionedMenuItem
+            icon={<Download />}
+            label="Download image"
+            shortcut={formatKeybind(KEYBINDS.IMAGE_DOWNLOAD)}
+            onClick={() => {
+              downloadDataUrl(image.dataUrl, image.downloadFilename);
+              props.contextMenu.close();
+            }}
+          />
+        </>
+      )}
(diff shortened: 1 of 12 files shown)


ammar-agent:
@codex review

ammar-agent:

- **Keyboard shortcuts for image actions** (`PRRT_kwDOPxxmWM6XrIno`): Added centralized `KEYBINDS.IMAGE_COPY` (mod+C) and `KEYBINDS.IMAGE_DOWNLOAD` (mod+S), handled by a shared `handleImageActionKeyDown` helper wired into both the image context menu (via a new `PositionedMenu.onKeyDown` prop) and the lightbox. Shortcut hints render in the menu items and lightbox buttons and are hidden on mobile (`hidden sm:inline`). mod+C defers to native copy when text is selected; mod+S suppresses the browser save dialog; both stop propagation so global handlers (e.g. vim-mode Ctrl+C stream interrupt) don't fire. "View full size" needs no extra bind — Enter/Space on the focused thumbnail button already opens the lightbox. Covered by new behavioral tests in `imageActions.test.ts`.
- **Touch long-press contract** (`PRRT_kwDOPxxmWM6XrInq`): Added the `ImageLongPressMenu` play story, which fires `touchStart`, waits past the 500ms long-press threshold for the menu to open, then verifies the follow-up click is suppressed (no lightbox on top of the menu).

@codex review

ammar-agent:
@codex review

Follow-up: the expanded lightbox now offers the same right-click/long-press context menu as the thumbnails, via a new shared `ImageActionsMenu` component (used by both surfaces so they can't drift). Escape layering verified by the new `LightboxContextMenu` play story: Escape closes the menu first, lightbox stays open.

ammar-agent:
Addressed the lightbox long-press finding (`PRRT_kwDOPxxmWM6Xr3XJ`), plus a user-reported alignment bug in the same surface:

- **Lightbox long-press contract**: `LightboxContextMenu` now reopens the menu via the touch path (real `Touch` instance, 500ms long-press) against the lightbox's full-size image, so the lightbox-specific touch wiring is exercised directly — the thumbnail contract no longer stands in for it. Notable detail: the synthetic touch events must populate `changedTouches`, because the Radix dialog mounts react-remove-scroll whose document-level touch listeners read `changedTouches[0]` and crash otherwise (this is why the touch path needed a lightbox-specific contract — the thumbnail story never hits that code path).
- **Cursor alignment fix**: `PositionedMenu` anchored via a `position: fixed` span, which resolves against the nearest transformed ancestor — inside `DialogContent` (centered with `translate(-50%,-50%)`) the menu opened hundreds of pixels away from the cursor. Replaced with a Radix `virtualRef` anchor measured in viewport coordinates, immune to caller transforms. The play function asserts menu-to-cursor alignment (100px tolerance for collision flipping) as a regression contract.

Validation: full Storybook test-runner passes locally (346 tests); alignment verified interactively (menu opens exactly at dispatch coordinates inside the dialog); `make static-check` + unit tests pass.

@codex review

PR #3833, opened 2026-08-11
🤖 fix: stop duplicating workflow report fields in result context messages

ThomasK33:
## Summary

`buildWorkflowResultContextMessage` — the single builder behind every workflow-completion context message (terminal-attention wakes, background continuations, ORPC continuations, `/workflow` slash command) — hoisted `reportMarkdown`/`structuredOutput` to the payload top level **and** embedded the full result object again under `result`. Since `agent()`-based workflows return exactly `{ reportMarkdown, structuredOutput }`, every workflow-completion wake message carried its payload twice, roughly doubling its token cost.

## Background

Observed in a real terminal-attention wake where a multi-KB gate report appeared verbatim at both the payload top level and nested under `result` (~1.6k wasted tokens for that single message). The duplication has existed since the original dynamic-workflows change (#3431). The `task_await` workflow branch (`buildWorkflowAwaitResult`) already avoids this by hoisting the report fields once and omitting the redundant copy — this PR brings the context-message builder in line.

## Implementation

- New `getResidualWorkflowResult` helper strips the hoisted fields from the embedded `result` object and returns `null` when nothing else remains, so the `result` key is dropped entirely for the standard report shape.
- Non-standard result shapes are preserved: results with extra fields keep only the residual fields under `result`; non-object results (strings, arrays) and results where nothing was hoisted pass through unchanged.
- All four call sites (`taskService.buildWorkflowTerminalPrompt`, `aiService.onBackgroundRunTerminal`, `router.sendWorkflowRunTerminalContinuation`, `chatCommands`) share this builder, so no caller changes were needed.

## Validation

- Three new behavioral tests parse the JSON payload out of the built message and assert: standard report shape → hoisted once with no `result` key; extra fields → `result` keeps only the residual; scalar result → kept verbatim.
- Caller suites pass: `chatCommands.test.ts` (81), `router.test.ts` (22), `taskService.test.ts -t workflow` (51).

## Risks

Low. The change only affects the model-facing workflow-result context message; UI cards and durable run records are untouched. Worst case for an exotic result shape is that the model sees the report fields once instead of twice — the fields themselves are always preserved.
---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$4.92`_

### DIFF
--- a/src/browser/utils/workflowRunMessages.test.ts
+++ b/src/browser/utils/workflowRunMessages.test.ts
@@ -10,6 +10,14 @@ import {
 import type { MuxMessage } from "@/common/types/message";
 import type { WorkflowRunRecord } from "@/common/types/workflow";
 
+function parseWorkflowResultPayload(message: string): Record<string, unknown> {
+  const match = /<mux_workflow_result>\n([\s\S]*)\n<\/mux_workflow_result>/.exec(message);
+  if (match?.[1] == null) {
+    throw new Error(`message does not contain a workflow result payload: ${message}`);
+  }
+  return JSON.parse(match[1]) as Record<string, unknown>;
+}
+
 describe("buildWorkflowRunCardMessage", () => {
   test("builds a stable workflow_run card message with the current durable run", () => {
     const run: WorkflowRunRecord = {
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

PR #3841, opened 2026-08-12
🤖 fix: never resume the parent as the compact agent after compaction

ibetitsmike:
## Summary

Parent auto-resume no longer adopts the compact agent's identity. `TaskService.resolveParentAutoResumeOptions` now treats `compact` as ineligible in both the stream-end hint and the history scan, so wake turns triggered around compaction resume with the real pre-compaction agent, its configured model, and its tools.

## Background

When a sub-agent report arrived while auto-compaction was streaming, the parent's wake turn dispatched with agentId `compact`: the resume-identity scan picked the compact summary assistant (the latest assistant in history), and the stream-end hint path could carry `compact` the same way for bare compactions. Since workspaces have no per-agent AI settings for `compact`, the model fell back to the hardcoded `DEFAULT_MODEL` (Opus) with thinking off, and the compact agent's deny-all baseline stripped every tool. The result was one maximally expensive, fully uncached provider call that then stalled because the turn had no tools.

## Implementation

`resolveParentAutoResumeOptions` is the single chokepoint for sub-agent report wakes, terminal-attention drains, and stream-end auto-resume. The fix ignores a `compact` hint and skips `compact` assistants during the bounded history scan, so the true pre-compaction agent wins; the existing exec recovery fallback is unchanged. Compaction summary turns themselves still run on the compact agent, and the compaction summary's own pending follow-up path is untouched (it always stages an explicit model and agent).

## Validation

- Three regression tests: report wake after compaction, bare compaction stream-end hint, and compact-only history fallback.
- Red-green verified with two independent toggles: reverting the production fix fails the tests, and disabling only the history scan (leaving the hint skip and fallback intact) also fails the strengthened scan test, which pins a `plan` identity so it cannot pass via the exec fallback.
- Full `taskService.test.ts` has 3 pre-existing failures that reproduce identically on the base commit; they are unrelated to this change.

## Risks

Low. The change narrows which history messages qualify as a resume identity source; the worst case for an exotic history (only compact assistants in the last 20 messages) is the existing deterministic exec fallback, which was already the behavior for empty histories.

---

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -3967,6 +3967,79 @@ describe("TaskService", () => {
     expect(snapshot?.error).toBeUndefined();
   });
 
+  test("bare compaction stream-end resumes the pre-compaction parent identity", async () => {
+    const config = await createTestConfig(rootDir);
+    const projectPath = path.join(rootDir, "repo");
+    const parentId = "parent-bare-compaction";
+    const childId = "child-bare-compaction";
+    const execModel = "openai:gpt-5.6-sol";
+
+    await saveWorkspaces(
+      config,
+      projectPath,
+      [
+        projectWorkspace(projectPath, "parent", parentId, {
+          aiSettingsByAgent: {
+            exec: { model: execModel, thinkingLevel: "high" },
+          },
+        }),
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #3459, opened 2026-06-04
🤖 fix: show delegated workflow activity in sidebar

ThomasK33:
## Summary

Adds delegated-work sidebar state for slash-launched workflows so launcher workspaces and collapsed task groups stay visibly active while descendant sub-agents are running, without reusing misleading assistant streaming copy.

## Background

Slash workflows run in the background and spawn child task workspaces, so the launcher workspace can otherwise look idle while the workflow is actively faning out. This change keeps the derived state frontend-local and rolls descendant task activity up through workspace metadata.

## Implementation

- Rolls descendant running/queued task state up by workspace id, including inherited workflow ownership and cycle/duplicate defenses.
- Gives delegated work lower precedence than own workspace question/error/stream/provisioning states.
- Shares the live-sidebar fallback with collapsed task-group counts so metadata lag does not make grouped rows look idle.
- Exposes delegated and task-group status text through `aria-describedby` while keeping action-oriented row labels stable.
- Keeps resumed descendants active even if they still carry stale `reportedAt` metadata from an earlier completed run.
- Keeps non-finalized interrupted descendants delegated-active while their live stream is still visible, but suppresses finalized completed reports.
- Adds real-row ProjectSidebar coverage. Dogfood screenshots/video were generated locally but intentionally left untracked because `dogfood-output/` is ignored.

## Validation

- `bun test src/browser/utils/ui/workspaceFiltering.test.ts src/browser/components/AgentListItem/AgentListItem.test.tsx src/browser/components/ProjectSidebar/TaskGroupListItem.test.tsx src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx`
- `TEST_INTEGRATION=1 bun x jest tests/ui/workspaces/subagents.test.ts --runInBand --silent=false`
- `make typecheck`
- `make fmt-check`
- `make lint`
- `make static-check`
- `MUX_ESLINT_CONCURRENCY=1 make static-check` after the default lint process was killed by local OOM.

## Dogfooding

Dogfooding was performed with local artifacts under `dogfood-output/workflow-sidebar/` (screenshots, video, console, errors). Those artifacts are not tracked in this PR per the repository ignore policy.

## Risks

Moderate UI-state risk in the sidebar: the roll-up is derived from existing workspace metadata and live sidebar state, but it changes status precedence and grouped-row styling. Targeted unit/integration coverage exercises the precedence, terminal-state, metadata-lag, resumed-task, interrupted-live, and accessibility branches.

---

<details>
[...]

PR #3846, opened 2026-08-13
🤖 fix: stop painting running sidebar task groups with the selected background

ibetitsmike:
## Summary

Sidebar workflow task-group headers (e.g. "Workflow · workflow") always rendered with the selected-row background while their run was active, so they looked permanently selected even when a different workspace was selected.

## Background

`TaskGroupListItem` applied `hasRunningWork && "bg-surface-secondary"`, the exact class used for `isSelected`. Regular workspace rows (`AgentListItem`) reserve that background for selection and hover; running state is conveyed by the status dot, the green kind icon, brighter title text, and the "Workflow running" / "N running" status line. The stray highlight was introduced in #3459.

## Implementation

- Remove the running-state background from the group header row; selection and hover backgrounds are unchanged, as are `data-running`, the status dot aggregate state, and the green icon/text treatment.
- Add a regression test asserting a running, non-selected group header does not carry the selected background while a selected one does.
- Fix a pre-existing `ProjectSidebar.test.tsx` suite-wide failure: #3835 added a `coder.svg?react` import to `ProviderIcon`, which the test's SVG mock list did not cover.

## Validation

Verified live in Storybook (`WorkflowRunGroups` story): running group headers no longer carry `bg-surface-secondary`, selecting a member workspace still highlights its group header, and header expand/collapse still works.

## Risks

Low: one class removed from one row component. Pixel snapshots of stories with running task groups will change background as intended.

---

### DIFF
--- a/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
+++ b/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
@@ -62,6 +62,7 @@ function installProviderIconSvgMocks() {
     "@/browser/assets/icons/moonshotai.svg?react",
     "@/browser/assets/icons/aws.svg?react",
     "@/browser/assets/icons/github.svg?react",
+    "@/browser/assets/icons/coder.svg?react",
   ] as const;
 
   for (const svgPath of providerIconSvgPaths) {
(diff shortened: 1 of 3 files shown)


ibetitsmike:
@codex review

PR #3853, opened 2026-08-14
🤖 fix: show hidden sub-agents summary while the coordinator streams

ibetitsmike:
## Summary

Fixes the just-merged "hide sub-agents in the sidebar" setting (#3847): the hidden-descendants summary never appeared in the primary scenario because it was gated on the parent workspace being idle. Coordinators stream for nearly the whole time their background sub-agents run, so users who enabled the setting saw hidden rows and no census, i.e. zero delegated-work visibility.

## Background

Field report from Mike: 4+ active background sub-agents, setting enabled, and the parent row showed only its todo-derived status. `shouldShowHiddenSubAgentsStatus` required `displayStreamingStatusPhase === null` and no armed bash monitor, which excludes exactly the orchestration window the summary exists for.

## Implementation

Drop the streaming and bash-monitor gates so the census outranks the row's own streaming/todo/bash-monitor status text whenever hidden descendants are active or queued. Questions, deletion, and errors still win, and the status dot still reflects the parent's own work. When all descendants go inactive, the row falls back to the normal status line as before.

## Validation

Flipped the old "streaming wins over summary" test to assert the summary renders while the coordinator streams with an `agentStatus` set (the reported scenario), and added a bash-monitor regression test.

## Risks

Low. One boolean gate change in `AgentListItem`; the summary only renders when the opt-in setting provides a `hiddenSubAgentsSummary`, so default-off behavior is untouched. Tradeoff: while any hidden descendant is active, the parent's todo status text is superseded by the census.

---

### DIFF
--- a/src/browser/components/AgentListItem/AgentListItem.test.tsx
+++ b/src/browser/components/AgentListItem/AgentListItem.test.tsx
@@ -607,16 +607,37 @@ describe("AgentListItem", () => {
     expect(rowView.queryByText("Workflow running")).toBeNull();
   });
 
-  test("keeps coordinator streaming copy ahead of the hidden sub-agents summary", () => {
-    mockWorkspaceSidebarState = createWorkspaceSidebarState({ canInterrupt: true });
+  test("shows the hidden sub-agents summary while the coordinator itself streams", () => {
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #3854, opened 2026-08-14
🤖 fix: clone read-only GitHub repos without write-capable credentials

ibetitsmike:
## Summary

Cloning a GitHub repo you can read but not write to could fail in the Clone dialog. This PR adds an HTTPS fallback when `owner/repo` shorthand SSH clones fail, and clarifies GitHub's misleading `remote: Write access to repository not granted.` error.

## Background

Two mechanisms broke read-only clones:

1. `owner/repo` shorthand expanded to `git@github.com:owner/repo.git` whenever `SSH_AUTH_SOCK` existed (near-universal on macOS). GitHub SSH requires a recognized key even for public repos, so readable repos failed with `Permission denied (publickey)` with no fallback.
2. GitHub answers HTTPS clones with `remote: Write access to repository not granted.` (403) when the presented credential cannot _read_ the repo (typically an under-scoped or fine-grained token). Mux surfaced that verbatim, which reads exactly like "cloning requires write access". Mux never requires write access to clone.

## Implementation

- `normalizeRepoUrlForClone` now returns an optional `fallbackCloneUrl`: shorthand still prefers SSH when an agent socket exists, but carries the HTTPS URL as a fallback. Explicit URLs are never rewritten.
- `cloneWithProgress` loops over the attempt URLs via a new private `runGitCloneAttempt` generator (single-attempt spawn/stream logic moved there; abort/cleanup/askpass semantics preserved per attempt). User-driven cancellations (rejected host key, cancelled credential prompt, abort) do not trigger the fallback; only organic clone failures do.
- `withCloneReadAccessHint` appends a clarification when stderr contains GitHub's "Write access to repository not granted": cloning needs only read access and the credential Git presented cannot read the repo.

## Validation

Behavioral tests cover: fallback ordering and success, no fallback for explicit SSH URLs, abort between attempts spawns no second clone and leaves no debris, and hint gating on the exact GitHub stderr (including a negative case where SSH `Permission denied` gets no hint).

## Risks

Clone flow only. The riskiest part is the retry loop refactor in `cloneWithProgress`; a regression there could double-spawn clones or leak temp dirs on abort, both covered by tests. Explicit-URL flows are pass-through and unaffected.

---

### DIFF
--- a/src/node/services/projectService.test.ts
+++ b/src/node/services/projectService.test.ts
@@ -53,20 +53,30 @@ async function withEnv<T>(
   }
 }
 
-async function writeFakeGitCloneShim(fakeGitPath: string): Promise<void> {
-  await fs.writeFile(
-    fakeGitPath,
-    `#!/bin/sh
+const DEFAULT_FAKE_GIT_SHIM = `#!/bin/sh
 printf '%s\n' "$@" > "$FAKE_GIT_ARGS_LOG"
 if [ "$1" = "clone" ]; then
   mkdir -p "$5/.git"
   exit 0
 fi
 exit 1
-`,
-    "utf-8"
-  );
+`;
+
+async function installFakeGit(tempDir: string, testCaseId: string, shimBody: string) {
+  const fakeBinDir = path.join(tempDir, `fake-bin-${testCaseId}`);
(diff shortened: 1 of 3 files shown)


ibetitsmike:
@codex review

PR #3855, opened 2026-08-14
🤖 fix: exclude sub-agent rows from sidebar agent counts

ibetitsmike:
## Summary

Sidebar agent counts (the `(N)` next to each project name, the Chats section, and the Multi-Project section) included sub-agent descendant workspaces, so a project with one root agent running seven sub-agents showed `(8)`. Counts now include only top-level agents.

## Background

Sub-agent rows render nested under their parent row (or are hidden entirely behind a parent summary), so counting them at the section level overstated how many agents the user actually has in a project.

## Implementation

Reuse the existing `excludeSubAgentRows` helper, which already powers the "hide sub-agents in sidebar" display filter, for all three section counts. It intentionally keeps promoted orphans and malformed-parent-cycle rows countable, matching what actually renders. The per-project filtered list is computed once and shared between the count and the hidden-row display path.

## Validation

Red-green: the new test (1 root + 2 children + 1 grandchild asserts the header shows `(1)`) fails with `demo-project(4)` when the fix is stashed.

## Risks

Low: display-only change scoped to sidebar section counts; row rendering, attention dots, and archive/removal counts are untouched.

---

### DIFF
--- a/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
+++ b/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
@@ -1881,6 +1881,53 @@ describe("ProjectSidebar multi-project completed-subagent toggles", () => {
   });
 });
 
+describe("ProjectSidebar project agent count", () => {
+  const demoProjectPath = "/projects/demo-project";
+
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #3858, opened 2026-08-15
🤖 feat: collapse scheduler machinery noise in the timeline panel

ibetitsmike:
## Summary

Collapses scheduler machinery in the Timeline tab so the timeline reads as a birds-eye narrative of the chat instead of watcher spam. Monitor wakes, background-work wakes, synthetic continue prompts, goal continuations, and heartbeat dispatches now group into a single expandable row per contiguous stretch, absorbing the `turn.completed` and compaction rules those turns produce.

## Background

Real timelines are dominated by loop machinery. Sampling recent workspaces: `turn.completed` is 34% of all events and `turn.monitor_wake` 16%; one chat rendered 155 individual "Monitor wake" rows (median 3.9 minutes apart) out of 618 visible items. The existing same-kind run collapse could not help because the dominant pattern is alternating `wake -> turn.completed -> wake`, and rule rows break same-kind runs.

## Implementation

Presentation-layer only; the recorder still persists every event, so existing timelines benefit retroactively and nothing is lost for debugging.

- `timelinePresentation.ts` gains `isMachineryKind` classifying the six machinery kinds.
- `collapseConsecutiveEvents` groups contiguous machinery stretches (2+) across kinds, absorbing completion/compaction rules produced by those turns. Deliberate boundaries (`context.reset`, `history.cleared`) still break groups. Timeline pages are newest-first, so the scan reverses to chronological order internally and back out.
- A `retry.abandoned(reason=aborted)` row adjacent to its `turn.interrupted` duplicates it and is dropped, except under the Errors filter where the interruption row is absent.
- Collapsed group rows count only machinery events and use a neutral "automatic" label for mixed-kind groups.

Measured on the noisiest sampled real timeline: 661 events, 618 visible items before, 304 after, with all 155 wake rows folded into expandable groups.

## Validation

Simulated the new algorithm against real persisted `timeline.jsonl` files from four recent workspaces to confirm the wake/turn churn actually folds (numbers above).

## Risks

Low: rendering-only change in the Timeline tab. The main behavioral risk is over-grouping (absorbing a rule a user cares about); expansion always reveals the full underlying rows, and deliberate context boundaries never collapse. Pixel snapshots for TimelinePanel stories change intentionally.

---

### DIFF
--- a/src/browser/features/RightSidebar/Timeline/TimelinePanel.tsx
+++ b/src/browser/features/RightSidebar/Timeline/TimelinePanel.tsx
@@ -35,6 +35,7 @@ import {
   getTimelineEventKind,
   getTimelineEventTitle,
   getTimelinePresentation,
+  isMachineryKind,
   type TimelineCategory,
 } from "./timelinePresentation";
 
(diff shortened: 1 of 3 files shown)


ibetitsmike:
@codex review

PR #3865, opened 2026-08-17
🤖 feat: add shared agent-foundation layer (event spine, journal kit, sandbox host, capability grants)

ThomasK33:
## Summary

Adds the **shared agent-foundation layer**: four substrates that both upcoming experiment tracks (Track 1: dsh-style plugin architecture; Track 2: prime-agent RLM / persistent kernel) build on — a typed event spine, a journal kit with a content-addressed blob store, a Sandbox Host Service with persistent mounts and an async capability bridge, and capability grants enforced at both the sandbox bridge and tool assembly.

## Background

Both experiment tracks need the same underlying mechanics: a hookable event pipeline around tool execution, durable append-only event logging that can replay byte-for-byte, a QuickJS sandbox whose state can outlive a single `code_execution` call (and a process restart), and a policy vocabulary that limits what a sandbox may reach. Implementing these once as a shared foundation keeps the two tracks decoupled from each other while giving them a common, tested substrate.

## Implementation

**Substrate 1 — Typed event spine** (`src/node/services/events/eventSpine.ts`)
- dsh-inspired three-way split: read-only **observer events** (workspace/session/stream/task lifecycle), mutating **waterfall hooks** (`tool.execute`, `request.assemble`, `compaction.prepare`), and **durable events** (persisted via the journal kit, not the spine itself).
- The `tool.execute` waterfall is an around-style middleware pipeline (`(ctx, next)`) with `useBefore`/`useAfter` sugar. Legacy `.mux/tool_hook` shell hooks inherently *wrap* execution, so they port cleanly onto the same primitive as the new `tool_pre`/`tool_post` hooks: `withHooks.ts` is now a built-in spine middleware. All 47 existing hook tests pass unmodified.
- Observer events are emitted from `WorkspaceService`, `TaskService`, `StreamManager`, and `AgentSession`; throwing observers are logged and swallowed so they can never break the emitting hot path.

**Substrate 2 — Journal kit + blob store** (`src/node/utils/journal/`)
- `Journal<T>`: append-only JSONL with monotonic `seq`, stable-ID dedupe, torn-tail newline repair, and malformed-line filtering (self-healing loads, per the crash-resilience doctrine).
- `BlobStore`: content-addressed sha256 store (`blobs/<hh>/<hash>`), atomic temp-file+rename writes, hash-verified reads that self-heal corrupt blobs.
- `DurableEventJournal` binds both to `durable-events.jsonl`; the record schema (`src/common/types/durableEvent.ts`) is a discriminated union over `kind` (`turn-envelope`, `refinement`, `result-handle`, `hook-context`, `sandbox-vars-snapshot`) with large payloads referenced by `BlobRef`. `chat.jsonl`/HistoryService is untouched.

**Substrate 3 — Sandbox Host Service** (`src/node/services/sandbox/sandboxHostService.ts`)
- `SandboxMount` with `ephemeral` (per-call) and `persistent` (per-workspace-session) lifetimes; persistent mounts share a `vars` namespace across evals/acquires, snapshot it to the durable journal on dispose, and restore it after restart.
- Host→guest event queue drained in-guest via `drainHostEvents()`.
- QuickJS runtime gains `registerPromiseFunction` (asyncified) and `registerSyncFunction` (plain sync). The split matters: quickjs-emscripten Asyncify can only suspend inside the initial `evalCodeAsync` stack, so bridges callable from post-`await` guest continuations must be sync-registered.
- Persistent mounts for `code_execution` are opt-in via `MUX_SANDBOX_PERSISTENT_MOUNTS=1` (full persistent-kernel UX is owned by Track 2).

**Substrate 4 — Capability grants** (`src/common/types/capabilityGrants.ts`, `src/common/utils/tools/capabilityGrants.ts`)
- `CapabilityGrants` (`bridgeTools` allow-list/`all`, `vars`, `hostEvents`) with `FULL_GRANTS` (session scope) and `LEAST_PRIVILEGE_GRANTS` (project scope).
- Enforced at two points with one vocabulary: the sandbox bridge (`ToolBridge` — denied tools excluded from both bridgeable and non-bridgeable sets, stubbed with a clear `Capability denied` guest error, re-checked at call time) and tool assembly (`applyCapabilityGrants` as a ceiling filter ahead of tool policy).

## Validation

- `make static-check` green (typecheck, lint, format, docs links).
- New suites: `eventSpine.test.ts`, `journal.test.ts`, `blobStore.test.ts`, `durableEventJournal.test.ts`, `sandboxHostService.test.ts` (registered in `isolated_unit_tests` — QuickJS-heavy), `capabilityGrants.test.ts`; extended `toolBridge.test.ts`, `quickjsRuntime.test.ts`. Existing touched suites (`withHooks`, `code_execution`, `WorkflowRunner`, `streamManager`, `agentSession`) pass.
- Dogfood gates run against a real workspace: legacy + new shell hooks executing through the spine; persistent `vars` shared across separate `code_execution` calls, snapshotted, and restored across a simulated restart; denied bridge capabilities produce catchable in-guest errors without crashing the sandbox.

## Risks

- **Tool execution path** (`withHooks` → spine middleware): highest-traffic surface touched. Mitigated by keeping all 47 existing hook tests unmodified and green, plus an empty-pipeline fast path (`hasMiddleware`) so workspaces without hooks skip context construction.
- **ToolBridge constructor signature** gained an optional `grants` param defaulting to `FULL_GRANTS` — existing callers are behaviorally unchanged.
- **New observer emissions** in stream/workspace/task hot paths are wrapped so listener errors cannot propagate.
- Journal/blob/sandbox services are new code, not yet wired into critical flows beyond opt-in paths; regression blast radius there is low.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$77.49`_

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -196,6 +196,7 @@ jobs:
             # when sharing the monolithic process; keep them in fresh isolated processes.
             src/node/services/workflows/WorkflowRunner.test.ts
             src/node/services/ptc/quickjsRuntime.test.ts
+            src/node/services/sandbox/sandboxHostService.test.ts
             src/node/orpc/router.test.ts
             src/browser/components/WorkspaceHeartbeatModal/WorkspaceHeartbeatModal.test.tsx
             src/browser/features/Messages/InlineSkillMarkdown.test.tsx
--- a/src/common/types/capabilityGrants.test.ts
+++ b/src/common/types/capabilityGrants.test.ts
@@ -1,0 +1,61 @@
+import { describe, expect, test } from "bun:test";
+import { tool } from "ai";
+import { z } from "zod";
+import {
+  FULL_GRANTS,
+  LEAST_PRIVILEGE_GRANTS,
+  isBridgeToolGranted,
+  resolveCapabilityGrants,
+  type CapabilityGrants,
+} from "./capabilityGrants";
+import { applyCapabilityGrants } from "@/common/utils/tools/capabilityGrants";
+
+function makeTool(): ReturnType<typeof tool> {
+  return tool({
+    description: "t",
+    inputSchema: z.object({}),
+    execute: () => Promise.resolve({}),
+  });
+}
+
+describe("capability grants", () => {
+  test("session scope resolves to full grants; project scope to least privilege", () => {
+    expect(resolveCapabilityGrants({ scope: "session" })).toEqual(FULL_GRANTS);
+    expect(resolveCapabilityGrants({ scope: "project" })).toEqual(LEAST_PRIVILEGE_GRANTS);
+  });
+
+  test("isBridgeToolGranted honors allow-all and allow-lists", () => {
+    expect(isBridgeToolGranted(FULL_GRANTS, "bash")).toBe(true);
+    expect(isBridgeToolGranted(LEAST_PRIVILEGE_GRANTS, "bash")).toBe(false);
+
+    const narrow: CapabilityGrants = {
+      version: 1,
+      bridgeTools: { allow: ["file_read"] },
+      vars: false,
+      hostEvents: false,
+    };
+    expect(isBridgeToolGranted(narrow, "file_read")).toBe(true);
+    expect(isBridgeToolGranted(narrow, "bash")).toBe(false);
+  });
+
+  test("applyCapabilityGrants filters assembled tools (grant/deny branches)", () => {
+    const tools = { file_read: makeTool(), bash: makeTool(), web_fetch: makeTool() };
+
+    // Allow-all passes through untouched (same reference: zero-cost ceiling).
+    expect(applyCapabilityGrants(tools, FULL_GRANTS)).toBe(tools);
+
+    const narrow: CapabilityGrants = {
+      version: 1,
+      bridgeTools: { allow: ["file_read", "web_fetch"] },
+      vars: false,
+      hostEvents: false,
+    };
+    expect(Object.keys(applyCapabilityGrants(tools, narrow)).sort()).toEqual([
+      "file_read",
+      "web_fetch",
+    ]);
+
+    // Least privilege denies everything.
+    expect(Object.keys(applyCapabilityGrants(tools, LEAST_PRIVILEGE_GRANTS))).toEqual([]);
+  });
+});
--- a/src/common/types/capabilityGrants.test.ts
+++ b/src/common/types/capabilityGrants.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from "bun:test";
-import { tool } from "ai";
+import type { Tool } from "ai";
 import { z } from "zod";
 import {
   FULL_GRANTS,
@@ -10,12 +10,12 @@ import {
 } from "./capabilityGrants";
 import { applyCapabilityGrants } from "@/common/utils/tools/capabilityGrants";
 
-function makeTool(): ReturnType<typeof tool> {
-  return tool({
+function makeTool(): Tool {
+  return {
     description: "t",
     inputSchema: z.object({}),
     execute: () => Promise.resolve({}),
-  });
+  };
 }
 
 describe("capability grants", () => {
--- a/src/common/types/capabilityGrants.ts
+++ b/src/common/types/capabilityGrants.ts
@@ -1,0 +1,67 @@
+/**
+ * Capability grants (substrate 4 of the shared agent foundation): one small
+ * vocabulary for "what can this guest / hook / tool-set do".
+ *
+ * Enforced at two points:
+ * - the sandbox bridge boundary (ToolBridge filters + re-checks per call), and
+ * - tool assembly (applyCapabilityGrants filter over the assembled tool set).
+ *
+ * Scope model (surfaces through existing Project Trust concepts):
+ * - `session` scope: model-authored code running in the user's own session
+ *   (PTC code_execution). Gets the full bridge — matches pre-grants behavior.
+ * - `project` scope: repo-controlled code (future Tier-1 plugin hooks).
+ *   Defaults to least privilege; an untrusted project always resolves to
+ *   least privilege regardless of requested scope.
+ */
+
+export interface CapabilityGrants {
+  version: 1;
+  /** Which mux.* bridge tools the guest may invoke. */
+  bridgeTools: { allow: "all" } | { allow: readonly string[] };
+  /** Guest-visible persistent `vars` namespace. */
+  vars: boolean;
+  /** Host→guest event queue + drain. */
+  hostEvents: boolean;
+}
+
+/** Everything the session's own agent code may already do today. */
+export const FULL_GRANTS: CapabilityGrants = {
+  version: 1,
+  bridgeTools: { allow: "all" },
+  vars: true,
+  hostEvents: true,
+};
+
+/** Deny-by-default: no bridge tools, no vars, no host events. */
+export const LEAST_PRIVILEGE_GRANTS: CapabilityGrants = {
+  version: 1,
+  bridgeTools: { allow: [] },
+  vars: false,
+  hostEvents: false,
+};
+
+export interface ResolveCapabilityGrantsOptions {
+  /** Who authored the code the grants govern. */
+  scope: "session" | "project";
+}
+
+/**
+ * Resolve default grants from code scope. Session-scoped (model-authored)
+ * code keeps the full bridge; project-scoped (repo-controlled) code is least
+ * privilege — capabilities must be granted explicitly (e.g. via a future
+ * plugin manifest gated on Project Trust), never implicitly.
+ */
+export function resolveCapabilityGrants(options: ResolveCapabilityGrantsOptions): CapabilityGrants {
+  if (options.scope === "session") {
+    return FULL_GRANTS;
+  }
+  return LEAST_PRIVILEGE_GRANTS;
+}
+
+/** Check whether a mux.* bridge tool is granted. */
+export function isBridgeToolGranted(grants: CapabilityGrants, toolName: string): boolean {
+  if (grants.bridgeTools.allow === "all") {
+    return true;
+  }
+  return grants.bridgeTools.allow.includes(toolName);
+}
--- a/src/common/types/durableEvent.ts
+++ b/src/common/types/durableEvent.ts
@@ -1,0 +1,123 @@
+/**
+ * Durable agent event record schema — the joint design constraint shared by
+ * the plugin-architecture track (turn envelopes, hook-injected context) and
+ * the RLM track (refinement journal, offloaded result handles).
+ *
+ * Rows are appended to per-session JSONL journals via the journal kit
+ * (src/node/utils/journal/). Invariants:
+ * - "model-visible ⟺ logged": anything the model can see must exist as a
+ *   durable row first (append-time materialization, never request-time
+ *   injection), so log replay can reconstruct requests byte-for-byte.
+ * - Large payloads (full prompt text, offloaded values, vars snapshots) are
+ *   content-addressed in the blob store and referenced by BlobRef.
+ * - Readers are tolerant: malformed lines and unknown kinds are skipped
+ *   (self-heal doctrine), so new kinds can be added without breaking readers.
+ */
+
+import { z } from "zod";
+import { JsonValueSchema } from "@/common/orpc/schemas/workflow";
+
+export const DURABLE_EVENT_VERSION = 1;
+
+/** Reference into a content-addressed blob store: `sha256:<64 hex chars>`. */
+export const BlobRefSchema = z.string().regex(/^sha256:[0-9a-f]{64}$/);
+export type BlobRef = z.infer<typeof BlobRefSchema>;
+
+/**
+ * Turn envelope: the request fingerprint for one stream turn. Full prompt text
+ * is content-addressed; the row stores only hashes so envelope comparison is
+ * cheap and the journal stays small.
+ */
+export const TurnEnvelopeDataSchema = z.object({
+  systemPromptHash: BlobRefSchema,
+  /** Sorted by tool name; schemaHash fingerprints the tool's input schema. */
+  toolsetManifest: z.array(z.object({ name: z.string(), schemaHash: z.string() })),
+  modelString: z.string(),
+  providerOptionsHash: z.string(),
+  thinkingLevel: z.string(),
+});
+
+/**
+ * Refinement journal entry: an auditable, invertible harness edit.
+ * `data.kind` is the refinement's own taxonomy (e.g. "skill-edit"); the
+ * envelope `kind` stays the discriminator, which is why payloads are nested.
+ */
+export const RefinementDataSchema = z.object({
+  kind: z.string(),
+  action: JsonValueSchema,
+  inverse: JsonValueSchema,
+  evidence: JsonValueSchema.optional(),
+  /** Envelope `id` of the entry this one rolls back. */
+  rollbackOf: z.string().optional(),
+});
+
+/**
+ * Offloaded result handle: a large tool result stored instead of returned.
+ * The row is the durable source of the model-visible text (handle + preview +
+ * size), so "model-visible ⟺ logged" holds for offloads.
+ */
+export const ResultHandleDataSchema = z.object({
+  handle: z.string(),
+  preview: z.string(),
+  blobHash: BlobRefSchema,
+  size: z.number().int().nonnegative(),
+});
+
+/**
+ * Hook-injected context: content a hook adds to a request. Materialized at
+ * append time; exactly one of `text` (small) or `blobHash` (large) is set.
+ */
+export const HookContextDataSchema = z
+  .object({
+    hookId: z.string(),
+    /** Where the content lands, e.g. "system-prompt" or "user-context". */
+    placement: z.string(),
+    text: z.string().optional(),
+    blobHash: BlobRefSchema.optional(),
+  })
+  .refine((data) => (data.text === undefined) !== (data.blobHash === undefined), {
+    message: "hook-context requires exactly one of text or blobHash",
+  });
+
+/** Sandbox persistent-mount `vars` snapshot (JSON text stored in blob store). */
+export const SandboxVarsSnapshotDataSchema = z.object({
+  scopeKey: z.string(),
+  blobHash: BlobRefSchema,
+  size: z.number().int().nonnegative(),
+});
+
+/** Envelope shared by all durable agent events (one JSONL row each). */
+const envelopeBase = {
+  v: z.literal(DURABLE_EVENT_VERSION),
+  /** Per-journal monotonic sequence, assigned at append. */
+  seq: z.number().int().nonnegative(),
+  /** Stable unique id; dedupe key on read. */
+  id: z.string().min(1),
+  /** Epoch milliseconds. */
+  ts: z.number(),
+  workspaceId: z.string().min(1),
+};
+
+export const DurableEventSchema = z.discriminatedUnion("kind", [
+  z.object({ ...envelopeBase, kind: z.literal("turn-envelope"), data: TurnEnvelopeDataSchema }),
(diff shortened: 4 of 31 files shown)


ThomasK33:
@codex review

ThomasK33:
Addressed all 5 Codex findings in 87c287139:

- **Persistent mounts bypass grants (P1)** — `toolAssembly` now forwards `opts.capabilityGrants` into `acquireMount`; additionally, reacquiring a scope whose effective grants changed disposes and rebuilds the mount under the new grants (`grantsKey` comparison), so a mount can never outlive its capability boundary. Regression test: `reacquiring with changed grants rebuilds the mount under the new grants`.
- **Stale bridge on reused mounts (P1)** — removed the `bridgeRegistered` latch; `code_execution` now re-registers the fresh per-request `ToolBridge` on every call (registration just overwrites the guest `mux` global). Regression test: `re-registering a narrower bridge on a reused runtime revokes previously exposed tools`.
- **Reset restores pre-reset vars (P2)** — added `SandboxHostService.discardScope()`: disposes without snapshotting and supersedes any prior snapshot with an empty one (append-only rotation); context reset now uses it. Regression test: `discardScope: context reset discards vars instead of restoring the last snapshot`.
- **Async observer rejections unhandled (P2)** — `subscribe` wrapper now captures promise return values and logs rejections; emission stays non-blocking. Regression test added.
- **Late-settling capability misattribution (P2)** — `registerPromiseFunction` captures `toolCalls`/`eventHandler` at call start, so fire-and-forget capabilities that settle after their eval returned attribute to the originating eval. Regression test: `attributes a late-settling fire-and-forget capability to its originating eval`.

Also live-dogfooded the fixed build: persistent `vars` shared across separate `code_execution` calls, snapshotted to `durable-events.jsonl` + blob store, and restored across a full server restart.

ThomasK33:
@codex review

ThomasK33:
Round-2 findings addressed in 5a3ff7185:

- **Emit workspace.created from every creation path (P2)** — now emitted from `createScratch()`, `createMultiProject()`, and `fork()` success paths in addition to `create()`.
- **Reapply request assembly hooks to refusal fallbacks (P2)** — the refusal-fallback path now runs the `request.assemble` waterfall over the rebuilt request (tools + system prompt, with token recount on mutation), mirroring the primary path.
- **Replace corrupted existing blobs on put (P2)** — `BlobStore.put` now reads and byte-compares an existing blob instead of trusting `access()`; corrupted files are rewritten via the same atomic temp+rename, so a later put of the original content repairs the poisoned ref. Regression test added.
- **Run compaction middleware before compacting (P2)** — this one appears to be a miss: the waterfall *is* invoked on both compaction entry points as of 87c287139 — `src/node/services/agentSession.ts:3035` (`reason: "on-send"`) and `src/node/services/agentSession.ts:3893` (`reason: "mid-stream"`). Resolving with this pointer; happy to revisit if there is a third compaction path I overlooked.

ThomasK33:
@codex review

ThomasK33:
Round-3 findings addressed in 5d3d06c0a:

- **Serialize access to each persistent runtime (P2)** — `SandboxHostService` now keeps a per-scope `AsyncMutex`: acquisition runs under the lock (concurrent first acquisitions can no longer both create runtimes), and the same lock backs the new `SandboxMount.exclusive()`, which `code_execution` uses to serialize the register → eval → persist sequence on persistent mounts. `disposeScope`/`discardScope`/`snapshotScope` take the lock too, so disposal waits for in-flight evaluation. Ephemeral runtimes stay lock-free (per-call). Regression tests: `concurrent first acquisitions share one mount`, `exclusive() serializes concurrent runs on a shared mount`.
- **Preserve denied tool names for bridge stubs (P2)** — the bridge is now constructed from the pre-grant policy-filtered toolset, so grant-denied tools land in `deniedToolNames` and guests get the documented catchable `Capability denied: mux.<tool>…` stub instead of `mux.<tool> is not a function`. The grants ceiling still applies to every model-visible set, including the exclusive-mode non-bridgeable set. Regression test: `grant-denied tools are hidden from the model but stubbed in the sandbox` (asserts the exact in-guest stub message).

ThomasK33:
@codex review

ThomasK33:
Round-4 findings addressed in 9d2e5c445:

- **Recreate mounts to revoke saved bridge functions (P1)** — persistent mounts are now keyed by the effective bridge configuration (`bridgeKey` = sorted bridgeable tool names, computed in tool assembly). When policy/grants narrow the sandbox-reachable toolset, `acquireMount` disposes and rebuilds the mount — destroying the runtime is the only reliable way to revoke guest-saved closures like `globalThis.savedBash = mux.bash`; `vars` survive via snapshot/restore. Regression test: `bridge narrowing rebuilds the mount, revoking guest-saved bridge references` (saves a bridge ref in a global, narrows, asserts the closure is gone and vars persisted).
- **Serialize late guest continuations with the mount (P2)** — new `IJSRuntime.setPendingJobGate`: when a fire-and-forget capability settles after its eval returned, `executePendingJobs()` is routed through the gate, which `SandboxMount` binds to its exclusive lock — a late continuation can no longer re-enter the shared context mid-eval. Regression test: `routes late-settlement continuations through the pending-job gate`.
- **Suppress lifecycle starts during stream replay (P2)** — the `stream.start` spine emission is skipped when `replay: true`; observers see exactly one start per stream.
- **Close lifecycle events on abort and failure (P2)** — `stream.end` moved from the normal-completion branch into `processStreamWithCleanup`'s `finally`, so every terminal outcome (completion, Escape/abort, provider failure) closes the lifecycle pair.
- **Apply fallback hooks before deriving sentinel tools (P2)** — `nextToolNamesForSentinel` is now computed after the fallback `request.assemble` run, matching the primary path ordering.

ThomasK33:
@codex review

ThomasK33:
Round-5 findings addressed in 4c67bba32:

- **Rebuild tool-search state after assembly hooks (P2)** — `rebuildToolSearchState` now runs immediately after the `request.assemble` waterfall in BOTH the primary path and the refusal-fallback path, so `allToolNames`, the deferred catalog, and activation sets always describe the post-hook toolset (`prepareStep` scoping and sentinel names read them). The rebuild mutates the state object in place, so StreamManager's held reference stays consistent.
- **Reconcile persistent vars after failed evaluations (P2)** — `code_execution` now persists `vars` after failed/timed-out/aborted evals too (best-effort with logged failure, never masking the eval result): the live guest keeps pre-throw mutations, so disk must match or a restart resurrects older state. Regression test: `persists vars mutated before a failed eval so memory and disk agree` (mutate → throw → simulated restart → asserts the mutated value).
- **Gate deferred settlement before touching QuickJS (P2)** — the entire VM-facing settlement (marshal + `deferred.resolve`/`reject` + `executePendingJobs`) now routes through the pending-job gate for late settlements. New eval-generation tracking distinguishes in-eval settlements (settled directly — the originating eval's resolve loop awaits them, and routing through the gate would deadlock on the mount lock that call already holds) from post-eval settlements (gated, so they cannot re-enter the shared context while a later eval runs).

ThomasK33:
@codex review

ThomasK33:
Round-6 findings addressed in a091f6b8c:

- **Dispose persistent mounts when removing workspaces (P2)** — new `SandboxHostService.dropScope()`: disposes the runtime and forgets journals with NO disk writes, and the per-scope lock makes it wait for any in-flight evaluation. Workspace removal calls it before deleting the session directory, so a late vars snapshot can no longer recreate the deleted directory and the QuickJS runtime no longer leaks in the process-wide singleton. Regression test: `dropScope: workspace removal disposes the mount without writing to disk` (asserts disposal + zero new files on disk).
- **Rebuild the mount when failure snapshots cannot persist (P2)** — when the post-eval snapshot rejects (e.g. the guest assigned `vars.self = vars` and threw), `code_execution` now disposes the mount so the next acquire rebuilds from the last durable snapshot, instead of keeping live state that disk can never reflect. Regression test: `recovers from unsnapshottable vars by rebuilding the mount from the last durable snapshot` (seed → poison with a cycle + throw → next call sees the durable value and no cycle).

ThomasK33:
@codex review

ThomasK33:
Round-7 findings addressed in a427d3283:

- **Hold the scope lease from acquisition through execution (P2)** — new `SandboxHostService.withPersistentMount(options, fn)`: acquires the per-scope lock ONCE and holds it from mount acquisition through `fn`'s completion. `code_execution` now runs its whole register → eval → persist sequence inside that single lease (the old acquire-then-`exclusive()` two-step is gone), so a concurrent grant/bridge change or scope disposal can no longer dispose the mount in the gap. Regression test: `withPersistentMount holds the lease: concurrent disposal waits for fn to finish`.
- **Rebuild code_execution after request tool filtering (P2)** — in supplement mode, when `request.assemble` changes the toolset, both the primary and fallback paths now rebuild `code_execution` via `applyToolPolicyAndExperiments` on the post-hook record: the fresh `ToolBridge` no longer exposes hook-removed tools, the regenerated description no longer advertises them, and the changed `bridgeKey` rebuilds the persistent mount (revoking guest-saved refs). Exclusive mode is unaffected since bridgeable tools aren't in the hook-visible record.
- **Preserve attribution for late guest continuations (P2)** — the gated continuation drain now swaps in the originating eval's `toolCalls`/`consoleOutput`/`eventHandler` for the duration of `executePendingJobs()` and restores the current eval's state afterwards, so an old continuation cannot append console output or emit events under a later eval's identity.

ThomasK33:
@codex review

ThomasK33:
Round-8 findings addressed in dc0a14bf0:

- **Rebuild PTC when a hook replaces a tool (P2)** — the post-hook `code_execution` rebuild now snapshots the pre-hook record and compares tool identities (added/removed names AND same-name replacements), so middleware that wraps a tool under the same name (e.g. audit/permission wrappers) also triggers the bridge rebuild. Applied to both the primary and refusal-fallback paths.
- **Persist vars after late guest continuations (P2)** — `SandboxMount`'s pending-job gate now persists `vars` after the gated continuation runs (a fire-and-forget `.then` can mutate vars after the originating call's snapshot), and disposes the mount on snapshot failure exactly like the normal execution path, so memory and disk always agree.

ThomasK33:
@codex review

ThomasK33:
@codex review

Round 9 fixes pushed in b59f5d3fb:

- **Preserve hook-replaced code_execution** (aiService.ts): `rebuildCodeExecutionAfterAssembleHook` now detects when `request.assemble` middleware replaced `code_execution` itself (identity comparison) and reinstates the middleware's replacement over the freshly rebuilt instance — a hook wrapper is never silently dropped. Bridge-relevant change detection excludes `code_execution` from the comparison so a hook-only replacement doesn't force an unnecessary rebuild.
- **Saved guest bridge references vs names-only bridgeKey** (quickjsRuntime.ts): `registerObject` now dispatches guest method calls through a live `registeredObjects` map at call time instead of binding hard closures. Re-registering `mux` on a reused persistent runtime retargets even guest-saved references (`globalThis.savedBash = mux.bash`) to the newest implementation, and removed tools fail closed with a clear error. This makes the names-only `bridgeKey` sufficient for same-name replacements; the toolAssembly comment documents the invariant. New unit tests cover retargeting, removal fail-closed, late-settlement attribution, and gated continuations.

ThomasK33:
@codex review

Round 10 fix pushed in 95c294086 — wrapper-over-stale-bridge closed for real this time:

- `code_execution` now reads its `ToolBridge` + mount runner **late-bound at call time** from a per-instance state box (`retargetableStates` WeakMap) instead of closure capture.
- New `retargetCodeExecutionTool(target, donor)` grafts the rebuilt post-hook bridge/mount onto the pre-hook instance before `rebuildCodeExecutionAfterAssembleHook` reinstates the middleware's replacement. A wrapper that delegates to the pre-hook instance — even one that captured the original `execute` function reference directly — now dispatches through the post-hook toolset, so a removed/replaced bridgeable tool can no longer be reached via `mux.*` in either the primary or fallback request path.
- Behavioral tests: captured-execute reference dispatches through the donor bridge after retarget (removed tool fails closed, surviving tool works); retarget refuses tools not created by the factory and leaves them functional.

ThomasK33:
@codex review

Round 11 fix pushed in 9fe721099:

- **Stale wrapper metadata** (toolAssembly.ts / aiService.ts): `reconcileHookReplacedCodeExecution` now reconciles a spread-style wrapper's model-facing description with the rebuilt toolset. When the wrapper inherited the pre-hook description verbatim (`{ ...tool, execute: wrapped }`), the rebuilt description — whose embedded TypeScript definitions reflect the post-hook toolset — is swapped in, so the model is no longer instructed that removed tools exist. Middleware that authored its own description keeps it (it took ownership of the model-facing contract). The middleware's `execute` wrapper is preserved in both cases, and round-10's bridge retargeting keeps its delegation path on the post-hook bridge.
- Tests: spread-wrapper gets the rebuilt description while keeping its wrapped execute; middleware-authored descriptions pass through untouched.

ThomasK33:
@codex review

Follow-up push a14259c9c: fixed a `static-check-full`-only lint failure (`as Tool` cast → typed const) in the new toolAssembly test. No production code changes since 9fe721099.

ThomasK33:
@codex review

Round 12 fix pushed in bd160c7a8:

- **Cross-eval promise consumption self-deadlock** (quickjsRuntime.ts): `settleInVm` now runs directly while *any* eval is active — that eval's resolve loop coordinates VM access and may itself be awaiting the settlement (later call returning `globalThis.p` stored by a prior call), so gate-routing deadlocked on the scope lock the awaiting eval holds. Attribution remains bound to the originating eval via the existing guarded state swap.
- **Gate-vs-eval lock race**: truly-late settlements (between evals) are queued before gate hand-off, and `eval()` drains the queue at start under the caller's mount lock. If the gate's lock acquisition loses the race to the next `code_execution` call, that call lands the settlement itself; the gate's eventual drain finds an empty queue and no-ops. `dispose()` drops queued settlements.
- Tests: settling a prior eval's stored promise mid-eval (mount lock held, mirroring `SandboxMount`'s gate) resolves with the value; a settlement queued while the next call already holds the lock is drained at eval start and the post-release gate drain no-ops.

ThomasK33:
@codex review

Round 13 fixes pushed in 906becdfb — both findings addressed by narrowing the direct-settlement window and splitting attribution by drain context:

- **Gate settlements during unrelated asyncified calls**: settlements may touch the VM directly only while `directSettleAllowed` — set strictly between `evalCodeAsync` completion and `eval()` return, i.e. exactly when the resolve loop coordinates and the VM is idle at the top level. A prior eval's settlement arriving while the current eval is suspended inside an asyncified `mux.*` call now queues instead of re-entering the unwound WASM stack; the resolve loop drains the queue (so a returned prior-eval promise still resolves) and the `finally` block hands leftovers to the gate (fire-and-forget settlements after a plain return/throw).
- **Attribute cross-eval reactions to the consuming eval**: when a queued settlement is drained while an eval is active, pending reactions include continuations that eval created (`return p.then(...)`), so the drain runs under the CURRENT eval's attribution state. Between evals, the originating-eval swap is preserved, and the eval-start drain runs before the new generation is marked active so pure late settlements keep originating attribution. The tool-call record itself always lands on the originating eval via the captured arrays.
- Tests: consuming eval owns the `.then()` reaction's console output (originating eval's output stays empty); a settlement arriving during an unrelated asyncified suspension queues without VM re-entry and the resolve loop lands it for consumption.

ThomasK33:
@codex review

Round 14 fix pushed in e4434ad8f — attribution is now resolved per-registration instead of per-drain, which closes the whole class of "whose reaction is this" findings:

- **Guest-side reaction tagging**: `Promise.prototype.then` is patched once per context (catch/finally delegate to then per spec) to capture the registering eval's generation and expose it in a guest global for the duration of each callback. Nested registrations inherit the enclosing reaction's owner.
- **Per-generation attribution contexts**: `eval()` registers its `{toolCalls, consoleOutput, eventHandler}` sinks in a bounded map (8 generations, oldest evicted). Host entry points — `console.*` and `registerPromiseFunction`'s call-time capture — route through `currentAttribution()`: the registering eval's context when inside a tagged reaction, else the current drain context.
- A reaction registered by eval 1 (`p.then(cb)`) that settles while eval 2's resolve loop drains now writes console output and nested capability records to **eval 1**; a reaction created by eval 2 on eval 1's stored promise (`return p.then(...)`) attributes to **eval 2**. The drain-context swaps in `settleInVm` remain only as fallback for untagged (await-registered) continuations, which use the engine-internal reaction path.
- Tests: prior-eval-owned reaction console output + nested capability records land on the registering eval (and not the draining eval); round-13 consumer-attribution and asyncified-suspension tests unchanged.

ThomasK33:
@codex review

Round 15 fixes pushed in 769abef1d — both findings closed structurally:

- **Await continuations now tagged**: capability promises are handed to the guest wrapped in a `Promise` subclass (`__muxWrapPromise`). `await` on a promise whose constructor is not `%Promise%` takes the spec's thenable path, which calls the **patched** `then` — so await continuations on capability promises are tagged at registration with the registering eval, same as explicit `.then` chains. Ownership propagates one hop to the continuation via deferred, epoch-guarded restore jobs (an earlier callback's restore cannot clear the owner before a later callback's queued continuations run), `eval()` drains its own leftover microtasks under its generation in its `finally`, and the host resets the owner global at every drain-batch boundary.
- **Refcount-based context retention**: every tagged registration retains its owner's attribution context via host endpoints (`__muxRetainGeneration`/`__muxReleaseGeneration`) and releases when the callback runs — contexts live exactly as long as outstanding reactions, not a fixed eval count. Idle (refs=0) contexts beyond a soft cap of 8 evict oldest-first; a hard cap of 64 bounds memory against never-settling registrations; the active generation is never evicted.
- Tests: a prior eval's fire-and-forget `await` continuation (engine-internal registration) attributes to the registering eval when settled during a later eval's resolve-loop drain; a `.then` reaction registered 12 evals earlier (beyond the old fixed cap) still attributes correctly. All 67 quickjsRuntime tests + sandbox/bridge/assembly suites green.

ThomasK33:
@codex review

Round 16 fix pushed in 2a2fc3ab2:

- **Leaked retains on opposite-branch settlements**: `.then(onFulfilled)` / `.then(undefined, onRejected)` reactions settling through the branch without a user handler never ran a wrapped callback, so the registration's retain leaked — repeated occurrences could push the 64-context hard cap into evicting genuinely-outstanding generations. Both branches are now wrapped unconditionally with the spec's identity/rethrow defaults for missing handlers, so exactly one wrapper runs at settlement and always releases the retain. Reactions with no function handler at all skip tagging entirely.
- Test: 70 single-handler reactions settling through the opposite branch across evals no longer evict a genuinely-outstanding generation registered before them — its late reaction still attributes to the registering eval.

ThomasK33:
@codex review

Round 17 fixes pushed in b0f5e7fdb:

- **Undefined receiver for tagged handlers**: the wrapper now invokes user handlers as plain calls (no receiver) instead of `fn.call(this, ...)`, matching native reaction semantics — a strict-mode handler sees `this === undefined` again.
- **Release retain on registration throw**: `origThen.call(...)` is wrapped in try/catch; when registration throws after `retain(owner)` (non-promise receiver, throwing `Symbol.species` constructor), the owner is released before rethrowing so completed generations can't accumulate toward the hard cap.
- Tests: strict-mode `this === undefined` through a tagged reaction; 70 evals with throwing registrations (`Promise.prototype.then.call({}, ...)`) no longer evict a genuinely-outstanding generation whose reaction still attributes to its registering eval.

ThomasK33:
@codex review

Round 18 fix pushed in 17e3600d4:

- **Leaked executePendingJobs results**: both settlement drain branches in `settleInVm` dropped the disposable result of `executePendingJobs()`, leaking QuickJS handles on every capability settlement (accumulating on persistent mounts). Extracted `drainPendingJobs()` which disposes the result in both the success and error cases; the eval-finally drain shares it, and the resolve loop already disposed correctly.

PR #3868, opened 2026-08-18
🤖 feat: port anti-slop lint rules and enable switch exhaustiveness checking

ThomasK33:
## Summary

Ports six "anti-slop" lint rules (from [dmmulroy/anti-slop](https://github.com/dmmulroy/anti-slop)) into our existing ESLint `local` plugin and enables `@typescript-eslint/switch-exhaustiveness-check`, so type evidence the compiler already has cannot be silently discarded or fabricated. No new toolchain: everything lives in `eslint.config.mjs`, and `make lint` is green with the rules enforced.

## Background

anti-slop is an oxlint ruleset targeting low-evidence TypeScript patterns common in AI-generated code. Adopting it wholesale would mean running a second linter for ~10 rules, several of which fight this repo's deliberate boundary-parsing idioms (`unknown` + zod at IPC/JSON edges). Instead, each rule was censused against the full `src/` tree with a real ESLint run, then ported/scoped only where it produces signal:

| Rule | Enforcement | Census |
| --- | --- | --- |
| `no-chained-type-assertions` (`x as unknown as T`) | error in prod; tests/stories exempt; 33-file shrink-only ratchet | 2,015 hits, 97% in tests (mock-cast idiom) |
| `no-known-value-widening` | error, assertion-position only (`checkAnnotations: false`) | 0 prod violations after scoping |
| `no-widen-then-assert` (`const t: unknown = x; t as Foo`) | error | 0 prod violations |
| `no-unknown-type-aliases` (`type X = unknown`) | error | 0 violations |
| `no-object-parameters` (bare `object` params) | error; 4-file shrink-only ratchet | 9 hits |
| `no-conditional-empty-object-spread` | implemented, **off** | 629 deliberate uses (omitted-vs-`undefined` semantics) |
| `@typescript-eslint/switch-exhaustiveness-check` | error (`considerDefaultExhaustiveForUnions`) | 2 hits, both fixed here |

## Implementation

- Rules are ported from upstream's oxlint AST to typescript-eslint's TSESTree (no `ParenthesizedExpression` nodes, `typeArguments`/`typeParameters` compat).
- Upstream's ~300-line module-level type-alias resolution is intentionally omitted. An AST census across all 2,148 src files found only 4 broad aliases repo-wide, zero assertion sites through them where the rules would fire, and zero cross-file usages (upstream's resolution is file-local anyway). Documented in the helper header with a revisit condition.
- Upstream's "anonymous object" widening target is dropped: inline object annotations are idiomatic here (296 prod sites) and `consistent-type-assertions` already restricts the assertion form.
- `no-known-value-widening` checks assertion flows only: the annotation-flow footprint (247 prod sites) is dominated by patterns this repo mandates or needs (`Record<Enum, Value>` exhaustive mappings, arbitrary-key lookup tables, `unknown`-returning boundary parsers).
- Two shrink-only ratchet blocks grandfather pre-existing violations (33 files chained assertions, 4 files object params); new files are enforced immediately.
- Exhaustiveness fixes: `SubAgentTasksDecoration` now narrows optional `taskExecutionStatus` before switching (fall-through to `taskStatus` is visible in the types); `chatCommands` declares an explicit `default` where a runtime `ReadonlySet` guard is invisible to the compiler.

## Validation

- Fixture tests: 26-case fixture per rule family (17 expected positives fired, 9 expected negatives clean), plus a negative test proving a fresh `as unknown as` in prod code fails `make lint`.
- Full-src censuses were run via real ESLint JSON reports, not grep.
- `make static-check` green; targeted tests for both touched source files pass (84 tests).

## Risks

Low. The two source-file changes are behavior-preserving control-flow rewrites covered by existing tests. The main risk is lint friction on future branches: pre-existing-violation files not in the ratchet lists would fail if someone adds a *new* violation there — which is the intended ratchet behavior.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$25.66`_

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1078,8 +1078,12 @@ const localPlugin = {
           FunctionDeclaration: checkParameters,
           FunctionExpression: checkParameters,
           TSDeclareFunction: checkParameters,
+          TSEmptyBodyFunctionExpression: checkParameters,
           TSMethodSignature: checkParameters,
           TSFunctionType: checkParameters,
+          TSCallSignatureDeclaration: checkParameters,
+          TSConstructSignatureDeclaration: checkParameters,
+          TSConstructorType: checkParameters,
         };
       },
     },
@@ -1137,6 +1141,13 @@ export default defineConfig([
       "react-hooks": reactHooks,
       tailwindcss,
       local: localPlugin,
+    },
+    linterOptions: {
+      // Grandfathered anti-slop violations are marked with per-site
+      // eslint-disable-next-line directives. Erroring on unused directives
+      // makes that a true shrink-only ratchet: fixing a site forces removal
+      // of its directive, and the directive cannot silently outlive the code.
+      reportUnusedDisableDirectives: "error",
     },
     settings: {
       react: {
@@ -1260,8 +1271,9 @@ export default defineConfig([
 
       // Anti-slop ports (see localPlugin). Chained assertions are banned in
       // production code; test/story/mock files are exempt (casting partial
-      // doubles through `unknown` is the standard mock idiom), and
-      // pre-existing violations are grandfathered in a ratchet block below.
+      // doubles through `unknown` is the standard mock idiom). Pre-existing
+      // violations carry per-site eslint-disable-next-line directives so new
+      // violations fail lint even in files that contain grandfathered ones.
       "local/no-chained-type-assertions": "error",
       // Implemented but not enforced: 629 existing occurrences across 145
       // files use `...(cond ? { x } : {})` deliberately to omit keys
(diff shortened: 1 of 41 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

All three findings addressed in af41a3349:

1. **Per-site ratchet** — replaced both file-wide rule-off blocks with `eslint-disable-next-line` directives at each of the 70 grandfathered sites (37 files), plus `linterOptions.reportUnusedDisableDirectives: "error"`. New violations in those files now fail lint, and fixing a site forces removal of its directive (shrink-only in both directions).
2. **Exhaustive workspace dispatch** — `WORKSPACE_ONLY_COMMAND_TYPE_LIST` is now an `as const` tuple exporting a `WorkspaceOnlyCommandType` union; `chatCommands` narrows via a type guard and the switch has no `default`, so a new workspace-only command type without a handler fails `switch-exhaustiveness-check` (verified by removing a case and observing the lint error).
3. **Callable coverage** — added `TSCallSignatureDeclaration`, `TSConstructSignatureDeclaration`, `TSConstructorType`, and `TSEmptyBodyFunctionExpression` visitors to `no-object-parameters`; verified `interface Handler { (value: object): void }` and `type Factory = new (value: object) => X` now report.

ThomasK33:
@codex review

Both findings addressed in 933d073bf:

1. **Tuple pinned to parser union** — `WorkspaceOnlyParsedCommand` now goes through a `SubsetOf<WorkspaceOnlyCommandType, NonNullable<ParsedCommand>["type"]>` constraint, so a typo or stale entry in `WORKSPACE_ONLY_COMMAND_TYPE_LIST` fails compilation (verified: probing with `"goal-cleer"` produces TS2344 at the constraint) instead of being silently dropped by `Extract`.
2. **Last block disable site-scoped** — restructured the `replay-history.ts` site so the chained assertion starts directly under a per-site `eslint-disable-next-line`, removing the disable/enable block that could shelter additional assertions.

ThomasK33:
@codex review

Addressed in 51763a8de: `TSNonNullExpression` and `TSSatisfiesExpression` are now traversed as transparent links in both directions of the chain walk (outermost-dedup upward, counting downward). Verified `(x as unknown)! as Foo` and `(x as unknown satisfies unknown) as Foo` each report exactly once, while `maybe! as Foo | null` and plain `satisfies` stay clean. Full static-check confirmed no previously-sheltered sites surfaced in src.

ThomasK33:
@codex review

Both findings addressed in 338d01905:

1. **Passthroughs unwrapped in `widenedBinding`** — the initializer is now run through the shared passthrough unwrapper before deriving the assertion, so `const tmp = (value as unknown)!; return tmp as Foo` reports like the unwrapped form (verified with a probe).
2. **Scoped aliases visited** — `no-unknown-type-aliases` now collects every `TSTypeAliasDeclaration` via visitor (function/block/namespace scope included) and evaluates at `Program:exit`. Duplicate names across scopes are treated as unresolvable (conservative) instead of guessing shadowing; verified `function parse() { type Payload = unknown; }` reports while a same-shape alias resolving to a real type stays clean.

ThomasK33:
@codex review

Addressed in 8a5c12b15: `knownValueEvidence` now applies the shared passthrough unwrapper before its direct-assertion check, so `const tmp: unknown = (value as Foo)!; return tmp as Foo` recognizes the wrapped `as Foo` as the evidence assertion and reports. Verified with a probe (wrapped form flags; un-evident value into a broad binding stays silent by design). Full static-check green.

ThomasK33:
@codex review

Both findings addressed in be13f2e2a:

1. **Destructured-param evidence** — `knownValueEvidence` now resolves the member type from an inline literal annotation on the owning `ObjectPattern` (shorthand and default-value forms). Verified: `({ value }: { value: Foo })` + widen + assert-back reports; a broad member (`{ value: unknown }`) and a named type reference (`Props`) both stay silent — named references are intentionally unresolvable rather than guessed.
2. **Test-harness exemption** — the value-flow exemption now includes the repo's test-support conventions (`**/*.testHarness.ts`, `src/**/test[A-Z]*.ts`, covering testUtils/testHelpers/testHistoryService et al.). The 5 grandfather directives in `agentSession.testHarness.ts` are removed — `reportUnusedDisableDirectives` would have flagged them as stale, which is the ratchet working as intended.

ThomasK33:
@codex review

Addressed in f6f8b6297: the destructured-evidence lookup now walks outward through nested Property/ObjectPattern levels, collects the member path innermost-first, and resolves it outermost-first through inline type literals. Verified `{ nested: { value } }: { nested: { value: Foo } }` + widen + assert-back reports, single-level still reports (regression), and both a broad nested member and a named intermediate type (`Inner`) stay silent by design. Array patterns and computed keys abort conservatively.

PR #3872, opened 2026-08-18
🤖 feat: make provider requests a pure function of the session log + sandboxed plugin hooks

ThomasK33:
## Summary

Implements Track 1 of the plugin-architecture work on top of the shared agent foundation (#3865): the provider request becomes a verifiable pure function of the append-only session log ("model-visible ⟺ logged"), with turn-envelope records, a byte-equality replay harness and cache-bust auditor, sandboxed Tier-1 plugin hooks (QuickJS), and graduation of the agent-plugins manifest into a unified `contributes` distribution unit with a per-workspace composition inspector.

## Background

Research into DeepSeek Harness ("everything is a plugin"; requests reconstructed purely from an append-only log) and OpenCode's plugin hook vocabulary identified two properties worth porting to Mux: log-purity of provider requests (optimal prompt-cache reuse, deterministic replay/fork/resume) and a safe plugin surface. Mux was already ~80% log-pure (history re-read from disk each turn, no in-memory transcript); this PR closes the remaining gaps and builds the plugin surface inside that guarantee. Builds directly on the event spine, journal kit, sandbox host, and capability grants from #3865.

## Implementation

Five phases, each implemented → quality-gated → adversarially reviewed → dogfooded via a durable Mux workflow (committed as `workflows/track1-implementation.js`):

- **p1 — Log purity**: file-change notifications and `@file` mention snapshots materialize into `chat.jsonl` at append time (before the request is built); the two request-time injectors in `messagePipeline.ts` are deleted. Old persisted histories still build.
- **p2 — Turn envelopes**: one `turn-envelope` durable event per assistant turn: content-addressed system prompt (BlobStore, dedupes across turns), name-sorted toolset manifest with schema hashes, model/thinking/providerOptions hashes (hash-only; raw options never persisted). Emission is observability — it never fails the turn.
- **p3 — Determinism harness**: `bun run debug replay-verify <ws>` reconstructs each turn's request from chat.jsonl + envelopes + blobs through the production pipeline and byte-compares against `devtools.jsonl`; `bun run debug cache-audit <ws>` diffs consecutive envelopes and attributes prompt-prefix busts with approximate token cost. Fixture-based CI tests keep the invariant enforced. Provider-defined tools (e.g. Anthropic web_search) fingerprint by wire identity (`id`+`args`) on both sides — their client-side inputSchema never crosses the wire.
- **p4 — Sandboxed plugin hooks**: `hooks.js` in agent-plugin containers loads into persistent QuickJS mounts and registers as event-spine middleware (`tool.execute.before` mutate/deny, `tool.execute.after` observe, `request.assemble` contribute context). Least-privilege capability grants (manifest-requested tool visibility, Project-Trust-gated project containers); hook-contributed context lands as durable `hook-context` rows before the prompt mutation, so replay stays byte-identical. Crash/timeout/malformed hooks are logged and skipped — only explicit denials become model-visible tool errors. Experiment-gated (`agent-plugins`).
- **p5 — Manifest graduation + inspector**: `plugin.json` gains `contributes` (skills, agents, workflows, MCP, slash commands, hooks) consumed through the existing per-scope loaders (no parallel loading paths); `plugin://` workflow scheme; plugin slash commands surface in the composer suggestions; `bun run debug plugins <ws>` plus one bulk oRPC endpoint print the effective composition by layer (`built-in | global | project | plugin:<name>`) including shadowing attribution.

## Validation

Beyond CI: each phase was dogfooded against a live dev-server sandbox with real provider turns —

- p1: `<system-file-update>` row durably appended before request start; request content byte-identical to log rows.
- p2: blob dedupe across identical turns; system-prompt blobs byte-identical to the wire request in `devtools.jsonl`.
- p3: live dogfooding caught the provider-tool fingerprint bug reviews missed; after the fix, `replay-verify` 3/3 PASS and `cache-audit` correctly attributed a deliberate AGENTS.md edit (~13.6k tokens re-processed).
- p4: demo plugin blocked `.env` reads (secret never reached the model), injected context landed as a `hook-context` row and in the wire request, deliberately crashing hooks were skipped with the turn completing, replay-verify stayed green.
- p5: live web-UI turn via a plugin-contributed slash command (agent-browser); inspector showed correct per-layer attribution and a project skill shadowing a plugin skill, matching what the production loaders enforced in the live system prompt.

## Risks

- **p1 changes the request-build path** (messagePipeline) — the highest-blast-radius area. Mitigated by the p3 harness (fixture tests fail on any reintroduced request-time injection), unchanged-behavior tests for old histories, and live byte-equality verification.
- **p4 executes third-party JS** — bounded by the QuickJS sandbox (no I/O, memory/time caps), least-privilege grants, experiment gate, and Project Trust for project-scoped containers. `hookService.ts` is the file to review hardest.
- p2/p3/p5 are additive (journal rows, debug CLI, discovery/inspection); envelope emission is fail-open by design.

## Pains

- The dev-server watch build (`build-main-watch`) crashes on intermediate edit states and can leave a stale/corrupt dist bundle serving `Cannot find module` errors; two workflow dogfooders were lost to this before the process pinned fresh rebuilds.
- `bun run debug send-message` looks like a turn driver but is display-only; live turns must be driven via the oRPC WS API. Cost several dogfood attempts before being diagnosed.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$336.75`_


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed all three round-1 findings in ca2b11f55:
- **Fallback envelope (P1)**: `modelFallback.prepare` now emits its own turn envelope with the rebuilt request identity (model, system prompt, toolset, provider options) before returning, so `pairSessionTurns`' last-envelope-per-sequence rule compares the request that actually streamed.
- **Dynamic import (P1)**: `hookService` statically imports `QuickJSRuntimeFactory`; the WASM stack still loads lazily at first runtime creation (`QuickJSRuntime.create`), so startup cost is unchanged and the `runtimeFactoryLoader` DI seam is preserved.
- **Experiment reactivity (P2)**: the composer subscribes via `useExperimentValue(EXPERIMENT_IDS.AGENT_PLUGINS)` and includes it in the load/clear effect, so toggling the experiment takes effect without remounting.

ThomasK33:
@codex review

Addressed all four round-2 findings in 3ee27c0e9:
- **Audit collapse (P1)**: `cache-audit` now collapses envelopes to the final row per `requestHistorySequence` via a shared `collapseEnvelopesToFinalPerSequence` helper (matching `pairSessionTurns`), so fallback turns are audited once with the identity that streamed; usage pairs once.
- **Step-0 tool fingerprints (P1)**: both the primary and fallback envelopes now fingerprint the first step's actual wire toolset — `forcedFirstStepToolNames` when present, else the tool-search active subset — instead of the full runtime map; the fallback's forced-name computation is hoisted and shared with the prepare return payload.
- **Provider config (P1)**: the `replay-verify` CLI passes `new ProviderService(defaultConfig).getConfig()` (the same view type the live build consumes) into `replayVerifySession`.
- **Hook load retry (P2)**: the stored reconcile fingerprint now includes only successfully-loaded candidates, so a transiently-failed hooks.js is retried on the next send; covered by a new test (fail once via spy → unchanged files → retried and active).

ThomasK33:
@codex review

Addressed all three round-3 findings in 81d507b14:
- **Shared journal (P1)**: added a process-wide `sharedDurableEventJournal(sessionDir)` registry (keyed by resolved path); aiService's envelope writer and SandboxHostService's snapshot writer (all sites, including `discardScope`) now obtain the same instance, so one seq counter owns each `durable-events.jsonl`. Covered by an interleaved-writers test asserting identity + strictly increasing seq.
- **Fallback envelope timing (P2)**: the superseding envelope moved out of `prepare()` into a new `PreparedModelFallback.onStreamConstructed` callback that StreamManager invokes only after `createStreamResult` succeeds (before consumption). Tested both branches: called exactly once on success, never when construction throws.
- **Agent sort precedence (P2)**: `discoverAgentDefinitions` now sorts same-ID duplicates as one group keyed by the winning (first-discovered) definition's display name, so composition shadowing attribution can't flip when a lower-precedence duplicate sorts earlier alphabetically. Test uses Zulu(project)/Alpha(global) to force the previously-broken ordering.

ThomasK33:
@codex review

Addressed both round-4 findings in the latest push:
- **Sentinel names (P2)**: turn envelopes now carry an optional `sentinelToolNames` field recorded independently of the (step-0-narrowed) wire manifest, populated on both the primary and fallback envelope paths from the same `toolNamesForSentinel` values the live pipeline used; `replayVerify` prefers it and falls back to manifest names only for legacy rows. Covered by an envelope round-trip test where the manifest is narrowed to the forced tool while the sentinel lists the full active set.
- **Args rewrite validation (P2)**: `tool.execute.before` rewrites are applied only when the value is a plain object; null/array/scalar shapes are logged and discarded, preserving original args (test: `{args: null}` rewrite → tool executes with original args).

ThomasK33:
@codex review

Addressed all three round-5 findings in the latest push:
- **Active user ID (P2)**: `streamWithHistory` now selects the last user matching `shouldUseUserMessageForRetry` (the same predicate startup recovery uses), so file-update notifications, `[CONTINUE]` sentinels, and snapshot rows can no longer become `activeStreamUserMessageId`.
- **Primary envelope timing (P2)**: primary emission moved into a new `onStreamConstructed` callback on `startStream`, invoked only after atomic stream creation + registration (identical boundary to the fallback's callback). Pre-construction aborts/setup errors leave no phantom row for cache-audit.
- **Pending thinking override (P2)**: the deferred emission folds `activeTurnThinkingOverride.pending` via a pure `computeRebuiltProviderOptions` helper factored out of the rebuild closure — same clamp/variant-swap gates, no `currentEffectiveLevelRef` mutation, so `prepareStep` still applies the override to the wire request while the envelope fingerprints the level/options that request will actually use.

ThomasK33:
@codex review

Addressed the round-6 finding in the latest push:
- **Message rebuild under pending level (P2)**: the PREPARING-window fold moved out of the envelope callback and now happens once before `startStream`: when `pending` yields an effective change, the first-step messages are re-prepared (`prepareProviderRequestMessages` + `prepareMessagesForProvider`) under the new effective level, the stream uses the rebuilt options/messages/level, the envelope records those same values, and `pending` is consumed with `applied` set (so `createStreamAtomically` seeds `streamInfo.thinkingLevel` and `prepareStep` has nothing to re-apply). Wire, envelope, and replay now all describe the same request identity, including the Anthropic signed-reasoning message transforms.

ThomasK33:
@codex review

Addressed the round-7 finding in the latest push:
- **Re-fold racing thinking changes (P2)**: the PREPARING-window fold is now a quiescence loop — each iteration consumes `pending`, folds against the level just applied (`streamThinkingLevel` as the running baseline), rebuilds the first-step messages under the new effective level, and then re-checks `pending`, so a change written during the awaited rebuild triggers another fold instead of being half-applied at step 0. No-op folds (`computeRebuiltProviderOptions` returning null) also re-check before exiting. The loop terminates because each iteration consumes the value and only repeats when a genuinely new pending was written.

ThomasK33:
@codex review

Addressed the round-8 finding in the latest push:
- **Close the post-quiescence race (P2)**: the fix now lives where pending is finally consumed. `prepareStep` at step 0, after `applyPendingThinkingOverride`, invokes a new `rebuildFirstStepForThinkingLevel` closure (threaded through `StreamRequestConfig`; primary and fallback variants each bound to their own build inputs) that re-runs `prepareProviderRequestMessages` + `prepareMessagesForProvider` under the applied level and emits a superseding turn envelope; prepareStep returns the rebuilt messages (with the same per-step strip/extract transforms) alongside the rebuilt provider options. Since replay pairing and cache-audit take the last envelope per `requestHistorySequence`, the superseding row makes wire, envelope, and replay agree even for writes racing `startStream`'s awaits. Rebuild failures fail open to the options-only behavior with a warning.

ThomasK33:
@codex review

Addressed the P2: the step-0 thinking rebuild in `prepareStep` now re-prepends the construction-time cached system row (`request.messages[0]` when `request.system` is undefined) before applying the per-step transforms, so Anthropic-cache requests keep their system prompt. Covers the fallback path too since it flows through the same `prepareStep`. Added a regression test (`step-0 message rebuild preserves the cached system row when the system prompt lives in messages`).

ThomasK33:
@codex review

Addressed the P2: after the step-0 message rebuild, `prepareStep` now re-invokes `request.onStepMessages` with the rebuilt transcript so consumers (advisor transcript ref in `aiService.ts`) track the messages actually sent, not the pre-rebuild batch. Regression test extended to assert the last `onStepMessages` batch equals the rebuilt step-0 messages.

ThomasK33:
@codex review

Addressed all three round-11 findings:

- **P1 (partial fallback replay)**: refusal-fallback envelopes now blob-store the partial continuation (`partialContinuationHash` on the turn-envelope row); `replay-verify` loads the blob and appends it via the same `replaceOrAppendMessageById` production uses, so partial-output fallbacks rebuild byte-identically. Covered by a new envelope round-trip test and an end-to-end `replayVerifySession` test.
- **P2 (plugin skill suggestions)**: the agent-skills loading effect in ChatInput now depends on the reactive agent-plugins experiment value, so toggling refetches `/skill` suggestions like it reloads plugin commands.
- **P2 (debug plugins CLI)**: `bun run debug plugins` now loads workspace metadata and mirrors the oRPC composition endpoint's runtime + host-checkout resolution (`createRuntimeContextForWorkspace`, `resolveWorkspaceRootPath`, `resolveAgentPluginsMcpContext`) with SSH/Docker off-host gating; multi-project workspaces report as unsupported instead of scanning the wrong host path.

ThomasK33:
@codex review

Addressed the P2: `buildWorkspaceComposition` now takes the nullable `hostCheckoutRoot` explicitly. When null (SSH/Docker), project/global plugin-container discovery is suppressed entirely — `discoverWorkspaceAgentPlugins` is not invoked with the remote `workspacePath`, and the null root also authoritatively gates `includeAgentPlugins` for the runtime-based skill/agent/workflow loaders (belt and braces beyond their internal `RemoteRuntime` class check). Both callers (oRPC composition endpoint and the debug CLI) pass the resolved nullable root. New regression test: a host-side plugin at the workspace path is NOT reported when `hostCheckoutRoot` is null.

Note: the `Test / Integration` failure on the previous run was `tests/ui/review/undo.test.ts` — untouched by this branch (last modified in PR #3061) and part of the same environment-load flake class diagnosed earlier (reproduced on origin/main in the same environment). This push supersedes that run.

ThomasK33:
@codex review

Addressed all three round-13 findings:

- **P2 (execution path)**: the composition endpoint and debug CLI now pass the workspace execution path as `workspacePath` (production loaders discover from there, including subProjectPath subdirectories); `hostCheckoutRoot` stays a separate nullable arg reserved for plugin-container discovery.
- **P2 (shell hooks via runtime)**: `collectShellHookEntries` now resolves both layers through the workspace `Runtime` (`runtime.stat` + `runtime.resolvePath("~")`, joined with hooks.ts's exported `joinPathLike`), mirroring production `hooks.ts` resolution instead of host `fs.stat`/`os.homedir()`.
- **P2 (preserve healthy mounts)**: `WorkspaceHookRegistration` now stores the FULL discovery fingerprint plus a `failedLines` set. On an unchanged fingerprint, only previously-failed candidates are retried (per-candidate `loadCandidateLocked`) — healthy siblings keep their persistent mounts and cross-turn guest state; no `teardownLocked` churn. New regression test asserts the healthy plugin's mount is neither re-created nor dropped while the failed sibling retries and recovers.

ThomasK33:
@codex review

Addressed both round-14 findings:

- **P1 (abort race after envelope write)**: after `await onStreamConstructed?.()`, `startStream` now re-checks the abort signal AND registration ownership before launching `processStreamWithCleanup()`. A canceled stream bails out (no post-abort `stream-start`), deletes the registration only if it still owns it (a replacement stream's slot is preserved), and resets `streamRegistered` so the finally block releases the never-processed stream's model/temp-dir resources. Regression test drives a real `startStream` with `stopStream` + a replacement registration inside the callback — verified it fails without the guard (stream-start emitted, replacement deleted) and passes with it.
- **P2 (nested plugin scan roots)**: the inspector's **skill** discovery now anchors plugin containers at the checkout root with checkout-rooted containment, exactly mirroring production (`skillStorageContext.buildProjectLocalRoots` anchors plugin skill containers at `checkoutRoot ?? projectRoot`). New test: a checkout-level plugin's skills are reported when the workspace executes in a subdirectory. **Agents and workflows intentionally keep execution-path defaults**: production derives their plugin containers from the discovery path (`streamContextBuilder` → `getDefaultAgentDefinitionsRoots(runtime, agentDiscoveryPath)`, `workflows.listScripts` → default roots from `workspacePath`) with `projectContainmentRoot: workspacePath` — checkout-anchoring them in the inspector would report plugin agents/workflows that production never loads, the same faithfulness bug class as the earlier off-host finding. Aligning production's agent/workflow plugin anchoring with skills is a candidate follow-up outside this PR's scope; the inspector documents the split inline.

ThomasK33:
@codex review

Fixed the `Test / Unit` CI failure: the round-5 construction-gating change moved primary turn-envelope emission into `startStream`'s `onStreamConstructed` callback, but the aiService test stub replaced `startStream` without invoking it — the turn-envelope test asserted an empty journal. The stub now mirrors the real contract (invokes the callback on successful construction, typed via `Parameters<StreamManager["startStream"]>`). All 101 aiService tests pass locally; no production code changed.

PR #3873, opened 2026-08-19
🤖 feat: advertise deferred MCP tool catalog in tool_catalog_search description

ThomasK33:
## Summary

Advertises the surface area of deferred MCP tools to the model by dynamically appending a compact, per-server catalog overview to the `tool_catalog_search` tool description, solving the "unknown unknowns" discovery problem while keeping full tool descriptions and parameter schemas deferred until activation.

## Background

With the tool-search experiment, MCP tools are deferred: they exist but are invisible to the model until discovered via `tool_catalog_search`. Nothing advertised *what* was deferred — no connected server names, tool counts, or domains — so the model had to guess search keywords blindly and could never know which capabilities existed.

## Implementation

- **Dynamic tool description (the core change):** `prepareToolSearch` now builds a deterministic overview via the new pure `buildToolCatalogOverview()` and appends it to `tool_catalog_search`'s static description on a shallow copy of the tool:

  ```
  Deferred tool catalog overview (search to activate):
  - github (12 tools): github_create_issue, github_get_pr, … (+4 more)
  - slack (2 tools): slack_list_channels, slack_send_message
  ```

  Tools are grouped by originating MCP server, sorted lexicographically, capped at 8 names per server with a `(+N more)` suffix. Only **names** are listed — descriptions and JSON schemas stay deferred until a tool is activated by search, so the token-bloat guarantee is preserved. The overview flows through every catalog (re)build path, including model-fallback and post-hook `rebuildToolSearchState`.

- **Server attribution plumbing:** `MCPServerManager.collectTools` now also returns a `toolServerNames` map (provider-safe namespaced tool name → server name), threaded through `getToolsForWorkspace` and `aiService` into the catalog inputs as `mcpToolServers`. This survives tool-name normalization/hash-suffixing, so grouping stays correct even when the prefix no longer equals the server name.

- **Server-aware search:** `searchToolCatalog` scores server-name token matches (+3, ranked between name and description hits) so a query like "github" surfaces that server's tools even after name normalization, and `ToolSearchMatch`/`ToolSearchToolResult` now report `serverName`.

## Validation

- New `catalog advertising` unit tests: server attribution in `buildToolCatalog`, deterministic per-server grouping, per-server eliding, empty-catalog/unknown-server cases, description augmentation in `prepareToolSearch`, deferred-tool records staying untouched, and server-name-only search hits.
- `make static-check` plus toolCatalog/toolSearch/toolDefinitions/mcpServerManager/streamManager suites green locally.

## Risks

Low. Deferral gating, activation semantics, and `prepareStep` scoping are unchanged; the only behavioral deltas are a longer `tool_catalog_search` description (one shallow-copied tool object per stream) and an extra optional `serverName` field on search results. Fallback/rebuild paths reuse the same `prepareToolSearch`, so descriptions stay consistent across model fallback.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `off` • Cost: `$5.32`_

### DIFF
--- a/src/common/types/tools.ts
+++ b/src/common/types/tools.ts
@@ -466,6 +466,7 @@ export type ToolSearchToolArgs = z.infer<typeof TOOL_DEFINITIONS.tool_catalog_se
 
 export interface ToolSearchToolResult {
   query: string;
+  /** serverName is a bounded display label, not the raw server-config key. */
   matches: Array<{ name: string; description: string; serverName?: string }>;
   /** Total deferred-catalog size, so the model/UI can see there are more undiscovered tools. */
   totalDeferred: number;
--- a/src/common/types/tools.ts
+++ b/src/common/types/tools.ts
@@ -466,7 +466,7 @@ export type ToolSearchToolArgs = z.infer<typeof TOOL_DEFINITIONS.tool_catalog_se
 
 export interface ToolSearchToolResult {
   query: string;
-  matches: Array<{ name: string; description: string }>;
+  matches: Array<{ name: string; description: string; serverName?: string }>;
   /** Total deferred-catalog size, so the model/UI can see there are more undiscovered tools. */
   totalDeferred: number;
 }
--- a/src/common/utils/tools/toolCatalog.test.ts
+++ b/src/common/utils/tools/toolCatalog.test.ts
@@ -176,6 +176,42 @@ describe("catalog advertising (overview + server names)", () => {
     expect(prep.tools.slack_send_message.description).toBe("Send a message to a Slack channel");
   });
 
+  test("re-running prepareToolSearch on an augmented toolset does not stack overviews", () => {
+    const first = prepareToolSearch({
+      tools: baseTools(),
+      mcpToolNames: MCP_NAMES,
+      mcpToolServers: MCP_TOOL_SERVERS,
+    });
+    // Rebuild paths (assembly hooks, model fallback) feed the augmented record back in.
+    const second = prepareToolSearch({
+      tools: first.tools,
+      mcpToolNames: MCP_NAMES,
+      mcpToolServers: MCP_TOOL_SERVERS,
+    });
+    const description = second.tools[TOOL_SEARCH_TOOL_NAME].description;
+    expect(description).toBe(first.tools[TOOL_SEARCH_TOOL_NAME].description);
+    const occurrences = String(description).split(
+      "Deferred tool catalog overview (search to activate):"
+    ).length;
+    expect(occurrences - 1).toBe(1);
+  });
+
+  test("rebuildToolSearchState replaces (not stacks) the overview on the augmented tool", () => {
+    const first = prepareToolSearch({
+      tools: baseTools(),
+      mcpToolNames: MCP_NAMES,
+      mcpToolServers: MCP_TOOL_SERVERS,
+    });
+    const state = first.state!;
+    const rebuilt = rebuildToolSearchState(state, {
+      tools: first.tools,
+      mcpToolNames: MCP_NAMES,
+      mcpToolServers: MCP_TOOL_SERVERS,
+    });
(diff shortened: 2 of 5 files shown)


ThomasK33:
@codex review

Addressed the double-append finding: `prepareToolSearch` now strips any previously appended overview (via the `Deferred tool catalog overview` header marker) before appending a fresh one, so post-hook `rebuildToolSearchState` and model-fallback re-invocations stay idempotent. Added regression tests covering repeated `prepareToolSearch` and `rebuildToolSearchState` on the augmented toolset (485a271cb).

ThomasK33:
@codex review

Fixed: `stripToolCatalogOverview` now removes only the generated block (header + contiguous bullet lines) and rejoins text before/after it, so `request.assemble` annotations appended after the overview survive re-preparation. Regression test added (11a205bca).

ThomasK33:
@codex review

Fixed: the generated overview is now delimited by an explicit end marker (`(end of deferred tool catalog overview)`), and stripping removes exactly the marker-delimited region — no content-shape heuristics — so bullet-form hook annotations survive. If the end marker is missing, the description is left untouched instead of guessing at block boundaries. Regression test added for a bullet-form suffix (202f8fba1).

ThomasK33:
@codex review

Both findings addressed in 4fa2193b2:
- **Bounded overview:** server labels are clamped to 64 chars, and the whole overview enforces a 2000-char budget — servers beyond it collapse into an `…and N more servers` line pointing at search. Tests cover a 200-char server name and a 200-server catalog.
- **Whitespace preservation:** stripping now removes exactly the marker-delimited region plus the separator we emitted before it, with no trimming — hook-authored suffixes including indented Markdown survive verbatim (regression test with an indented block).

ThomasK33:
@codex review

Fixed: server labels are now sanitized before rendering — both reserved marker strings are scrubbed and newlines flattened — so a label can neither terminate nor forge the generated block; marker-based stripping stays exact and rebuilds stay idempotent. Tool names need no scrubbing (provider-safe `[a-zA-Z0-9_-]`). Regression test with an adversarial server name embedding the end marker + newlines (0edcb108b).

ThomasK33:
Correction: the marker-sanitization fix landed as 7def2fcac (SHA typo in the previous comment).

ThomasK33:
@codex review

Fixed in fa8cc8593: label scrubbing now iterates to a fixpoint — the loop repeats until the sanitized label contains neither reserved marker, so deletion passes can no longer synthesize a marker by joining fragments. Termination is guaranteed because each pass strictly shortens the string. Regression test added with the exact fragment-joining construction from the finding.

ThomasK33:
@codex review

Fixed in b3eff1c6b: search matches now return a bounded display label via a shared `displayServerLabel` (same sanitize + 64-char clamp as the overview), so an unbounded server-config key can no longer be repeated across up to 25 matches. Raw server names stay internal for grouping/scoring only; `ToolSearchMatch`/`ToolSearchToolResult` docs updated. Regression test with a 300+ char server name added.

ThomasK33:
@codex review

Fixed in af0f40af1: `displayServerLabel` clamps the input to a 512-char prefix before the fixpoint scrub, bounding the worst case regardless of key size (a partial marker at the cut is harmless since stripping matches only complete markers), and marks truncation with an ellipsis. Regression test builds a ~195KB key with 5000 nested markers and asserts bounded runtime plus marker-free, ≤64-char output in both the overview and search matches.

ThomasK33:
@codex review

Both findings addressed in 783667119 with one change: `searchToolCatalog` now tokenizes the bounded `displayServerLabel` (the exact string the overview advertises) instead of the raw config key, memoized once per distinct server per search call.
- Bounded work: the huge-key regression test now spreads the ~195KB nested-marker key across 50 catalog entries and still asserts bounded runtime.
- Consistent scoring: new test truncates a 100-char single-token server name to its advertised 63-char+ellipsis label, queries with that visible label, and asserts the server's tool matches.

PR #3852, opened 2026-08-14
[task-service] 🤖 preserve nested workspace-turn correlation

coadler:
## Summary

Preserve workspace-turn correlation when a spawned workspace receives nested agent reports.

## Background

A spawned workspace can create a nested agent.

The nested agent can send progress or a terminal report.

Those reports start a stream in the spawned workspace.

The stream lacked the active workspace-turn metadata.

TaskService then marked the parent workspace turn as interrupted or incomplete.

An app restart could also lose the correlation before the continuation finished.

Older terminal report rows can also exist without the new metadata.

## Implementation

- Resolve the active workspace-turn handle from the durable handle store.
- Pass exact metadata through progress report wake-ups.
- Pass exact metadata through terminal attention resumes.
- Persist exact metadata on new synthetic terminal report history rows.
- Backfill metadata only when an existing terminal report lacks correlation.
- Preserve valid correlation from an earlier workspace turn.
- Settle a correlated turn when progress wake dispatch is canceled or fails before streaming.
- Persist exact metadata in startup retry options.
- Re-derive the correlation from history during startup retry recovery.
- Detect queued continuations with the same handle, owner, and turn IDs.
- Keep manual messages and unrelated queued work as superseding input.

## Validation

- `make static-check`
- Full `src/node/services/taskService.test.ts`
- Focused workspace-turn regression tests
- `src/node/services/messageQueue.test.ts`
- `src/node/services/agentSession.workspaceTurnInheritance.test.ts`

## Risks

The change touches stream-end and queued-message lifecycle logic.

Exact metadata matching limits the change to the affected workspace turn.

History updates preserve the original history sequence.

Failed correlated report dispatches now settle instead of leaving a turn running.

Existing uncorrelated stream behavior remains covered by regression tests.

---

_Generated with `mux` • Model: `openai:gpt-5.6-luna` • Thinking: `max` • Cost: `$7.74`_

PR #3874, opened 2026-08-19
🤖 fix: settle workspace-turn queue-cut supersedes as interrupted, not error

ThomasK33:
## Summary

Settle delegated workspace-turn queue cuts (finishReason `tool-calls`) as `interrupted` with a human-readable supersede reason instead of a red task failure that leaks internal finishReason jargon to the owner.

## Background

When a parent delegates a turn to another workspace (`task(kind=workspace)`) and someone types a manual message into that target workspace, the queued input dispatches at the next tool boundary and soft-cuts the in-flight stream with finishReason `tool-calls` (the `hasQueuedMessages("tool-end")` stop condition). Same-turn continuations (bash-monitor wakes, report wake-ups) already defer settlement (#3797, #3852) — but any *other* queued input supersedes the delegated turn, and `buildTerminalWorkspaceTurnRecordFromEvent` lumped that cut into the generic non-`stop` truncation branch:

> **1 task failed** — `Workspace turn ended before completion (finishReason: tool-calls)`

That is wrong on both axes: the delegated work did not fail (the child keeps working under the new input), and `finishReason: tool-calls` is internal streaming vocabulary that should never surface as an erroneous state to the owner. The alarming failure wake also baits the parent into re-delegating a duplicate turn.

## Implementation

- `buildTerminalWorkspaceTurnRecordFromEvent`: a `tool-calls` finish on a correlated delegated turn is by construction a queue cut whose same-turn continuations were already deferred by the caller — settle it as `interrupted` with the reason *"Workspace turn superseded by new input in the target workspace; the workspace continues under that input and this delegated turn will not report"*. The task UI renders `interrupted` with the calm tone (no danger banner, no force-expand), and terminal-attention wakes report outcome `interrupted`.
- Genuine truncation finishes (`length`, etc.) keep the existing `error` settlement.
- The new supersede interrupt stays self-heal eligible in `isSelfHealEligibleSettledWorkspaceTurn` (it previously was, via `status: "error"`), so a false supersede is still corrected by late correlated evidence of the same turn.

## Validation

Reconstructed the reported incident from live session data (`chat.jsonl`, `task-handles/wst_*.json`, terminal-attention records): a manual user message queued into the delegated child cut the stream at a tool boundary and the handle settled `error` with the finishReason message. Updated the two tests that pinned the old error behavior; `workspace-turn`/`resettle`/`self-heal`/`revive` test groups pass. The 3 remaining monolithic-run failures in `taskService.test.ts` reproduce identically on the base commit (cross-test interference; they pass in isolation).

## Risks

Low severity, scoped to delegated workspace-turn settlement. Owners now see `interrupted` instead of `error` for queue-cut supersedes; waiter promises still reject (message text changed). Truncation/error paths and same-turn continuation deferral are untouched.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$8.81`_

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -4859,6 +4859,20 @@ describe("TaskService", () => {
     expect(snapshot?.deferredMessageIds).toBeUndefined();
   });
 
+  test("repeat interrupt of an already-interrupted workspace turn is a no-op", async () => {
+    // A queue-cut supersede settles the handle interrupted while the target
+    // workspace keeps streaming under the new input. A stale task_stop for the
+    // settled handle must not stop that unrelated stream.
+    const { parentId, taskService, aiMocks } = await startWorkspaceTurnForTest();
+    const first = await taskService.interruptWorkspaceTurn(parentId, "wst_handle");
+    expect(first.success).toBe(true);
+    aiMocks.stopStream.mockClear();
+
+    const repeat = await taskService.interruptWorkspaceTurn(parentId, "wst_handle");
+    expect(repeat).toEqual(Ok({ workspaceId: "childworkspace" }));
+    expect(aiMocks.stopStream).not.toHaveBeenCalled();
+  });
+
   test("workspace-turn stale recovery skips deferred pre-handoff stream-end history", async () => {
     const { config, parentId, projectPath, taskService, historyService } =
       await startWorkspaceTurnForTest();
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -5587,6 +5587,61 @@ describe("TaskService", () => {
     });
   });
 
+  test("snapshot history repair does not upgrade an error settlement to a supersede", async () => {
+    // A tool-calls turn correctly settled as error (no live queue-cut evidence,
+    // e.g. a required-tool stop) must stay an error even after unrelated user
+    // messages land in child history: order is not causal queue-cut evidence.
+    const { config, parentId, taskService, historyService } = await startWorkspaceTurnForTest();
+    const muxMetadata = {
+      type: "workspace-turn-task" as const,
+      taskHandleId: "wst_handle",
+      ownerWorkspaceId: parentId,
+      turnId: "turn",
+    };
+    expect(
+      (
+        await historyService.appendToHistory(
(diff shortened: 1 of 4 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed: `task_await` now surfaces the persisted interrupt reason via a shared `interruptedWorkspaceTurnResult` builder (all four interrupted branches), with a test covering the supersede case.

ThomasK33:
@codex review

Addressed: `interruptWorkspaceTurn` now short-circuits already-`interrupted` handles as an idempotent no-op (no stream stop, no side effects), so a stale `task_stop` cannot abort the superseding stream in the target workspace. Regression test added.

ThomasK33:
@codex review

Addressed both findings:
1. Supersede classification now requires queue-cut evidence: live queued/preparing/streaming successor in the finalize path (`hasLiveQueueCutSupersedeEvidence`), or a newer user message in the history repair scans. Without evidence, `tool-calls` finishes keep the original truncation error handling (test reverted accordingly).
2. Direct-parent delivery gates (`workspaceTurnRequiresDirectParentDelivery`, `deliverPersistentChildWorkspaceTurnResultUnlocked`) now accept the distinguished supersede interrupt and deliver it as `workspace_turn_superseded`; explicit cancellations stay silent. Behavioral test added for both the delivery and the silence.

ThomasK33:
@codex review

Addressed: the snapshot history-repair path now treats the persisted supersede classification as queue-cut evidence when rebuilding from the SAME correlated final (`event.messageId === record.messageId`), so the race window before the superseding user message lands cannot downgrade the record to a truncation error. A different correlated final (contradictory same-turn evidence) can still resettle. Regression test added and verified to fail without the fix.

ThomasK33:
@codex review

Addressed: history scans no longer infer queue-cut evidence from message order. The repair path passes evidence only when preserving an already-persisted supersede rebuilt from the same correlated final; stale recovery settles tool-calls finals as errors (still self-heal eligible), since genuine supersedes are classified live at stream-end. Added a regression test for the false-upgrade case (verified to fail without the fix) alongside the existing no-downgrade test.

PR #3875, opened 2026-08-19
🤖 fix: retain reusable persistent sub-agents

ammar-agent:
## Summary

Makes persistent sub-agent retention the default so parents preserve valuable idle context and reuse or retitle suitable children instead of routinely deleting and replacing them.

## Background

The lifecycle prompt treated completed work and PR boundaries as cleanup signals, which encouraged irreversible removal even though inactive children are inexpensive and can carry useful context into later turns. A deep-review-specific instruction reinforced the same behavior.

## Implementation

- instruct parents to consider relevant inactive standalone children before spawning replacements
- encourage reawakening and retitling when prior context or expertise fits the new responsibility
- require repository-dependent reuse to verify or synchronize the retained checkout when its snapshot may be stale
- reserve `task_remove` for explicit user requests or clearly obsolete context with no plausible future value
- remove the deep-review cleanup override and align conditional `task_list` guidance
- regenerate the system-prompt docs and built-in skill artifacts

## Validation

- targeted lifecycle, task-tool, and system-message tests
- generated docs and built-in skill synchronization checks
- TaskToolCall UI test
- `make static-check`

## Risks

Low runtime risk: this changes model-facing lifecycle policy and generated documentation, not task persistence mechanics. Model compliance remains probabilistic, while destructive removal and spawning a fresh child remain explicitly available when warranted.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `xhigh` • Cost: `$7.08`_


ammar-agent:
@codex review

Please review the retention-first persistent sub-agent lifecycle policy, including the removal of the deep-review cleanup override.

ammar-agent:
Addressed the stale-checkout concern by clarifying that reawakening preserves the child checkout. Repository-dependent reuse now requires the retained snapshot to be appropriate or an explicit instruction to verify and synchronize before acting; otherwise the parent should spawn a new child.

@codex review

PR #3876, opened 2026-08-19
🤖 fix: align Docker server MUX_ROOT persistence contract

### LINKED ISSUE #3802: 🤖 fix: align Docker server MUX_ROOT persistence contract
## Summary

The Docker server image currently creates and mounts `/root/.mux`, but exports `MUX_HOME=/root/.mux`. The application resolver (`getMuxHome()` in `src/common/constants/paths.ts`) recognizes `MUX_ROOT`, not `MUX_HOME`.

The root image works today only because `os.homedir()` resolves to `/root`. Under `docker run --user ...`, that implicit fallback can select a different home directory, bypass the mounted volume, or fail on writes.

## Proposed contract

- Make `MUX_ROOT` the explicit Docker image contract: set `MUX_ROOT=/root/.mux` in `Dockerfile`.
- Retain the root default and the existing `docker-compose.yml` persistence target, `mux-data:/root/.mux`, so existing volumes and bind mounts remain compatible.
- Do not add `MUX_HOME` as an application alias: `MUX_ROOT` is the existing resolver override and dual-variable precedence would create an unnecessary compatibility surface.
- Treat non-root execution as an explicit operator configuration: users running the image with `--user` must provide a writable, mounted `MUX_ROOT` path with compatible ownership. The image should not infer a different home, chown mounted volumes, or migrate state.

## Implementation and validation

1. Replace the Dockerfile `MUX_HOME` environment declaration with `MUX_ROOT=/root/.mux`, retaining directory creation and the current volume location.
2. Add a behavioral unit test in `src/common/constants/paths.test.ts` for `getMuxHome()` proving `MUX_ROOT` is returned exactly and restoring mutated environment variables afterward. If practical, also prove `MUX_HOME` alone does not alter resolution.
3. Run the focused test, Dockerfile lint/static checks, and a local image build/smoke test when Docker is available.

Existing CI already has Docker image startup smoke coverage in `.github/workflows/pr.yml`, but it checks health/version rather than persistence-root resolution. The resolver test should protect the actual contract without adding a brittle filesystem-only container assertion.

## Context

A prior search found no exact duplicate. #3699 may be a user-facing symptom but is unconfirmed. PR #849 introduced the current image convention; PR #1587 is related precedent for non-root runtime handling.

---

_Generated with `mux` • Model: `openai:gpt-5.6-terra` • Thinking: `high` • Cost: `$0.04`_

ibetitsmike:
Fixes #3802

## Summary

The Docker server image declared `ENV MUX_HOME=/root/.mux`, but the application resolver `getMuxHome()` (`src/common/constants/paths.ts`) only honors `MUX_ROOT`, so `MUX_HOME` was dead configuration. This PR makes `MUX_ROOT=/root/.mux` the explicit image contract and adds behavioral tests pinning the resolver's precedence.

## Background

The image worked only because `os.homedir()` happens to resolve to `/root` in the root container, so state landed in the mounted volume (`mux-data:/root/.mux`) by coincidence. Under `docker run --user ...`, the implicit homedir fallback can resolve elsewhere, silently bypassing the mounted volume or failing on writes.

Per the issue's contract: `MUX_HOME` is intentionally NOT added as an application alias (no dual-variable precedence), directory creation and the volume location are unchanged so existing volumes stay compatible, and non-root execution remains an explicit operator concern (no chown/migration logic).

## Validation

- New tests in `src/common/constants/paths.test.ts`: `MUX_ROOT` is returned exactly; `MUX_HOME` alone does not alter resolution; `-dev` suffix branch covered; mutated env vars restored. Red-green verified: removing the `MUX_ROOT` branch from `getMuxHome()` fails exactly the precedence test.
- Local image build + smoke: `docker inspect` shows `MUX_ROOT=/root/.mux` and no `MUX_HOME`; container `/health` returns OK; server state (`sessions/`, `logs/`, `server.lock`) is written under `/root/.mux`.
- Remote dogfood UAT was unavailable for this Docker-image scenario (no agent-enabled template provisions coder/mux); the local image smoke above stands in as the scenario evidence.

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -130,7 +130,7 @@ RUN mkdir -p /root/.mux
 
 # Default environment variables
 ENV NODE_ENV=production
-ENV MUX_HOME=/root/.mux
+ENV MUX_ROOT=/root/.mux
 
 # Expose server port
 EXPOSE 3000
--- a/src/common/constants/paths.test.ts
+++ b/src/common/constants/paths.test.ts
@@ -7,10 +7,10 @@ import {
   rmSync,
   writeFileSync,
 } from "fs";
-import { tmpdir } from "os";
+import { homedir, tmpdir } from "os";
 import { join } from "path";
-import { afterEach, describe, expect, test } from "bun:test";
-import { cleanupObsoleteMuxBinArtifacts } from "./paths";
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { cleanupObsoleteMuxBinArtifacts, getMuxHome } from "./paths";
 
 const tempDirs: string[] = [];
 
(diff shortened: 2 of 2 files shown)


ibetitsmike:
@codex review

PR #3877, opened 2026-08-19
🤖 fix: detect devcontainer CLI installed as .cmd shim on Windows

### LINKED ISSUE #3717: Windows: devcontainer CLI not detected when spawned without executable extension
`mux` appears to fail detecting an installed `devcontainer` CLI on Windows.

At `src/node/runtime/devcontainerCli.ts` (around line 181), the code runs:

```ts
spawn("devcontainer", ["--version"], { ... })
```

Source: https://github.com/coder/mux/blob/main/src/node/runtime/devcontainerCli.ts#L181

On Windows, `devcontainer` is often resolved through extensions such as `.bat`/`.cmd`/`.ps1`. With plain `child_process.spawn(...)` (without shell resolution or a cross-platform wrapper), command lookup can fail even when `devcontainer` is installed and available in `PATH`.

### Steps to reproduce
1. Use Windows (PowerShell or CMD), with `devcontainer` CLI correctly installed and callable from terminal.
2. Run mux in a flow that triggers the devcontainer version check.
3. Observe that mux reports `devcontainer` as missing/unavailable.

### Expected behavior
Mux should detect `devcontainer` successfully on Windows when the CLI is installed and in `PATH`.

### Actual behavior
Detection fails because `spawn("devcontainer", ...)` does not reliably resolve the Windows command variant.

### Environment
- OS: Windows (10/11)
- Repo: `coder/mux`
- File: `src/node/runtime/devcontainerCli.ts`
- Observed near: line 181
- Commit referenced: `370cb3b97f5a173fc6c8ea88edcd0f1b16bf6e9f`

### Possible fix
Either of these approaches should make detection cross-platform-safe:
1. Use `spawn(..., { shell: true })` for this command check, or
2. Replace direct `spawn` with a wrapper like `cross-spawn`, which handles Windows command resolution (`PATHEXT`, `.cmd`/`.bat`, etc.) more reliably.

### Additional notes
Happy to test a patch on Windows if helpful.

ibetitsmike:
## Summary

On Windows, mux reported the devcontainer CLI as missing even when it was installed and on PATH. All three devcontainer spawn sites now resolve and launch the CLI through a shared, platform-aware helper; POSIX behavior is unchanged, and no new dependency is added.

Fixes #3717

## Background

npm installs the devcontainer CLI on Windows as `devcontainer.cmd`/`devcontainer.ps1` shims. Node's `spawn` without `shell: true` does not consult PATHEXT, so `spawn("devcontainer", ...)` fails with ENOENT at:

- `checkDevcontainerCliVersion()` (`src/node/runtime/devcontainerCli.ts`)
- `devcontainerUp()` (`src/node/runtime/devcontainerCli.ts`)
- `DevcontainerRuntime.exec` (`src/node/runtime/DevcontainerRuntime.ts`)

Constraints that ruled out the simpler shapes:

- Node >= 20.12 (Electron 40 bundles Node 22) throws `EINVAL` when spawning a `.cmd`/`.bat` path without a shell (CVE-2024-27980 hardening), so resolving the shim path and spawning it directly is not viable.
- Blanket `shell: true` is a quoting/injection hazard: Node does not escape args under `shell: true`, and the exec site passes arbitrary payloads like `-- bash -c "cd '<path>' && <command>"` through to cmd.exe.
- cross-spawn (the usual answer) only applies its double metachar escaping to `node_modules\.bin\*.cmd` paths. Global npm shims (`%APPDATA%\npm\devcontainer.cmd`) are generated by the same cmd-shim library and re-expand `%*` through a second cmd parse, so cross-spawn under-escapes them and metachar payloads can be mangled or executed on the host (flagged by Codex review on an earlier revision of this PR that dispatched through cross-spawn).

## Implementation

One shared `spawnDevcontainer()` helper in `devcontainerCli.ts`, used by all three sites:

- POSIX: native `spawn("devcontainer", args, options)`, byte-for-byte previous behavior, no lookup subprocess.
- win32: resolve the real entry via `where.exe` (bounded by a 5s timeout; successful resolutions are cached so repeat spawns skip the lookup subprocess entirely), then:
  - `.exe`/`.com`: spawn the resolved path directly.
  - `.cmd`/`.bat`: wrap in `cmd.exe /d /s /c` with cross-spawn's escape algorithm (vendored, MIT, ~25 lines) with double escaping always on, since devcontainer `.cmd`/`.bat` entries are npm cmd-shims that re-expand `%*`. `windowsVerbatimArguments: true` prevents Node re-quoting the pre-escaped line.
  - Nothing spawnable found or lookup failure: fall back to the bare name, keeping the status-quo "CLI not detected" failure surface.
- The helper takes an injectable deps seam (`platform`, lookup, spawn, resolution cache) so both branches are unit-testable on any host without mutating `process.platform` or module mocks.
- Options (`stdio`, `timeout`, `cwd`, `detached`, `windowsHide`) pass through unchanged at every site.

## Validation

- Unit tests pin the exact final dispatch, including the full escaped `cmd.exe /d /s /c` command line for a metachar-heavy `bash -c` payload (spaces, `&&`, quotes, `|`, `%`). The expected string is reference output generated from cross-spawn@7.0.6's `escape.argument(arg, doubleEscapeMetaChars=true)`, so the vendored copy is verified against the battle-tested original rather than against itself.
- Further coverage: `.exe` direct dispatch, no-spawnable-candidate fallback, lookup-failure fallback, success-only caching (lookup runs once), failure non-caching, and POSIX native-spawn with zero lookups.
- Red-green verified twice: removing the win32 branch fails the win32 tests; removing only the second escape pass fails the pinned-escaping test. POSIX baseline passes in both toggles.

**Windows verification gap:** no Windows host was available, so the win32 path is verified by unit tests only, not end-to-end on Windows. The issue reporter kindly offered to test a patch on Windows — testing help on a real Windows setup is very welcome.

## Risks

Low on POSIX (dispatch is byte-for-byte the previous call). On Windows the CLI now actually launches via an explicitly escaped cmd.exe shim invocation; process-tree cleanup at the exec site already uses `taskkill /T` semantics (`killProcessTree`), which covers the extra shim layer. The vendored escaping is the highest-risk code, mitigated by pinning its output against cross-spawn's reference behavior. Worst case on unusual Windows setups (e.g. `.ps1`-only installs) degrades to the current "CLI not detected" behavior.

---


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Both round-1 findings addressed in b5c937bb0:

- **P1 (global npm shim under-escaping):** dropped cross-spawn entirely. The helper now dispatches `.cmd`/`.bat` through `cmd.exe /d /s /c` itself, using cross-spawn's vendored escape algorithm with double escaping always applied (devcontainer `.cmd` entries are npm cmd-shims that re-expand `%*`). The test now pins the exact final escaped command line for a metachar-heavy `bash -c` payload, with the expected string generated as reference output from cross-spawn's `escape.argument(arg, true)`. `.exe`/`.com` candidates spawn directly with no wrapper.
- **P2 (blocking unbounded lookup per spawn):** the `where.exe` lookup now has a 5s timeout and successful resolutions are cached, so repeat spawns (including every `exec`) skip the lookup subprocess entirely; failures are not cached.

Net dependency delta is now zero (cross-spawn and @types/cross-spawn removed again).

ibetitsmike:
@codex review

Both round-2 findings addressed in d6ce9fbce:

- **P1 (orphaned CLI under the cmd.exe wrapper):** `devcontainerUp`'s abort and timeout handlers now call `terminateDevcontainerProc`, which kills the full tree via `killProcessTree` (taskkill /T) on win32 and keeps the graceful direct-child SIGTERM on POSIX. The redundant spawn-level `timeout` option is removed since Node's built-in kill signals only the wrapper; the explicit timer is now the single timeout mechanism. Covered by unit tests on the platform-branch dispatch.
- **P2 (blocking where.exe on every availability check when the CLI is absent):** lookup misses are now negatively cached for 30s (`missedAtMs` + injected clock), so repeated `checkRuntimeAvailability` calls hit the cache instead of re-running the synchronous lookup; a later successful resolution clears the negative entry. Covered by TTL-expiry and cache-clearing tests with a fake clock.

ibetitsmike:
@codex review

Round-3 finding addressed in d084ba62a: `checkDevcontainerCliVersion` now uses an explicit timer that calls `terminateDevcontainerProc` (tree-kill on win32), force-closes the stdio pipes via `forceCloseStdio`, and settles immediately, so a surviving CLI descendant can neither keep running nor hold `checkRuntimeAvailability` pending. The spawn-level `timeout` option is gone from all devcontainer spawn sites now; swept the class: `docker rm` is a direct child on all platforms with its own timeout, and the exec site already settles via DisposableProcess cleanup (killProcessTree + forceCloseStdio).

PR #3879, opened 2026-08-19
🤖 fix: preserve legacy MCP OAuth binding fields for downgrade compatibility

### LINKED ISSUE #3823: 🤖 MCP OAuth: preserve legacy @ai-sdk/mcp credential binding fields for downgrade compatibility
## Background

PR #3822 migrated Mux's MCP client from `@ai-sdk/mcp` to the official `@modelcontextprotocol/client` SDK v2. During review, Codex flagged a P1 in `src/node/services/mcpOauthService.ts`: https://github.com/coder/mux/pull/3822#discussion_r3743928013

This was explicitly **deferred** from the migration PR because it only affects a *downgrade* back to an `@ai-sdk/mcp`-based Mux version, and downgrade compatibility was not a requirement for that work.

## Problem

When a user authorizes (or refreshes) an MCP OAuth server after the SDK switch, the official SDK supplies its `issuer`-based credential shape but not the legacy `authorization_server` / `token_endpoint` binding pair. `saveTokens()` and `saveClientInformation()` persist those objects wholesale, so credentials issued or refreshed by SDK v2 carry only the v2 `issuer` stamp. Legacy bindings that existed before the upgrade are round-tripped, but newly issued/refreshed credentials never gain them.

## Symptom

After downgrading Mux to a version that uses `@ai-sdk/mcp`, persisted OAuth credentials for affected servers lack the binding fields the legacy client requires for refresh-token use, forcing an interactive re-login for each such MCP server.

## Proposed fix

Persist the legacy `authorization_server` / `token_endpoint` binding pair alongside the v2 `issuer` whenever credentials are issued or refreshed by the new SDK (derivable from the resolved authorization server metadata), rather than only preserving bindings that happened to exist before the upgrade.

---

_Generated with `mux` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$123.35`_

ibetitsmike:
Fixes #3823

## Summary

Newly issued or refreshed MCP OAuth credentials now carry the legacy `@ai-sdk/mcp` `authorization_server`/`token_endpoint` binding pair alongside the SDK v2 `issuer` stamp, so downgrading Mux to an `@ai-sdk/mcp`-based version no longer forces an interactive re-login per MCP server.

## Background

PR #3822 migrated the MCP client to the official `@modelcontextprotocol` SDK v2. Its `saveTokens()`/`saveClientInformation()` persist the SDK-supplied objects wholesale, and SDK v2 objects carry only the v2 `issuer` stamp, never the legacy binding pair that `@ai-sdk/mcp`'s `auth()` requires before it will use a stored `refresh_token`. Worse than the issue states: a background token refresh on a pre-v2 credential replaced `creds.tokens` wholesale and wiped an existing binding pair from disk. Codex flagged this as a P1 on #3822 (deferred there because downgrade compatibility was out of scope for the migration): https://github.com/coder/mux/pull/3822#discussion_r3743928013

## Implementation

- Both OAuth providers (interactive flow + background refresh) now implement the SDK's `saveDiscoveryState` hook, caching the discovery result in memory (on the flow object / in the provider closure). Only the write hook is implemented; the `discoveryState()` read-back is intentionally omitted so SDK discovery behavior is unchanged.
- `saveTokens()`/`saveClientInformation()` accept an optional legacy binding derived from that discovery state (`authorizationServerUrl` + `authorizationServerMetadata.token_endpoint`, atomic pair or nothing) and stamp it onto the persisted object without touching the v2 `issuer`.
- When no fresh binding is available (e.g. metadata discovery failed), a pre-existing complete pair on the old entry is preserved instead of being wiped. Preservation is skipped when the defensive server-URL-mismatch branch fires, since old bindings for a different server must not leak onto new credentials.

## Validation

- New behavioral store tests drive the real service through the public provider seam and assert the persisted `mcp-oauth.json`: fresh stamping from discovery state, refresh preserving an existing pair (fails on main, the refresh-wipe regression), `saveClientInformation` stamping, and partial discovery never producing a half-pair.
- Dogfood UAT recorded as not applicable: persistence-layer fix with no UI surface (explicit opt-out); the credential-store unit tests are the acceptance evidence.

## Risks

Low. Scope is limited to the two credential-save paths in `mcpOauthService.ts`; the stamped fields are already round-tripped by the store parser and ignored by SDK v2. Worst case regression is a malformed binding pair on saved credentials, which the existing self-healing read path (`parseAuthorizationServerBinding`) drops as a pair.

---

### DIFF
--- a/src/node/services/mcpOauthService.test.ts
+++ b/src/node/services/mcpOauthService.test.ts
@@ -11,7 +11,20 @@ import {
   probeServerForBearerChallenge,
   resolveOAuthScope,
 } from "./mcpOauthService";
+import type { OAuthDiscoveryState } from "@modelcontextprotocol/client";
 import type { MCPOAuthClientInformation, MCPOAuthTokens } from "@/common/types/mcpOauth";
+
+interface StoredCredentialsSnapshot {
+  serverUrl: string;
+  updatedAtMs: number;
+  clientInformation?: MCPOAuthClientInformation;
+  tokens?: MCPOAuthTokens;
+}
+
+interface StoreSnapshot {
+  version: 2;
+  entries: Record<string, StoredCredentialsSnapshot>;
+}
 
 function getStoreFilePath(muxHome: string): string {
   return path.join(muxHome, "mcp-oauth.json");
@@ -26,6 +39,17 @@ describe("McpOauthService store", () => {
 
   const serverName = "test-server";
   const serverUrl = "https://example.com";
+  const authorizationServerUrl = "https://auth.example.com/";
+  const tokenEndpoint = "https://auth.example.com/token";
+  const discoveryState: OAuthDiscoveryState = {
+    authorizationServerUrl,
+    authorizationServerMetadata: {
+      issuer: authorizationServerUrl,
+      authorization_endpoint: "https://auth.example.com/authorize",
+      token_endpoint: tokenEndpoint,
+      response_types_supported: ["code"],
+    },
+  };
 
   beforeEach(async () => {
     muxHome = await fs.mkdtemp(path.join(os.tmpdir(), "mcp-oauth-home-"));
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

Issue #3819, opened 2026-08-08
🤖 Docker credential sharing is not portable across non-root users and host-specific Git helpers


alecsg77:
## Summary

Docker workspace credential sharing currently assumes that the host global Git configuration can be copied into a container unchanged. That assumption fails in two related ways:

1. `DockerRuntime.setupCredentials()` copies the host config to `<container>:/root/.gitconfig`, while Docker workspace images can run as a non-root default user.
2. Even when the file is written to the effective container user’s home directory, a host Git config can reference credential helpers, includes, commands, sockets, files, keyrings, or network configuration that are not available inside the container.

The first is a concrete bug. The second is a broader design question about the supported contract for Git and SSH credential forwarding in Docker and Dev Container workspaces.

## Current behavior

- `src/node/runtime/DockerRuntime.ts` calls `setupCredentials()` both after fresh provisioning and when `postCreateSetup()` reuses an already-valid forked container.
- `setupCredentials()` currently uses `docker cp` to copy the host config to `/root/.gitconfig`, then runs `gh auth setup-git` when `GH_TOKEN` is available.
- Fresh provisioning detects the container uid/gid/home before credential setup, but the reuse/fork branch invokes credential setup before `containerHome` has been cached.
- `DevcontainerRuntime.setupCredentials()` already uses a more user-correct transport: it reads the host config and writes it through `cat > "$HOME/.gitconfig"` as the configured container user. It also forwards a broader set of Coder-related environment/mount inputs.

Relevant prior work to re-evaluate:

- #1458 — initial credential sharing
- #1506 — reuse lifecycle
- #1587 — non-root user support
- #2980 — uid/gid hardening

## Example: Coder workspace credentials

In some Coder-backed environments, the host Git config contains a credential helper that depends on Coder-specific executables, agent state, environment variables, or mounts. Copying that config into a generic Docker workspace does not make those dependencies available. The helper may fail, block a supported fallback, or point Git at paths that do not exist in the container.

This is not unique to Coder. Similar examples include OS keychain helpers, Git Credential Manager, libsecret, password-manager helpers, enterprise helpers, custom shell helpers, and helpers backed by local files or sockets.

## Why copying a global Git config is not a portable credential transport

Potentially non-portable configuration includes:

- `credential.helper`, including multiple helpers and URL-scoped helpers;
- shell or absolute-path helpers;
- helper state files, daemons, sockets, and browser/OAuth flows;
[...]

PR #3881, opened 2026-08-19
🤖 fix: share gitconfig with non-root Docker container users

ibetitsmike:
Part of #3819.

## Summary

`DockerRuntime` now writes the shared host `~/.gitconfig` into the container as the container's default user, targeting that user's `$HOME`, instead of `docker cp`-ing it to a hardcoded `/root/.gitconfig`. Container user info is also detected and cached before credential setup on the container reuse/fork path, matching the fresh-provisioning path.

## Background

Two concrete bugs from #3819:

- `setupCredentials()` copied the host gitconfig to `<container>:/root/.gitconfig`. Images whose default `USER` is non-root (e.g. `coder` in codercom images, `node` in node images) read `$HOME/.gitconfig` and cannot even traverse `/root` (mode 700), so credential sharing silently did nothing for them.
- The reuse/fork branch (`case "skip"` in `postCreateSetup()`) ran credential setup while `this.containerHome` was still undefined; only fresh provisioning detected the container user first, so `~` resolution fell back to `/root` on reused containers.

This PR fixes only these concrete bugs. The broader design question about portability of credential helpers referenced *inside* the copied gitconfig (host-specific helper paths, `gh auth setup-git` reachability, etc.) intentionally remains open in #3819.

## Implementation

- `setupCredentials()` mirrors `DevcontainerRuntime.setupCredentials()`: it streams `readHostGitconfig()` into `this.exec('cat > "$HOME/.gitconfig"')`. `docker exec` runs as the image's default user, so `$HOME` resolves in-container and the file gets that user's ownership; root-user containers still end up with `/root/.gitconfig`. Failures now throw `RuntimeError` (Devcontainer parity) instead of being silently ignored.
- The `gh auth setup-git` step already ran as the default container user (it was not affected by the `/root` bug); it is converted to the same `this.exec` form for parity and testability.
- New `ensureContainerUserInfo()` helper (extracted from `ensureReady()`) detects and caches container user info; the reuse/fork path now calls it before credential setup.
- `getHostGitconfigPath()` prefers `$HOME` over `os.homedir()` (matches git's own config resolution order and keeps the new tests deterministic across platforms); `hasHostGitconfig` is dropped in favor of the existing `readHostGitconfig` and the path helper is no longer exported (DockerRuntime was the only external consumer).
- Tests drive the real `postCreateSetup()` skip path through a subclass with canned exec streams (no Docker daemon): root and non-root home fixtures, detection-before-write ordering, streamed gitconfig bytes, `GH_TOKEN` gating, and `shareCredentials` off.

## Validation

Live smoke test on a real Docker daemon against a non-root image (`FROM node:20` / `USER node`), driving the real `postCreateSetup()`:

- Pre-fix behavior reproduced: `docker cp` to `/root/.gitconfig` is unreadable by the default user (`Permission denied`, `git config --get user.name` empty).
- With this branch, both the fresh-provisioning pass and a second reuse/skip pass (new runtime instance, `containerHome` undefined) produce `/home/node/.gitconfig` owned `node:node`, `git config --get user.name` returns the shared identity as the non-root user, `resolvePath("~")` returns `/home/node` on both paths, and `/root/.gitconfig` is absent.

## Risks

Low and scoped to Docker workspaces with `shareCredentials`. Behavior changes: gitconfig write failures now fail workspace provisioning loudly instead of silently continuing without credentials, and root containers get `$HOME/.gitconfig` resolved in-container (same `/root/.gitconfig` result). The reuse-path user detection adds three cheap `docker exec` calls that fresh provisioning already performed.

---

### DIFF
--- a/src/node/runtime/DockerRuntime.test.ts
+++ b/src/node/runtime/DockerRuntime.test.ts
@@ -1,6 +1,9 @@
-import { describe, expect, it } from "bun:test";
-import { DockerRuntime, getContainerName } from "./DockerRuntime";
-import type { InitLogger } from "./Runtime";
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import * as fs from "fs/promises";
+import * as os from "os";
+import * as path from "path";
+import { DockerRuntime, getContainerName, type DockerRuntimeConfig } from "./DockerRuntime";
+import type { ExecOptions, ExecStream, InitLogger, WorkspaceInitParams } from "./Runtime";
 
 const noopInitLogger: InitLogger = {
   logStep: () => {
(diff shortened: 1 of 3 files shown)


ibetitsmike:
@codex review

PR #3885, opened 2026-08-20
🤖 fix: bound the persistent sub-agent bench

ammar-agent:
## Summary

Bounds the long-lived roster of reusable standalone sub-agents while preserving the value of persistent context. Each parent should aim for a stable bench of at most seven distinct roles and keep it below ten unless a deliberate grouped parallel run temporarily needs more.

## Background

PR #3875 corrected overly eager deletion by making inactive children cheap to retain. In practice, an entirely retention-first policy can accumulate dozens of inactive children. This follow-up balances both concerns: retain a small useful organizational bench, but prune temporary or redundant work.

## Implementation

- define shared target and exclusive-limit constants for the reusable standalone bench
- prefer reawakening and retitling relevant bench members before creating new standalone roles
- require a genuinely distinct responsibility above the target, with consolidation before the limit
- treat grouped best-of candidates as temporary and remove them after results/artifacts are consumed when no same-candidate follow-up is expected
- expose persisted `bestOf` metadata in `task_list` so grouped candidates remain distinguishable after compaction or restart
- prune overlapping, obsolete, or least-useful inactive roles while prohibiting blanket end-of-turn cleanup
- align spawn, removal, and task-list guidance and regenerate system-prompt documentation

## Validation

- TaskService grouped-metadata projection tests
- task-list output-schema and forwarding tests
- targeted lifecycle, task-tool, and system-message tests
- generated docs and built-in skill synchronization checks
- `make static-check`

## Risks

The roster bound remains model-facing orchestration policy rather than a hard scheduler limit. Deliberate grouped runs can exceed it. The optional grouping field is additive and scoped to descendant rows; it contains internal grouping identifiers but no prompts, reports, paths, or model settings.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `xhigh` • Cost: `$18.24`_


ammar-agent:
@codex review

Please review the bounded standalone sub-agent bench policy, especially the distinction between long-lived roles and temporary grouped candidates.

ammar-agent:
Addressed the post-restart classification gap by exposing persisted `bestOf` grouping metadata in descendant `task_list` rows. Grouped candidates are now observable after compaction/restart, the retention note explains that they are temporary rather than bench members, and service/tool tests cover grouped versus standalone rows.

@codex review

ammario (approved):

PR #3888, opened 2026-08-20
🤖 fix: stop EPIPE error dialog loop when stdout/stderr pipe dies

### LINKED ISSUE #3082: An unexpected error occurred: write EPIPE
Hi Mux Team,

I faced with an annoying issue: 
The desktop AppImage shows an **Application Error** dialog with the message:

`An unexpected error occurred: write EPIPE`

This happens in the desktop app and triggers an error popup that repeats indefinitely.

## Error Message

`write EPIPE`

## Screenshot

Application Error screenshot

## Environment

```bash
$ uname -a
Linux hpms 6.17.0-19-generic #19~24.04.2-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar  6 23:08:46 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux

$ uname -r
6.17.0-19-generic

$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 24.04.4 LTS
Release:        24.04
Codename:       noble
```

Currently i have mux-0.22.1-nightly.3-x86_64.AppImage but it also appears on prev releases
I'm running it in --no-sandbox mode

it seems this happens when the screen goes blank in 10 mins of inactivity, but my system doesn't allow any suspend/sleep modes, only screen off.

ibetitsmike:
## Summary

Prevents the endless "Application Error: write EPIPE" modal loop in packaged Linux builds by swallowing EPIPE errors on `process.stdout`/`process.stderr` and excluding EPIPE from the crash-dialog paths.

Fixes #3082

## Background

When the terminal that launched the AppImage exits (or the pipe reader otherwise goes away), every console write emits an async `error` event with code `EPIPE` on the stdio stream. With no `error` listener attached, Node promotes it to an `uncaughtException`, whose handler shows Electron's "Application Error" modal and logs to the same dead pipe — so dismissing one dialog immediately triggers the next, forever, making the app unusable.

The existing `log.ts` EPIPE protection only catches synchronous write failures; the async stream `error` events were unguarded.

## Implementation

- `src/desktop/utils/epipeGuard.ts`: `isEpipeError` classifier plus `installEpipeGuard`, which attaches an `error` listener that swallows EPIPE and rethrows anything else (so genuinely unexpected stream failures still surface).
- `src/desktop/main.ts`: installs the guard on stdout/stderr at the very top, before anything logs; additionally filters EPIPE out of the `uncaughtException`/`unhandledRejection` dialog paths as defense-in-depth for EPIPEs surfacing from other fds.

## Validation

- Empirically confirmed on Linux that writes to a dead pipe emit `EPIPE` on every subsequent write (never `ERR_STREAM_DESTROYED`), so the EPIPE-only classifier fully covers the loop.
- Unit tests cover the classifier and the swallow-vs-rethrow branching of the stream guard.

## Risks

Low. EPIPE on stdio means the reader is gone and nothing is actionable; all other error codes keep the existing crash-dialog behavior.

---

### DIFF
--- a/src/desktop/main.ts
+++ b/src/desktop/main.ts
@@ -2,6 +2,14 @@
 // Must be set before any filesystem operations occur.
 process.umask(0o077);
 
+import { installEpipeGuard, isEpipeError } from "./utils/epipeGuard";
+
+// When the launching terminal exits (packaged Linux/AppImage), stdout/stderr
+// become dead pipes and every console write raises EPIPE. Swallow those before
+// anything logs so they can never crash the app or loop the error dialog (#3082).
+installEpipeGuard(process.stdout);
+installEpipeGuard(process.stderr);
+
 // Enable source map support for better error stack traces in production
 import "source-map-support/register";
 import * as fs from "node:fs";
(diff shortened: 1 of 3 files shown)


ibetitsmike:
@codex review

PR #3889, opened 2026-08-20
🤖 fix: don't re-fire browser notifications when replaying notify results

### LINKED ISSUE #2547: Mux keeps re-notifying about past events (browser notifications, Safari)
I've noticed weird behavior for Mux notifications for a while. And today I realized I can easily reproduce this simply by changing between chats. When I select one Mux will re-notify the last message from that chat and constantly keep doing so if I jump back and forth between two chats.

Just reloading Mux will also re-notify for the active chat.

ibetitsmike:
## Summary

In browser mode, Mux re-fires OS notifications for past `notify` tool calls every time the user switches workspaces or reloads the app. This gates the notification side effect so only live tool results notify; history hydration and reconnect replay stay silent.

Fixes #2547

## Background

In browser mode the `notify` tool always returns `notifiedVia: "browser"` (Electron is unavailable server-side), delegating delivery to the frontend. That routing metadata is persisted in chat history. `StreamingMessageAggregator.processToolResult` fired the Web Notification unconditionally, and it runs on two replay paths in addition to live tool-call-end events:

- `loadHistoricalMessages` re-processes every historical tool result to rebuild derived state. Since only the active workspace holds the `onChat` subscription, every workspace switch and reload triggers full history hydration, re-firing every historical browser notification. This is exactly the reported repro (re-notify on chat switch, re-notify on reload).
- Reconnect replay (`replayStream`) re-emits `tool-call-end` with `replay: true` for already-completed calls of an active stream, which flowed into the same handler with a fresh live timestamp.

The response-complete producer (App.tsx) is not affected: `stream-end` is never replayed for completed streams, and the background-activity completion path requires observing a live streaming true→false transition.

## Implementation

`processToolResult` now takes an explicit `liveContext` parameter (renamed from `messageContext`) that is present only for live, non-replay tool results:

- The `notify` → browser-notification side effect requires `liveContext`.
- `handleToolCallEnd` omits `liveContext` when the event carries `replay: true`, so reconnect replay behaves like history hydration. As a side benefit this also stops replayed `review_pane_update` results from re-arming the assisted-review "new" badge with a fresh timestamp, matching the documented intent of that badge.

## Validation

- Red-green on both guards independently: removing the `liveContext` gate fails the hydration and replay tests; reverting the `handleToolCallEnd` replay check fails exactly the replayed tool-call-end test.
- New behavioral tests: live notify results fire once; replayed `tool-call-end` and history hydration stay silent; a genuinely new notify after hydration still fires.

## Risks

Low severity, scoped to notification delivery. The behavioral change is that notify results reconstructed from history or reconnect replay no longer notify; a notify that completes while the workspace is live-subscribed still does. Notifications for workspaces without an active `onChat` subscription were already not delivered through this path (activity-snapshot-driven response-complete notifications are unchanged).

---

### DIFF
--- a/src/browser/utils/messages/StreamingMessageAggregator.test.ts
+++ b/src/browser/utils/messages/StreamingMessageAggregator.test.ts
@@ -141,6 +141,7 @@ function endToolCall(
     result: unknown;
     timestamp?: number;
     parentToolCallId?: string;
+    replay?: boolean;
   }
 ): void {
   aggregator.handleToolCallEnd({
@@ -152,6 +153,7 @@ function endToolCall(
     result: options.result,
     timestamp: options.timestamp ?? Date.now(),
     parentToolCallId: options.parentToolCallId,
+    replay: options.replay,
   });
 }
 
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #3892, opened 2026-08-20
🤖 fix: consult agent definition ai defaults when spawning sub-agent tasks

### LINKED ISSUE #3038: Agent definition ai.model is not consulted when spawning sub-agent tasks
# Agent definition `ai.model` is not consulted when spawning sub-agent tasks

## Summary

The `ai.model` field in agent `.md` definition files is documented and parsed but **not used** when the task service creates sub-agent workspaces. Sub-agents always inherit the parent session's model, regardless of what `ai.model` is set to in the agent definition.

This makes it impossible for custom agent definitions to pin themselves to a specific model — e.g., running review agents on `openai:gpt-5.4` while the parent session uses `anthropic:claude-opus-4-6`.

## Reproduction

1. Create a custom agent definition with a pinned model:
   ```yaml
   ---
   name: GPT Review
   base: exec
   subagent:
     runnable: true
   ai:
     model: openai:gpt-5.4
     thinkingLevel: high
   ---
   A review agent pinned to GPT-5.4.
   ```

2. In a session running on `anthropic:claude-opus-4-6`, spawn the agent:
   ```
   task({ agentId: "gpt-review", prompt: "Review this file...", title: "test" })
   ```

3. **Expected**: Sub-agent runs on `openai:gpt-5.4` as specified in its definition.
4. **Actual**: Sub-agent runs on `anthropic:claude-opus-4-6` (inherited from parent).

## Root cause

There are two independent model resolution paths, and the one that matters for sub-agents doesn't read the agent definition:

### Task Service path (`taskService.ts` ~line 1053-1068)

This runs when `task` tool creates the workspace. Resolution order:

1. `cfg.agentAiDefaults[agentId].modelString` — from **app config** (Settings UI)
2. `args.modelString` — parent's `MUX_MODEL_STRING` env var
3. `parentAiSettings.model` — workspace-level settings
4. `defaultModel` — hardcoded fallback

**The agent descriptor's `aiDefaults.model` (from the `.md` file) is not in this chain.**

### ACP Session Loader path (`resolveAgentAiSettings.ts` ~line 77-119)

This runs when the client connects to a workspace. It **does** consult `descriptor.aiDefaults` as a fallback — but by then, `aiSettings` have already been persisted by the task service with the parent's model, so the `.md` value is shadowed.

### `thinkingLevel` has the same gap

The task service's thinking level resolution also skips the agent descriptor's `ai.thinkingLevel`. The same fix should cover both fields.

## Suggested fix

In the task service model resolution, insert the agent descriptor's `aiDefaults` between the app config check and the parent model fallback. The descriptor is already parsed and available — it just needs to be wired in.

Proposed resolution order:
1. `args.modelString` — explicit call-site request from parent
2. `cfg.agentAiDefaults[agentId]` — app config / Settings UI
3. **`agentDescriptor.aiDefaults`** — from agent `.md` definition ← **add this**
4. `parentAiSettings.model` — inherited ambient model
5. `defaultModel`

This preserves the app config as the highest-priority override (Settings UI wins), makes agent definitions work as documented, and falls back to parent model inheritance when no model is specified.

## Workaround

Per-agent model overrides can be configured via the `agentAiDefaults` key in `~/.mux/config.json`. This path is consulted by the task service. However, this is a global setting that applies to all sessions, and it's not portable — agent definitions should be the authoritative source for agent-specific defaults.

## Environment

- Mux version: v0.21.0
- Agent definitions with `ai.model` set in frontmatter
- Sub-agents spawned via `task` tool

ibetitsmike:
## Summary

Sub-agent tasks now honor `ai.model` / `ai.thinkingLevel` declared in an agent definition's frontmatter. Previously the task service's AI-settings resolution consulted app config and parent settings but never the agent definition, so definition-pinned models were silently ignored and sub-agents always inherited the parent's model.

Fixes #3038

## Precedence

`resolveTaskAISettings` resolves each field (model, thinkingLevel) independently in this order:

1. Explicit per-spawn override (task tool `model` / `thinking` params)
2. `subagentAiDefaults[agentId]` (Settings UI, sub-agent defaults)
3. `agentAiDefaults[agentId]` (Settings UI, agent defaults)
4. **Agent definition frontmatter `ai`** (new)
5. Parent runtime AI settings (`MUX_MODEL_STRING` / `MUX_THINKING_LEVEL`)
6. Parent workspace persisted settings
7. App default

Rationale: an explicit call-site choice must always win; user settings stay above definition defaults per the documented contract (docs/agents: "Per-agent AI defaults (overridable by user settings)"); and definition defaults beat silent inheritance of the parent's model so pinned agents work as documented.

## Implementation

- Both task-creation paths (`create` and `createMany`) already resolve the agent's frontmatter (base chain applied) to validate runnability; the resolution of AI settings now happens after that step and receives `frontmatter.ai`.
- `ai.model` may be an abbreviation per docs (e.g. `sonnet`); it is resolved with the same alias logic as the explicit task-tool override (`normalizeModelInput`). An invalid value is logged and falls through to parent inheritance instead of failing the spawn, so a typo in a `.md` cannot brick the agent.
- The plan→exec same-workspace continuation is intentionally unchanged: it cannot distinguish an explicit spawn-time model from ambient inheritance, so consulting exec's definition defaults there could override an explicit spawn choice.
- Queued task relaunches reuse the settings persisted at creation time, so they pick up the fix automatically.
- Built-in impact: only `desktop.md` declares `ai` defaults (`thinkingLevel: medium`), which now actually applies to desktop sub-agents as advertised.

## Validation

- New behavioral tests: definition defaults win over parent inheritance (including parent runtime settings); explicit caller override still wins; config agent defaults still win; abbreviations resolve; model-only `ai` blocks inherit thinkingLevel field-wise. Red-checked: the two new-behavior tests fail on the base commit, the two precedence-guard tests hold on both.
- `make static-check-full` green locally. The two `terminal recovery` test failures in `taskService.test.ts` reproduce on the base commit and are unrelated.

## Risks

Low. Behavior changes only for sub-agent spawns where the agent definition declares `ai` defaults and no higher-precedence source exists, which is exactly the reported defect. Agents without `ai` defaults keep the existing inheritance chain unchanged.

---

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -287,6 +287,32 @@ async function saveLocalParentWorkspace(
   return { parentId, projectPath };
 }
 
+/** Write a runnable exec-derived custom agent definition into the project's .mux/agents. */
+async function writeCustomAgentDefinition(
+  projectPath: string,
+  extraFrontmatterLines: string[] = []
+): Promise<void> {
+  const agentsDir = path.join(projectPath, ".mux", "agents");
+  await fsPromises.mkdir(agentsDir, { recursive: true });
+  await fsPromises.writeFile(
+    path.join(agentsDir, "custom.md"),
+    [
+      "---",
+      "name: Custom",
+      "description: Exec-derived custom agent for tests",
+      "base: exec",
+      "subagent:",
+      "  runnable: true",
+      ...extraFrontmatterLines,
+      "---",
+      "",
+      "Test agent body.",
+      "",
+    ].join("\n"),
+    "utf-8"
+  );
+}
+
 function stubStableIds(config: Config, ids: string[], fallbackId = "fffffffff0"): void {
   let nextIdIndex = 0;
   const configWithStableId = config as unknown as { generateStableId: () => string };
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #3893, opened 2026-08-20
🤖 fix: strip empty-string args for optional MCP tool params

### LINKED ISSUE #2887: bug: MCP tool calls fail when LLM emits empty strings for optional parameters
## Problem
When Mux invokes MCP tools (e.g. `gitlab_mcp_list_issues`), the LLM may supply empty strings or placeholder values for optional parameters:
```json
{
  "project_id": "42332",
  "assignee_id": "",
  "author_id": "",
  "author_username": "",
  "iteration_id": "",
  "search": ""
}
```
Mux forwards these arguments verbatim to MCP servers via `wrapMCPTools` in `src/node/services/mcpServerManager.ts`. Downstream APIs that require optional parameters to be **omitted** (not empty) reject the request. For example, GitLab returns:
```
400 Bad Request
author_id, author_username are mutually exclusive
assignee_id should be an integer, none, or any, however got ""
iteration_id should be an integer, none, any, or current, however got ""
```
From a user perspective, this surfaces as a failed tool call with no obvious workaround.
## Root Cause
1. LLMs commonly emit empty strings (`""`) for optional fields they don't have values for, rather than omitting them.
2. `wrapMCPTools` wraps tool execution for timeout/abort handling and activity tracking, but passes `args` through without any sanitization.
3. Many REST-backed MCP servers (GitLab, and likely others) treat empty strings as present-but-invalid, causing 400 errors.
## Proposed Fix
Introduce light argument normalization in `wrapMCPTools` before invoking the original `execute` function. Specifically:
- **Strip keys with empty string values (`""`)** from the arguments object before forwarding. Empty strings for optional parameters are almost always LLM artifacts, not intentional values.
- **Do NOT strip `null` values** — some MCP servers may use explicit `null` to mean "clear/unset this field."
- **Do NOT strip empty arrays (`[]`)** — these are generally treated as "no filter" by APIs and are benign.
### Example normalization (conceptual)
```typescript
function sanitizeArgs(args: Record<string, unknown>): Record<string, unknown> {
  const sanitized: Record<string, unknown> = {};
  for (const [key, value] of Object.entries(args)) {
    if (value === "") continue; // Skip empty strings
    sanitized[key] = value;
  }
  return sanitized;
}
```
This could be applied inside `wrapMCPTools` before calling `originalExecute(args, context)`.
### Considerations
- **Different MCP servers may interpret empty vs. missing differently.** The normalization should be conservative (empty strings only).
- **Opt-out escape hatch (future):** If any MCP server legitimately needs empty strings passed through, a per-server config flag like `sanitizeArgs: false` could be added later.
- **Nested objects:** For v1, top-level empty string stripping is sufficient. Deep/recursive sanitization can be considered if needed.
## Affected Component
`src/node/services/mcpServerManager.ts` — `wrapMCPTools` function (line ~131)
## Impact
This affects any MCP server backed by a strict REST API that distinguishes between "parameter omitted" and "parameter present but empty." GitLab MCP is one confirmed case, but the issue is generic.

ibetitsmike:
## Summary

MCP tool calls now drop top-level empty-string arguments that the model emitted for optional parameters before the call is forwarded to the server. Fixes #2887.

## Background

LLMs often fill optional tool parameters with `""` instead of omitting them. `wrapMCPTools` forwarded arguments verbatim, so strict REST-backed MCP servers (GitLab MCP is the confirmed case) rejected the call with `400 Bad Request` (e.g. `assignee_id should be an integer, none, or any, however got ""`), leaving the user with a failed tool call and no workaround.

## Implementation

Sanitization happens in `wrapMCPTools` (argument path only), just before `originalExecute` is invoked. Design decisions:

- **Schema-aware, not unconditional:** every MCP tool mux builds carries its server-declared JSON schema via the AI SDK's `jsonSchema()` wrapper (`tool.inputSchema.jsonSchema`), so the top-level `required` list is available at zero extra cost. `""` is stripped only for parameters not listed in `required`; a genuinely required empty string is never silently dropped (the server stays the arbiter of its validity). If the schema is missing or unreadable (not the case for tools built by `mcpClient`), all top-level `""` values are stripped, matching the issue's intent since the reported failure mode is far more common than a legitimately empty required value.
- **Top-level only:** matches the reported failure; no evidence found that nested empty strings cause this class of error. Recursive stripping can be added later if needed.
- **`null` and `[]` pass through untouched:** some servers assign them meaning ("clear this field" / "no filter"). Nothing else on the MCP argument path normalizes null vs. undefined today (the repo's `.nullish()` convention applies to mux's own tool schemas), so this change stays consistent by not introducing any null handling.

Kept deliberately clear of the result path (`transformMCPResult`) that #3891 touches; the only shared-file overlap is the `wrapMCPTools` body, which #3891 does not modify.

## Validation

Red-green verified: with the sanitization call bypassed, the three stripping tests fail and the two pass-through tests still pass.

## Risks

Low severity, scoped to MCP tool invocation. The behavior change is invisible unless a model emits `""` for an optional parameter; the main theoretical regression is a server that intentionally distinguishes optional-`""` from omitted, which the issue triage considered an acceptable tradeoff (a per-server opt-out can be added later if one surfaces).

---

### DIFF
--- a/src/node/services/mcpServerManager.test.ts
+++ b/src/node/services/mcpServerManager.test.ts
@@ -23,7 +23,7 @@ import type { Runtime } from "@/node/runtime/Runtime";
 import { DevcontainerRuntime } from "@/node/runtime/DevcontainerRuntime";
 import { RemoteRuntime } from "@/node/runtime/RemoteRuntime";
 import { DisposableTempDir } from "@/node/services/tempDir";
-import type { Tool } from "ai";
+import { jsonSchema, type Tool } from "ai";
 
 interface MCPServerManagerTestAccess {
   workspaceServers: Map<string, unknown>;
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #3894, opened 2026-08-20
🤖 fix: pass ProxyCommand to cmd.exe verbatim on Windows (ssh2 premature close)

### LINKED ISSUE #3110: Coder Runtime: ssh2-proxy-socket Premature close on Windows with self-hosted Coder deployment
# Coder Runtime: `ssh2-proxy-socket` Premature close on Windows with self-hosted Coder deployment

## Description

The Coder Runtime fails to establish an SSH connection to a self-hosted Coder workspace on Windows. Every connection attempt results in `ERR_STREAM_PREMATURE_CLOSE` on the `ssh2-proxy-socket`, immediately going into backoff and eventually timing out.

Normal SSH via `coder ssh <workspace>` works perfectly from the same machine. The SSH Runtime workaround (using `coder config-ssh` host aliases with system OpenSSH) also works fine. The issue is isolated to Mux's internal Node.js `ssh2` library when used with the Coder Runtime's ProxyCommand on Windows.

## Environment

- **Mux version:** (latest stable as of 2026-04-01)
- **OS:** Windows 10/11
- **Coder server:** v2.31.6 (self-hosted on NKP/Kubernetes)
- **Coder CLI (system):** v2.31.6 (Slim build)
- **Workspace template:** Kubernetes-based

## Steps to Reproduce

1. Install Mux on Windows
2. Configure Coder Runtime pointed at a self-hosted Coder deployment
3. Create a workspace using the Coder Runtime
4. Attempt to start an agent session

## Expected Behavior

Mux connects to the Coder workspace via SSH and the agent session starts.

## Actual Behavior

Connection fails immediately with:

```
Stream Error: runtime_start_failed
Failed to reach SSH host: SSH2 connection failed: SSH connection to <workspace>.mux--coder
is in backoff and maxWaitMs exceeded. Last error: Premature close
```

Logs show repeated `ssh2-proxy-socket` errors:

```
Stream error {"label":"ssh2-proxy-socket","code":"ERR_STREAM_PREMATURE_CLOSE","message":"Premature close"}
Error: Premature close
    at Socket.onclose (node:internal/streams/end-of-stream:171:30)
```

## Diagnosis

The ProxyCommand in `~/.ssh/config` (written by Mux) is correct and works via system OpenSSH:

```
# --- START MUX CODER SSH ---
Host *.mux--coder
    ProxyCommand "C:\Users\mathomas\coder.exe" "ssh" "--stdio" "--hostname-suffix" "mux--coder" "%h"
# --- END MUX CODER SSH ---
```

Verified the ProxyCommand works directly:

```powershell
PS> & "C:\Users\mathomas\coder.exe" ssh --stdio --hostname-suffix mux--coder max-test.mux--coder
SSH-2.0-Go    # ← banner exchange succeeds
```

And `coder ssh` works perfectly:

```powershell
PS> coder ssh max-test -- echo "hello"
hello
```

The issue is that Mux's Node.js `ssh2` library receives the proxy stream but the socket closes before the SSH handshake completes. No key exchange, algorithm negotiation, or auth errors appear in logs — just immediate `Premature close` on every attempt.

### Additional context: version mismatch in earlier logs

Earlier log entries also revealed that Mux's Coder Runtime uses an internal Coder CLI (v2.30.5 in the `CoderSSHRuntime.ts` code path) that is separate from the system Coder CLI. This caused version mismatch errors against the self-hosted server even after the system CLI was updated. It's unclear whether the Premature close issue is caused by or related to the internal CLI version.

## Workaround

Switch from **Coder Runtime** to **SSH Runtime** using the host alias from `coder config-ssh`:
- Runtime: SSH
- SSH Host: `coder.<workspace-name>`

This bypasses Mux's internal `ssh2` library and uses the system OpenSSH client, which handles the ProxyCommand correctly.

ibetitsmike:
## Summary

Fixes the Coder Runtime on Windows: the ssh2 ProxyCommand child process was spawned through `cmd.exe /d /s /c` **without** `windowsVerbatimArguments`, so Node's default MSVCRT-style argument escaping rewrote the embedded quotes and cmd.exe received a mangled command line. The proxy process died instantly and every connection attempt failed with a bare `ERR_STREAM_PREMATURE_CLOSE` on `ssh2-proxy-socket`.

Fixes #3110

## Background

Mux's SSH config writer double-quotes every ProxyCommand argument (e.g. `"C:\Users\me\coder.exe" "ssh" "--stdio" ...`). System OpenSSH parses that fine, but `SSH2ConnectionPool.spawnProxyCommand` passed the same string as an argv element to `spawn(cmd.exe, ["/d", "/s", "/c", command])`. Without `windowsVerbatimArguments`, Node escapes each embedded `"` as `\"` on win32 — cmd.exe does not understand backslash-escaped quotes, so the coder CLI never launched, matching the reporter's symptoms (instant premature close, no handshake, while `coder ssh` and system OpenSSH both work).

## Implementation

- `getProxyShellArgs` now mirrors Node's own `shell: true` handling on win32: `cmd.exe /d /s /c "<command>"` with `windowsVerbatimArguments: true`, so cmd.exe sees the original ProxyCommand byte-for-byte (`/s` strips the added outer quotes). POSIX (`/bin/sh -c`) is unchanged.
- Error surfacing: the proxy's stderr is now retained as a bounded tail (2 KiB), and connection errors append the proxy's exit status + stderr when the proxy has died. A proxy that exits before the handshake now fails fast with that description instead of hanging until `readyTimeout` or surfacing a bare "Premature close".
- The `ssh2` dynamic import moved above the proxy spawn so spawn → handler attachment runs synchronously, eliminating a window where an instantly-dying proxy could emit `error`/`close` with no listeners attached.
- Added `SSH2ConnectionPool.test.ts` to the Windows CI smoke list; the spawn tests execute a real double-quoted-argv0 ProxyCommand on both platforms.

## Validation

- `bun test src/node/runtime/SSH2ConnectionPool.test.ts` (plus sshConfigParser/muxSshConfigWriter suites) green.
- Cannot run Windows locally; the mechanism was verified from Node.js win32 spawn semantics and the test asserts the verbatim payload contract. The new tests also run on the `Test / Windows` CI job.

## Risks

Low and scoped to the ssh2 ProxyCommand path (Coder Runtime / SSH configs with ProxyCommand). Main behavioral changes: Windows spawn semantics (previously broken for any quoted command, which mux always writes), and connect-time failure now fires on proxy death instead of waiting for the ready timeout.

---

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -397,7 +397,7 @@ jobs:
       - name: Run unit tests (Windows smoke)
         shell: pwsh
         run: |
-          bun test src/node/utils/paths.test.ts src/node/utils/main/bashPath.test.ts src/node/utils/main/resolveLocalPtyShell.test.ts
+          bun test src/node/utils/paths.test.ts src/node/utils/main/bashPath.test.ts src/node/utils/main/resolveLocalPtyShell.test.ts src/node/runtime/SSH2ConnectionPool.test.ts
       - name: Run Windows integration smoke tests
         shell: pwsh
         run: |
--- a/src/node/runtime/SSH2ConnectionPool.test.ts
+++ b/src/node/runtime/SSH2ConnectionPool.test.ts
@@ -1,0 +1,75 @@
+import { describe, expect, it } from "bun:test";
+import { getProxyShellArgs, spawnProxyCommand } from "./SSH2ConnectionPool";
+
+const PROXY_TOKENS = { host: "example.test", port: 2222, user: "alice" };
+
+describe("getProxyShellArgs", () => {
+  // Shape mux's SSH config writer emits: every argument double-quoted.
+  const quotedCommand =
+    '"C:\\Program Files\\coder\\coder.exe" "ssh" "--stdio" "--hostname-suffix" "mux--coder" "my-ws.mux--coder"';
+
+  it("passes the command to cmd.exe verbatim so embedded quotes survive (win32)", () => {
+    const spec = getProxyShellArgs(quotedCommand, "win32");
+
+    // Node's default escaping rewrites embedded quotes as \" which cmd.exe
+    // cannot parse; the proxy process then dies instantly (#3110).
+    expect(spec.windowsVerbatimArguments).toBe(true);
+    expect(spec.args.slice(0, 3)).toEqual(["/d", "/s", "/c"]);
+
+    // cmd.exe /s strips the first and last quote of the /c payload; what
+    // remains must be the original ProxyCommand byte-for-byte.
+    const payload = spec.args[3];
+    expect(payload.startsWith('"')).toBe(true);
+    expect(payload.endsWith('"')).toBe(true);
+    expect(payload.slice(1, -1)).toBe(quotedCommand);
+  });
+
+  it("runs the command through /bin/sh -c unchanged on POSIX", () => {
+    const spec = getProxyShellArgs(quotedCommand, "linux");
+
+    expect(spec.command).toBe("/bin/sh");
+    expect(spec.args).toEqual(["-c", quotedCommand]);
+    expect(spec.windowsVerbatimArguments).toBe(false);
+  });
+});
+
+describe("spawnProxyCommand", () => {
+  it("executes a ProxyCommand with a double-quoted binary path and substitutes tokens", async () => {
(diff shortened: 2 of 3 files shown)


ibetitsmike:
@codex review

Issue #3186, opened 2026-04-23
Extremely slow (3 t/s) and can't setup llama.cpp + a couple of errors


sudo-saif:
Hey there, I have just tried Mux, and I face two problems that make the app non-usable at the moment.

1. Getting extremely low speeds with Qwen 3.6 35b a3b:
- Using OpenAI compatible URL for LM Studio, I get 3 t/s speed (note: my average LM Studio speed is 72 t/s)

2. Using Qwen 3.5 9b, it works fast but I get a couple of "Invalid type for 'input'." errors during chats:
- Speed: LM Studio ~ 86 t/s while Mux is ~80 t/s

<img width="991" height="737" alt="Image" src="https://github.com/user-attachments/assets/52b735dd-9dfc-4645-94a2-1aa86320b43b" />

3. When I set llama.cpp up via OpenAI compatible URL, then try to chat, it keeps showing "Cannot determine type of 'item'
" error:

<img width="920" height="211" alt="Image" src="https://github.com/user-attachments/assets/8faefa11-b445-4bac-903f-3dd47c1934fd" />

PR #3899, opened 2026-08-20
🤖 fix: make llama-like OpenAI-compatible endpoints work out of the box

### LINKED ISSUE #2438: Custom OpenAI compatible endpoints are broken
I have tried adding custom OpenAI compatible endpoints from both llama.cpp and vLLM, both are throwing errors in Mux, but they do work correctly with other tools. It would be great to improve support for them for handling of non-cloud based environments.

Also it would be nice to be able to add more than one OpenAI compatible endpoint.

ibetitsmike:
## Summary

Makes llama.cpp / vLLM / LM Studio OpenAI-compatible endpoints work smoothly (#2438): custom `openai-compatible` provider base URLs without a path get `/v1` appended automatically, users who point the built-in `openai` provider at a llama-like server get an actionable hint instead of only the server's opaque error, and the custom-provider streaming path gains CI regression coverage that needs no external server.

## Background

Fixes #2438. Users adding llama.cpp/vLLM endpoints hit two failure modes:

1. Overriding `baseUrl` on the built-in `openai` provider sends Responses API requests (`/v1/responses`), which llama-like servers reject with opaque errors such as `Cannot determine type of 'item'` (#3186). The existing `wireFormat` setting that fixes this was undocumented and undiscoverable at the point of failure.
2. A custom provider `baseUrl` without a path (e.g. `http://localhost:8080`) silently missed the `/v1` prefix most servers require (the `anthropic` provider already normalizes this).

Multiple simultaneous custom OpenAI-compatible providers (the other half of #2438) already work via #3202; this PR pins that behavior with tests.

## Implementation

- `normalizeOpenAICompatibleBaseURL` in `providerModelFactory.ts`: append `/v1` only when the URL has an empty path, mirroring the Anthropic normalization; explicit proxy paths stay untouched.
- `getOpenAIResponsesBaseUrlHint` (new pure helper): appends a hint to stream errors only for the built-in `openai` provider with a non-`api.openai.com` base URL in `responses` mode failing with HTTP 400/404/405. No wire-format auto-fallback: silently downgrading would hide feature loss for Azure/gateway users.
- Settings: helper text under the OpenAI Base URL field (shown only when a custom base URL is set or being edited) pointing at the Wire format setting and custom providers.
- Docs: `wireFormat` documented, custom-provider section cross-linked.
- Tests: unit coverage for both helpers plus a new Jest IPC suite (`tests/ipc/providers/openaiCompatible.test.ts`) with a mock SSE server covering streaming, error guidance, chatCompletions mode, a real bash tool round-trip, and two simultaneous custom providers.

## Validation

- Live end-to-end UAT against a real llama.cpp server (Qwen3-0.6B) through the app UI: all five scenarios passed, including no-`/v1` base URL, two simultaneous providers, multi-step tool calling, the error hint, and chatCompletions success.
- The new IPC suite reproduced the #3186 failure mode against a real llama.cpp server before the fix and passes hermetically after it.

## Risks

Low. The `/v1` normalization only fires for custom `openai-compatible` providers whose URL has no path, a configuration that previously always failed against standard servers. The error hint is append-only on an existing error path. The built-in OpenAI default behavior is unchanged.

---


ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
Closing and recreating this PR: all 6 `@codex review` triggers here were acknowledged but never produced a review (account-level silent drop), so a fresh PR is being opened for the identical head `03e9b66ce` to get a working review context.

> Xum acted on Mike's behalf.

PR #3912, opened 2026-08-21
🤖 fix: preserve auto-compaction follow-ups

coadler:
## Summary

Fix auto-compaction when a synthetic snapshot follows the compaction request. The completion path now preserves the compacted boundary and sends the pending continuation.

## Background

Auto-compaction can persist a synthetic file-change snapshot after its request. The completion handler previously treated that snapshot as the latest user message. It then skipped compaction completion and did not send `Continue`.

## Implementation

- Track the active compaction request across synthetic snapshot rows.
- Pass the exact request ID to the compaction completion handler.
- Resolve pending follow-up data from the correlated request.
- Add regression tests for boundary creation and follow-up preservation.

## Validation

- `make static-check`
- `bun test src/node/services/agentSession.autoCompaction.test.ts src/node/services/compactionHandler.test.ts`

## Risks

The change affects compaction request detection and completion. The fallback behavior remains unchanged for callers without request correlation.

---

_Generated with `mux` • Model: `openai:gpt-5.6-sol` • Thinking: `xhigh` • Cost: `$0.01`_

### DIFF
--- a/src/node/services/agentSession.autoCompaction.test.ts
+++ b/src/node/services/agentSession.autoCompaction.test.ts
@@ -131,6 +131,51 @@ describe("AgentSession on-send auto-compaction snapshot deferral", () => {
     session.dispose();
   });
 
+  test("tracks a compaction request when a synthetic snapshot follows it", async () => {
+    const model = "openai:gpt-4o";
+    const { session } = await createSessionHarness({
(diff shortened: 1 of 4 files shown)


coadler:
@codex review

PR #3913, opened 2026-08-21
🤖 ci: skip existing xvfb installation

coadler:
## Summary

Skip the slow APT update when the Linux runner already has Xvfb.

## Background

The Linux E2E job spent 7 minutes 42 seconds in `apt-get update`.
The runner already had the required Xvfb package installed.

## Implementation

Check for both `Xvfb` and `xvfb-run` before installing the package.
Run APT only when either command is missing.
Apply the same guard to the performance profile workflow.

## Validation

- `make static-check`
- `git diff --check`

## Risks

If a runner lacks either command, the workflow uses the existing APT install path.

---

_Generated with `mux` • Model: `openai:gpt-5.6-luna` • Thinking: `medium` • Cost: `$0.07`_

### DIFF
--- a/.github/workflows/perf-profiles.yml
+++ b/.github/workflows/perf-profiles.yml
@@ -25,6 +25,11 @@ jobs:
       - uses: ./.github/actions/setup-mux
       - name: Install xvfb
         run: |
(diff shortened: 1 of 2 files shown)


coadler:
@codex review

Please review the guarded Xvfb installation in both Linux workflows.

PR #3917, opened 2026-08-21
🤖 fix: make TitleBar tooltip CTAs clickable links

### LINKED ISSUE #2330: Link update/install tooltips
UX papercut -- can we link these CTAs in the install and Mux gateway tool tips?

https://github.com/user-attachments/assets/f161eee4-5367-44ac-90de-0d2dff1a5188

ibetitsmike:
## Summary

Makes the call-to-action text in the TitleBar tooltips actually clickable: the version tooltip's "Click for more details" now opens the About dialog, and the gateway/provider route tooltip's "Click to open gateway settings" / "Click to open provider settings" now opens Settings → Providers with the active provider expanded.

Fixes #2330

## Background

Both tooltips ended in CTA copy that read like a link but was inert plain text, so clicking the words did nothing (UX papercut). The actions were only reachable by clicking the trigger buttons themselves.

## Implementation

- The two CTAs are now link-styled `<button>` elements inside the shared `TooltipContent`, invoking the exact same handlers as their triggers (`openAboutDialog`, `openSettings("providers", { expandProvider: activeRouteProvider })`). Shared link styling lives in a local `TOOLTIP_CTA_CLASSNAME` constant.
- Clicking inside tooltip content works because the app-level `TooltipProvider` keeps Radix hoverable-content enabled (pointer can travel from trigger into the tooltip).
- Both affected tooltips are now controlled (`open`/`onOpenChange`) so the CTA click closes the tooltip before the About dialog / Settings surface opens; otherwise the z-indexed tooltip lingers on top.

## Validation

- Remote dogfood UAT (browser-mode dev server at this exact head): PASS. Verified pointer travel into both tooltips, CTA clicks open the About dialog / Settings → Providers with the right provider expanded (OpenRouter via real UI flow; Xum Gateway branch incl. Balance/Concurrent rows in logged-out state), no lingering tooltips after any click (DOM-checked), trigger-click regressions unaffected, other TitleBar tooltips unchanged, no overflow at 375px.
- Known minor UAT observations (accepted, not fixed): tooltip CTA is pointer-only (Radix tooltips close on Tab; the focusable trigger performs the identical action), and the hover color shift is gated by `@media (hover: hover)` like all other hover styles.

---

### DIFF
--- a/src/browser/components/TitleBar/TitleBar.tsx
+++ b/src/browser/components/TitleBar/TitleBar.tsx
@@ -47,6 +47,10 @@ import { WORKSPACE_DEFAULTS } from "@/constants/workspaceDefaults";
 // Update check interval
 const UPDATE_CHECK_INTERVAL_MS = 4 * 60 * 60 * 1000; // 4 hours
 
+// Tooltip CTAs mirror their triggers so pointer users can act on the copy directly.
+const TOOLTIP_CTA_CLASSNAME =
+  "text-muted hover:text-foreground cursor-pointer border-0 bg-transparent p-0 text-left underline underline-offset-2 transition-colors";
+
 interface VersionRecord {
   git?: unknown;
   git_describe?: unknown;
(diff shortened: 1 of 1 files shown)


ibetitsmike:
@codex review

PR #3922, opened 2026-08-22
🤖 fix: support "/" in git branch names when creating workspaces

### LINKED ISSUE #407: Support "/" character in branch names
Having `/` in Git branch name is legal, but cmux does not accept it. I suspect this is because it might be interpreted as a system path.

Image

cmux now (Workspace name can only contain lowercase letters, digits, underscore, and hyphen):

Image

ibetitsmike:
## Summary

Workspace creation now accepts Git branch names containing `/` (e.g. `feature/foo`). The Git branch keeps its slashes while the workspace/worktree folder name is derived by replacing `/` with `-`; if a slash branch sanitizes to a workspace name that already exists, creation fails with an explicit conflict error. Validation errors for the name input now render in a tooltip attached to the input instead of cramped inline text.

Fixes #407

## Background

`/` is legal in Git branch names, but the creation flow used one identifier for the branch, worktree folder, workspace name, and session directory, so validation rejected slashes outright. Per the maintainer guidance in #407: replace `/` with `-` for path purposes and fail on conflicts, plus clean up the cramped validation-error display with a tooltip.

## Implementation

- `workspaceValidation.ts`: new `validateWorkspaceBranchName` (slash-separated segments of the existing `[a-z0-9_-]` charset; rejects leading/trailing/consecutive slashes), `sanitizeBranchNameForWorkspace` (`/` → `-`; valid branch names always sanitize to valid workspace names), and a conflict helper that reports when a slash branch maps onto an existing workspace name.
- `WorkspaceService.create` / `createMultiProjectWorkspace`: track the branch name and sanitized workspace name separately; pass `branchName` (real branch) and `directoryName` (sanitized) to `runtime.createWorkspace`, which already supported the split. Persisted metadata/config use the sanitized name, and runtime hooks (`finalizeConfig`, `validateBeforePersist`) receive it since Coder-derived names cannot contain `/`.
- Conflict policy: slash branches fail fast with an explicit error naming both the branch and the conflicting workspace name; slash-free names keep the existing random-suffix retry behavior.
- UI: `useWorkspaceName` validates with the branch-name rules; the name input's tooltip shows validation/transport errors (forced open, error styling, `aria-invalid`) while generation errors keep their richer box below the input. The error border stays red while the input is focused.

## Validation

- Remote dogfood UAT (2 rounds, PASS): created `feature/uat-407` → worktree folder `feature-uat-407` with `git branch --show-current` = `feature/uat-407` and no nested directory; duplicate slash branch rejected with the explicit conflict error and no suffixed workspace; invalid names (`Feature Foo`, `/foo`, `foo//bar`, `foo/`) show the error tooltip incl. at ~375px width; slash-free creation, duplicate suffixing, and wand auto-naming regressions pass.
- Behavioral tests: branch-name validation/sanitizer invariants, conflict helper, service-level conflict rejection, and a WorktreeManager slash-branch worktree test (sanitized path + persisted directory→branch mapping).

## Risks

Creation-path change shared by all runtimes. For slash-free names (the overwhelmingly common case) the flow is unchanged: same validation charset, same collision suffixing. Slash branches are new input that was previously rejected, so regression risk concentrates in the new sanitized-name path; rename/fork/sub-agent naming intentionally keep the stricter slash-free validation. Worktree removal/rename for slash branches rides the existing persisted directory→branch mapping, covered by tests.

---

### DIFF
--- a/src/browser/features/ChatInput/CreationControls.stories.tsx
+++ b/src/browser/features/ChatInput/CreationControls.stories.tsx
@@ -420,7 +420,6 @@ export const NameGenerationAuthError: Story = {
     ),
 };
 
-/** Manual naming validation error for characters outside the workspace-name policy. */
 export const NameValidationError: Story = {
   render: () =>
     renderCreationControls(
@@ -430,7 +429,7 @@ export const NameValidationError: Story = {
         error: {
           kind: "validation",
           message:
-            "Workspace names can only contain lowercase letters, numbers, hyphens, and underscores",
+            'Branch names can only contain lowercase letters, numbers, hyphens, underscores, and "/" separators',
         },
       })
     ),
--- a/src/browser/features/ChatInput/CreationControls.tsx
+++ b/src/browser/features/ChatInput/CreationControls.tsx
@@ -776,7 +776,9 @@ function CreationControlsContent(props: CreationControlsProps) {
                      field-sizing-content focus:border focus:bg-bg-dark focus:outline-none 
                      disabled:opacity-50 max-w-[50vw] sm:max-w-[40vw] lg:max-w-[30vw]`,
                     nameState.autoGenerate ? "text-muted" : "text-foreground",
-                    nameState.error && "border-red-500"
+                    // focus:border-red-500 keeps the error border visible while typing;
+                    // focus:border-accent would otherwise override it.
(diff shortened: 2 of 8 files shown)


ibetitsmike:
@codex review

PR #3419, opened 2026-05-29
🤖 refactor: remove experimental image tools

ThomasK33:
## Summary

Remove the failed experimental Image Tools feature from Mux. This deletes the `image_generate` / `image_edit` tools, the Image Tools experiment/config UI, generated-image transcript presentation, image-tool backend runtime wiring, the built-in `imagegen` skill, and the related ADR/docs entries while preserving generic image attachment, screenshot, lightbox, and token metadata support.

## Background

We experimented with Image Tools in Mux, but we were not happy with how the first implementation turned out, and the value from the feature was not clear enough to keep carrying it forward right now.

We're not opposed to having image tools in the future, but this first iteration was not good enough, so this PR removes the experiment.

Closes #3270

## Implementation

- Removed the Image Tools experiment from settings, send options, persisted task experiment schemas, and stream options.
- Removed image tool definitions, runtime implementations, image artifact helpers, provider image-model creation paths, and generated/edited image display rows.
- Removed imagegen-specific built-in skill filtering and regenerated the built-in skill snapshot after deleting the `imagegen` skill and image ADR docs.
- Left generic image surfaces intact: attachments, screenshots, `ToolResultImages`, `ImageLightbox`, raster resizing, and upstream token/model metadata.

## Validation

- `make static-check`
- Targeted tests for tool definitions, config/API schemas, stream context, agent skill loading, message rendering/aggregation, provider model factory, and model stats.
- Dogfooded a clean dev-server sandbox and verified Settings → Experiments no longer shows Image Tools or image model/upload controls.

## Pains

- `make test` was attempted locally; it did not complete cleanly because of unrelated existing/suite-order failures. The previously failing hook and UI tests passed in isolation, while `WorkspaceService fork > cleans up init state when orchestrateFork rejects` reproduced independently of this change.

## Risks

Low-to-medium. This removes a product experiment and many direct callsites; old persisted image tool outputs now render as generic tool rows rather than first-class image preview rows. Generic image attachment/screenshot previews were intentionally preserved.

---

<details>
<summary>📋 Implementation Plan</summary>

# Plan: remove failed Image Tools experiment and close PRD issue

## Context and evidence

- User intent: treat the Image Tools experiment as failed, remove Image Tools/image-tool-related code from Mux first, then comment on and close the related GitHub PRD issue.
[...]

PR #3936, opened 2026-08-23
🤖 fix: bundle darwin-x64 sharp runtimes in macOS builds

### LINKED ISSUE #3338: Mux 0.25 doesn't work on Mac x64

ibetitsmike:
## Summary
Fixes #3338. Intel (x64) macOS builds shipped without the darwin-x64 sharp native runtime, so `sharp` could not load on Intel Macs. This PR makes macOS packaging install both Darwin sharp runtimes before electron-builder copies `node_modules`, and hardens the packaged-app smoke check so a missing per-arch runtime fails CI instead of shipping.

## Background
macOS release builds run on arm64 runners where bun installs only host-platform optional dependencies, and electron-builder (`npmRebuild: false`) packs the same `node_modules` tree into both arch bundles. Verified against the latest release: `mux-0.28.2-x64.zip` contains only `@img/sharp-darwin-arm64` and `@img/sharp-libvips-darwin-arm64` under `app.asar.unpacked/node_modules/@img`, with no darwin-x64 packages.

In v0.25.0 (the report in #3338) this crashed startup because the since-removed experimental image tools (#3282, removed in #3419) imported sharp eagerly. On current main sharp is lazy-loaded, so the app starts but `attach_file` raster image processing fails at runtime on Intel Macs. The existing smoke test (#3088) only asserted that some sharp binary exists in one bundle, so CI never caught the broken x64 artifact. PR #3434 took the same approach in June but went stale and no longer applies to main.

## Implementation
- `Makefile`: new `ensure-mac-sharp-runtime-deps` target runs `bun install --frozen-lockfile --os=darwin --cpu=x64` and `--cpu=arm64` (additive on bun >= 1.3; `bun.lock` already pins all platform variants), then a host-native install to restore `node_modules/.bin` symlinks, then asserts all four `@img` darwin package dirs exist. The guard matters because bun < 1.3 silently ignores `--os`/`--cpu`. Wired into `dist-mac`, `dist-mac-release`, `dist-mac-x64`, and `dist-mac-arm64`.
- `Makefile`: `ESBUILD_BIN` now prefers `@esbuild/<host-platform>` (derived from `uname`) before the wildcard fallback; with darwin esbuild packages installed on a non-mac host, the old `$(firstword $(wildcard ...))` picked `darwin-arm64` alphabetically and broke every esbuild build step.
- `scripts/checkMacAttachFileRuntime.ts`: verifies every built app bundle instead of one, detects each bundle's architectures via `lipo -archs`, and requires the exact `@img/sharp-darwin-<arch>` binding (`*.node`) and `@img/sharp-libvips-darwin-<arch>` dylib per architecture. In default mode it also fails if either the x64 or arm64 bundle is missing entirely. The runtime attach_file smoke run stays native-arch only.
- `.github/workflows/pr.yml`: Build / macOS now runs `make check-mac-attach-file-runtime` after `make dist-mac`, so PR CI proves both bundles carry their runtimes (previously only release builds ran the check).

## Validation
- Inspected the v0.28.2 x64 release zip to confirm the missing darwin-x64 packages (root cause proof).
- Verified on bun 1.3.5 (CI's pinned version) that the cross-arch installs are additive and leave `bun.lock` untouched, and that bun 1.2.15 silently no-ops (hence the guard).
- Red-green checked the `ESBUILD_BIN` fix: with a fake `@esbuild/darwin-arm64` present, the old expression resolves to the darwin binary, the new one resolves to the host binary.
- Remote dogfood UAT (3 rounds, Linux): ensure target exit 0 with all four darwin packages added and lockfile unchanged; guard fails loudly on old bun; `attach_file` tests 19/19; `make dev-server` boots cleanly after the ensure target; static-check green.
- Not verifiable without darwin-x64 hardware: actually executing the packaged x64 app. The strengthened smoke check in the Build / macOS job is the closest CI proxy (per-arch binary presence plus `lipo` verification plus a native-arch runtime run).

## Risks
Release-pipeline only; no app runtime code changes. The mac dist targets now mutate `node_modules` before packaging, which grows both mac bundles by the extra Darwin sharp/libvips packages (a few MB each). If a future bun regresses `--os`/`--cpu` handling, the explicit directory guard and the CI smoke check fail the build rather than shipping a broken artifact.

---

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -538,6 +538,8 @@ jobs:
       - name: Build macOS distributables
         # Retry for transient Apple timestamp-server failures during code signing.
         run: ./scripts/retry.sh 3 30 make dist-mac
+      - name: Smoke test packaged macOS attach_file runtime
+        run: make check-mac-attach-file-runtime
       - name: Upload x64 artifact
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ include fmt.mk
 .PHONY: build-renderer version build-icons build-static build-docker-runtime verify-docker-runtime-artifacts
 .PHONY: lint lint-fix typecheck static-check static-check-full
 .PHONY: test test-unit test-integration test-watch test-coverage test-e2e test-e2e-perf smoke-test
-.PHONY: dist dist-mac dist-win dist-linux install-mac-arm64 check-appimage-icons check-mac-attach-file-runtime
+.PHONY: dist dist-mac dist-win dist-linux install-mac-arm64 ensure-mac-sharp-runtime-deps check-appimage-icons check-mac-attach-file-runtime
 .PHONY: vscode-ext vscode-ext-install
 .PHONY: docs-server check-docs-links
 .PHONY: storybook storybook-run storybook-build test-storybook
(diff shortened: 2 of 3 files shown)


ibetitsmike:
@codex review

PR #3942, opened 2026-08-24
🤖 fix: preserve MCP OAuth discovery across callbacks

ethanndickson:
## Summary

Preserve MCP OAuth authorization-server discovery across the redirect and callback legs so providers such as Linear can complete the authorization-code exchange without weakening the SDK's issuer mismatch protection.

## Background

`@modelcontextprotocol/client` verifies that the authorization server discovered during the callback matches the server that initiated the redirect before sending the authorization code and PKCE verifier. Xum saved that discovery state on the interactive flow but did not expose it through `OAuthClientProvider.discoveryState()`, so the SDK treated the redirect issuer as unavailable and rejected Linear's callback.

## Implementation

- Return the redirect-leg discovery state from the interactive OAuth provider.
- Add an integration-style regression test with separate resource-server and authorization-server origins, matching Linear's OAuth topology.
- Verify that the callback reaches the token endpoint with the authorization code and PKCE verifier.

## Validation

- `bun test src/node/services/mcpOauthService.test.ts`
- `make static-check`
- Confirmed the new regression test fails with the original implementation using the SDK's `Authorization server changed between redirect and callback` error, then passes with discovery-state read-back enabled.

## Risks

Low. The behavior change is limited to interactive MCP OAuth flows and supplies state that Xum already retained in memory. The SDK continues to perform its authorization-server identity check before exchanging credentials.

---

_Generated with `xum` • Model: `openai:gpt-5.6-sol` • Thinking: `high` • Cost: `$12.28`_

### DIFF
--- a/src/node/services/mcpOauthService.test.ts
+++ b/src/node/services/mcpOauthService.test.ts
@@ -600,7 +600,7 @@ describe("probeServerForBearerChallenge", () => {
   });
 });
 
-describe("McpOauthService.startDesktopFlow", () => {
+describe("McpOauthService OAuth flows", () => {
   let xumHome: string;
   let projectPath: string;
   let config: Config;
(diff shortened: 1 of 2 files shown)


ethanndickson:
@codex review

Please review the OAuth discovery-state callback fix and its regression coverage.

PR #3945, opened 2026-08-24
🤖 fix: keep compaction identity across startup auto-retry

coadler:
## Summary

Startup auto-retry of an interrupted compaction lost the compaction-request metadata, so the resumed stream was no longer tracked as a compaction. The model's summary reply became a plain assistant response and the chat never compacted.

## Background

When a stream is interrupted (crash, restart), `deriveStartupAutoRetryRequest` rebuilds send options from persisted state. Its compaction branch rebuilt model/agentId/toolPolicy but dropped `muxMetadata`, because `StartupRetrySendOptions.muxMetadata` only allowed `workspace-turn-task` metadata.

On resume, `resolveCompactionRequest` scans history backward for the compaction-request row. Without the stream identifying as a compaction request, any trailing synthetic user row — a `<system-file-update>` file-change notice or the `[CONTINUE]` sentinel — aborts the scan. `activeCompactionRequest` stayed undefined, `compactionHandler.handleCompletion` treated the stream-end as a normal turn, and no boundary summary was written.

## Implementation

- Widen `StartupRetrySendOptions.muxMetadata` to also accept `{ type: "compaction-request" }`.
- In the compaction branch of `deriveStartupAutoRetryRequest`, carry the original `muxMetadata` through so the resumed stream still identifies itself; `resolveCompactionRequest` then resolves via the options path regardless of trailing synthetic rows.

## Validation

- New regression test seeds durable history with a compaction-request row plus a trailing synthetic row, runs the real startup auto-retry path (`scheduleStartupAutoRetryIfNeeded`), and asserts the boundary lands. Fails before the fix, passes after.
- Added a control test proving plain synthetic agent-initiated sends (the parent-to-child guidance path) compact correctly end-to-end.
- Added a characterization test documenting that family-payload (`preTurnMessages`) sends intentionally skip on-send compaction.
- `make static-check`, targeted compaction suites green. Full `src/node/services` run shows only failures verified pre-existing on the pristine baseline (unrelated hooks/browser suites).

## Risks

Low. The metadata now flows through the same option field `resumeStream` already consumes; `normalizeGatewaySendOptions` spreads it through untouched. The widened union is additive: existing `workspace-turn-task` behavior is unchanged, and non-compaction retries never set the field.

---

_Generated with `xum` • Model: `openrouter:stealth/ox-alpha` • Thinking: `high` • Cost: `$0.04`_

### DIFF
--- a/src/common/types/message.ts
+++ b/src/common/types/message.ts
@@ -93,8 +93,16 @@ export type StartupRetrySendOptions = Pick<
   | "disableWorkspaceAgents"
   | "allowAgentSetGoal"
 > & {
-  /** Correlation for a delegated workspace turn that must survive restart recovery. */
-  muxMetadata?: Extract<MuxMessageMetadata, { type: "workspace-turn-task" }>;
+  /**
+   * Correlation metadata that must survive restart recovery: delegated
+   * workspace turns and interrupted compaction requests. A resumed compaction
+   * stream needs its original metadata so trailing synthetic rows cannot break
+   * compaction detection (resolveCompactionRequest only skips synthetic rows
+   * when the stream itself identifies as a compaction request).
+   */
+  muxMetadata?:
+    | Extract<MuxMessageMetadata, { type: "workspace-turn-task" }>
+    | Extract<MuxMessageMetadata, { type: "compaction-request" }>;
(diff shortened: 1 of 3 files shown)


coadler:
@codex review

PR #3951, opened 2026-08-24
🤖 perf: reuse instruction sources during stream startup

yanncabral:
## Summary

Load structured instruction sources once during stream startup and reuse that snapshot for both system-message assembly and tool-scoped instruction extraction. Existing `buildSystemMessage()` and `readToolInstructions()` wrappers remain available for callers that need load-and-build behavior.

## Background

Stream startup previously loaded the same AGENTS.md/instruction sources while building the system message, then loaded them again while extracting tool instructions. Besides duplicate I/O, the two consumers could observe different file contents if instructions changed between reads.

## Implementation

- add pure helpers that assemble a system message and extract tool instructions from an existing `InstructionSources` value
- return the loaded snapshot from `buildStreamSystemContext()` and reuse it for tool extraction
- thread the snapshot through policy/model-driven system-context rebuilds within the same startup
- retain the existing loading APIs and source-ordering/scoping behavior

## Validation

- TDD RED: the new stream-context regression initially failed because `extractToolInstructionsFromSources` was not exported
- TDD GREEN: focused regression now verifies one snapshot produces both the expected prompt content and bash tool instructions
- `bun test src/node/services/systemMessage.test.ts src/node/services/streamContextBuilder.test.ts src/node/services/aiService.test.ts --timeout 30000` — 150 passed
- targeted ESLint on all five changed files — passed (split into two invocations after the combined process was OOM-killed)
- `bun run tsgo --noEmit -p tsconfig.main.json` — passed
- Prettier check/write and `git diff --check` — passed
- full `make typecheck` attempted: main-process typecheck passed, broad renderer typecheck was SIGKILL/OOM

## Risks

The change touches startup context plumbing. Regression coverage retains the existing system-message matrix for sub-project, multi-project, global/compat, local, model/mode, Xum-only, and tool-scoped semantics, and asserts identity reuse at the AI service boundary.

---

_Generated with `xum` • Model: `openai:gpt-5.6-sol` • Thinking: `medium` • Cost: `unknown`_

### DIFF
--- a/src/node/services/aiService.test.ts
+++ b/src/node/services/aiService.test.ts
@@ -304,10 +304,16 @@ function stubCommonStreamMessageDependencies(args: {
   useRequestedModelString?: boolean;
   onPlanPayloadMessageIds?: (messageIds: string[]) => void;
   onBuildStreamSystemContext?: (
-    args: Parameters<typeof streamContextBuilder.buildStreamSystemContext>[0]
+    args: Parameters<typeof streamContextBuilder.buildStreamSystemContext>[0],
+    instructionSources: Awaited<
+      ReturnType<typeof streamContextBuilder.buildStreamSystemContext>
+    >["instructionSources"]
   ) => void;
   onPrepareMessagesForProvider?: (
     args: Parameters<typeof messagePipeline.prepareMessagesForProvider>[0]
+  ) => void;
+  onExtractToolInstructions?: (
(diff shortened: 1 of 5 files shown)


ThomasK33:
Closing as superseded by #4432, which builds on this PR's load-once, reuse-per-turn helpers. `readToolInstructionsMs` is now 0 on SSH workspaces, and median time-to-stream-start at 80 ms RTT dropped from 13.1 s to 7.3 s together with #4433. Thank you for the approach, which is credited in #4432.

---

PR #3952, opened 2026-08-24
🤖 perf: load workspace context concurrently

yanncabral:
## Summary

Loads independent workspace context concurrently after agent resolution, reducing startup latency before a model stream can begin.

## Implementation

- Starts goal, workspace MCP override, and scratchpad/additional-context reads together.
- Preserves MCP and scratchpad fallback behavior and startup phase timings.
- Awaits MCP overrides before server listing and tool assembly, the goal before goal-tool availability, and scratchpad context before plan instruction construction.
- Adds a regression test that holds the goal read pending and verifies the other independent reads have already started.

## Validation

- Confirmed the new focused test fails against the serial implementation and passes after the change.
- `bun test src/node/services/aiService.test.ts` (105 pass, 0 fail)
- `bunx prettier --check src/node/services/aiService.ts src/node/services/aiService.test.ts`
- `bunx eslint src/node/services/aiService.ts src/node/services/aiService.test.ts`
- `bun x tsgo --noEmit -p tsconfig.main.json`

## Risks

Low-to-moderate risk in stream startup ordering. The resolved values are still awaited at their original consumption boundaries, and fallback behavior remains covered by the existing AIService test suite.

---

_Generated with `xum` • Model: `unknown` • Thinking: `unknown` • Cost: `unknown`_

### DIFF
--- a/src/node/services/aiService.test.ts
+++ b/src/node/services/aiService.test.ts
@@ -65,6 +65,8 @@ import { ExperimentsService } from "./experimentsService";
 import type { DevToolsService } from "./devToolsService";
 import { TelemetryService } from "@/node/services/telemetryService";
 import type { WorkspaceGoalService } from "./workspaceGoalService";
+import type { WorkspaceMcpOverridesService } from "./workspaceMcpOverridesService";
+import * as additionalSystemContext from "./additionalSystemContext";
 import * as agentResolution from "./agentResolution";
 import * as streamContextBuilder from "./streamContextBuilder";
 import * as messagePipeline from "./messagePipeline";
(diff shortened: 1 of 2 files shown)


yanncabral:
@codex review

ThomasK33:
Closing after measurement. On an 80 ms RTT SSH workspace, the three reads this PR runs in parallel cost about 320 ms (MCP overrides) and about 0 ms (additional system context). Running them in parallel would save close to nothing, while the MCP-override read has timeout, abort and authoritative semantics that parallelism would need to preserve. The related stream-startup work landed in #4432/#4433. Thank you for the contribution.

---

PR #3953, opened 2026-08-24
🤖 fix: clamp model selector dropdown height to stay below the top bar

ibetitsmike:
## Summary

The model selector dropdown could render partially or fully under the top bar, hiding its search input and top items. The panel now measures the space above its trigger when it opens and clamps its height so it always fits below the bar, with only the model list shrinking and scrolling.

## Background

The dropdown opens upward (`absolute bottom-full`) with a fixed 280px list plus a search header and footer, and nothing constrained it to the viewport. In short windows, or wherever the composer sits high (for example the new-workspace page), the panel top extended past the window top and under the title/menu bar. The auto-focused search input became invisible, and in desktop mode the bar is a window-drag region, so the overlapped part could not even receive clicks.

## Implementation

- `ModelSelector` runs a layout effect when the dropdown opens: it reads the trigger's viewport position and caps the panel `max-height` to the space above it, keeping clear of the titlebar height plus a small gap. It re-measures on window resize.
- The panel is now a flex column; the search header and footer are `shrink-0`, so only the model list shrinks and scrolls. The existing 280px list cap still applies when space allows.
- A 150px floor keeps the panel usable in pathological layouts (slight bar overlap is preferred over an unusably short panel).
- `useDesktopTitlebar.ts` exports the titlebar height as a numeric constant alongside the existing Tailwind classes.

The sibling pickers (`AgentModePicker`, `ThinkingSelector`) share the upward-opening pattern but are short enough not to hit this in practice; they are left unchanged to keep the diff minimal.

## Validation

Reproduced and verified visually in Storybook (`ModelSelectorDropdownOpen` story) at a 900x400 viewport, where the old code placed the panel top ~38px above the window: the full panel (search input, scrollable list, footer) now fits between the top bar and the trigger.

No unit test added: happy-dom cannot do layout, so the only possible assertion would restate the clamp arithmetic against a mocked rect.

---

### DIFF
--- a/src/browser/components/ModelSelector/ModelSelector.tsx
+++ b/src/browser/components/ModelSelector/ModelSelector.tsx
@@ -8,6 +8,7 @@ import React, {
   useState,
   useRef,
   useEffect,
+  useLayoutEffect,
   useCallback,
   useImperativeHandle,
   forwardRef,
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

Issue #3960, opened 2026-08-25
🤖 perf: frontend WebSocket transfers 73 MB and blocks the UI for minutes on large deployments


ThomasK33:
## Summary

On a large deployment (246 live workspaces, 1,513 total in config.json, 100+ with sub-agent children), the single `/orpc/ws` connection transferred **73.34 MB** and the UI stayed non-interactive for minutes. Investigated live (probe = headless Chromium loading the real frontend, attributing every WS frame to its oRPC procedure). Full report: `docs/perf/websocket-traffic-investigation.md` on branch `ws-traffic-perf`.

**Measured on the live deployment (build `93244e0c7`):**

- Fresh client boot: **8.6–9.1 MB**, time-to-interactive **131–162 s**.
- Idle (no workspace selected): **1.2 MB/min**.
- Workflow workspace open: **5.2 MB/min (85 KB/s)** steady state.

## Attribution (boot, measured)

| bytes | frames | procedure |
| ---: | ---: | --- |
| 4,488,126 | 4 | `projects/list` (1.12 MB/call — embeds all 1,513 workspaces, **1,267 archived = 84% of payload**; re-fetched on every `config.onConfigChanged`; 8 calls = 9.3 MB in one 4-min session) |
| 2,893,100 | 1 | `workspace/activity/list` (13,901 entries from never-pruned `extensionMetadata.json`; server took **~59 s** to answer) |
| 882,295 | 244 | `workspace/getSessionUsage` (one RPC per workspace; `getSessionUsageBatch` already exists) |
| 214,572 | 1 | `workspace/list` |
| < 100 KB | — | everything else (subscription deltas are well-behaved) |

Compounding factors: no `perMessageDeflate` on the WS server (`src/node/orpc/server.ts:1628` — `ws` defaults off, so all of this is uncompressed JSON), and the `<App />` mount gate (`AppLoader.tsx:219-276`) waits on responses that queue behind this flood on one ordered socket — that's the multi-minute lockout.

Workflow layer: contributing (~14% while viewing a workflow workspace), not primary — `WorkflowRunToolCall.tsx:1536-1544` polls the **full growing run record** (105 KB and counting) every 2 s even when the card is collapsed, and `workflows.subscribe` pushes the full record per durable write (O(N²) cumulative). Tracked with the anomaly issue below.

## Recommended first move

**Stop serializing dead workspaces** (server-side only, protocol-compatible):

[...]

Issue #3959, opened 2026-08-25
🤖 perf: workspace.activity.list ships 13,895 stale entries (2.9 MB, ~59 s) — extensionMetadata.json never prunes


ThomasK33:
## Summary

Separable live anomaly found during the WS-traffic investigation (see the main issue): on a long-lived deployment, `~/.mux/extensionMetadata.json` contains **13,895 workspace entries** while only 246 workspaces are live (1,513 in config including archived). Entries for removed workspaces/sub-agents are never pruned.

## Impact (measured live)

- `workspace.activity.list` returns **13,901 entries = 2.89 MB** of uncompressed JSON on every client boot/reconnect (the sidebar needs ~246).
- The call took the server **~59 s** to answer: `getActivityList()` (`src/node/services/workspaceService.ts:12521-12599`) unions `extensionMetadata.getAllSnapshots()` keys and then bootstraps `getActiveWorkflowRunIds(workspaceId)` — a per-workspace disk probe — for all 13,901 ids (`workspaceService.ts:12551`, `3737-3762`).
- That response shares the single ordered `/orpc/ws` socket with the boot-gate-critical responses, contributing directly to the minutes-long "Loading Xum" lockout.

## Suggested fix

- Intersect `getActivityList()` ids with config-known (live) workspaces instead of every snapshot key on disk, keeping the workflow/bash-monitor tombstone semantics for known workspaces.
- Prune `extensionMetadata.json` entries when workspaces are removed (plus a one-time lazy cleanup at load for pre-existing deployments). Keep the on-disk file backward-compatible (filter at read) per upgrade↔downgrade policy.

A healthy/fresh deployment does not exhibit this: activity.list is a few KB and instant. This is a state leak that grows monotonically with deployment age.

Related: #MAIN_ISSUE

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$1.01`_

ThomasK33:
Main investigation issue: https://github.com/coder/xum/issues/3960

PR #3969, opened 2026-08-25
🤖 perf: exclude archived workspaces from projects.list response

ThomasK33:
Part (a) of #3960.

## Summary

`projects.list` no longer serializes archived workspaces. This is a read-side/API projection only: archived entries remain untouched in `config.json` (upgrade↔downgrade safe), they are simply excluded from the response that `ProjectContext` re-fetches on every `config.onConfigChanged` signal. On the reference deployment (1,513 embedded workspaces, 84% archived, 1.12 MB per call, re-fetched 8× in a 4-minute session) this removes the top byte source on the frontend WebSocket.

## Audit: who consumes the embedded `workspaces` arrays?

Exactly three frontend read sites consume `project.workspaces` from `projects.list`:

| Consumer                                                                        | Needs archived entries?                                                                                                        | Outcome               |
| ------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------ | --------------------- |
| `buildSortedWorkspacesByProject` (`src/browser/utils/ui/workspaceFiltering.ts`) | No — looks up `ws.id` against live metadata from `workspace.list`, which already excludes archived; archived IDs never matched | unchanged             |
| Sub-project removal confirmation (`ProjectSidebar.handleRemoveSection`)         | No — informational count for a non-destructive action; now counts live workspaces only (matches the sidebar view it describes) | comment added         |
| Project delete confirmation (`ProjectSidebar.handleRequestProjectRemoval`)      | **Yes** — derived active/archived counts locally to build the type-to-confirm dialog                                           | **converted** (below) |

The archived-workspace UI (project page “Archived Workspaces” section, restore/delete flows) was already on-demand via `workspace.list({ archived: true })` and is unaffected.

Dropping the embedded bodies entirely was considered but rejected for now: the sidebar’s first-pass grouping still uses the live arrays (incl. multi-project workspace buckets), and live entries are a small fraction of the payload.

## Implementation

- **`ProjectService.list()`**: projects each project to a copy whose `workspaces` array is filtered through the existing `isWorkspaceArchived` helper. New arrays are built (never mutating the loaded config), so persisted state keeps archived entries.
- **Project delete flow**: since blocker counts can no longer be derived client-side, the sidebar calls a new **read-only** `projects.getRemovalBlockers` preflight, which reports the same blocker counts `remove()` enforces (own bucket + cross-project references) without `remove()`’s side effects (stale-entry pruning, deletion of empty projects). The dialog therefore shows authoritative counts — _more_ accurate than the old local count, which missed cross-project references. A monotonic sequence guard discards stale preflight responses so rapid Delete clicks on different projects can’t cross wires. Empty projects are still removed immediately (same as before).
- **`remove(force)` hardening** (from review): the force cascade now fails fast on cross-project references _before_ deleting anything, so a mixed project can no longer have its directly-owned workspaces deleted only for the operation to fail on external references afterwards.

Not touched (per #3959 coordination): `getActivityList()` / extension-metadata store.

## Validation

- Unit tests: projection filters archived, preserves live + re-unarchived entries, keeps multi-project workspaces in their bucket, and leaves persisted config unmodified (`projectService.test.ts`); delete flow shows backend counts in the dialog and removes empty projects without a dialog (`ProjectSidebar.test.tsx`).

### Dogfood (dev-server sandbox, seeded config: 1 project, 100 workspaces = 10 live + 90 archived)

Payload of one `projects.list` response (uncompressed JSON, same seed config):

|                        |     bytes | embedded workspaces |
| ---------------------- | --------: | ------------------: |
| before (parent commit) |    60,613 |                 100 |
| after (this PR)        |     5,713 |                  10 |
| **reduction**          | **10.6×** |                     |

Extrapolated to the reference deployment (84% archived + per-entry overhead): ~1.12 MB → ~0.1 MB per call, ~9.3 MB → ~0.9 MB for the measured 4-minute session.

UI flows verified in the browser against the patched sandbox:

- Sidebar lists the 10 live workspaces; project page shows “Archived Workspaces (90)” loaded on demand.
- Restore: archived workspace moves to sidebar, count 10 → 11.
- Archive: workspace leaves sidebar, count 11 → 10.
- Delete project: the backend blocker counts resolve to `{activeCount: 10, archivedCount: 90}` and the type-to-confirm dialog renders “This will permanently delete 100 workspaces (10 active, 90 archived)” — correct counts despite `projects.list` containing no archived entries. (Dogfooded pre-review via the non-forced `projects.remove` probe; the flow now uses the read-only `getRemovalBlockers` preflight covered by unit + story tests.)

| Archived section (90) + sidebar (10)                                                                             | Delete dialog with backend counts                                                                                              | Restored workspace in sidebar (11)                                                                                        |
| ---------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------- |
| archived | delete-dialog | restored |

Screen recording of the full flow (restore → archive → delete dialog → cancel): [dogfood-flow.webm](https://raw.githubusercontent.com/coder/xum/assets/projects-list-projection-evidence/dogfood-flow-fixed.webm)

_(Evidence lives on the temporary `assets/projects-list-projection-evidence` branch because attachment upload requires an interactive SSO session; safe to delete after merge.)_

## Risks

- **Delete-project UX**: clicking “Delete…” now issues a read-only `projects.getRemovalBlockers` RPC before the dialog (no destructive effects, no pruning). Severity: low.
- **`remove(force)` behavior change**: projects with cross-project references now fail before (rather than after) cascade deletion; previously such removals destroyed owned workspaces and then failed anyway, so this strictly reduces damage. Severity: low.
- **Sub-project removal dialog** counts only live workspaces now; the action itself is non-destructive.
- Any external API consumer reading archived workspaces out of `projects.list` must use `workspace.list({ archived: true })` instead (the UI already did).

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$0.00`_

### DIFF
--- a/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
+++ b/src/browser/components/ProjectSidebar/ProjectSidebar.test.tsx
@@ -639,11 +639,15 @@ function installProjectSidebarTestDoubles() {
     ((props: {
       isOpen: boolean;
       projectName: string;
+      activeCount: number;
+      archivedCount: number;
       onConfirm: () => void;
       onCancel: () => void;
     }) =>
       props.isOpen ? (
-        <div data-testid="project-delete-confirmation-modal">{props.projectName}</div>
+        <div data-testid="project-delete-confirmation-modal">
+          {`${props.projectName}:${props.activeCount}:${props.archivedCount}`}
+        </div>
       ) : null) as unknown as typeof ProjectDeleteConfirmationModalModule.ProjectDeleteConfirmationModal
   );
   installProviderIconSvgMocks();
@@ -2368,11 +2372,21 @@ describe("ProjectSidebar project actions menu", () => {
     expect(view.getByRole("button", { name: "Edit name" })).toBeTruthy();
   });
 
-  test("menu actions route to settings and delete confirmation", () => {
+  test("menu actions route to settings and delete confirmation", async () => {
+    const removeProjectCalls: Array<{ path: string; options?: { force?: boolean } }> = [];
     projectContextValue = createProjectContextValue({
       userProjects: new Map([
         [demoProjectPath, { workspaces: [{ path: `${demoProjectPath}/ws-1` }] }],
       ]),
+      removeProject: (path, options) => {
+        removeProjectCalls.push({ path, options });
+        // projects.list excludes archived workspaces, so the delete flow gets
+        // blocker counts from the backend's non-forced rejection.
+        return Promise.resolve({
+          success: false,
+          error: { type: "workspace_blockers", activeCount: 2, archivedCount: 3 },
+        });
+      },
     });
 
     const view = renderSidebar();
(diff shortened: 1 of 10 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
Addressed all three review findings in 3f95b79a5:

1. **Side-effectful removal before confirmation** → the delete flow no longer probes via `remove()`. A new read-only `projects.getRemovalBlockers` RPC reports the same blocker counts (own bucket + cross-project references) without pruning stale entries or deleting empty-looking projects. Covered by a unit test seeding a workspace entry with a missing checkout directory and asserting config is unmodified after the preflight.

2. **Confirmed deletion vs cross-project blockers** → `remove(force)` now counts cross-project references *before* starting the cascade and fails fast with `workspace_blockers`, so a mixed project can no longer lose its directly-owned workspaces to a doomed removal. Covered by a unit test asserting no `workspaceService.remove` calls and an intact config.

3. **Concurrent removal preflights** → `handleRequestProjectRemoval` uses a monotonic sequence token; stale responses are discarded (both success and error paths). Covered by a component test racing two preflights and asserting the slower earlier response cannot overwrite the newer dialog.

ThomasK33:
@codex review

ThomasK33:
Addressed all three round-2 findings:

1. **Manual memoization** → `getRemovalBlockers` is now defined inline in the provider value memo (like the adjacent inline actions; `api` is already a memo dependency), no `useCallback`.
2. **Stale error anchor** → `handleRequestProjectRemoval` captures the anchor coordinates before awaiting the preflight; `showProjectRemoveError`/`removeProjectWithFeedback` now take precomputed coordinates instead of an element.
3. **Duplicated IPC shape** → `ProjectWorkspaceCountsSchema` now lives in `src/common/utils/projectRemoval.ts` as the single source of truth; the `ProjectWorkspaceCounts` type is inferred from it and the `projects.getRemovalBlockers` output schema imports it.

ThomasK33:
@codex review

PR #3972, opened 2026-08-25
🤖 fix: commit compaction boundary after crash-recovery resume

coadler:
## Summary

Keeps a pending compaction request correlated with its resumed stream after an app crash mid-compaction, so the compaction boundary is committed instead of silently dropped.

## Background

When Xum crashes mid-compaction (for example on OOM), startup recovery finds a trailing non-partial assistant row and appends a synthetic `[CONTINUE]` user sentinel before re-running the turn. The recovered stream then produces a valid summary, but `resolveCompactionRequest` stopped scanning history at that sentinel unless the current stream itself was a compaction send. The regenerated summary persisted as an ordinary assistant message and no compaction boundary was ever committed; the user had to run compaction again.

Observed in workspace `dffb219cd6`: attempt 3 completed at the provider level (`finishReason: stop`, 262k input tokens) but produced no `compaction.completed` timeline event.

## Implementation

- `resolveCompactionRequest` now skips any synthetic user row (`metadata.synthetic === true`) regardless of send options. Correlation still stops at the first real user row, so normal sends after a compaction request never mis-correlate.
- Added a behavioral test covering the exact crash shape: compaction request → orphaned assistant summary → synthetic `[CONTINUE]`, plus a guard that a real user message stops correlation.

## Validation

- Reproduced the failure shape from the affected workspace in a unit test; it fails against the old logic and passes now.
- `bun test src/node/services/agentSession.autoCompaction.test.ts src/node/services/compactionHandler.test.ts`: 83 pass, 0 fail.
- `make typecheck`, `make lint`, and `make static-check` all pass.

## Risks

Low. The change only widens which synthetic rows are skipped during backward correlation; stopping at the first real user row is unchanged, so ordinary conversation flow cannot mis-correlate a stale compaction request. Worst case for a pathological history is correlating a summary stream with an older pending compaction request, which commits the boundary that request asked for anyway.

---

_Generated with `xum` • Model: `openrouter:stealth/ox-alpha` • Thinking: `high` • Cost: `$0.01`_

### DIFF
--- a/src/node/services/agentSession.autoCompaction.test.ts
+++ b/src/node/services/agentSession.autoCompaction.test.ts
@@ -151,8 +151,9 @@ describe("AgentSession on-send auto-compaction snapshot deferral", () => {
         muxMetadata: compactionMetadata,
       }
     );
-    const snapshot = createMuxMessage("file-change", "user", "<file-change />", {
+    const snapshot = createMuxMessage("file-snapshot", "user", "@foo.ts contents", {
       synthetic: true,
+      fileAtMentionSnapshot: ["t0"],
     });
     const internals = session as unknown as {
       resolveCompactionRequest: (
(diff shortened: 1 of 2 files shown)


coadler:
@codex review

coadler:
Addressed in cbd307de2. Agreed — skipping every synthetic user row was too broad: a stale request left by a failed summary (no boundary committed) could be claimed by an unrelated later turn that began with some other synthetic row.

Traversal now stops at any synthetic user row except the three recognized producer shapes:

- file @-mention prompt snapshots (metadata.fileAtMentionSnapshot != null)
- turn-start <system-file-update> file-change notifications (appended after the compaction request within the same send, before resolveCompactionRequest runs)
- the crash-recovery [CONTINUE] sentinel

Everything else — goal continuations, task wakes, real user rows — stops correlation. Added a regression test for the stale-request scenario.

coadler:
@codex review

Please take another look.

PR #3973, opened 2026-08-25
🤖 fix: JSON-normalize code_execution output and clamp giant stream errors

ibetitsmike:
## Summary

Fixes a mid-stream crash (`AI_InvalidPromptError: messages do not match the ModelMessage[] schema`) caused by `code_execution` tool results containing non-JSON values (`undefined`), and clamps the resulting multi-hundred-KB error payloads that were persisted, shipped over IPC, and rendered in chat.

## Background

A workspace stream died mid-turn with a schema validation error. Forensics from session logs:

1. A `code_execution` script logged an expression that evaluated to `undefined` in the QuickJS guest.
2. `ctx.dump()` marshals that as a real in-memory `undefined` inside the tool result (`consoleOutput[0].args = [undefined]`).
3. On the next step of the same stream, the AI SDK embeds that live object verbatim into a tool-result message and validates it against its strict `JSONValue` schema. `undefined` is not a JSON value, so the stream is killed.
4. Automatic retry succeeded because it rebuilt messages from JSON-rehydrated history where `undefined` had been normalized to `null`: a heisenbug that crashes live but works after reload.

Separately, the `InvalidPromptError`'s cause chain embeds the entire serialized prompt (~150KB+), which `getErrorMessage()`'s cause walk appended to the persisted/rendered error.

## Implementation

- **Root cause:** `code_execution` output is cloned through `jsonSafeClone()` into the exact JSON shape persistence produces (`undefined` in arrays → `null`, `undefined` object values → dropped, `toJSON` honored) before crossing into the AI SDK, so live and rehydrated behavior are identical.
- **New `jsonSafeClone()` util** (`src/common/utils/jsonSafeClone.ts`) with a hardened fallback for values `JSON.stringify` outright rejects: BigInt → decimal string (magnitude-gated at 10^4096 with an O(1) comparison so hostile sandbox values cannot block the host on `toString()`), cycles → `"[Circular]"`, throwing getters dropped, throwing `toJSON` getters treated as absent, sparse-array holes filled with `null`, literal `__proto__` keys kept as own data properties (`Object.fromEntries`), and objects with throwing reflective traps (Proxy `ownKeys`/revoked proxies) degraded to `"[Unserializable]"`.
- **Error inflation:** stream error payloads are clamped to 8KB (head + tail preserved so the actionable validation detail after the payload dump survives) inside `persistStreamError()`, the single choke point every payload passes through before persist/IPC/render, covering early-return branches like `ModelRefusalError` whose `fallbackNote` can embed another error's message.
- Limits centralized in `src/constants/errors.ts` and `src/constants/json.ts`.

## Validation

- Regression test reproducing the exact crash vector (`console.log(undefined)` plus a returned `{a: undefined, arr: [undefined, 1], nan: NaN}`) asserting the tool output equals its own JSON round-trip.
- Fallback-hardening regression tests: oversized/boundary BigInts, sparse arrays, throwing `toJSON` getters, `__proto__` prototype integrity, throwing Proxy traps, revoked proxies.
- `code_execution`, `streamManager`, `errors`, and `jsonSafeClone` suites pass; full `make static-check` green.

## Risks

Low-to-moderate, scoped to `code_execution` result shape and stream error formatting:

- `jsonSafeClone` runs on every `code_execution` result; a behavior difference vs. raw output would surface as subtly different tool results. Mitigated by matching `JSON.stringify` semantics exactly (the shape every result already assumed after rehydration).
- Error clamping truncates the middle of giant messages; worst case is losing detail in a >8KB error body, which was previously unreadable in chat anyway.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$23.08`_

### DIFF
--- a/src/common/utils/errors.test.ts
+++ b/src/common/utils/errors.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "bun:test";
-import { getErrorMessage } from "./errors";
+import { clampErrorMessage, getErrorMessage } from "./errors";
 
 describe("getErrorMessage", () => {
   it("returns string representation of non-Error values", () => {
@@ -110,5 +110,29 @@ describe("getErrorMessage", () => {
     const err = new Error("self");
     err.cause = err;
     expect(getErrorMessage(err)).toBe("self");
+  });
+});
+
+describe("clampErrorMessage", () => {
+  it("returns short messages unchanged", () => {
+    expect(clampErrorMessage("boom", 100)).toBe("boom");
+  });
+
+  it("clamps oversized messages while keeping head and tail", () => {
+    const head = "Invalid prompt: schema mismatch. ";
+    const tail = " Error message: expected string, received undefined";
+    const message = head + "x".repeat(100_000) + tail;
+
+    const clamped = clampErrorMessage(message, 1_000);
+
+    expect(clamped.length).toBeLessThan(1_100);
+    expect(clamped.startsWith(head)).toBe(true);
+    expect(clamped.endsWith(tail)).toBe(true);
+    expect(clamped).toContain("chars omitted");
+  });
+
+  it("clamps at the default bound", () => {
+    const clamped = clampErrorMessage("y".repeat(500_000));
+    expect(clamped.length).toBeLessThan(10_000);
   });
 });
--- a/src/common/utils/errors.ts
+++ b/src/common/utils/errors.ts
@@ -1,3 +1,5 @@
+import { ERROR_MESSAGE_CLAMP_MAX_CHARS } from "@/constants/errors";
+
 /**
  * Extract a string message from an unknown error value.
  * Handles Error objects and other thrown values consistently.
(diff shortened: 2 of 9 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Addressed all 5 findings in 8a64ae7c0:
- **P1 clamp constant**: moved `ERROR_MESSAGE_CLAMP_MAX_CHARS` to `src/constants/errors.ts`.
- **Sparse arrays**: fallback now index-loops (holes and throwing index getters become `null`, matching JSON.stringify), with a regression test.
- **Refusal fallback clamp**: clamp moved into `persistStreamError()`, the single choke point every payload (including the ModelRefusalError early return) passes through before persist/emit.
- **toJSON guard**: the `toJSON` property is read once inside a try/catch and invoked via `.call(value)`; a throwing getter degrades to "no toJSON", with a regression test.
- **__proto__**: fallback objects are built via `Object.fromEntries`, so a literal `__proto__` key stays an own data property; regression test asserts prototype integrity.

ibetitsmike:
@codex review

Addressed in 7f9b31d9d: the fallback's try/finally is now try/catch/finally, so reflective failures on the current value (throwing ownKeys/length traps, revoked proxies) degrade to an "[Unserializable]" placeholder leaf instead of rejecting the whole normalization. Regression tests cover a throwing-ownKeys Proxy and a revoked proxy.

ibetitsmike:
@codex review

Addressed the BigInt host-DoS finding in 4e7055eba: BigInt magnitude is now gated by an O(1) comparison against 10^4096 (precomputed limit, comparison short-circuits on digit-count mismatch) before any toString() call; oversized values become the bounded placeholder "[BigInt: >4096 digits]". Boundary regression tests cover 10^4096 (placeholder, both signs) and 10^4095 (exact conversion).

On the broader "traversal and output budgets" suggestion: the fallback's traversal is linear over the same graph the primary JSON.stringify path already walks for every tool result, and code_execution separately bounds oversized return payloads at the kernel layer (vars handles), so the BigInt expansion was the only cost vector unique to this fallback.

PR #2413, opened 2026-02-13
🤖 feat: incremental workspace.onChat subscription (cursor + live-only)

ThomasK33:
## Summary

Add incremental `workspace.onChat` subscription support so clients can reconnect with a cursor instead of replaying full history. This eliminates redundant data transfer after transient disconnects and enables live-only subscriptions.

## Background

Today `workspace.onChat` always does a full history replay on subscribe/reconnect — re-sending all messages from `chat.jsonl`, replaying active streams from scratch, and replaying init state. For browser/remote connections, this resends a lot of redundant data after transient disconnects. This PR adds three subscription modes:

1. **`full`** (default): current behavior, unchanged.
2. **`since`** (with cursor): replay only messages newer than the client's cursor, plus incremental stream replay filtered by timestamp.
3. **`live`**: skip all history replay; receive only future events.

### Shared schemas (`src/common/orpc/schemas/stream.ts`, `api.ts`, `types.ts`)
- New `OnChatHistoryCursorSchema`, `OnChatStreamCursorSchema`, `OnChatCursorSchema`, `OnChatModeSchema`
- `CaughtUpMessageSchema` extended with optional `replay` (enum) and `cursor` fields for backward compatibility
- `workspace.onChat.input` now accepts optional `mode` parameter

### Backend (`router.ts`, `agentSession.ts`, `streamManager.ts`, `aiService.ts`)
- Router threads `input.mode` to `session.replayHistory()`
- `AgentSession.emitHistoricalEvents()` implements mode-aware replay:
  - **live**: skips all history, stream, and init replay
  - **since**: validates cursor against persisted history (`id + historySequence`), filters messages, and passes `afterTimestamp` for incremental stream replay
  - Invalid cursors fall back to full replay
- `caught-up` event now always includes `{ replay, cursor }` — the server declares which strategy it actually used
- `StreamManager.replayStream()` accepts optional `afterTimestamp` to filter stream parts while always emitting `stream-start`
- Init replay only runs for `replay === "full"`

### Frontend (`StreamingMessageAggregator.ts`, `WorkspaceStore.ts`)
- `StreamingMessageAggregator`:
  - New `getOnChatCursor()` method computes cursor from max `historySequence` + active stream state
  - `loadHistoricalMessages()` supports `{ mode: "append" }` for incremental loads (skips clearing)
  - `handleStreamStart()` is replay-idempotent: won't wipe existing parts when `data.replay === true`
- `WorkspaceStore`:
[...]

PR #3979, opened 2026-08-26
🤖 fix: stop silent full-history re-transfers on workspace switch-back (onChat since cursor)

ThomasK33:
## Summary

Fixes the silent full-history re-transfer over the WebSocket on nearly every workspace switch-back: the client now reuses the **server-issued** onChat reconnect cursor verbatim instead of recomputing the cursor fingerprint from its divergent in-memory representation, and since-mode replays reconcile the transcript suffix authoritatively. Adds permanent observability so since→full downgrades can never hide again.

## Background

`workspace.onChat` supports incremental reconnects (`since` mode, #2413), but a representation mismatch silently defeated it:

- The server persists **one part per streaming delta**; the client **compacts parts on append** and keeps its own assembled parts (plus client-side timestamps) at stream-end.
- The since-cursor fingerprint (FNV-1a over `JSON.stringify(message.parts)` + metadata of every row below the anchor) was recomputed client-side from that divergent representation.
- Whenever a live-assembled assistant row sat *below* the cursor anchor (≥2 turns completed while watching), fingerprints diverged → the server silently downgraded to a **full active-epoch replay** (median ~400 KB, p90 ~1.7 MB, pathological 161 MB) → full wire transfer + full aggregator rebuild. Nothing logged the downgrade.

## Implementation

- **Phase 0 (observability):** `agentSession.emitHistoricalEvents` classifies every since→full downgrade by the first failing predicate (`cursor-row-missing | oldest-mismatch | fingerprint-mismatch | history-read-failed`), logs one `onChat replay` line per replay with row counts, and carries `downgradeReason` on the `caught-up` payload (additive optional schema field; upgrade/downgrade safe).
- **Phase 1 (fix):** `StreamingMessageAggregator` stores the server-issued history cursor from each caught-up (`setServerHistoryCursor`) and `getOnChatCursor()` returns it verbatim; client-side fingerprint computation is deleted. Because the stored anchor can be older than the client's cached transcript, since-replay handling treats replayed rows as the **authoritative suffix** for `seq >= anchor` (`reconcileSinceReplay`): stale rows are deleted, rewritten rows replaced by persisted forms (even with fewer parts), optimistic rows (no `historySequence`) untouched, and the server-confirmed active stream keeps the richer local assembly via an attempt-scope-gated carve-out. Derived state (todos, agent status, skills, review pins) is deterministically rebuilt from the replay window whenever a discarded row could have contributed to it.
- **WorkspaceStore:** the requested anchor is attempt-scoped and routed through the same abort-guarded event path as chat events; a since caught-up without a usable anchor triggers a hard fallback (clear stored cursor + replay buffers, abort → full resubscribe).
- No wire-protocol changes; `OnChatModeSchema` untouched; server validation below the anchor unchanged (fail-safe).

**Accepted trade-off:** rows persisted during the current viewing session sit above the stored anchor and are re-transferred once on the next resubscribe (bounded: a few turns; the anchor ratchets forward on each caught-up — measured steady state re-sends only the anchor row).

## Validation

**Red-first contract test** (`agentSession.sinceReplayContract.test.ts`, real server replay ⇄ real aggregator). On pre-fix code:

```
error: expect(received).toBeUndefined()
Received: "fingerprint-mismatch"
(fail) onChat since-mode reconnect cursor contract > reconnects incrementally after two live-streamed turns
```

Green with the fix (replay stays `since`, no downgrade, only rows ≥ anchor re-sent, transcript converges exactly to persisted state).

**Live dogfooding** (isolated `dev-server-sandbox`, real provider turns, driven with agent-browser; `XUM_LOG_LEVEL=debug`):

| Scenario | Before (Phase-0-only build) | After (this branch) |
|---|---|---|
| 2 live turns → switch away/back | `requestedMode=since replayMode=full downgradeReason=fingerprint-mismatch sentRowCount=4/4` | `replayMode=since`, no downgrade |
| Second switch-back | always full | `since`, `sentRowCount=1/4` (anchor ratchet) |
| Mid-stream switch-back | — | `since`; stream continues, carve-out preserves assembled content |
| Interrupt (Stop) → switch-back | — | `since`; interrupted marker intact |
| 10× rapid A↔B toggle | — | every reconnect `since` `sentRowCount=1`; transcript intact; no console errors |
| `/compact` → switch-back | — | one logged `cursor-row-missing` full replay of the tiny fresh epoch, then `since` again |

Zero downgrades in the entire after-fix session log. Suites: `StreamingMessageAggregator.test.ts`, `WorkspaceStore.test.ts`, `agentSession.preStreamError.test.ts`, `agentSession.sinceReplayContract.test.ts`, `onChatCursorFingerprint.test.ts` — 303 tests green; `make static-check` green (re-run after rebase onto `main`).

## Risks

- **Reconciliation deletes local rows** (`seq >= anchor` not re-sent). This is the new authority contract; wrong anchors would drop visible rows. Mitigated by attempt-scoping the anchor through the abort-guarded event path, a hard full-resubscribe fallback when the anchor is missing, and dedicated ghost-removal/boundary-dedup/optimistic-row unit tests. Severity if wrong: visible transcript corruption in the chat pane until a full replay.
- **Active-stream carve-out** could preserve stale content if mis-gated; gated on server-confirmed stream id == the attempt cursor's stream id, with tests for match/mismatch/stream-end-right-after-caught-up.
- **Derived-state rebuild** resets todos/status/skills/review pins when a suffix source row was discarded; scoped to the server replay window so `loadOlderHistory` pages stay excluded. Worst case: todo/status UI briefly reflects only surviving rows (matches persisted reality).
- Cursor staleness below the anchor (edits/truncation/compaction while away) keeps today's fail-safe behavior: server validation downgrades to full replay — now logged and observable.

---

<details>

# Fix silent full-history re-transfer over the WebSocket on workspace switch

## Problem statement (investigated & empirically confirmed)

`workspace.onChat` already supports incremental reconnects (`since` mode with a history cursor +
fingerprint, PR #2413), but a representation mismatch silently defeats it in the most common usage
pattern, causing **full active-epoch re-transfers on nearly every workspace switch-back**:

- **Server persists one part per streaming delta** (`streamManager.ts` `~3042-3048`: "Append each
  delta as a new part (merging happens at display time)"; persisted verbatim at `~3590-3599` via
  `updateHistory`). Real rows on disk have 5,000+ parts.
- **Client compacts parts on append** (`StreamingMessageAggregator.handleStreamDelta` `~2095-2110`,
  `handleReasoningDelta` `~2811+`) and keeps its own assembled parts at stream-end (only tool
  outputs adopted from `data.parts`, `~2146-2166`). Client assistant `metadata.timestamp` is a
  client-side `Date.now()` (`~2064-2066`), while the persisted row carries server metadata.
- The since-cursor fingerprint is `FNV-1a over JSON.stringify(message.parts)` + metadata of every
  row **below** the cursor anchor (`src/common/orpc/onChatCursorFingerprint.ts`). The client
  computes it from its in-memory representation (`StreamingMessageAggregator.getOnChatCursor()`
  `~1324-1395`); the server recomputes from persisted rows (`agentSession.ts` `~2568-2626`).
- **Result:** whenever a live-assembled assistant row sits *below* the cursor anchor (i.e. ≥2 turns
  completed while watching, or 1 turn + any later row such as a queued send/sub-agent report), the
  fingerprints diverge → the server **silently downgrades to full replay** → the client does
  `loadHistoricalMessages(mode: "replace")` (`WorkspaceStore.ts:4531`) → full wire transfer + full
  aggregator rebuild. Nothing logs the downgrade.

**Empirical proof** (real production modules driven in a scratch script, this machine):

| Scenario | Client fp | Server fp | Outcome |
|---|---|---|---|
| 2 live turns → switch away → back | `d2697240` | `cd3859e7` | MISMATCH → silent full replay |
| 1 live turn (control) | `0167240c` | `0167240c` | MATCH → since works (boundary row is re-sent `>=` in persisted form) |

**Measured payload cost** (1,545 sessions on this host): full replay = whole active epoch in one
shot (`getHistoryFromLatestBoundary(skip=0)`, `agentSession.ts:2541`; no within-epoch wire
pagination). Median epoch ~400 KB, p90 ~1.7 MB, top workspaces 6–23 MB, pathological 161 MB
(43k-row loop workspace with no compaction boundary). Tool outputs ≈60% of bytes, JSON overhead
≈25–30% (largely per-delta parts bloat).

## Goals

1. Make since-mode reconnects actually work after live-streamed turns (the common case).
2. Add observability so since→full downgrades can never hide again.
3. Regression-test the client↔server cursor contract end-to-end (red-first: the new contract test
   must fail on current `main`).

## Non-goals (documented follow-ups, separate PRs — see "Follow-up roadmap")

- Server-side part compaction at persist time (disk/wire JSON bloat).
- Within-epoch pagination of full replay (first paint = newest N rows).
- Runaway boundary-less workspaces (the 161 MB case).
- Frontend render-cost work (RightSidebar remount, markdown re-parse, aggregator LRU).

## Chosen design: the client reuses the **server-issued cursor** verbatim

**Key insight:** the server already computes and sends an authoritative cursor (anchor +
`oldestHistorySequence` + `priorHistoryFingerprint` over *persisted* rows) in every `caught-up`
event (`agentSession.ts` `~2669-2688` builds `serverCursor` for both full and since replays;
`CaughtUpMessageSchema.cursor`). The client currently uses only fragments of it
(`WorkspaceStore.ts:4468-4479`) and **recomputes** the fingerprint from its own divergent in-memory
representation. That recomputation is the root cause.

**Fix:** stop client-side fingerprint recomputation entirely. The aggregator stores the last
server-issued history cursor and `getOnChatCursor()` returns it verbatim (plus the existing
client-computed *stream* cursor, which is advisory and clamped server-side at
`agentSession.ts:2605-2615`).

**Required invariant — suffix reconciliation (closes the stale-row hole):** the server-issued
anchor can be *older* than the client's cached transcript (rows streamed live since the last
caught-up sit above it). Server validation (anchor row exists, epoch-oldest match, fingerprint)
only protects rows **below** the anchor. Rows in `[anchor, client-max]` could be deleted or
rewritten server-side while the client is unsubscribed (e.g. auto-retry rewrites, background
truncation), and plain append semantics would retain them as stale ghosts. Today's design is
implicitly safe here only because the anchor always equals client-max. Therefore, since-replay
handling must treat the replayed rows as the **authoritative suffix for `seq >= requestedAnchor`**
— see Phase 1, "since-replay suffix reconciliation".

Why this is correct and robust:

- **Representation-independent.** Client memory layout (compact-on-append, adopted tool outputs,
  client timestamps) no longer participates in the contract. The client keeps its
  memory-efficient compact parts.
- **Fail-safe below the anchor.** The server validates anchor row existence, epoch-oldest match,
  and prior-rows fingerprint on every subscribe (`agentSession.ts:2568-2603`) — a stale stored
  cursor (edit/truncation/compaction while away) degrades to full replay, which is today's
  behavior.
- **Authoritative at/above the anchor.** Since-replay re-sends every current row `>=` anchor in
  persisted form; with suffix reconciliation the client converges exactly to persisted state
  (stale rows removed, rewritten rows replaced, live-assembled rows replaced by canonical forms).
- **Covers all divergence sources at once** — text/reasoning delta compaction, reasoning
  `providerOptions`/`signature` stashed only server-side (`streamManager.ts:3067-3110`),
  client-generated `metadata.timestamp`, and the abort/error paths (`stream-abort` carries no
  parts, so client/persisted representations diverge for interrupted turns too).

**Accepted trade-off:** rows persisted *during* the current viewing session sit above the stored
anchor and are re-transferred once on the next resubscribe (bounded: a few turns, typically tens
of KB; vs. multi-MB full epochs today). A cursor-ratchet optimization is listed as follow-up 3d.

### Rejected alternatives

| Approach | Why rejected | est. net LoC |
|---|---|---|
| **A. Client adopts server parts at stream-end** (mirror reconnect path `~2204-2224`) | Doesn't cover `stream-abort`/`stream-error` (no parts on those events); `metadata.timestamp` still client-generated → still mismatches; regresses client memory (per-delta parts retained); keeps the fragile byte-equality contract | +50–100 |
| **B. Server-side part compaction at persist time only** | Insufficient alone: client parts still lack reasoning `providerOptions`/`signature` and carry client timestamps → still mismatches. Valuable independently for the 25–30% JSON bloat → follow-up 3a | +80–150 |
| **C. Server-computed per-row hashes carried on every terminal event; client hashes-of-hashes** | Structurally sound but strictly more machinery (new schema fields on stream-end/abort/error/message events, client hash bookkeeping) for the same outcome as the chosen design | +150–250 |

**Chosen design est. net LoC (product code): +120 to +250** (cursor storage/reuse ~30; suffix
reconciliation + derived-state rebuild ~80–160; attempt-scoped anchor plumbing + full-resubscribe
fallback ~30–60; minus ~40 of removed client-side fingerprint computation).

---

## Phase 0 — Observability first (land before the fix; est. +40–70 LoC)

Purpose: prove the bug live in dogfooding "before" evidence, and permanently guard against silent
downgrades.

Changes:

1. `src/node/services/agentSession.ts` (`emitHistoricalEvents`, `~2546-2638`):
   - Classify the downgrade reason when `mode.type === "since"` fails validation:
     `"cursor-row-missing" | "oldest-mismatch" | "fingerprint-mismatch" | "history-read-failed"`.
     The existing checks (`matchedHistoryCursor`, `oldestHistoryMatches`, `priorHistoryMatches`)
     already compute each predicate — derive the reason from the first failing one.
   - `log.debug` one line per replay with `{ workspaceId, requestedMode, replayMode,
     downgradeReason, epochRowCount, sentRowCount }`. Row counts only — no `JSON.stringify` byte
     accounting on the hot path.
2. `src/common/orpc/schemas/stream.ts` (`CaughtUpMessageSchema`): add optional
   `downgradeReason: z.enum([...]).optional()`, populated only when a requested `since` was
   downgraded. Enables behavioral tests + client-side visibility without log scraping.
   (Additive optional field → upgrade/downgrade safe.)
3. `src/browser/stores/WorkspaceStore.ts` (caught-up handler `~4461`): `console.debug` the replay
   mode + downgrade reason (dev observability; no UI).

Tests (behavioral, not tautological): reason **classification branching** — given a cursor whose
anchor row is missing vs. whose fingerprint mismatches vs. whose oldest mismatches, the caught-up
event carries the matching reason and `replay: "full"`. Extend the existing agentSession replay
test suite.

**Gate 0:** targeted tests green; sandbox dogfood shows `fingerprint-mismatch` downgrades on
unfixed code (capture logs/screenshots as "before" evidence).

## Phase 1 — Authoritative server cursor reuse + suffix reconciliation (the fix; est. +120 to +250 LoC)

**Step 0 (red-first):** write the round-trip contract test from Phase 2 item 1 **after Phase 0
has landed but before implementing Phase 1** (the test asserts `downgradeReason`, which Phase 0
introduces; on raw `main` the equivalent red assertion is simply `replay !== "since"`). Confirm it
fails on the Phase-0-only code with `replay: "full"`, `downgradeReason: "fingerprint-mismatch"`.
Write the test against the **existing black-box surfaces** (server: `replayHistory` with a
`since` mode input; client: the current `getOnChatCursor()` output after replay + live events) so
it compiles before any Phase 1 API additions land. Record the red run in the PR description.

Changes:

1. `src/browser/utils/messages/StreamingMessageAggregator.ts`:
   - New field `lastServerHistoryCursor: OnChatHistoryCursor | null = null` + setter
     `setServerHistoryCursor(cursor: OnChatHistoryCursor | null)`.
   - Rewrite `getOnChatCursor()` (`~1324-1395`): return
     `{ history: this.lastServerHistoryCursor ?? undefined, stream: <existing stream-cursor logic> }`;
     return `undefined` when there is no stored history cursor (→ full mode, as today for fresh
     aggregators). Keep the defensive `activeStreams.size > 1 → undefined` guard.
   - Delete the client-side `computePriorHistoryFingerprint` usage and the min/max sequence scan;
     drop the now-unused browser import (the function itself stays in `src/common/orpc/` — the
     server still uses it).
   - **Keep `establishedOldestHistorySequence` as-is** for now (it interacts with
     `loadOlderHistory` pagination); defer any removal to a follow-up once suffix
     reconciliation/pagination behavior has settled in production.
   - **New: since-replay suffix reconciliation.** Dedicated method, e.g.
     `reconcileSinceReplay(requestedAnchorSeq, replayedMessages, preservedActiveStreamMessageId)`:
     1. Delete every local message with a **defined** `historySequence >= requestedAnchorSeq`
        whose id is **not** in the replayed set and is **not** the preserved active-stream
        message. Rows **without** a `historySequence` (optimistic sends awaiting ack) are never
        touched.
     2. Apply replayed rows with **replace-by-id semantics** (persisted rows are authoritative for
        finalized turns — do *not* route them through `addMessage`'s "prefer richer content" guard,
        which could keep stale local content when a rewrite has fewer parts) — **except** for the
        preserved active-stream message, which keeps the richer local assembly: the persisted
        placeholder row has empty/partial parts and stream replay only re-sends deltas after the
        stream cursor, so hard-replacing it would lose already-rendered content. (This carve-out
        is exactly what the existing `addMessage` guard at `~1116-1125` protects.)
     3. **Active-stream carve-out gating:** preserve the active-stream row **only when the
        server-confirmed active stream (`caught-up.cursor.stream.messageId`) matches the locally
        active stream that existed when this attempt's cursor was built**. On mismatch or when no
        local active stream exists, apply no carve-out — replayed rows + subsequently replayed
        stream events rebuild the stream message (the existing stale-stream cleanup at
        `WorkspaceStore.ts:4487-4500` already clears mismatched local contexts before this point).
     4. **Derived-state rebuild:** after the suffix mutation, deterministically rebuild derived
        state from the post-reconcile message set rather than trusting per-delete invalidation.
        Factor the existing replace-mode derived rebuild (`loadHistoricalMessages` replace path
        rebuilds todos, loaded skills/errors, detected links, recency, caches) into a shared
        `rebuildDerivedState()` used by both replace-mode load and reconciliation, preserving
        active-stream transient state where the carve-out applies.
     5. Assert (defensive style): `requestedAnchorSeq` is a non-negative finite number; every
        replayed row with defined seq satisfies `seq >= requestedAnchorSeq`.
2. `src/browser/stores/WorkspaceStore.ts`:
   - `runOnChatSubscription` (`~3904-3922`): **attempt-scope the requested anchor.** Capture
     `{ anchor: mode.cursor.history.historySequence, localActiveStreamMessageId }` in the attempt
     context (alongside the existing `attemptController`), not as a bare shared transient field.
     The existing abort guards (`signal.aborted || attemptSignal.aborted` microtask checks,
     `~3974-3979`, plus `clearReplayBuffers` on switch-away) must provably prevent a caught-up
     from a stale attempt consuming a newer attempt's anchor; route the anchor through the same
     attempt-guarded event path so reconciliation only ever sees its own attempt's anchor.
   - Caught-up handler (`~4461-4540`):
     - `aggregator.setServerHistoryCursor(data.cursor?.history ?? null)` on **every** caught-up
       (full and since). A caught-up without a cursor (server replay failure sets
       `serverCursor = undefined`, `agentSession.ts:2732`) **clears** the stored cursor → next
       subscribe is full (safe).
     - When `replay === "since"`: call `reconcileSinceReplay(...)` with the attempt-scoped anchor
       and carve-out decision, instead of plain `loadHistoricalMessages(mode: "append")`.
     - **Missing-anchor hard fallback:** assert the attempt-scoped anchor exists when the server
       reports since (it always must — the client sent it). If it is somehow missing in
       production: `console.warn`, **clear the stored server cursor, clear the current replay
       buffers (`clearReplayBuffers`) so stale buffered rows cannot carry into the forced full
       replay, and abort/restart the subscription attempt so the next attempt performs a full
       replay** — never append (stale rows could survive, violating suffix authority) and never
       replace-mode load (drops rows below the anchor). The existing deferred-reset behavior (`resetChatStateForReplay` runs
       only after the new iterator is established, `~3953-3958`) keeps the current UI visible
       until the full replay lands.
     - Clear the attempt-scoped anchor after processing each caught-up.
3. No server changes beyond Phase 0. No wire-protocol changes. `OnChatModeSchema` untouched.

Edge cases (all verified against current code):

- **First open / fresh aggregator:** no stored cursor → full replay (unchanged).
- **Empty epoch:** caught-up may carry no history cursor → stored cursor cleared → trivial full.
- **Edit/truncation/compaction below the anchor while away:** server-side validation fails
  (anchor row missing / oldest mismatch / fingerprint mismatch) → full replay. Same as today;
  now logged via Phase 0.
- **Deletion/rewrite at or above the anchor while away:** handled by suffix reconciliation —
  missing rows are removed, rewritten rows replaced by persisted forms (see Phase 1 change 1).
- **Long viewing session (N turns since caught-up):** rows `>=` anchor re-sent once on
  resubscribe; anchor ratchets forward with the new caught-up. Bounded; follow-up 3d can ratchet
  per stream-end if it proves worth it.
- **Active stream at reconnect:** the server-confirmed active-stream message keeps the richer
  local assembly (placeholder rows have empty parts; stream replay only re-sends deltas after the
  stream cursor). If the stream ended while away, the server sends the finalized persisted row
  and no stream cursor → replace-by-id applies and stale stream context is already cleared by the
  existing `caught-up` handling (`WorkspaceStore.ts:4487-4500`).
- **Client pagination (`loadOlderHistory`):** irrelevant to the cursor — the server cursor's
  `oldestHistorySequence`/fingerprint were computed over the epoch (skip=0) server-side, matching
  exactly what server validation compares against.
- **Subscription retry loop (`runOnChatSubscription` attempts):** same path, stored cursor reused;
  a successful since replay refreshes it via its own caught-up.

## Phase 2 — Regression tests

1. **Round-trip contract test (the money test)** — written red-first in Phase 1 Step 0; fails on
   `main`, passes with the fix.
   Location: extend `agentSession` replay tests (or a focused new colocated test file) using
   `createTestHistoryService()` per the HistoryService doctrine:
   - Persist a small history where an assistant row has **per-delta parts** (as `streamManager`
     writes them: multiple `{type:"text",text,timestamp}` parts, server metadata incl. timestamp).
   - Run `replayHistory(listener, {type:"full"})`; feed all emitted events + caught-up into a real
     `StreamingMessageAggregator` (mirroring the `WorkspaceStore` caught-up handling for cursor
     storage).
   - Simulate a second live turn: append the persisted row via the history service **and** feed the
     matching live events (stream-start/deltas/stream-end with per-delta `parts`) to the aggregator.
   - Resubscribe with `mode: {type:"since", cursor: aggregator.getOnChatCursor()}` → assert
     `caught-up.replay === "since"`, no `downgradeReason`, and only rows `>=` anchor were emitted.
2. **Aggregator unit tests** (`StreamingMessageAggregator.test.ts`):
   - `getOnChatCursor()` returns the stored server cursor verbatim; returns `undefined` before any
     caught-up; cleared when `setServerHistoryCursor(null)`. Update the two existing cursor tests
     (`~2185`, `~2583`) to the new semantics.
   - **Suffix reconciliation:** (a) a local row above the anchor absent from the replayed set is
     removed; (b) a rewritten row (same id, different persisted parts — including *fewer* parts
     than the local compacted copy) is replaced by the persisted form; (c) optimistic rows without
     `historySequence` survive; (d) the boundary row is not duplicated; (e) derived state
     regression: derived state sourced from a removed suffix row (e.g. detected links / todo
     state) is gone after reconciliation.
   - **Active-stream carve-out gating:** (a) server and local active stream match → locally
     assembled content preserved; (b) mismatched or absent local active stream → no carve-out,
     stream message rebuilt from replayed rows + replayed stream events; (c) stream ends
     immediately after caught-up → the later `stream-end` finalizes the message correctly.
3. **WorkspaceStore tests** (`WorkspaceStore.test.ts`): existing since-reconnect suites
   (`~5116-5332`) keep passing — including "preserves pagination state across since reconnect
   retries" (`~5116`), which guards the `loadOlderHistory` interaction; add: second subscription
   after caught-up sends `since` with the server-issued cursor (mock asserts the exact cursor
   payload), a caught-up without cursor forces the next subscribe to full, and a since caught-up
   whose replayed set omits a previously-cached row removes that row from the rendered state.
4. **Phase 0 classification tests** as described above.
5. `onChatCursorFingerprint.test.ts` unchanged (server-side use is untouched).

Note: `workspaceService.test.ts` has one pre-existing failure at HEAD in this environment
("accepted history suppresses redelivery…", eager mockRejectedValue) — unrelated; don't chase it.

## Phase 3 — Follow-up roadmap (out of scope for this PR; file as issues or subsequent PRs)

- **3a. Persist-time part compaction** (`streamManager` stream-end/partial finalization): merge
  adjacent text/reasoning delta parts before `updateHistory`, preserving reasoning
  `providerOptions`/`signature` merge semantics. Cuts ~25–30% of epoch bytes on disk and wire.
  Old per-delta rows remain readable (readers already merge at display). Est. +80–150 LoC.
- **3b. Within-epoch replay pagination:** first paint needs only the newest ~64 rows
  (`MAX_DISPLAYED_MESSAGES` limits DOM, not wire); page older rows on demand like cross-boundary
  `history.loadMore`. Defuses the 6–23 MB epochs and the 161 MB pathological case. Est. +200–400.
- **3c. Boundary-less loop workspaces:** periodic forced boundaries or replay caps (mostly
  subsumed by 3b).
- **3d. Cursor ratchet:** server maintains a running FNV fingerprint per session (FNV extends
  sequentially over seq-sorted rows) and emits a refreshed cursor on each stream-end, eliminating
  the bounded re-transfer of the chosen design. Only if dogfooding shows it matters.

---

## Dogfooding & verification (quality gates)

Environment: use the `dev-server-sandbox` skill (isolated `XUM_ROOT` + free ports) so real user
data is untouched; drive the UI with `agent-browser` (snapshot → click/fill → re-snapshot).
Capture screenshots at every step and record a video of the switch sequence; attach via
`attach_file` for review evidence. This workspace's Coder host is headless — `agent-browser`
handles that; do not rely on Storybook for this validation.

**"Before" evidence (after Phase 0, before Phase 1):**
1. Start a sandboxed dev server; create a project + two workspaces A and B.
2. In A, run ≥2 short real turns while watching (provider keys are available in this env; the
   debug CLI `bun run debug send-message <workspace-id>` is a fallback).
3. Switch A → B → A. Grep the sandbox server log for the Phase 0 replay line: expect
   `requestedMode=since replayMode=full downgradeReason=fingerprint-mismatch` and
   `sentRowCount == epochRowCount`.
4. Screenshot the workspaces + save the log excerpt.

**"After" evidence (after Phase 1):**
5. Repeat the same scenario: expect `replayMode=since`, no downgrade reason, `sentRowCount` ≈
   rows since the anchor (small). Confirm the transcript renders identically after switch-back
   (no missing/duplicated rows), including: an interrupted turn (Esc during stream), an edited
   message (truncation → logged full replay once → since again afterwards), and a compaction
   (`/compact`) while watching.
6. Quick-switch stress: rapidly toggle A↔B ~10×; verify no assertion failures in console, no
   transcript corruption, memory stable.
7. Record the video of steps 5–6; attach screenshots + video.

**Validation commands (each phase):** `make static-check` (set `MUX_ESLINT_CONCURRENCY=1` if
memory-constrained) and targeted suites:
`bun test src/browser/utils/messages/StreamingMessageAggregator.test.ts src/browser/stores/WorkspaceStore.test.ts src/node/services/agentSession*.test.ts src/common/orpc/onChatCursorFingerprint.test.ts`.

## Acceptance criteria

1. Switching back to a workspace where ≥2 turns completed live triggers `replay: "since"` (not
   full), verified by the round-trip contract test **and** live dogfooding logs.
2. The contract test fails on unfixed `main` (red-first proof) and passes with the fix.
3. Every since→full downgrade is logged with a machine-readable reason; `caught-up` carries
   `downgradeReason` when a requested since was downgraded.
4. No wire-protocol or persisted-format breaking changes (additive optional schema field only);
   old client ↔ new server and new client ↔ old server both degrade gracefully to full replay.
5. Transcript correctness preserved across switch-back for: normal turns, interrupted turns,
   edits/truncations, compaction, empty epochs (dogfood checklist above).
6. **Suffix authority:** a row deleted or rewritten server-side *above* the stored anchor while
   the client was unsubscribed is respectively removed/replaced after a since reconnect (no stale
   ghosts); verified by the reconciliation unit tests.
7. **Pagination integrity:** `loadOlderHistory` before switch-back neither pollutes the cursor nor
   loses pagination state across a since reconnect; no boundary-row duplication.
8. `make static-check` green; all listed test suites green (modulo the documented pre-existing
   `workspaceService.test.ts` failure).

## Risks & mitigations

- **Stored cursor goes stale below the anchor in ways not anticipated** → server-side validation
  is unchanged and fail-safe there (worst case = today's full replay), now with logging to detect
  any residual downgrade hot spots.
- **Stale rows at/above the anchor** → closed by suffix reconciliation (the central new invariant);
  covered by dedicated unit tests including the delete-above-anchor and rewrite-with-fewer-parts
  cases.
- **Suffix reconciliation wipes active-stream content** → explicit carve-out for the
  server-confirmed active-stream message + a dedicated test; the existing `addMessage` richer-
  content guard remains for that path.
- **Existing tests assume client-computed fingerprints** → update them to server-cursor semantics;
  server-side fingerprint tests are untouched, so safety coverage does not shrink.
- **Behavior change in re-sent rows (`>=` anchor) surfacing duplicate-handling bugs** → since
  replay already re-sends the boundary row today; reconciliation replace-by-id semantics are
  covered by new unit tests and the round-trip contract test.
- **`requestedSinceAnchor` lifecycle bugs (retry loops, aborted attempts)** → anchor is recorded
  per subscription attempt and cleared on each caught-up; asserted present when the server reports
  `replay: "since"`; production fallback is today's append behavior (never replace).

</details>

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$105.52`_

### DIFF
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -5406,6 +5406,129 @@ describe("WorkspaceStore", () => {
       expect(clearedAbortReason).toBe(true);
     });
 
+    it("reuses the server-issued cursor verbatim on the next subscription", async () => {
+      const workspaceId = "cursor-reuse-workspace";
+      const serverCursor = {
+        messageId: "history-7",
+        historySequence: 7,
+        oldestHistorySequence: 3,
+        priorHistoryFingerprint: "0badf00d",
+      };
+      const firstSubscription = createReleaseGate();
+      const requestedModes: unknown[] = [];
+      let subscriptionCount = 0;
+
+      mockOnChat.mockImplementation(async function* (
+        input?: { workspaceId: string; mode?: unknown },
+        _options?: { signal?: AbortSignal }
+      ): AsyncGenerator<WorkspaceChatMessage, void, unknown> {
+        subscriptionCount += 1;
+        requestedModes.push(input?.mode);
+        if (subscriptionCount === 1) {
+          yield createHistoryMessageEvent("history-7", 7);
+          yield caughtUpEvent({ replay: "full", cursor: { history: serverCursor } });
+          await firstSubscription.wait;
+          return;
+        }
+        yield createHistoryMessageEvent("history-7", 7);
+        yield caughtUpEvent({ replay: "since", cursor: { history: serverCursor } });
+      });
+
+      createAndAddWorkspace(store, workspaceId);
+
+      const hydrated = await waitUntil(
+        () => store.getAggregator(workspaceId)?.getAllMessages().length === 1
+      );
+      expect(hydrated).toBe(true);
+      expect(requestedModes[0]).toBeUndefined();
+
+      firstSubscription.release();
+
+      const resubscribed = await waitUntil(() => subscriptionCount >= 2);
+      expect(resubscribed).toBe(true);
+      // The client must not recompute any cursor fields; the server-issued cursor is
+      // reused verbatim (fingerprint included).
+      expect(requestedModes[1]).toEqual({
+        type: "since",
+        cursor: { history: serverCursor },
+      });
+    });
+
+    it("forces a full resubscribe when caught-up carries no cursor", async () => {
+      const workspaceId = "cursor-missing-workspace";
+      const firstSubscription = createReleaseGate();
+      const requestedModes: unknown[] = [];
+      let subscriptionCount = 0;
+
+      mockOnChat.mockImplementation(async function* (
+        input?: { workspaceId: string; mode?: unknown },
+        _options?: { signal?: AbortSignal }
+      ): AsyncGenerator<WorkspaceChatMessage, void, unknown> {
+        subscriptionCount += 1;
+        requestedModes.push(input?.mode);
+        // Replay failures make the server omit the cursor from caught-up.
+        yield createHistoryMessageEvent("history-1", 1);
+        yield caughtUpEvent({ replay: "full" });
+        if (subscriptionCount === 1) {
+          await firstSubscription.wait;
+        }
+      });
+
+      createAndAddWorkspace(store, workspaceId);
+
+      const hydrated = await waitUntil(
+        () => store.getAggregator(workspaceId)?.getAllMessages().length === 1
+      );
+      expect(hydrated).toBe(true);
+
+      firstSubscription.release();
+
+      const resubscribed = await waitUntil(() => subscriptionCount >= 2);
+      expect(resubscribed).toBe(true);
+      expect(requestedModes[1]).toBeUndefined();
+    });
+
+    it("removes rows the server did not re-send from a since replay", async () => {
+      const workspaceId = "since-suffix-ghost-workspace";
+      const firstSubscription = createReleaseGate();
+      const getSubscriptionCount = mockChatReconnectScript((subscriptionCount) =>
+        subscriptionCount === 1
+          ? [
+              createHistoryMessageEvent("history-1", 1),
+              createHistoryMessageEvent("history-2", 2),
+              fullCaughtUpEvent(2, "history-2"),
+              Promise.resolve(),
+              // Row streamed live after caught-up: it sits above the stored anchor.
+              createHistoryMessageEvent("history-3", 3),
+              firstSubscription.wait,
+            ]
+          : [
+              // Server deleted history-3 while the client was unsubscribed, so the
+              // since replay re-sends only the anchor row.
+              createHistoryMessageEvent("history-2", 2),
+              sinceCaughtUpEvent(2, "history-2"),
+            ]
+      );
+
+      createAndAddWorkspace(store, workspaceId);
+
+      const seeded = await waitUntil(
+        () => store.getAggregator(workspaceId)?.getAllMessages().length === 3
+      );
+      expect(seeded).toBe(true);
+
+      firstSubscription.release();
+
+      const ghostRemoved = await waitUntil(() => {
+        const ids = store
+          .getAggregator(workspaceId)
+          ?.getAllMessages()
+          .map((message) => message.id);
+        return getSubscriptionCount() >= 2 && JSON.stringify(ids) === '["history-1","history-2"]';
+      });
+      expect(ghostRemoved).toBe(true);
+    });
+
     it("clears stale auto-retry status when full replay reconnect replaces history", async () => {
       const workspaceId = "task-created-workspace-auto-retry-reset";
       const firstSubscription = createReleaseGate();
--- a/src/browser/stores/WorkspaceStore.test.ts
+++ b/src/browser/stores/WorkspaceStore.test.ts
@@ -488,6 +488,19 @@ const sinceCaughtUpEvent = (
     cursor: { history: { messageId, historySequence }, ...(stream ? { stream } : {}) },
   });
 
+// Full replay caught-up carrying a server cursor, as the real server issues on every
+// non-empty replay. The client reuses this cursor verbatim, making the next
+// subscription a legitimate since request.
+const fullCaughtUpEvent = (
+  historySequence = 1,
+  messageId = `history-${historySequence}`,
+  stream?: { messageId: string; lastTimestamp: number }
+): WorkspaceChatMessage =>
+  caughtUpEvent({
+    replay: "full",
+    cursor: { history: { messageId, historySequence }, ...(stream ? { stream } : {}) },
+  });
+
 const streamEndEvent = (
   workspaceId: string,
   messageId: string,
@@ -1923,12 +1936,24 @@ describe("WorkspaceStore", () => {
         subscriptionCount += 1;
 
         if (subscriptionCount === 1) {
-          yield { type: "caught-up" };
+          // Server issues a reconnect cursor on every caught-up; the client reuses
+          // it verbatim, so the second subscription legitimately requests since.
+          yield {
+            type: "caught-up",
+            cursor: {
+              history: {
+                messageId: "reconnect-pending-start",
+                historySequence: 1,
+              },
+            },
+          };
           await Promise.resolve();
           yield createUserMessageEvent("reconnect-pending-start", "hello", 1, 1_000);
           return;
         }
 
+        // Since replay re-sends the cursor-boundary row before caught-up.
+        yield createUserMessageEvent("reconnect-pending-start", "hello", 1, 1_000);
         yield {
           type: "caught-up",
           replay: "since",
@@ -2035,11 +2060,14 @@ describe("WorkspaceStore", () => {
       const requestedModel = "openai:gpt-4o-mini";
       let subscriptionCount = 0;
 
+      // Empty-history caught-ups carry no cursor, so every resubscribe legitimately
+      // requests (and receives) a full replay; the server never reports since for a
+      // full request.
       mockChatStreamFor(workspaceId, function* () {
         subscriptionCount += 1;
         yield {
           type: "caught-up",
-          replay: subscriptionCount === 1 ? "full" : "since",
+          replay: "full",
         };
       });
 
@@ -5178,13 +5206,14 @@ describe("WorkspaceStore", () => {
       const getSubscriptionCount = mockChatReconnectScript((subscriptionCount) =>
         subscriptionCount === 1
           ? [
-              caughtUpEvent(),
+              createHistoryMessageEvent("history-1", 1),
+              fullCaughtUpEvent(),
               Promise.resolve(),
               bashOutputEvent(workspaceId, "call-bash-4", "stale-output\n"),
               taskCreatedEvent(workspaceId, "call-task-4", "child-workspace-4", 2),
               firstSubscription.wait,
             ]
-          : [sinceCaughtUpEvent()]
+          : [createHistoryMessageEvent("history-1", 1), sinceCaughtUpEvent()]
       );
 
       createAndAddWorkspace(store, workspaceId);
@@ -5216,7 +5245,13 @@ describe("WorkspaceStore", () => {
       const getSubscriptionCount = mockChatReconnectScript((subscriptionCount) =>
         subscriptionCount === 1
           ? [
-              caughtUpEvent(),
+              // Cursor anchored at the row the in-flight stream below will finalize,
+              // so the second subscription legitimately requests since.
+              caughtUpEvent({
+                cursor: {
+                  history: { messageId: "msg-old-stream-missing-local", historySequence: 1 },
+                },
+              }),
               Promise.resolve(),
               streamStartEvent(workspaceId, "msg-old-stream-missing-local", {
                 startTime: 1_000,
@@ -5236,7 +5271,9 @@ describe("WorkspaceStore", () => {
               firstSubscription.wait,
             ]
           : [
-              sinceCaughtUpEvent(1, "history-1", {
+              // Since replay re-sends the cursor-boundary row before caught-up.
+              createHistoryMessageEvent("msg-old-stream-missing-local", 1),
+              sinceCaughtUpEvent(1, "msg-old-stream-missing-local", {
                 messageId: "msg-new-stream-missing-local",
                 lastTimestamp: 2_000,
               }),
@@ -5273,9 +5310,11 @@ describe("WorkspaceStore", () => {
       const getSubscriptionCount = mockChatReconnectScript((subscriptionCount) =>
         subscriptionCount === 1
           ? [
-              caughtUpEvent(),
+              createHistoryMessageEvent("history-1", 1),
+              fullCaughtUpEvent(),
               Promise.resolve(),
               streamStartEvent(workspaceId, "msg-old-stream", {
+                historySequence: 2,
                 startTime: 1_000,
                 model: "claude-3-5-sonnet-20241022",
               }),
@@ -5286,6 +5325,8 @@ describe("WorkspaceStore", () => {
               firstSubscription.wait,
             ]
           : [
+              // Since replay re-sends the cursor-boundary row before caught-up.
+              createHistoryMessageEvent("history-1", 1),
               sinceCaughtUpEvent(1, "history-1", {
                 messageId: "msg-new-stream",
                 lastTimestamp: 2_000,
--- a/src/browser/stores/WorkspaceStore.ts
+++ b/src/browser/stores/WorkspaceStore.ts
@@ -271,6 +271,20 @@ export interface WorkspaceSidebarState {
 type DerivedState = Record<string, number>;
 
 /**
+ * Per-attempt context for an onChat subscription. Carries the since-mode anchor the
+ * attempt's cursor requested (so caught-up reconciliation only ever sees its own
+ * attempt's anchor) plus an abort handle used to force a full resubscribe when a
+ * since caught-up arrives without a usable anchor.
+ */
+interface OnChatAttemptContext {
+  abort: () => void;
+  since?: {
+    requestedAnchorSequence: number;
+    localActiveStreamMessageId: string | undefined;
+  };
+}
+
+/**
  * Usage metadata extracted from API responses (no tokenization).
  * Updates instantly when usage metadata arrives.
  *
@@ -3908,6 +3922,15 @@ export class WorkspaceStore {
         const aggregator = this.aggregators.get(workspaceId);
         let mode: OnChatMode | undefined;
 
+        // Attempt-scoped reconnect context: since-replay reconciliation must only ever
+        // see the anchor from its own attempt's cursor. Routed through the same
+        // attempt-guarded event path as chat events (the abort checks below drop
+        // events from stale attempts), so a caught-up from a previous attempt can
+        // never consume a newer attempt's anchor.
+        const attemptContext: OnChatAttemptContext = {
+          abort: () => attemptController.abort(),
+        };
+
         if (aggregator) {
           const cursor = aggregator.getOnChatCursor();
           if (cursor?.history) {
@@ -3917,6 +3940,10 @@ export class WorkspaceStore {
                 history: cursor.history,
                 stream: cursor.stream,
               },
+            };
+            attemptContext.since = {
+              requestedAnchorSequence: cursor.history.historySequence,
+              localActiveStreamMessageId: cursor.stream?.messageId,
             };
           }
         }
@@ -3979,7 +4006,7 @@ export class WorkspaceStore {
             if (signal.aborted || attemptSignal.aborted) {
               return;
             }
-            this.handleChatMessage(workspaceId, data);
+            this.handleChatMessage(workspaceId, data, attemptContext);
           });
         }
 
@@ -4452,7 +4479,11 @@ export class WorkspaceStore {
     );
   }
 
-  private handleChatMessage(workspaceId: string, data: WorkspaceChatMessage): void {
+  private handleChatMessage(
+    workspaceId: string,
+    data: WorkspaceChatMessage,
+    attemptContext?: OnChatAttemptContext
+  ): void {
     // Aggregator must exist - workspaces are initialized in addWorkspace() before subscriptions run.
     const aggregator = this.assertGet(workspaceId);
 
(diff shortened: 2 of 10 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex security review

PR #3980, opened 2026-08-26
🤖 fix: stop one corrupt persisted row from bricking workspace fetch (onChat)

ibetitsmike:
## Summary

One malformed persisted assistant row permanently bricked `workspace.onChat` for an entire workspace: a `code_execution` part recorded a nested kernel call made with zero arguments, `JSON.stringify` dropped the `input: undefined` key on persist, and the strict `NestedToolCallSchema` then rejected the row during replay output validation — killing the event iterator on every subscription (`Event iterator validation failed`). This PR heals such rows at the schema level, makes replay self-healing against any row that still fails wire validation, and normalizes zero-arg nested calls at write time so new rows can never be persisted without `input`.

## Background

Hit in production (server `v0.28.3-nightly.99`): a workspace's chat could no longer be fetched at all — the replay died after 5 events on a single row (seq 65) containing `mux.linear_get_issue_status()` called with no arguments. The zero-arg call also produced a `Cannot read properties of undefined (reading 'length')` TypeError inside the tool handler at call time, and the live `tool-call-start` event failed the same way mid-stream. One bad line in `chat.jsonl` should never brick a workspace (see AGENTS.md self-healing rules).

## Implementation

Four layers, from data compatibility to prevention:

1. **Schema heal** (`src/common/orpc/schemas/message.ts`, `src/common/types/message.ts`): `NestedToolCallSchema.input` is now optional, so existing rows missing the key validate on read. Consumers treat a missing `input` as `{}`.
2. **Replay self-healing** (`src/node/services/agentSession.ts`): every replayed history row is gated through `safeParse` against the wire schema; a row that fails is skipped and logged instead of propagating a fatal iterator validation error. A single corrupt row now costs that row, not the workspace.
3. **Live-event normalization** (`src/node/services/streamManager.ts`): nested calls with missing args stream and persist as `{}`, so the live `tool-call-start` path and newly persisted `nestedCalls` always carry an input.
4. **Kernel boundary** (`src/node/services/ptc/toolBridge.ts`): a zero-arg capability invocation is treated as `{}`, matching provider semantics — tools return readable validation errors instead of throwing TypeErrors.

## Validation

- New tests: replay skips a corrupt row while delivering the remaining rows plus `caught-up`; legacy nested-call rows without `input` replay successfully; streamManager and toolBridge normalization paths.
- Reproduced the original failure over the wire against a live server, then verified end-to-end (dev-server seeded with a copy of the real broken session) that replay delivers all rows — including the previously fatal 414 KB one — and reaches `caught-up`.
- `make static-check` and the touched test suites (`agentSession.replaySelfHealing`, `streamManager`, `toolBridge`) pass locally.

## Risks

Low. The schema change is a strict loosening (accepts previously rejected rows; all previously valid rows behave identically). The replay `safeParse` gate only alters behavior for rows that would previously hard-fail the whole subscription; worst case a corrupt row is omitted from the transcript instead of bricking fetch, with a log line for diagnosis. Affected area: onChat replay/streaming of nested tool calls.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$29.04`_

### DIFF
--- a/src/common/orpc/schemas/message.ts
+++ b/src/common/orpc/schemas/message.ts
@@ -60,7 +60,12 @@ const MuxToolPartBase = z.object({
 export const NestedToolCallSchema = z.object({
   toolCallId: z.string(),
   toolName: z.string(),
-  input: z.unknown(),
+  // Optional: kernel guests can invoke a capability with zero arguments
+  // (mux.tool()), and JSON.stringify drops the resulting `input: undefined`
+  // key on persist. Requiring the key here made onChat replay of such
+  // persisted rows fail oRPC output validation and permanently brick
+  // workspace fetch (one corrupt row killed the whole subscription).
+  input: z.unknown().optional(),
   output: z.unknown().optional(),
   state: z.enum(["input-available", "output-available", "output-redacted"]),
   failed: z.boolean().optional(),
--- a/src/common/types/message.ts
+++ b/src/common/types/message.ts
@@ -1156,10 +1156,12 @@ export type DisplayedMessage =
       /** Durable workflow run attachment recovered from partial history. */
       workflowRun?: MuxToolPart["workflowRun"];
       // Nested tool calls for code_execution (from PTC streaming or reconstructed from result)
+      // input is optional to mirror NestedToolCallSchema: zero-arg kernel calls
+      // persist without an input key.
       nestedCalls?: Array<{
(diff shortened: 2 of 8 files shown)


ibetitsmike:
@codex review

PR #3981, opened 2026-08-26
🤖 fix(pr-status): resilient PR badge lookup fallbacks

ibetitsmike:
## Problem

The workspace PR badge resolves its PR with an argument-less `gh pr view`, which matches by local branch name only. A PR whose head was pushed with an explicit refspec from a differently named local branch (e.g. an auto-generated worktree branch pushed to `refs/heads/mike/<topic>` with no upstream tracking) is invisible: the lookup returns nothing and the badge stays blank.

## Change

`detectWorkspacePR` now runs a single combined bash script (`DETECT_PR_SCRIPT`, still one `executeBash` round trip, unchanged JSON contract) that falls through three lookups:

1. `gh pr view --json ...` for the current branch (unchanged primary).
2. The tracked merge ref: `git config branch.<cur>.merge`, stripped of `refs/heads/`, skipped when unset or equal to the local name.
3. Commit SHA: `gh api repos/{owner}/{repo}/commits/<HEAD>/pulls` filtered to **open** PRs via gh's built-in `--jq`, resolved back through `gh pr view <number>` so the payload shape stays identical.

The SHA lookup is the only fallback that works with zero local hints after a refspec push without tracking. The open-only filter prevents a false-positive badge on fresh branches whose HEAD is an ancestor of main (the endpoint would otherwise return the merged PR that introduced that commit). Timeout raised 15s -> 30s for the worst-case chain.

## Tests

`PRStatusStore.detectScript.test.ts` executes the exported script with real bash in a temp git repo against a logging `gh` shim: argument-less hit, tracked-ref hit, same-name-ref skip, SHA-fallback hit, and all-fail -> `{"no_pr":true}` with exit 0. `make static-check` and `bun test PRStatusStore` (24 pass) are green.

Dogfood UAT ran remotely via Coder Agents and passed: all four badge scenarios (same-name regression, refspec-push-without-tracking SHA fallback, tracked-ref fallback, fresh-main false-positive guard) verified in the served Xum UI with screenshots, git ground truth, and console captures at this exact head.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$32.26`_

### DIFF
--- a/src/browser/stores/PRStatusStore.detectScript.test.ts
+++ b/src/browser/stores/PRStatusStore.detectScript.test.ts
@@ -1,0 +1,162 @@
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import { chmod, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { DETECT_PR_SCRIPT } from "./PRStatusStore";
+
+const LOCAL_BRANCH = "local-branch";
+const FIELDS =
+  "number,url,state,mergeable,mergeStateStatus,title,isDraft,headRefName,baseRefName,statusCheckRollup";
+const GH_SHIM = [
+  "#!/usr/bin/env bash",
+  `printf '%s\n' "$*" >> "$GH_LOG"`,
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #3991, opened 2026-08-27
🤖 feat: promote Claude Mythos 5.1 as the mythos model

ThomasK33:
## Summary

Promotes Mythos 5.1 (`anthropic:claude-mythos-5-1`) to the `MYTHOS` known model, so the `mythos` alias and model picker route to the new id. Mythos 5 stays usable as the custom model string `anthropic:claude-mythos-5` and keeps its metadata entry and tokenizer approximation. This is the Mythos twin of the Fable 5.1 promotion (#3988) and is stacked on `mike/fable-5-1-drop` (this PR contains only the Mythos-side changes).

Anthropic has announced Fable 5.1 / Mythos 5.1 ([What's new](https://platform.claude.com/docs/en/models/fable-5-1/whats-new-fable-5-1)), so the original merge gate is lifted and all previously-assumed numbers below are verified. The corresponding Fable 5.1 cache-read pricing fix landed on the base branch; this PR applies the same announced price to Mythos 5.1.

## Background

Fable 5 and Mythos 5 shipped together in ba36c818f; #3988 prepped the Fable 5.1 promotion and this PR is the Mythos twin, originally merge-gated as speculative prep and now verified against the announcement.

## Verified against the announcement

- **API id**: `claude-mythos-5-1` ✓ (docs list it as the successor to Mythos 5, Project Glasswing–restricted).
- **Envelope**: native 1M context (default and maximum), 128K max output ✓.
- **Thinking/effort**: adaptive thinking always on, disabled thinking rejected, effort parameter with the full ladder ✓ — matches the existing Mythos-class wildcard matchers and off→low clamp.
- **Tokenizer**: same as Fable 5 (Opus 4.7 tokenizer, unpublished upstream) ✓ — the Opus 4.5 approximation stays.
- **Pricing**: $10/M input, $50/M output, cache write 1.25× ✓ — except cache reads dropped to 0.025× input ($0.25/M) from the assumed 0.1×; applied here for `claude-mythos-5-1` (the `claude-fable-5-1` fix is on the base branch).

### Breaking-changes audit (announcement vs Mux)

- **Forced tool use returns 400** (`tool_choice: any/tool`): Mux never forces tool choice on Anthropic models — the only `toolChoice: "required"` path (`streamManager` first-step forcing) is xAI-only, and `workspaceTitleGenerator` deliberately relies on prompt + `tools.require`/`stopWhen` enforcement instead. No change needed.
- **Thinking-block model binding / prefix-edit invalidation**: enforced only for accounts created on/after Aug 31, 2026, otherwise opt-in via the `thinking-binding-controls-2026-08-01` beta header, which Mux does not send; unreadable blocks on model switches are dropped server-side without error. No change needed for the model drop; adopting the beta controls (and `display: "updates"`, per-message effort, turn-scoped system messages) is follow-up feature work.
- **Refusal fallback**: Mythos remains without a default chain (safeguards lifted; no Mythos key in `DEFAULT_MODEL_FALLBACKS`), so no `defaultModelFallbacksSeededMythos51` migration is needed and the migrations schema is untouched — trivially coexisting with `defaultModelFallbacksSeededFable51` and any sibling flags.

## Implementation

Mirrors #3988 surface-for-surface, adapted to Mythos:

- `knownModels.ts`: `MYTHOS` → `claude-mythos-5-1` (keeps the `mythos` alias; stays unwarmed since most users cannot access it and its tokenizer override is already warmed via FABLE). `LEGACY_TOKENIZER_MODEL_OVERRIDES` retains `anthropic:claude-mythos-5`, exactly like the retired `claude-fable-5` / `claude-opus-4-8` ids.
- `models-extra.ts`: new `claude-mythos-5-1` entry with announced pricing (incl. $0.25/M cache reads); Mythos 5 entry retained.
- `models.ts`: `claude-mythos-5-1` added to the native-1M patterns (anchored; the existing `claude-mythos-5` pattern tolerates date suffixes but not a `-1` minor).
- No wire-format changes for thinking/effort/web-fetch: the Mythos-class wildcard matchers already match the new id; tests pin that.
- Docs model table and built-in skill content regenerated, not hand-edited.
- No WorkflowRunner fixture change: no fixture pins the `mythos` alias.

## Validation

- Behavioral coverage for the new id: native-1M classification (direct + gateway-prefixed), display formatting (`Mythos 5.1`), 5-level thinking policy with off-clamp, provider options (adaptive + summarized display, no disabled thinking), native web-fetch support.
- `make static-check` green; targeted suites (tokens, models, modelDisplay, policy, providerOptions, tools) and `config.test.ts` pass.

## Risks

Low: additive registry/metadata changes. The user-visible effect is that the `mythos` alias routes to Mythos 5.1 (now live at the API) with announced pricing. Existing Mythos 5 selections keep working as custom model strings with unchanged token accounting.

---

### DIFF
--- a/docs/config/models.mdx
+++ b/docs/config/models.mdx
@@ -14,7 +14,7 @@ Xum ships with curated models kept up to date with the frontier. Use any custom 
 | Model                  | ID                            | Aliases                                                      | Default |
 | ---------------------- | ----------------------------- | ------------------------------------------------------------ | ------- |
 | Fable 5.1              | anthropic:claude-fable-5-1    | `fable`                                                      |         |
-| Mythos 5               | anthropic:claude-mythos-5     | `mythos`                                                     |         |
+| Mythos 5.1             | anthropic:claude-mythos-5-1   | `mythos`                                                     |         |
 | Opus 5                 | anthropic:claude-opus-5       | `opus`                                                       | ✓       |
 | Sonnet 5               | anthropic:claude-sonnet-5     | `sonnet`                                                     |         |
 | Haiku 4.5              | anthropic:claude-haiku-4-5    | `haiku`                                                      |         |
--- a/src/common/constants/knownModels.ts
+++ b/src/common/constants/knownModels.ts
@@ -42,15 +42,16 @@ const MODEL_DEFINITIONS = {
     // same text, so real usage can run ~1.0-1.3x higher than this estimate.
     tokenizerOverride: "anthropic/claude-opus-4.5",
   },
-  // Claude Mythos 5 - released June 9, 2026 alongside Fable 5. Same underlying model,
-  // specs, and pricing as Fable 5 ($10/M input, $50/M output) but with safeguards lifted
-  // in some areas. Limited availability: restricted to approved Project Glasswing /
-  // trusted-access customers (no self-serve sign-up). API id `claude-mythos-5`.
+  // Claude Mythos 5.1 - successor to Mythos 5 (released June 9, 2026 alongside Fable 5)
+  // as the restricted-access, safeguards-lifted twin of Fable 5.1, assumed at unchanged
+  // pricing ($10/M input, $50/M output) and availability (approved Project Glasswing /
+  // trusted-access customers, no self-serve sign-up). API id `claude-mythos-5-1`;
+  // Mythos 5 stays usable as the custom model string `anthropic:claude-mythos-5`.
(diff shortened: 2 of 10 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

PR #3993, opened 2026-08-27
🤖 feat: promote Claude Opus 5.1 as the opus model

ThomasK33:
## Summary

Prepares Xum for the Claude Opus 5.1 drop: promotes Opus 5.1 (`anthropic:claude-opus-5-1`) to the `OPUS` known model, so the `opus` alias, the `/opus` command, tokenizer warming, the default model, the mux-gateway first-time model list, and the CI agent workflows all route to the new model. Opus 5 stays usable as the custom model string `anthropic:claude-opus-5` and keeps its metadata entry and tokenizer approximation.

> [!IMPORTANT]
> **Merge gate: hold until Anthropic officially announces Opus 5.1**, then verify the assumptions below against the announcement before merging. Merging early would break `opus`-alias sends (the API id would not exist yet). This PR stays a **draft** until then.

Originally stacked on #3988 (Fable 5.1 promotion); now rebased onto `main` after that PR merged and Fable 5.1 / Mythos 5.1 shipped (Sept 1), which upgraded two of the three assumptions below from pure guesses to precedent-backed guesses.

## Background

Opus 5.1 is **still unannounced**; this PR is prepared ahead of the drop (same play as the Opus 5 promotion in #3750 and the Fable 5.1 promotion in #3988), with assumptions to confirm on release day — all remain provisional until Anthropic's official announcement/docs:

- **API id**: `claude-opus-5-1` (dash form, dateless). Now precedent-backed: Fable 5.1 shipped exactly this convention (`claude-fable-5-1`, `claude-mythos-5-1`), but unconfirmed for Opus.
- **Pricing**: $5/M input, $25/M output, cache write 1.25× input (unchanged from Opus 5), cache read **0.025× input** ($0.125/M) — mirroring the cache-read cut Fable 5.1 / Mythos 5.1 shipped with (they went 0.1× → 0.025×; Opus 5 is currently 0.1×). This is the most speculative number: if Opus 5.1 keeps 0.1× cache reads, `models-extra.ts` needs a one-line fix before merge.
- **Envelope**: unchanged from Opus 5 — native 1M context, 128K max output, full effort ladder with native xhigh and max, thinking on by default but *disable-able* (unlike Mythos-class models, "off" remains a valid level), and no Fable-style safeguard classifiers (so Opus 5.1 stays a refusal-fallback *target*, not a source). Consistent with Fable 5.1 shipping zero wire-format changes, but unconfirmed.

## Implementation

- `knownModels.ts`: `OPUS` -> `claude-opus-5-1` (keeps the `opus` alias, warming, and the Opus 4.5 tokenizer approximation). The retired `anthropic:claude-opus-5` id joins `LEGACY_TOKENIZER_MODEL_OVERRIDES` (alongside main's `claude-fable-5` / `claude-mythos-5` entries) so exact-id lookup keeps its approximation instead of warning and falling back to the generic per-provider tokenizer.
- `models-extra.ts`: new `claude-opus-5-1` pricing/limits entry; the Opus 5 entry is retained.
- `models.ts`: `claude-opus-5-1` added to the native-1M patterns — the existing `claude-opus-5` pattern only tolerates date suffixes, not `-1` (the exact trap #3988 documented for Fable).
- No wire-format changes needed for thinking/effort: `anthropicSupportsNativeXhigh` already matches any Opus 5+ (`claude-opus-[5-9]`, unanchored), `anthropicRejectsDisabledThinking` correctly does NOT match (Mythos-class only), and the native web-fetch parser reads `claude-opus-5-1` as major 5 / minor 1. Tests pin all three plus display formatting (`Opus 5.1`).
- **Fallback target migration** (the Opus-side counterpart of #3988's `defaultModelFallbacksSeededFable51`): promoting `OPUS` moves the *target* of the shipped default chain (Fable 5.1 → Opus 5 becomes Fable 5.1 → Opus 5.1), and already-seeded configs would keep the stale target forever — a key-gap re-seed can't help because no source key changed (and re-running it would resurrect user-deleted chains). A new one-shot flag (`defaultModelFallbacksSeededOpus51`) guards a target migration in `src/node/config/index.ts`: a chain that still deep-equals the superseded shipped default (`{ models: ["anthropic:claude-opus-5"] }`, no enabled/triggers customization, canonical key) moves to the current default exactly once; any deviation — edits, deletions, gateway-prefixed keys — is user intent and is never touched. The migration runs after the key-gap seed in the same load pass, so freshly seeded chains (already on the new target) are no-ops, and it is idempotent if an old build strips the flag.
- **Downgrade byte-stability**: `LEGACY_DEFAULT_MODEL_FALLBACKS`' Fable 5 chain now pins the literal `anthropic:claude-opus-5` target instead of `KNOWN_MODELS.OPUS.id` (which would have silently drifted to Opus 5.1). The legacy chain exists for downgraded pre-5.1 builds whose own shipped default was Fable 5 → Opus 5; pinning keeps the seeded bytes identical regardless of which build ran the original seed pass, and the target migration deliberately skips it.
- Gateway first-time defaults (`providerService.ts`), CI agent workflows (auto-cleanup, terminal-bench), terminal-bench leaderboard `MODEL_METADATA` (Codex round 1), docs model table, `docs/guides/github-actions.mdx`, and built-in skill content regenerated — mirroring #3750's surface set.

## Validation

- Behavioral coverage for the new id: native-1M classification (incl. gateway-prefixed form), 6-level thinking policy with "off" retained, provider options (adaptive + summarized display, native xhigh/max effort), native web-fetch support, display formatting, alias/tokenizer-override registry wiring, and the config migration semantics (exact-default chain migrated once with the legacy chain byte-identical on disk, customized/deleted/gateway-keyed chains untouched, no-op re-run after downgrade, fresh installs carrying all three flags).
- Audited remaining `anthropic:claude-opus-5` fixtures: all use it as an explicit custom model string, which stays valid; default-model-derived fixtures (CLI `--help`, workspace creation, e2e display name) were updated.
- Rebase onto `main` (post-#3988-merge) re-verified: conflicts resolved additively (Mythos 5.1 promotion, gemini 3.8 fixture churn, `config.ts` → `config/index.ts` split) and `make static-check` green. The full unit suite was re-run locally on the pinned Bun in fail-closed chunks; every failure was classified as pre-existing: a host git `init.templateDir` quirk (passes with `GIT_TEMPLATE_DIR` set), Storybook snapshot-budget and BackupRepoCache failures that reproduce identically on clean `origin/main`, and suite-load flakes that pass in isolation. CI on this head is fully green and is the authoritative gate.

## Risks

Low-to-medium: additive registry/metadata changes plus one config migration. The main user-visible effect is that the `opus` alias, `/opus`, and the default model route to Opus 5.1; until the model exists at the API, those sends would fail, which is why the merge gate above matters. The target migration rewrites exactly one chain shape and only when it deep-equals the superseded shipped default — a user who deliberately hand-wrote that exact default shape is indistinguishable from the seed and will be migrated (accepted tradeoff, same class as any default migration). Existing Opus 5 selections and user-edited fallback chains are unaffected. Cost tracking uses the provisional 0.025× cache-read price until the announcement confirms it.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$49.52`_


ThomasK33:
@codex review

ThomasK33:
@codex review

Addressed round 1: registered `anthropic/claude-opus-5-1` in `MODEL_METADATA` (display name "Claude Opus 5.1", folder `Claude-Opus-5.1`) in 2259d651e. Please take another look.

ThomasK33:
@codex review

Rebased onto main after #3988 merged: conflicts resolved (Mythos 5.1, config.ts → config/index.ts split), target migration relocated, and the provisional Opus 5.1 cache-read price now mirrors the shipped Fable 5.1 precedent (0.025× input). Final review round for this draft.

PR #3996, opened 2026-08-28
🤖 refactor: cut the taskService/workspaceService dependency cycle at a typed seam

ibetitsmike:
## Summary

Cuts the TaskService/WorkspaceService concrete-class dependency cycle with one-directional typed ports, then pays for the new contract by deleting the code and test scaffolding the ports make obsolete. The full branch is a net simplification: **+676/−1010 (net −334)** across 11 files.

## Background

TaskService and WorkspaceService imported each other's concrete classes, so neither side had a stated contract: every public member of each class was implicitly reachable from the other, and tests smuggled partial fakes through `as unknown as` casts. An architecture change must shrink the codebase, so the seam had to delete more than it declares.

## Implementation

- `taskWorkspaceSeam.ts` is the single home for the two ports and every type they share:
  - `WorkspaceHost` (~38 methods): what TaskService may call; WorkspaceService `implements` it.
  - `AgentTaskIntegration` (12 methods): what WorkspaceService may call; TaskService `implements` it, wired via `setAgentTaskIntegration` (the old `setTaskService` is deleted).
  - `ArchiveWorkspaceOptions`, `SendMessageInternalOptions` (formerly a 56-line inline literal), `WorkspaceLiveActivity`, `AgentTaskStatus`, and the archive untracked-path helpers moved here instead of being duplicated; `StreamErrorRecoveryOutcome` is imported from its canonical `agentSession` export. The `AgentTaskStatus` re-export shim in taskService is gone; importers use the seam.
- Host call placement is byte-position-identical to main. The call sites are race-hardened sink rechecks and admission-gate interleavings, so placement is behavior; this PR only changes what the callers are allowed to know about each other.
- Invariants moved into types: method-level optional chaining on both ports (`integration?.method?.()`) and the "test harnesses mock a narrow surface" hedges are deleted because the interfaces now guarantee the methods exist.

## Test shrink (where the net-negative lives)

- Typed fakes replace all `as unknown as TaskService/WorkspaceService` casts: `makeAgentTaskIntegrationFake()` (shared testUtils) and a `satisfies WorkspaceHost` central factory.
- Deleted tests the port makes unwritable or that only asserted wiring: a phantom `cleanupReportedDescendantsAfterArchive` guard (the method never existed in production), an archive lock pass-through assertion, and four private `updateAgentStatus` non-invocation spies whose positive behavior tests remain.
- Near-identical siblings collapsed into `test.each` tables (send/resume interrupted-task lifecycle, winding-down gate, auto-resume reset, foreground-wait backgrounding); every case is preserved.
- `createWorkspaceServiceMocks` lost its dead stubs (`waitForIdle`, `deleteWorktree`, `updateAgentStatus` never existed on the port), its hand-written 37-line return annotation, and unconsumed return entries; 23 copy-pasted `createWorkspace` mock blocks collapsed into two shared helpers typed via `Parameters<WorkspaceHost["create"]>`; redundant conditional-spread forwarding in two harnesses became direct pass-through.

## Validation

- `make static-check` green; typecheck is the enforcement mechanism for the seam (it caught two false "unused symbol" audit claims during development).
- Full taskService/workspaceService/heartbeatService suites: 1022 tests, failures limited to the 4 pre-existing flakes reproduced identically on untouched trunk in a clean worktree.
- Structural greps: zero remaining cross-references between the two services; the seam imports neither.

## Risks

The production diff is receiver-renames, type moves, and deleted dead plumbing; runtime control flow is unchanged (one `.bind` indirection replaced by a direct guarded call). Regression risk concentrates in the test consolidation: a `test.each` translation error could weaken a race guard. Mitigated by preserving each original case row and assertion set, and by the suites passing with an unchanged test count on the task side.

---

### DIFF
--- a/src/browser/features/ChatInput/index.tsx
+++ b/src/browser/features/ChatInput/index.tsx
@@ -78,7 +78,8 @@ import {
 import {
   prepareCompactionMessage,
   processSlashCommand,
-  type SlashCommandContext,
+  type CommandAction,
+  type SlashCommandEnv,
 } from "@/browser/utils/chatCommands";
 import {
   addWorkflowRunCardMessageForRun,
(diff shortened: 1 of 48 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

Branch reworked per reviewer directive: the seam now pays for itself. Net diff vs main is +676/−1010 (−334). New commits dedupe seam types into taskWorkspaceSeam.ts, encode port invariants in types instead of optional chaining, and delete seam-obsoleted test scaffolding (phantom-method guard, wiring-only tests, private-spy tests, copy-pasted mock blocks, dead stubs).

PR #4025, opened 2026-08-31
🤖 refactor: convert memory operations to Effect internals behind Promise facades

ThomasK33:
## Summary

Phase 1 of the progressive Effect migration: the memory subsystem's route operations are now Effect-native end-to-end — `Effect.gen` internals, typed domain errors, and all unary `memory.*` oRPC procedures running through the `@orpc/experimental-effect` `handlerGen` bridge — while every pre-Effect caller and existing test keeps working unchanged through thin `Effect.runPromise` facades.

## Background

The Effect + oRPC spike (#4022) landed the bridge (`handlerGen`, `WithEffectContext`, `EffectSchemaToJsonSchemaConverter`), converted `MemoryMetaService` internals, and proved the pattern on a single procedure (`memory.setPinned`). This PR completes the memory subsystem (Phase 1 scope): the remaining `memoryOperations.ts` functions, the remaining `memory.*` router handlers, and promotion of spike-era naming/comments to durable production docs. Providers/gateway/streaming (Phase 2) and background workers (Phase 3) are untouched.

## Implementation

- **`memoryOperations.ts`** — every operation (`list`, `read`, `save`, `delete`, `setPinned`, `consolidationStatus`, `consolidate`) is an `Effect.gen` pipeline exported as `*Effect`, plus a Promise facade with the original name/signature. Failures the client renders stay in the success channel as the unchanged wire `{ success: false }` unions; the Effect error channel carries only typed errors a caller can branch on:
  - `MemoryWorkspaceNotFoundError` (new, `Schema.TaggedError`): raised by the shared `resolveMemoryScope` helper; each route catches the tag and maps it onto its own union shape (plain string vs `MemorySaveError`).
  - `MemoryMetaWriteError` (existing): still mapped to the legacy string error in `setPinned`.
  - Every exported effect is fully handled (`Effect.Effect<A, never>`), so `handlerGen` handlers are one-line delegations.
- **`router.ts`** — all unary memory procedures now run via `handlerGen` (client aborts interrupt the handler fiber; wire contracts unchanged). `memory.onChange` intentionally stays an AsyncGenerator subscription: it returns an event iterator, which `handlerGen` cannot produce — an Effect `Stream` bridge is a later-phase concern (comment in code).
[...]

PR #4027, opened 2026-08-31
🤖 refactor: convert retryManager and muxGatewayOauthService internals to Effect

ThomasK33:
## Summary

Phase 2a of the progressive Effect migration: converts the internals of `RetryManager` and `MuxGatewayOauthService` to Effect while keeping their public APIs and observable behavior byte-identical via thin facades. Both existing test suites pass **unchanged**.

## Background

Follows #4022 (Effect v4-rc + oRPC 1.14 spike, `effectBridge`/`handlerGen`) and #4025 (memory subsystem conversion). Same house pattern: `Effect.gen` internals, `Schema.TaggedError` domain errors where callers branch, `Effect.runPromise` Promise facades so pre-Effect callers (`agentSession`, `taskService`, oRPC routes) stay untouched. Later phases own `providerService`, `providerModelFactory`, and `streamManager`.

## Implementation

**`retryManager.ts`** — the hand-rolled `setTimeout`/`clearTimeout` timer bookkeeping is replaced by a forked retry fiber: `Effect.sleep` for the backoff delay, `Fiber.interrupt` for cancellation. `Effect.runFork` executes synchronously up to the sleep and interruption clears the timer synchronously, so the exact observable ordering (synchronous event emission, timer registration/cancellation) is preserved — verified by the pre-existing tests, which mock global `setTimeout`/`clearTimeout` and assert synchronous sequencing. The generation guard is retained for purely synchronous re-entrancy (a status callback cancelling mid-emission), which fiber interruption alone cannot cover because interrupts only land at yield points.

**On `Schedule`** (informs Phase 2b/2c/3): Effect `Schedule` did **not** fit here and was deliberately not used. `Schedule` drives *the runtime retrying an effect it owns*; `RetryManager` is an externally-driven state machine — attempts are triggered by stream failure/success events arriving from outside, and the "retry" is a fire-and-forget callback whose outcome arrives as a later event. The backoff policy is already a pure shared one-liner (`calculateBackoffDelay`, also used by the frontend types), so expressing it as a `Schedule` would mean effectful stepping machinery to recompute `min(1000·2^n, 60000)`. Schedule remains the right tool where the effect being retried lives inside the Effect world (e.g. Phase 3 background workers/heartbeats).

**`muxGatewayOauthService.ts`** — every fallible pipeline (`getAccountStatus`, `startDesktopFlow`, server/desktop callback exchange, token exchange) is now an `Effect.gen` program with typed errors:

- `MuxGatewaySessionExpiredError` — the one tag a caller genuinely branches on: `getAccountStatus` catches it to clear stored credentials before surfacing the fixed session-expired message.
- `MuxGatewayOAuthError` — everything else; `reason` carries the exact user-facing string, folded back onto the wire `Result<_, string>` by a single `toWireResult` helper at the facades.

The desktop background task (callback → exchange → finish) is now a forked fiber. `Scope` was **not** forced in: flow lifetime (loopback server, timeout, cancellation) is owned by the shared promise-based `OAuthFlowManager` (used by 4 OAuth services), so an acquire/release conversion belongs to a later phase that migrates that shared manager, not to this leaf.

One subtlety: `Effect.tryPromise` thunks around `fetch` are `async` so non-Promise returns (tests' synchronous fetch mocks) coerce exactly like the old `await` did.

## Validation

- `bun test src/node/services/retryManager.test.ts` — 14/14 pass, file untouched.
- `bun test src/node/services/muxGatewayOauthService.test.ts` — 10/10 pass, file untouched.
- `agentSession.preStreamError`, `agentSession.startupAutoRetry`, `streamManager` suites green.
- `taskService.test.ts`: 2 failures reproduce identically on unmodified `main` in this environment (pre-existing baseline, unrelated terminal-recovery mock plumbing); the suite is otherwise green.
- No new tests: the error-channel tags are exercised through the existing exact-error-string assertions; adding tag-only tests would be tautological.

## Risks

Low-to-moderate: `RetryManager` sequencing is intricate (re-entrant cancellation, stale-generation guards) and regressions would affect stream auto-retry UX. Mitigated by the unchanged test suite that pins synchronous ordering, timer counts, and exact delays, plus a probe verifying Effect v4's runtime registers/clears plain `setTimeout` timers synchronously under `runFork`/interrupt. The OAuth service conversion is mechanical: every error string and branch is preserved 1:1.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$9.79`_

### DIFF
--- a/src/node/services/muxGatewayOauthService.ts
+++ b/src/node/services/muxGatewayOauthService.ts
@@ -1,4 +1,22 @@
+/**
+ * Xum Gateway OAuth + balance service.
+ *
+ * Internals are Effect-native: each fallible pipeline is an `Effect.gen`
+ * program whose error channel carries typed domain errors, and the public
+ * Promise methods are thin `Effect.runPromise` facades that fold those errors
+ * back into the wire `Result<_, string>` shape, so pre-Effect callers (oRPC
+ * routes, tests) keep working unchanged. Only `MuxGatewaySessionExpiredError`
+ * gets its own tag because a caller genuinely branches on it (clearing stored
+ * credentials); every other failure is a `MuxGatewayOAuthError` whose `reason`
+ * is already the exact user-facing string.
+ *
+ * Desktop flow bookkeeping (deferreds, loopback server lifetime, timeouts)
+ * stays promise-based in `OAuthFlowManager`, which is shared with the other
+ * OAuth services — converting that lifecycle to Effect `Scope` is deferred
+ * until the shared manager itself migrates.
+ */
 import * as crypto from "crypto";
+import { Effect, Schema } from "effect";
 import type { Result } from "@/common/types/result";
 import { Err, Ok } from "@/common/types/result";
 import {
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex security review

PR #4032, opened 2026-08-31
🤖 refactor: adopt handlerGen as the oRPC router default and convert gateway OAuth procedures

ThomasK33:
## Summary

Final phase of the progressive Effect migration roadmap: makes `handlerGen` the documented default for future oRPC procedures, converts the two remaining router procedures whose backing service is already Effect-native (`muxGateway.getAccountStatus`, `muxGatewayOauth.startDesktopFlow`), and delivers the streamManager placement decision plus a migration completion audit (below). Deliberately progressive, not wholesale: procedures backed by Promise services are left untouched — the convention is to convert the service surface first, never to wrap Promises in Effect at the router.

## Background

Phases 0–3 landed in #4022 (spike/effectBridge), #4025 (memory), #4027 (retryManager + gateway OAuth internals), #4028 (providerService), #4030 (providerModelFactory), #4031 (heartbeat/idle workers on Schedule+Scope). `router.ts` has ~315 handler sites; 11 already rode `handlerGen` (7 `memory.*`, 4 `providers.*` mutations). This PR audits the remaining 300+, converts exactly the ones whose backing pipelines already exist as Effect, and encodes the go-forward convention as a module doc in `router.ts`.

## Implementation

- **`muxGatewayOauthService`**: the Effect pipelines from #4027 were private behind `Effect.runPromise` facades. They are now exposed as wire-shaped public Effect methods (matching the `providerService.setConfigEffect` house pattern):
  - `getAccountStatusEffect()` — left **interruptible**: the balance fetch is a pure read, and the session-expired credential clear is a single best-effort promise that runs to completion even if the fiber is interrupted while awaiting it (JS promises are not cancelled by fiber interruption).
  - `startDesktopFlowEffect()` — wrapped in `Effect.uninterruptible` (mirrors `asAtomicMutation` in providerService): a client abort between loopback-server acquisition and `desktopFlows.register` would otherwise leak the server with nothing left to close it.
  - The Promise facades remain (thin `runPromise` wrappers) so the existing service tests stay byte-identical; the router no longer calls them.
- **`router.ts`**: the two procedures ride `handlerGen`; a module doc codifies the convention (handlerGen default for new unary procedures; plain handlers only for Promise-backed services pending conversion, event-iterator subscriptions, and trivial sync reads; audit abort-atomicity before converting mutations). No lint rule: no cheap existing rule expresses "async handler in this one file is suspect" without flagging the ~280 legitimately deferred sites, and building lint infrastructure is out of scope.

## Conversion audit

[...]

PR #4033, opened 2026-09-01
🤖 refactor: convert OAuthFlowManager flow lifecycle to Effect per-flow Scope

ThomasK33:
## Summary

Phase 5 of the progressive Effect migration (first phase of Wave 2, the wave's gating item deferred since #4027): converts `OAuthFlowManager` internals to Effect with real resource safety. Every registered desktop OAuth flow now owns a per-flow `Scope` whose release finalizers guarantee cleanup (registration-timeout clear, deferred settlement, loopback-server close) on every termination path — finish, cancel, caller-timeout race, duplicate registration, `shutdownAll`, and defects. The Promise-based public API is preserved as thin `Effect.runPromise` facades, so the three not-yet-converted OAuth services (`coderOauthService`, `codexOauthService`, `muxGovernorOauthService`) and all existing tests work unchanged.

## Background

Wave 1 (#4022, #4025, #4027, #4028, #4030, #4031, #4032) established the house pattern: `Effect.gen` internals, thin `runPromise` facades, `handlerGen` for oRPC procedures. #4027 converted `muxGatewayOauthService` but explicitly deferred the shared flow-lifecycle manager: its resources (loopback `http.Server`, registration `setTimeout`, result deferred) were cleaned up via ad-hoc `try/catch` + fire-and-forget `void closeServer(...)`, and a defect while resolving the deferred silently skipped the server close. This PR is the acquire/release case that deferral pointed at, and unblocks Phase 6 (batch conversion of the sibling OAuth services).

## Implementation

**Per-flow Scope design** — `register` creates a `Scope.makeUnsafe()` per flow and moves ownership of the caller-acquired resources into it via one `Effect.acquireRelease` per resource (a combined acquisition would install its finalizer only after every step succeeded, leaking earlier resources on a later defect — the #4031 Codex P2 lesson). Release runs in reverse acquisition order, preserving the pre-Effect `finish` ordering: clear registration timeout → settle deferred (waiters unblock before the async close) → close loopback server (awaited).

**Deferred settlement via finalizer** — each `ActiveFlow` carries a mutable `finalResult` staged by the terminating path (finish/cancel/shutdown/replace); the settle finalizer resolves the caller's deferred with it. Settlement is therefore scope-guaranteed rather than an ad-hoc `resolve` call, with a defensive fallback result so waiters can never hang.

**Caller-facing timeout race** — `waitFor` maps to `Effect.timeout` over `Effect.promise` on the shared deferred: the local wait timer is fiber-managed (interruption clears it), stays separate from the registration-time timeout, and on any error result runs `finish` for shared cleanup. The cleanup's synchronous bookkeeping (map removal, completed-result recording) runs before `waitFor` resolves — exact parity with the old sync prefix — while the async release runs in an `Effect.forkDetach` fiber, replacing the old `void this.finish(...)` fire-and-forget with a supervised fiber that survives the caller's completion (verified by a live-runtime probe: detached fibers outlive the parent, `runFork`/`runPromise` execute synchronously to first suspension, and a throwing finalizer does not skip its siblings).

**shutdownAll contract** — preserved as async (`Promise<void>` facade): `serviceContainer.dispose` awaits it, and loopback-server closes are bounded by the server's force-finish socket handling. It never rejects; release defects are caught (`Effect.catchDefect`) and logged at debug level, per the startup/shutdown-must-never-crash rule.

**Effect-native surface** — `waitForEffect` / `cancelEffect` / `finishEffect` / `cancelAllEffect` / `shutdownAllEffect` are public (wire-shaped, never-failing — same shape as #4032's Effect surfaces). `muxGatewayOauthService`'s Effect pipeline now yields `finishEffect` directly instead of `Effect.promise(() => …finish(...))`, and its registration-timeout callback uses `Effect.runFork(finishEffect(...))` instead of `void finish(...)`.

**Not converted to Effect `Deferred`** — the result deferred's identity is part of the public caller-owned `OAuthFlowEntry` (the three unconverted services construct entries with `createDeferred`), so swapping it would break the "existing callers unchanged" contract; revisit when Phase 6 converts entry construction.

## Validation

- All 18 pre-existing `oauthFlowManager` tests pass byte-identical, plus all OAuth service suites (194 tests: coder/codex/muxGateway/muxGovernor/mcp/copilot/codexOauthAuth) and loopback-server/oauthUtils suites.
- Two new behavioral tests for the genuinely-new guarantees: (1) server close + timeout clear still happen when the deferred `resolve` throws (the pre-Effect code skipped the close — this test fails on the old implementation), and (2) the detached cleanup fiber completes after `waitFor` has already returned on the timeout path (guards against accidental child-fiber supervision, where the release would be interrupted with the caller).
- A standalone Effect v4 runtime probe validated the semantics the design relies on (finalizer independence under defects, reverse sequential release order, eager sync-prefix execution of `runPromise`/`runFork`, `forkDetach` outliving the parent, `Effect.timeout` + `Effect.catch` over `Effect.promise`).
- `make static-check` green.

## Risks

Low-to-moderate: this is shared lifecycle code under four OAuth login flows (Gateway, Governor, Codex, Coder). The public API, observable ordering (map removal before `finish` resolves, deferred settlement before server close, synchronous `register`), and error strings are preserved exactly; regressions would surface as leaked loopback listeners, hung `waitFor` calls, or unsettled deferreds — all covered by the existing + new suites.

## Lessons for Phase 6

Phase 6 is the batch conversion of `coderOauthService`, `codexOauthService`, `muxGovernorOauthService`, `copilotOauthService`, plus their ~20 router sites. Notes to make it mechanical:

- The manager now exposes never-failing, wire-shaped `waitForEffect`/`cancelEffect`/`finishEffect`/`shutdownAllEffect`, so converted service pipelines can yield them directly (see `desktopCallbackPipeline` in `muxGatewayOauthService` as the template), and registration-timeout callbacks should use `Effect.runFork(manager.finishEffect(...))`.
- `beginFinish`'s sync-bookkeeping/async-release split is the pattern to reach for wherever a service needs "unregister now, release in background" semantics.
- Each sibling's `startDesktopFlow` should become uninterruptible like the gateway's (#4032): a client abort between loopback acquisition and `register` would otherwise leak the server.
- `coderOauthService` is the outlier: it has extra commit-path liveness checks (`has`) and multi-step persist/commit finish calls (~10 `desktopFlows.*` sites vs ~5 in the others) — expect most of the Phase 6 effort there.
- **Recommendation: two PRs.** PR A: codex + governor + copilot service internals (near-identical DesktopFlow shape, mechanical) together with their router procedures moving to `handlerGen` (the `waitFor`/`cancel` handlers for the gateway can join here — the router comment at `muxGatewayOauth` already points at this). PR B: `coderOauthService` alone — its commit/persist liveness semantics deserve isolated review, and a combined PR would bury it under the mechanical churn.
---

_Generated with [`mux`](https://github.com/coder/mux) • Model: `anthropic:claude-fable-5` • Thinking: `xhigh`_

### DIFF
--- a/src/node/orpc/router.ts
+++ b/src/node/orpc/router.ts
@@ -579,7 +579,7 @@ export const router = (authToken?: string) => {
       // startDesktopFlow rides handlerGen; its service pipeline is
       // uninterruptible (see startDesktopFlowEffect) so a client abort cannot
       // leak the loopback server. waitFor/cancel stay plain handlers until the
-      // promise-native OAuthFlowManager grows an Effect surface.
+      // batch OAuth-service conversion migrates the remaining procedures.
       startDesktopFlow: t
         .input(schemas.muxGatewayOauth.startDesktopFlow.input)
         .output(schemas.muxGatewayOauth.startDesktopFlow.output)
--- a/src/node/services/muxGatewayOauthService.ts
+++ b/src/node/services/muxGatewayOauthService.ts
@@ -11,9 +11,9 @@
  * is already the exact user-facing string.
  *
  * Desktop flow bookkeeping (deferreds, loopback server lifetime, timeouts)
- * stays promise-based in `OAuthFlowManager`, which is shared with the other
- * OAuth services — converting that lifecycle to Effect `Scope` is deferred
- * until the shared manager itself migrates.
+ * lives in `OAuthFlowManager`, which is shared with the other OAuth services
+ * and is itself Effect-native (per-flow `Scope` lifecycle) — Effect callers
+ * here use its `*Effect` surface directly.
  */
 import * as crypto from "crypto";
 import { Effect, Schema } from "effect";
@@ -295,7 +295,9 @@ export class MuxGatewayOauthService {
         // Keep server-side timeout tied to flow lifetime so abandoned flows
         // (e.g. callers that never invoke waitForDesktopFlow) still self-clean.
         timeoutHandle: setTimeout(() => {
-          void self.desktopFlows.finish(flowId, Err("Timed out waiting for OAuth callback"));
+          Effect.runFork(
+            self.desktopFlows.finishEffect(flowId, Err("Timed out waiting for OAuth callback"))
+          );
         }, DEFAULT_DESKTOP_TIMEOUT_MS),
       });
 
@@ -353,7 +355,7 @@ export class MuxGatewayOauthService {
         result = Err(`Xum Gateway OAuth error: ${callbackOrDone.error}`);
       }
 
-      yield* Effect.promise(() => self.desktopFlows.finish(flowId, result));
+      yield* self.desktopFlows.finishEffect(flowId, result);
     });
   }
 
--- a/src/node/utils/oauthFlowManager.test.ts
+++ b/src/node/utils/oauthFlowManager.test.ts
@@ -293,6 +293,74 @@ describe("OAuthFlowManager", () => {
   });
 
   // -----------------------------------------------------------------------
+  // guaranteed cleanup (per-flow scope)
+  // -----------------------------------------------------------------------
+
+  describe("guaranteed cleanup", () => {
+    it("closes the server and clears the timeout even when the deferred resolve throws", async () => {
+      // Per-flow scope guarantee: each resource has an independent release, so
+      // a defect while settling the deferred cannot leak the loopback server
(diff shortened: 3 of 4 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex security review

PR #4037, opened 2026-09-01
🤖 feat: render background workflow-result wakes as compact machine events

ThomasK33:
## Summary

Background workflow completions now render in the transcript as a compact, collapsible machine event ("Background workflow finished") instead of dumping the raw `<mux_workflow_result>` payload into a full-width user bubble.

## Background

When a background workflow reaches a terminal state, the terminal-attention drain (`taskService.drainTerminalAttention`) wakes the owner workspace with one coalesced synthetic user prompt built from `buildWorkflowResultContextMessage`. Because that prompt can bundle several runs (plus workspace-turn wakes), it carries no per-run `workflow-result` metadata — so the transcript's metadata-based hiding for in-stream workflow continuations never applies, and the raw model-facing payload rendered as if the user had typed it ("Original workflow command: … `<mux_workflow_result>` …").

Other machine-authored wakes (sub-agent completions, workspace-turn terminals, bash monitor wakes) already get the collapsed `CollapsibleMachineMessage` treatment, matched by their fixed opening sentences.

## Implementation

`getBackgroundWorkWakeSummary` now also matches `WORKFLOW_RESULT_MESSAGE_OPENING_SENTENCE` ("The workflow below has finished.") and returns "Background workflow finished", routing the row through the same `BackgroundWorkWakeMessage` → `CollapsibleMachineMessage` path as the other background wakes. The full model-facing prompt stays inspectable behind the expand toggle.

- Only synthetic rows get this treatment (`message.isSynthetic === true` guard in `MessageRenderer`); a user-typed lookalike still renders as a normal user message.
- The timeline already classifies this prompt as `workflow.result` ("Workflow finished") from the same opening sentence, so labels stay consistent.
- Mixed drains that lead with the workspace-turn section keep their existing "Background workspace turn finished" treatment.
- In-stream workflow continuations (with `workflow-result` metadata) remain hidden entirely; the run card shows those results.

## Validation

- New behavioral test: a synthetic wake built with the real `buildWorkflowResultContextMessage` renders collapsed, hides the payload, and reveals it on expand; existing tests cover the non-synthetic fallback.
- `SyntheticAutoResumeMessages` story extended with a coalesced workflow wake for visual/Pixel coverage.

## Risks

Low. The change is a single additional prefix match in the wake-summary classifier, gated behind the existing synthetic-row check. Worst case for a false positive would be a synthetic prompt that legitimately starts with the workflow opening sentence — which is exactly the drain's workflow prompt, i.e. the intended target.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `xhigh` • Cost: `$6.87`_

### DIFF
--- a/src/browser/features/Messages/BackgroundWorkWakeMessage.tsx
+++ b/src/browser/features/Messages/BackgroundWorkWakeMessage.tsx
@@ -2,6 +2,7 @@ import type { ReactElement } from "react";
 import { BellRing } from "lucide-react";
 import type { DisplayedMessage } from "@/common/types/message";
 import { BACKGROUND_WORK_WAKE_OPENINGS } from "@/common/utils/machineTurnPrompts";
+import { WORKFLOW_RESULT_MESSAGE_OPENING_SENTENCE } from "@/common/utils/workflowRunMessages";
 import { CollapsibleMachineMessage } from "./CollapsibleMachineMessage";
 
 interface BackgroundWorkWakeMessageProps {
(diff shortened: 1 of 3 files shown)


ThomasK33:
@codex review

PR #4041, opened 2026-09-01
🤖 fix: release preparing after superseded monitor wakes

coadler:
## Summary

Keep bash-monitor wake dispatches from leaving a workspace permanently stuck in `PREPARING` when the level-triggered reconciler supersedes a queued wake before acceptance.

## Background

The reconciler can queue a match wake while a model turn is active, then cancel it when process settlement produces a newer wake. `AgentSession.sendQueuedMessages()` claims `PREPARING` before dispatch and relies on shared cancellation state to return to idle when cancellation wins. Bash-monitor dispatches passed the abort signal but omitted that state, so the canceled send returned successfully without starting a stream or releasing `PREPARING`. The newer settlement wake then deferred forever because the session appeared busy.

## Implementation

- Create one cancellation-state object for each bash-monitor wake dispatch and pass it alongside the reconciler's abort signal.
- Extend the existing WorkspaceService wiring test to require that state on queued monitor wakes. Existing AgentSession queue tests already cover the downstream cancellation and idle transition.

## Validation

- `make static-check`

## Risks

Low. The change only adds bookkeeping to the existing cancelable monitor-wake path. Accepted wakes and non-monitor queue entries keep their current behavior.

---

_Generated with `xum` • Model: `openai:gpt-5.6-sol` • Thinking: `xhigh` • Cost: `$0.00`_

### DIFF
--- a/src/node/services/workspaceService.test.ts
+++ b/src/node/services/workspaceService.test.ts
@@ -784,9 +784,16 @@ describe("WorkspaceService bash monitor wake reconciler wiring", () => {
       runtimeConfig: { type: "local" },
     });
     let queuedMode: string | undefined;
+    let queuedCancelState: { canceledBeforeAcceptance: boolean } | undefined;
     const sendMessage = mock(
-      (_workspaceId: string, _prompt: string, options: { queueDispatchMode?: string }) => {
+      (
+        _workspaceId: string,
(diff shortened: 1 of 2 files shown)


coadler:
@codex review

PR #4046, opened 2026-09-01
🤖 fix: render native macOS shortcut labels (⌘P, not ⌘·P)

ibetitsmike:
## Summary

On macOS, every keyboard shortcut label rendered with a middle-dot separator and Command first ("⌘·P", "⌘·⇧·P"). `formatKeybind` now emits the native macOS form: one run of symbols with no separator, modifiers ordered ⌃⌥⇧⌘ ("⌘P", "⇧⌘P", "⌃C"). Windows/Linux output ("Ctrl+Shift+P") is unchanged.

## Background

The collapsed-sidebar hamburger tooltip showed "Open sidebar (⌘·P)". The same formatter feeds the command palette, Settings → Keybinds, sidebar collapse buttons, composer hints and every other shortcut hint, so the fix is in the one shared function. `DirectoryPickerModal` had hardcoded `isMac() ? "⌘O"` to dodge the dot; that workaround is removed now that the formatter produces the same string.

## Implementation

`ctrl` still means Command on macOS (matching `matchesKeybind`) unless `macCtrlBehavior: "control"` or the keybind also sets `meta`, in which case it renders as ⌃. Only the presentation changed; matching logic and the KEYBINDS registry are untouched.

## Validation

- Unit tests cover separator, ⌃⌥⇧⌘ ordering, ctrl-as-Command vs ctrl-as-Control, multi-character keys, and the unchanged Linux/Windows format.
- Remote dogfood UAT (browser build, macOS spoofed via `navigator.userAgentData`, then plain Linux): hamburger tooltip, 70 command-palette labels and 60 Settings → Keybinds labels contain no middle dot and keep ⌃⌥⇧⌘ order; Cmd+P / Cmd+Shift+P still work; Linux labels unchanged.

## Risks

Low. Display-only change in one function; no keybind matching or registry changes.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$0.00`_

### DIFF
--- a/src/browser/components/DirectoryPickerModal/DirectoryPickerModal.tsx
+++ b/src/browser/components/DirectoryPickerModal/DirectoryPickerModal.tsx
@@ -13,7 +13,7 @@ import { Input } from "@/browser/components/Input/Input";
 import type { FileTreeNode } from "@/common/utils/git/numstatParser";
 import { DirectoryTree } from "../DirectoryTree/DirectoryTree";
 import { useAPI } from "@/browser/contexts/API";
-import { formatKeybind, isMac } from "@/browser/utils/ui/keybinds";
+import { formatKeybind } from "@/browser/utils/ui/keybinds";
 import { getErrorMessage } from "@/common/utils/errors";
 
 const OPEN_KEYBIND = { key: "o", ctrl: true };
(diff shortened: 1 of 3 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

PR #4047, opened 2026-09-02
🤖 feat: Login with Coder from a browser on a remote Xum server

ThomasK33:
## Summary

"Login with Coder" now works from a browser connected to a Xum server, including a remote one. The authorization redirect lands on the Xum server's own `/auth/coder/callback` route instead of a desktop loopback listener, following the pattern the Gateway/MCP OAuth flows already use. The Settings UI no longer disables the login control on remote hosts.

## Background

The Coder OAuth flow (authorization-code + PKCE, RFC 7591 dynamic client registration with exact redirect-URI matching) only had a desktop implementation: it bound an ephemeral `127.0.0.1:<port>` listener and registered that as the client's redirect URI. A browser talking to a remote Xum server can never reach that listener, so `ProvidersSection` hid the control behind `isRemoteServer` with the message "Login with Coder requires the desktop app or a locally hosted Xum server".

That limitation was self-imposed. Coder's OAuth2 provider accepts non-loopback redirect URIs (HTTPS required for non-localhost hosts — verified against `codersdk/oauth2_validation.go`), and Xum already hosts server-side OAuth callbacks for other providers. Since DCR lets Xum register any redirect URI, the server can register its own public callback URL.

## Implementation

- **`coderOauthService.ts`** — the callback source is abstracted as a `CoderCallbackChannel`: the existing `LoopbackServer`, or a new `ServerCallbackChannel` (`server: null`) for browser mode. `startServerFlow({ deploymentUrl, flowId, redirectUri })` runs the same launch pipeline as `startDesktopFlow` with the server channel; `handleServerCallback({ state, code, error, errorDescription })` delivers the redirect to its flow (state = flow ID) and resolves with the login outcome for the route to render. Delivery is one-shot and raced against the flow's own completion so a Cancel/timeout mid-exchange settles the browser request instead of hanging it. Server flows live in the same `OAuthFlowManager`, so `waitForDesktopFlow`, `cancelDesktopFlow`, `disconnect`, the stored-client lease, the cross-process commit lock, and revocation apply unchanged.
- **`orpc/server.ts`** — `GET /auth/coder/start?deploymentUrl&flowId` (authenticated; the redirect URI is built server-side from the validated public host incl. app-proxy prefix and is never accepted from the client — there is deliberately no oRPC surface that takes a redirect URI) and `ALL /auth/coder/callback` (unauthenticated navigation, added to the origin-bypass set, GET + `form_post`). The byte-identical Gateway/MCP callback HTML is extracted into one `sendOAuthCallbackPage` helper rather than adding a fourth copy; Governor keeps its own template.
- **`ProvidersSection.tsx`** — branches on `isDesktop` like the Gateway flow: desktop → oRPC loopback flow; browser (local or remote) → `/auth/coder/start`, then the existing oRPC wait/cancel. The remote-server paragraph and its gate on the palette hint are removed. The flow ID (which doubles as the OAuth `state`) is now derived from `crypto.getRandomValues` because `crypto.randomUUID` is undefined outside secure contexts (plain-HTTP remote origins), and "Copy & Open Coder" opens before the best-effort clipboard write for the same reason (`navigator.clipboard` is undefined there).
- **Docs** (`providers.mdx` + regenerated built-in skill content) describe remote login and the HTTPS requirement.

## Validation

Dogfooded end-to-end in a `dev-server-sandbox` served at a non-localhost hostname (`xum.127.0.0.1.nip.io`) against a mock Coder deployment implementing buildinfo, RFC 8414 discovery, DCR, an explicit Allow/Deny consent page, token exchange, revocation, and the AI Gateway listings:

- DCR registered `http://xum.127.0.0.1.nip.io:<port>/auth/coder/callback` (the Xum server's origin, built from the request host); the authorize URL carried that redirect URI plus an S256 challenge.
- Allow → redirect to `/auth/coder/callback` → PKCE exchange (verifier present) → catalog discovery → Settings shows **Connected** with Re-login / Refresh models / Disconnect; the callback tab auto-closed.
- Deny → server-rendered "Login failed: access_denied: The user declined" page; Settings shows the inline error and the previous login stays connected.
- Dogfooding surfaced the two insecure-context bugs above (`randomUUID`, `navigator.clipboard`), which would have silently broken login on plain-HTTP remote servers.

New tests: service (full server flow with exact HTTPS redirect URI in DCR + exchange + persistence + replay refusal; OAuth error redirect; cancel mid-exchange settles the pending callback and revokes; invalid redirect URI rejected pre-network), routes (auth required on start, app-proxy-aware redirect URI, callback success/failure rendering, cross-origin POST on the new callback), UI (browser-mode hint starts via `/auth/coder/start` and continues on `waitForDesktopFlow`; remote host renders the login control and targets its own origin).

## Risks

Medium, scoped to the Coder provider login. The desktop path is only refactored to consume the channel interface (the loopback `LoopbackServer` is structurally compatible; all 88 pre-existing service tests pass unchanged). The new unauthenticated callback route only looks up a live flow by `state` and hands it an authorization code that is useless without this process's PKCE verifier and client secret. Gateway/MCP callback routes are refactored to the shared page helper with byte-identical output (existing route tests cover them).

Known limitation: Coder rejects `http://` redirect URIs for non-localhost hosts, so a remote Xum server reachable only over plain HTTP gets a clear DCR error from the deployment rather than a login.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$29.72`_


ThomasK33:
@codex review

ThomasK33:
@codex review

PR #4052, opened 2026-09-02
🤖 fix: terminal wakes skip the compaction request's disable-all tool policy

ibetitsmike:
## Summary

Terminal-attention wakes (workflow results, workspace-turn wakes) no longer inherit the compaction request's disable-all tool policy. `resolveTerminalWakeCallerSendRestrictions` now skips `compaction-request` rows while walking history for the caller's restrictions, so the wake adopts the row that actually defines the conversation's tool policy.

## Background

After a compaction, the newest user row carrying a `toolPolicy` is the synthetic compaction-request row, whose `[{ regex_match: ".*", action: "disable" }]` policy exists only for its own summary turn. The wake walk treated it as the caller's restriction, so every terminal wake until the next manual send ran with zero tools (`tool configuration ... "toolNames":[]` in `mux.log`). The agent could only answer with text ("Let me inspect the checkout...") and stalled. Observed live on a dogfood workspace whose `improve-codebase-architecture` workflow finished after a compaction; a paused goal meant nothing re-prompted it.

## Implementation

Same doctrine as `resolveParentAutoResumeOptions`, which already excludes the `compact` identity: the compaction row is a per-turn policy, not a caller restriction, so the backward walk skips it and keeps going.

## Validation

- New test: "workflow wakes skip the compaction request's own disable-all tool policy". It seeds a restricted manual row (`^bash$`) followed by a compaction row and asserts the wake carries the manual restriction; a second run seeds an unrestricted manual row plus compaction and asserts the wake has no `toolPolicy`.
- Red-green: with the skip stubbed out, the test fails with the incident's exact signature (received the `.*` disable policy instead of `^bash$`).

## Risks

Low. The change only affects which user row the terminal-wake walk adopts, and only when a compaction row sits between the wake and the last manual send. Workspaces on older builds still hit the stall after their next compaction until upgraded; a manual message clears it.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$8.76`_

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -2582,6 +2582,102 @@ describe("TaskService", () => {
     expect(liftedOptions.toolPolicy).toBeUndefined();
   });
 
+  test("workflow wakes skip the compaction request's own disable-all tool policy", async () => {
+    const config = await createTestConfig(rootDir);
+    const { parentId } = await saveLocalParentWorkspace(config, rootDir);
+    const restrictedPolicy = [{ regex_match: "^bash$", action: "disable" as const }];
+    const compactionPolicy = [{ regex_match: ".*", action: "disable" as const }];
+    const runStore = new WorkflowRunStore({ sessionDir: path.join(config.sessionsDir, parentId) });
+    const createRun = async (runId: string) => {
+      await runStore.createRun({
+        id: runId,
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

PR #4090, opened 2026-09-05
🤖 refactor: deslop pass 2, remove unused re-exports and single-call alias wrappers

ibetitsmike:
## Summary

Pass 2 of the behaviour-preserving deslop (stacked on #4089, pass 1): needless re-exports and redundant wrappers. Removes 191 re-export entries that nothing imports through their barrel, the 3 shadcn sub-components, 3 orphaned exports and 8 import bindings that only existed to feed those entries, and inlines 2 single-call alias functions. 34 files, +24 / -354 lines. No behaviour changes.

## Stack context

- #4089 (pass 1, base): dead exports and unreachable files.
- **this PR (pass 2)**: unused re-exports, dead barrel-only symbols, single-call alias wrappers.
- pass 3 (planned): useless tests. pass 4 (planned): comments and defensive noise.

## Why

knip reports an export as unused when no module imports it, and a barrel re-export is exactly that kind of export: a name listed in `export { ... } from` that no consumer ever imports through the barrel. After pass 1 the repo's own `make check-deadcode` output was still dominated by these entries (588 findings, most of them barrel lines and story exports), which hides real dead code. This pass removes the entries that are provably unused so the check regains signal.

## Implementation

1. Candidates: every knip "unused export" (round 3 run on top of pass 1) whose declaration is an `export { ... }` specifier rather than a definition: 227 entries in 30 files.
2. Usage proof per entry with a TypeScript-API pass over every `src`, `tests`, `scripts`, `vscode`, `packages` and `.storybook` file: resolve each import/re-export specifier to its file, and mark an entry used if any importer of that barrel imports the name (named import, `export { name } from`, `export *`, dynamic `import()`), or references it as a member of a namespace import (`import * as schemas` then `schemas.Name` or `typeof schemas.Name`). knip misses the namespace-member case: it flagged 36 `schemas.ts` entries that `src/common/orpc/types.ts` reads via `import type * as schemas`. Those 36 are kept. 181 entries passed, and a second round after removing them found 10 more re-exports whose only consumer had been the removed barrel line (`schemas/api.ts` -> `schemas.ts`, `tabRegistry.tsx` -> `Tabs/index.ts`).
[...]

PR #4089, opened 2026-09-04
🤖 refactor: deslop pass 1, remove dead exports and unreachable files

ibetitsmike:
## Summary

Pass 1 of a multi-pass, behaviour-preserving deslop of the codebase: dead code only. Deletes 96 top-level declarations (functions, types, schemas) that nothing imports and 3 unreachable files: 45 files, +12 / -1,559 lines. No behaviour changes, no renames, no test deletions.

## Background

Requested by Mike as a whole-repo slop hunt, worked in passes ordered safest-first (dead code, then redundant wrappers, then useless tests, then comments/defensive noise), each pass capped at roughly 1,500 deleted lines and landed through green CI plus a clean Codex verdict before the next starts. This PR is the first pass. I currently judge the repo needs **4 passes** (a 5th, optional, for production code reachable only from tests, is listed under Deferred).

## Implementation

Detection used two tools, then every candidate was verified independently:

1. `make check-deadcode` (the repo's ts-prune wrapper) and an ad hoc `knip@5` run (config kept out of the repo; entries: all `src` entrypoints, tests, stories, scripts, workflows, vscode, packages). knip reported 1,189 unused exports/types outside stories.
2. Repo-wide `git grep -w` for every candidate identifier across all tracked files (scripts, workflows, .md, vscode, packages included) to catch consumers knip cannot see.
3. Dead-cluster experiment: in a scratch copy, stripped the `export` modifier from every candidate with zero external references and let `tsgo --noEmit` (`noUnusedLocals`) report what became unused, iterating until it converged. Only symbols the compiler proved unreferenced were deleted; the export-strip itself was discarded. The cascade was repeated after each deletion round and after the file deletions (three knip rounds total).
4. Every deleted name was checked against the added lines of all 46 open PR branches updated in the last 60 days plus the local `openai-ws-transport` and `mike/server-self-update` branches: no in-flight branch adds a usage of any deleted symbol.

## Evidence per deletion

Verification for each entry below unless stated otherwise: knip reports no importer; `git grep -w` finds no reference outside the definition; `tsgo --noEmit` flags it as unused once the `export` keyword is removed; both tsconfigs typecheck after deletion.

<details>
<summary>Unreachable files (3)</summary>

| File | What it was | Why unused |
| --- | --- | --- |
[...]

PR #4092, opened 2026-09-05
🤖 tests: deslop pass 3, remove duplicate and tautological test cases

ibetitsmike:
## Summary

Pass 3 of the behaviour-preserving deslop (stacked on #4090, pass 2; #4089 is pass 1): useless tests. Deletes 37 test cases (11 of them as 2 whole test files) that assert the language or a constant rather than our behaviour, or duplicate another case byte for byte, plus the one no-op export those tests were the last reader of. 21 files, +0 / -331 lines. No production behaviour changes; total unit test count goes from 14,383 to 14,346.

## Stack context

- #4089 (pass 1): dead exports and unreachable files.
- #4090 (pass 2): unused re-exports, dead barrel-only symbols, single-call alias wrappers.
- **this PR (pass 3)**: useless tests.
- pass 4 (planned): comments and defensive noise.

## Rules applied

A test was deleted only if it meets one of these, verified by reading the test and its neighbours in full:

1. **Asserts the language or a constant, not our behaviour**: a constant equals its own literal, `typeof X === "string"` on a typed constant, a constructor does not throw, a no-op function does not throw, `expect(true).toBe(true)`.
2. **Byte-for-byte duplicate** of another case in the same file (identical callback text after whitespace normalisation, confirmed by reading both), so the identical code path with identical inputs and assertions runs twice. The surviving case is the one whose title matches what the body actually asserts.

Nothing was deleted for being slow, flaky-looking, over-mocked, or hard to read. Candidates found by the scan that failed the rules are listed under "Reviewed and kept".

## Implementation

A TypeScript-API scan over all 970 test files hashed every `it`/`test` callback body (duplicates), and classified every `expect` chain (subjects that are imported constants compared to literals; `typeof` checks; `not.toThrow` as the only assertion). Every hit was then read in full before deciding; 16 duplicate-body hits, 6 constant-literal hits, 9 existence-only hits and 26 no-throw-only hits were reviewed. Deletions were applied by removing whole `it`/`test`/`describe` statements with the TypeScript AST, followed by removing imports that became unused (`tsgo` `noUnusedLocals`).

## Evidence per deletion

<details>
<summary>Rule 2: byte-for-byte duplicate cases (14 cases, 10 files)</summary>

| File | Deleted case | Identical to (kept) |
| --- | --- | --- |
| `src/browser/utils/vim.test.ts` | first `describe("Navigation")` block: `w moves to next word`, `b moves to previous word`, `$ moves to end of line`, `0 moves to start of line` (4 cases) | the second `describe("Navigation")` block, which repeats all four verbatim and then continues with more cases |
| `src/browser/features/ChatInput/draftAttachmentsStorage.test.ts` | `parsePersistedChatAttachments preserves legacy provider attachments` | `parsePersistedChatAttachments returns attachments for valid items` (same input, same expected output; no legacy shape was involved) |
| `src/browser/features/Settings/Sections/ModelsSection.test.ts` | `keeps non-required OpenAI models visible without OAuth` | `shows GPT-5.5 when OpenAI OAuth is not configured` (same `shouldShowModelInSettings(KNOWN_MODELS.GPT.id, false)` call) |
| `src/browser/stores/GitStatusStore.test.ts` | `getStatus() returns same reference for same workspace when no changes` | `getStatus() returns same reference when status hasn't changed` |
| `src/common/utils/compaction/autoCompactionCheck.test.ts` | Edge Cases > `missing context usage returns safe defaults` | `returns false when no context usage data` |
| `src/common/utils/thinking/policy.test.ts` | second `returns 5 levels including xhigh for gpt-5.5` | first case with the same title |
| `src/node/services/compactionHandler.test.ts` | `should not emit DeleteMessage events during append-only compaction` | `should not emit delete events when compaction is append-only` |
| `src/node/services/desktop/DesktopBridgeServer.test.ts` | `bridges binary traffic in both directions for a valid token` | `handleUpgrade bridges binary traffic when mounted on an external HTTP server` (both mount via `listenUpgradeServer`; the kept title describes the body) |
| `src/node/services/sshPromptService.test.ts` | `emits request event only for first caller`; `returns immediately after responder released` | `deduped waiters all resolve on respond` (asserts the single request and all three resolutions); `returns immediately with empty response when no responders` |
| `src/node/services/workspaceService.test.ts` | `unarchive() rolls back unarchivedAt when snapshot restore fails` | `unarchive() returns Err when snapshot restore fails`; the deleted title promised a rollback assertion the body never made |

</details>

<details>
<summary>Rule 1: constant, type, or no-op assertions (23 cases, 2 of them whole files)</summary>

| File | Deleted | Why it asserted nothing about behaviour |
| --- | --- | --- |
| `src/constants/workspaceDefaults.test.ts` (file) | all 9 cases | every case restates the `WORKSPACE_DEFAULTS` literal (`agentId` is `"exec"`, `input` is `""`), its `typeof`s, or that `Object.freeze` froze it. The constant is exercised by 26 production files and their tests. |
| `src/common/telemetry/client.test.ts` (file) | both cases | `initTelemetry and shutdownTelemetry are no-ops` and `trackEvent silently forwards ... without throwing` only assert `not.toThrow()` on no-op / fire-and-forget functions |
| `src/cli/argv.test.ts` | `describe("CLI_GLOBAL_FLAGS")` (2 cases) | `toContain("--help")` etc. and `toHaveLength(4)` on the constant; how the flags behave is covered by the `isElectronLaunchArg` cases in the same file ("returns false for CLI flags in packaged mode (--help, --version)") |
| `src/common/compat/legacyMux.test.ts` | `prefers xum while retaining mux as an accepted alias` | `toEqual(["xum", "mux"])` on `SUPPORTED_XUM_PROTOCOL_SCHEMES`; the `xum://` and `mux://` handling that consumes it is covered by `deepLink.test.ts`, `xumProtocolRegistration.test.ts`, and `argv.test.ts` |
| `src/common/types/thinking.test.ts` | `MAX_THINKING_INDEX is 9` | `toBe(9)` on the constant; `parseThinkingInput`, the only consumer, has its own numeric-index cases in the same file |
| `src/common/utils/copilot/modelRouting.test.ts` | `COPILOT_MODEL_PREFIXES exports the shared Copilot model family filters` | `toEqual([...])` on the constant; `isCopilotRoutableModel` cases cover its use |
| `src/node/utils/gitNoHooksEnv.test.ts` | `describe("GIT_NO_HOOKS_ENV")` (2 cases) | `toEqual({...})` restating the constant and `typeof value === "string"` on a `Record<string, string>`; `gitNoHooksPrefix` cases cover the env's effect |
| `src/node/runtime/DockerRuntime.test.ts` | `should accept image name`, `should accept registry image` | `new DockerRuntime({...})` does not throw; the next two cases construct the same runtimes and assert real getters |
| `src/node/services/ptc/quickjsRuntime.test.ts` | `abort method exists and can be called`; `supports Symbol.dispose` | `runtime.abort()` does not throw (the preceding case exercises abort semantics); `typeof runtime[Symbol.dispose] === "function"` (the preceding case disposes twice) |
| `src/node/utils/paths.test.ts` | `falls back to original path if cygpath unavailable` | body is `expect(true).toBe(true)` with a comment saying it cannot be tested |

</details>

<details>
<summary>Production code deleted as a consequence (1)</summary>

| Symbol | File | Why |
| --- | --- | --- |
| `shutdownTelemetry` | `src/common/telemetry/client.ts` | no-op "kept for API compatibility"; its barrel re-export went in pass 2 and the deleted `client.test.ts` was its only remaining caller (`git grep -w shutdownTelemetry` now returns nothing) |

The imports that the deleted cases left unused (`CLI_GLOBAL_FLAGS`, `SUPPORTED_XUM_PROTOCOL_SCHEMES`, `MAX_THINKING_INDEX`, `COPILOT_MODEL_PREFIXES`, `GIT_NO_HOOKS_ENV`) were removed from the test files; all five constants remain in use by production code.

</details>

## Reviewed and kept

- 23 of the 26 "only asserts `not.toThrow`" hits are real behaviour: idempotent cleanup (`sshAskpass cleanup is idempotent`, `double-release is safe`, `end() multiple times`), positive validation (`validateRequiredParams does not throw when ...` x5, `assertValidSkillId accepts valid kebab-case`), defensive-path handling (`init-output without init-start`), `installEpipeGuard swallows EPIPE`, `models without cleanup are safe`, `is JSON + structured-clone safe`.
- `eventStore.test.ts` > `should not throw when deleting non-existent persisted state` ends in `expect(true).toBe(true)` but exercises a distinct error-handling branch; kept.
- `GitStatusStore.test.ts` > `subscribe and unsubscribe` asserts `typeof unsubscribe` but also that a second unsubscribe is safe; kept.
- The two `tests/ui/storybook/coverage.test.ts` duplicate-body hits are loop-generated cases over different file lists, not duplicates.
- 286 cases whose only assertions are on mocks were sampled; the sampled ones render a component or drive a service and assert a callback prop or dependency was invoked, which is behaviour coverage. Not deleted.
- Two permanently skipped UI tests (`tests/ui/chat/readMore.test.ts` "expansion state persists across tab switches", `tests/ui/compaction/compaction.test.ts` "force compaction ... only ONE notification") document intended coverage the harness cannot run; left in place as deferred.

## Validation

- `make static-check`: green.
- Every modified test file re-run in isolation: 1,099 pass, 0 fail.
- Full unit suite, run in CI's layout (the 13 order-sensitive files each in their own process, the rest shared): every isolated file exits 0; the shared run is green apart from two pre-existing order-dependent items that reproduce identically on origin/main: the three `WorkspaceFooterBar last prompt` cases (on CI's isolation list) and `workflow_run duplicate guard > serializes overlapping launches`, which fails and then hangs whenever `workflow_run.test.ts` runs without the default `bun test src` file order (verified on a clean origin/main worktree: same failure, same hang). Neither file is touched by this PR.
- A first plain `bun test src` run segfaulted Bun inside `WorkflowRunner.test.ts` ("Bun has crashed. This indicates a bug in Bun"), the QuickJS-heavy file CI already isolates for that reason; untouched by this PR.

## Risks

None to runtime behaviour: only test code and one no-op function with no callers are removed. Coverage risk is nil for duplicates (the identical case still runs) and for constant assertions (the constants are exercised through their consumers' tests).

## Declined review findings

(none yet)

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$61.60`_


ibetitsmike:
@codex review

ibetitsmike:
@codex review

Please review current rebased head 2016e1db4d7a0aafd53c286c2e8bd98c9e10c4df. The summary marks this head completed, but the explicit no-findings comments still reference the pre-rebase commit; we need a current-head verdict before merging.

PR #4104, opened 2026-09-06
🤖 tests: establish agent loop lifecycle contracts

ThomasK33:
## Summary
Establish deterministic lifecycle contracts before progressively moving the agent loop onto Effect. Queue tests now prove admission reserves synchronously and successor persistence waits for failed-head cleanup; stream tests prove completion readers observe finalized history and abort partial cleanup. Idle tests distinguish an idle phase from pending work.

Existing stop-versus-success coverage now gates the real final history write instead of using a timer. No production behavior changes.

## Validation
- 349 tests passed across 11 lifecycle, queue, chaos, workspace-turn, and reconnect suites.
- Full local `make static-check` passed with installed tool paths; `git diff --check` passed.
- Local Hadolint used native 2.14.0 because CI-pinned 2.12.0's macOS x86 binary exited 139 on arm64. CI retains its pinned version. Existing formatting target skipped Nix formatting because Nix is unavailable locally.

## Risks
Test-only change. Controlled barriers exercise real HistoryService I/O and preserve existing failure behavior. This is Phase 0 of the approved loop lifecycle refactor.

---

### DIFF
--- a/src/node/services/agentSession.queueDispatch.test.ts
+++ b/src/node/services/agentSession.queueDispatch.test.ts
@@ -61,6 +61,79 @@ async function waitForCondition(condition: () => boolean, timeoutMs = 500): Prom
 }
 
 describe("AgentSession queued message tool-call dispatch", () => {
+  test.each(["returned error", "rejection"] as const)(
+    "reserves queued startup synchronously and drains after %s cleanup",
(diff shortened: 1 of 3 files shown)


ThomasK33:
@codex review

PR #4111, opened 2026-09-06
🤖 refactor: supervise agent turn work with scoped Effect lifetimes

ThomasK33:
Turn admission can write history before a stream exists, and terminal accounting can continue after the engine finishes. The session now supervises these lifetimes so app shutdown waits for already admitted work before tearing down its dependencies.

The synchronous TurnCoordinator remains the source of admission and ownership decisions. A private Effect interpreter owns terminal-policy fibers and scoped resources, while physical execution leases cover preparation, queue callbacks, and turn-owned recovery/continuation work. A guardian registered in the app scope accepts late engine completions during shutdown, joins the original Promise-backed work, then closes the session scope. Engine completion remains independent of session policy settlement.

This also fixes an adjacent failure path: rejected abort bookkeeping still reports its error, delivers the abort once, and finalizes only the original turn, preventing an edit waiting for idle from blocking shutdown. Reentrant replacement turns retain their ownership and thinking state.

Validation includes held pre-preparation history writes, background startup and queue cancellation callbacks, late engine completion with blocked accounting, construction after app scope closure, retired physical work, abort-bookkeeping failures, and real engine partial persistence before bridge teardown. Final local checks: exact static-check passed; 591 affected backend tests and 2 AI completion tests passed; the rebuilt send-mode UI suite passed all 5 scenarios with unchanged timeouts.

Regression risk is concentrated in interruption, edits, and shutdown ordering. Existing Promise I/O is joined rather than treated as canceled. The app's bounded shutdown remains unchanged; speculative continuous-compaction jobs and service work before session entry retain their existing separate ownership for later migration phases.

---

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -8,6 +8,7 @@
  *   xum run --runtime "ssh user@host" "Deploy changes"
  */
 
+import { EffectRunnerTag } from "@/node/services/di/effectRunner";
 import { Command } from "commander";
 import { resolveXumEnvironmentValue } from "@/common/compat/legacyMux";
 import { tool } from "ai";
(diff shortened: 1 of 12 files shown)


ThomasK33:
@codex review

ThomasK33:
@codex review

ThomasK33:
@codex review

PR #4117, opened 2026-09-07
🤖 refactor: centralize turn preparation and queue admission

ThomasK33:
Turn preparation now uses explicit coordinator admission results and one preparation lifecycle for direct sends, queued sends, background startup, and resume. This closes races where a stale send could change its replacement's retry state, a queue observer could replace the entry being dispatched, or a direct send could write history while an edit owned it.

The coordinator claims ownership and installs correlation before publishing PREPARING. Queued work adopts that exact owner, revalidates the queue head before dequeue, and settles failure callbacks before releasing preparation. Preparation tracks rollback eligibility, durable rows, acceptance, and background/provider handoff separately. Edit replacement requires its exact reservation token and retains that reservation through correlated failure settlement, including background startup; payload validation runs before truncation. These changes preserve the Promise facade and the existing scoped Effect execution introduced in #4111.

Validation covers real-history rollback and held edit writes, reentrant queue mutation/shutdown, stale thinking and retry updates, and transient or persistent failure-callback rejection. All AgentSession/coordinator/MessageQueue/service-container suites pass (514 tests), along with the WorkspaceService admission/queue/resume/shutdown subset (104 tests). The full static checks, Electron main build, and real Electron send-mode workflow all pass (5 integration tests).

Risk is concentrated in admission ordering, cancellation, and accepted pre-stream failures: a regression could strand queued work or associate history with the wrong turn. Deterministic barriers exercise these boundaries, while full CI and the real Electron send-mode workflow cover integration.

---

### DIFF
--- a/src/node/services/agentSession.preStreamError.test.ts
+++ b/src/node/services/agentSession.preStreamError.test.ts
@@ -296,7 +296,11 @@ describe("AgentSession pre-stream errors", () => {
       coordinator: TurnCoordinator;
     };
 
-    privateSession.coordinator.prepare();
+    privateSession.coordinator.prepare({
+      kind: "fresh",
+      intent: "handoff",
+      expectedTurnId: privateSession.coordinator.turnId,
+    });
     aiEmitter.emit("runtime-status", {
       type: "runtime-status",
       workspaceId,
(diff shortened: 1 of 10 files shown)


ThomasK33:
@codex review

PR #4118, opened 2026-09-07
🤖 refactor: make turn cancellation and session disposal awaitable

ThomasK33:
Cancellation now finishes the captured stream attempt's partial persistence and terminal delivery before a replacement starts. Session and workspace disposal await in-flight work, while application and CLI shutdown keep their existing bounded teardown budget.

Abort persistence moves from AIService event forwarding into StreamManager and its mock counterpart. The first cancellation owns its reason and completion; identity checks prevent old cleanup from touching a replacement, and failed commits retain recoverable partial data. Startup cancellation joins the same completion barrier without waiting on an envelope callback that may itself await stop.

AgentSession separates synchronous closing from awaitable disposal. WorkspaceService tracks destructive cleanup outside continuation callbacks to avoid joining the callback that requested removal, keeps subscriptions through captured terminal delivery, and joins cleanup within the app scope. Closing cancels idle and queue waits separately from physical cleanup. Empty service-graph rollback remains synchronous so constructor failures retain their original error.

Test fixtures now await teardown. Policy-only fake handles retire on their own session's closing signal; lifecycle tests retain independent completion barriers.

Validation: 512 session/coordinator/queue/hooks tests, 421 engine/history/AI/mock tests, 1,162 workspace/task/container tests, and 20 CLI/edit tests passed. The full static check and main-process build passed. Isolated UI suites passed for interruption (1 test) and send-mode/queue behavior (5 tests). Deterministic regressions hold disk writes, envelope callbacks, session work and task-tree locks to exercise concurrent stop, replacement, cleanup errors, reentrant disposal, and shutdown bounds.

Risk: incorrect cleanup ordering could stall interruption or workspace removal, or lose interrupted output. Raw engine events retain their role, and session policy settlement stays separate from engine completion to avoid circular waits. Compaction policy and retry ownership remain for later phases.

---

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -1580,7 +1580,9 @@ async function main(): Promise<number> {
         // Interrupt + await the runtime's supervised fibers while their
         // dependencies are still alive (same slot as ServiceContainer.dispose).
         { name: "appFiberScope.close", run: () => closeScopeBounded(appFiberScope) },
-        { name: "session.dispose", run: () => session.dispose() },
+        // The service guardian joins disposal inside closeScopeBounded. After a timeout,
+        // only initiate cleanup here: a second unbounded join would prevent CLI exit.
+        { name: "session.dispose", run: () => session.beginDispose() },
         { name: "mcpServerManager.dispose", run: () => mcpServerManager.dispose() },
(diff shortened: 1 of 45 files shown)


ThomasK33:
@codex review

PR #4120, opened 2026-09-07
🤖 refactor: scope retry and startup recovery ownership

ThomasK33:
Retry cancellation now follows the same attempt through scheduling and startup, so a reentrant observer or delayed pricing lookup cannot launch or mutate a replacement retry. Startup recovery uses one checkpointed runner: successful follow-up work is not replayed when later history reads fail.

The retry manager acquires one logical claim before publishing the starting event. Effect cancellation fences startup until delivery, while original Promise work remains joined by the session's execution scope. Recovery waits for edit reservations and raw stream activity to clear; shutdown cancels idle/backoff waits and joins all active disk reads.

Updates to the retry preference file are serialized. An older marker cleanup cannot delete a newly persisted user opt-out, and an admitted update still commits when its retry generation retires.

### Validation

- 1,696 affected backend tests passed across session/retry, workspace/task/turn-manager/container, and CLI suites.
- Full `make static-check` and `make build-main` passed.
- Deterministic barriers and the Effect test clock cover reentrant scheduling, cancellation during pricing, stale startup derivation, held edits, bounded history retries, shutdown and preference write ordering.
- All six real-backend chat UI tests passed: stream interrupt and send-mode/queue behavior.

### Risk

This changes cancellation and startup-recovery timing, which can affect interrupted-workspace recovery and manual-input priority. Backoff calculation, provider event timing and persisted schemas are unchanged; the new regressions exercise the relevant ownership and durability boundaries.

---


ThomasK33:
@codex review

PR #4121, opened 2026-09-07
🤖 refactor: coordinate compaction and durable continuations

ThomasK33:
## Summary

Compaction can finish after its turn has been stopped, cleared, or replaced. Give observations and continuation handoffs explicit coordinator ownership, and persist Stop independently of transcript cleanup. Late work cannot clear a replacement's state or revive canceled work after restart. Temporary admission holds resume recovery when released.

## Implementation

- The coordinator owns compaction intent, observation stages, and handoff tokens. Semantic retirement wakes waiters without joining obsolete physical work; summary jobs and cleanup retain execution leases through shutdown. Failed cleanup remains bounded and retryable outside the live-session registry.
- Continuation admission and cleanup validate exact ownership under the shared history write lock. Summary handoffs check row identity, pending payload, permitted tail, and cancellation; source-less handoffs carry the original stream's durable publication generation. Legacy, heartbeat, and continuous summary publication and partial cleanup use the same generation fence. Pending-context preparation, malformed-state cleanup, and failed-publication rollback carry that generation through their queued filesystem writes; a rejected old summary cannot overwrite or delete a successor's context.
- A versioned atomic cancellation sidecar survives transcript failures. Each explicit Stop gets a distinct nonce and advances the durable publication generation before publishing. Failed writes retain their predecessor and retry only that frontier; concurrent replacement reads share a retry, and obsolete retries preserve newer journals. Full clear retains cancellation until actual explicit replacement, preventing late summaries from repopulating cleared work. Reset rows and reset-containing batches, full deletion, and trims that remove active provider context advance the shared generation inside their actual write lock after admission checks and before mutation. Failed generation publication refuses the context write; no-op, rejected, and inactive-history-only trims preserve the journal.
- Acceptance occurs at the irrevocable row frontier before goal synchronization or observers. Every manual append branch and explicit Retry compares captured cancellation identity under the accepting write lock. Internal handoff snapshots and their user receipt commit atomically, including with token budgets disabled. Skipped writes cannot acknowledge acceptance; genuine I/O failures remain errors.
- Exact replacement witnesses in active or archived history retire matching cancellation. Failed unlink is ancillary debt and does not block accepted work. Live dispatch and cleanup reconcile these witnesses without erasing a newer Stop. Automatic retries cannot witness replacement and recheck unresolved Stop before scheduling, after backoff, and before provider entry.
[...]

PR #4129, opened 2026-09-08
🤖 fix: retain eager compaction execution through reset

ThomasK33:
Reset can clear the active eager-compaction job while its preparation or summary request is still running. Session shutdown then loses track of that detached work and can finish too early.

Acquire the existing coordinator execution lease when the eager job starts, and release it in that same promise's `finally`. Semantic reset and normal turn admission retain their existing behavior.

This is the first small replacement for closed #4121. It only fixes physical lifetime ownership; coordinator extraction and persistence hardening are separate changes.

Validation covers held preparation and summary work across completion, rejection, reset, shutdown, and replacement by another eager job. The tests retain real coordinator leases and HistoryService storage.

Risk: shutdown now correctly waits for eager work that ignores cancellation until it settles. Existing cancellation and semantic idle behavior are preserved.

---

### DIFF
--- a/src/node/services/agentSession.continuousCompaction.test.ts
+++ b/src/node/services/agentSession.continuousCompaction.test.ts
@@ -24,6 +24,7 @@ import {
   type AgentSessionHarness,
 } from "./agentSession.testHarness";
 import type { ContinuousCompactor } from "./continuousCompactor";
+import type { TurnCoordinator } from "./turnCoordinator";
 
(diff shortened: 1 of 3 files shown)


ThomasK33:
@codex review

Please review this bounded physical-lifetime fix, including reset/replacement and shutdown draining. It replaces only the eager-job execution ownership portion of closed #4121.

---

PR #4133, opened 2026-09-08
🤖 fix: serialize compaction journal publication and cleanup

ThomasK33:
Compaction journal writes and history folding use different serialization boundaries. A delayed fallback or cleanup can therefore operate on a journal that has already been replaced, and a failure after a committed history rename can still look like an uncommitted compaction to its caller.

Serialize journal publication and folding with the existing cross-process history lock. Compare the captured generation and exact journal, and publish ownership/commit receipts synchronously with the final rename. Keep a committed boundary's pending-state success latch set even if later cleanup fails.

Explicit reset remains a separate operation with its established local queue, clearing and invalidation behavior, including an unread startup journal. It does not acquire the shared history lock, so lock contention cannot abandon reset unlink. The advisory existence probe also uses the local queue; actual journal reads and mutations keep the shared lock. Asynchronous owned cleanup uses an exact receipt; completing an apply does not perform a second global clear. Invalid, stale, source-mismatched and durably applied journals preserve their recovery outcome when cleanup fails, leaving exact cleanup retryable. These boundaries let the change ship independently of later destructive-history fencing, attachment-persistence and durable Stop protocols.

This is the journal-publication slice replacing closed #4121 and starts a separate persistence stack.

Validation: the broader implementation suite passed 370 targeted tests; the final cleanup changes passed all 168 affected tests and `make static-check`. Regressions cover interrupted startup recovery, reset under a held history lock, failed cleanup and later recovery, successor preservation, queued fallback/reset ordering, cross-instance publication, commit receipts before cleanup, and no rollback or duplicate folding after a committed boundary. Independent local review approved the implementation and review fixes. Nix formatting was skipped because Nix is unavailable.

Risk centers on journal recovery and lock ordering. Under-lock operations avoid re-entering the journal queue; tests use real HistoryService storage and deterministic barriers. The ordered file protocol does not add a new power-loss durability guarantee.

---

### DIFF
--- a/src/common/orpc/schemas/continuousCompaction.ts
+++ b/src/common/orpc/schemas/continuousCompaction.ts
@@ -75,6 +75,8 @@ const attachment: z.ZodType<PostCompactionAttachment> = z.discriminatedUnion("ty
 export const ContinuousCompactionJournalSchema = z
   .object({
     version: z.literal(1),
+    // Legacy journals are valid until a durable publication generation is advanced.
+    publicationGeneration: z.string().optional(),
     boundary: row,
     staticCopies: z.array(row),
     liveTailCopySpec: z.object({
--- a/src/constants/continuousCompaction.ts
+++ b/src/constants/continuousCompaction.ts
@@ -4,3 +4,5 @@ export const TAIL_MIN_TOKENS = 4_000;
 export const TAIL_MAX_TOKENS = 60_000;
 export const MIN_HEAD_TOKENS = 8_000;
 export const SUMMARIZER_INPUT_FRACTION = 0.7;
+export const CONTINUOUS_COMPACTION_JOURNAL_FILE = "continuous-compaction.json";
+export const CONTINUOUS_COMPACTION_GENERATION_FILE = "continuous-compaction-generation";
--- a/src/node/services/agentSession.continuousCompaction.test.ts
+++ b/src/node/services/agentSession.continuousCompaction.test.ts
@@ -25,6 +25,8 @@ import {
 } from "./agentSession.testHarness";
 import type { ContinuousCompactor } from "./continuousCompactor";
 import type { CompactionToken, TurnCoordinator } from "./turnCoordinator";
+import * as fileLock from "@/node/utils/concurrency/fileLock";
+import { historyWriteLockPath } from "./workspaceRemoval";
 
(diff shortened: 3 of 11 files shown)


ThomasK33:
@codex review

Please review the journal/publication ownership invariant, shared-lock ordering, synchronous receipts, and preservation of explicit reset behavior. Destructive-history generation activation and durable Stop/replacement are separate follow-up layers.

---

ThomasK33:
Addressed both findings:

- 3957153014: explicit reset retains the legacy locally queued unlink, avoiding the newly introduced history-lock timeout. Exact owned publication/cleanup still takes the shared lock. A held real file-lock regression verifies physical removal and unsuccessful recovery by a fresh compactor. Durable cross-backend Stop remains a separate layer.
- 3957153018: malformed-journal deletion logs failure and returns null. A filesystem-failure regression verifies recovery stays usable and cleanup succeeds when storage recovers.

Both regressions failed before the fix. All 370 targeted tests and static checks pass; independent review approved the bounded fix.

@codex review

---

ThomasK33:
Addressed 3957400943: the advisory exists probe now uses the local queue. Journal reads, publication and folding still use the history lock. The regression holds a real history lock during an ordinary send's recovery, verifies the absent-journal probe completes, then verifies message persistence and stream startup after release.

The regression reproduced the lock timeout before the fix. All 146 relevant tests and final static checks pass; independent review approved the two-file change. Scope reassessment kept this fix to one production call site and its rationale comment.

@codex review

---

ThomasK33:
Addressed the cleanup findings as one consistent recovery invariant:

- 3957707739: schema-invalid and stale-generation journals share best-effort discard. Failed unlink returns null, preserves the current history, and permits later cleanup and fresh publication.
- 3957803214: successful fold and already-present-boundary recovery retain success when exact cleanup times out. Local retirement remains generation guarded, and retained journals can be cleaned on recovery without duplicate folding or completion.
- Independent review found the equivalent source/fingerprint-mismatch path; it now uses the same best-effort exact cleanup and stays rejected without blocking recovery.

Six lock-failure regression cases failed before these changes. All 168 targeted tests and final static checks pass; independent review approved the four-file fix. Nix formatting was skipped because Nix is unavailable.

@codex review

---

PR #4153, opened 2026-09-08
🤖 fix: prevent message edits from hanging during active streams

ammar-agent:
## Summary

Fix message edits hanging during an active stream by reserving the edit before awaiting interruption. Terminal policy can no longer dispatch a queued successor while `stopStream` settles, leaving the edit waiting on the wrong turn.

## Implementation

- Use the existing owned edit reservation for the entire interruption → truncation → replacement-startup path.
- Remove the separate queue-deferral flag and its try/finally synchronization; preserve completion cleanup, startup preemption, and admission guards.
- Allow the edited startup's own emergency context-budget recovery through its reservation without admitting competing turns.

## Validation

- Deterministic completion/abort races failed before the fix and pass afterward: the edit retains ownership, restores queued input, and persists only the edited user turn.
- Added a regression for emergency budget recovery during edited startup.
- After rebasing onto current main, all 667 agent-session/coordinator tests passed (~11s).
- All 6 stream-completion IPC tests pass under Node 22 (~6s); replaced the old private-flag polling with an explicit edit-wait barrier.

## Subtractive ledger

Production: **26 lines added, 48 removed; net −22**. Removed duplicate edit-deferral state, its synchronization, and a redundant queue-drain guard. Existing reservation ownership and cleanup now enforce the same safety boundary across the whole operation. No unrelated offsets.

## Risks

This changes edit exclusion timing across stream completion, queue dispatch, and compaction. Existing completion/admission/queue tests remain green, and owned budget-recovery coverage protects the edited request from blocking itself.

---

_Generated with `xum` • Model: `openai:gpt-6-astra` • Thinking: `xhigh` • Cost: `$3.79`_

### DIFF
--- a/src/node/services/agentSession.tokenBudget.test.ts
+++ b/src/node/services/agentSession.tokenBudget.test.ts
@@ -1279,6 +1279,32 @@ describe("AgentSession token-budget lifecycle", () => {
     estimate: 127_000,
     hardCeiling: 119_808,
   };
+  test("an edited request can recover its own pre-handle budget overflow", async () => {
+    const h = await setup({ failure: (attempt) => (attempt === 1 ? exceeded : undefined) });
+    await seedHistory(h, 20_000);
+    await h.historyService.appendToHistory(
+      workspaceId,
+      createMuxMessage("edit-target", "user", "Original request")
+    );
+
+    expect(
+      (
+        await h.session.sendMessage("Edited request", {
+          ...options,
+          editMessageId: "edit-target",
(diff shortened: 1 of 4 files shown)


ammar-agent:
@codex review

ammar-agent:
@codex review

Updated the integration test to await the actual edit-completion barrier instead of polling the removed private flag. All six IPC completion tests and local static checks pass; production code is unchanged from the approved review.

PR #4155, opened 2026-09-08
🤖 fix: bound the origin fetch during worktree workspace creation

### LINKED ISSUE #4122: 🤖 fix: bound workspace creation when Coder Git authentication waits indefinitely
Creating a Git worktree workspace can wait indefinitely when Coder's GitHub external authentication is missing or expired. The UI appears to do nothing, while the app remains healthy and workspace/list continues to succeed. The user confirmed that reconnecting GitHub resolved the incident.

### Reproduction

1. Run Xum inside a Coder workspace with an HTTPS GitHub origin and Coder's Git askpass helper, but without valid GitHub external authentication.
2. Create a worktree workspace from `main`, either through the UI or `workspace/create`.
3. Observe that creation never finishes and no workspace appears in the list. A direct API caller in this incident reached its own 60-second timeout; this was not a captured server timeout or HTTP 504.
4. Reconnect GitHub through Coder external authentication. Pending requests may then proceed.

### Confirmed cause and evidence

`WorktreeManager` awaits a best-effort `git fetch origin main` before adding the worktree and persisting workspace metadata. This fetch has no deadline. Coder's askpass helper waits for external authentication, so the entire creation operation remains pending.

- Four app-owned fetch processes were observed waiting on Coder askpass for 41–48 minutes. The helper was requesting a username for `https://github.com`.
- Coder's external-auth access-token command reported that GitHub authorization was required.
- The test workspace had no corresponding branch, worktree, or config entry when inspected. No creation retry was needed to reproduce the stall.
- A deterministic subprocess reproduction showed the same behavior on main `373257bbc68f5f3fe2009e2abd7433563287bd7e` and the pre-agent-loop-refactor baseline `5cca5bf5f`. This predates the Effect refactor.
- Installed code from `42863ff6b` also contained the unbounded fetch. The app reported `v0.28.5-nightly.41-2-g42863ff6b-dirty` during the original report.

### Expected behavior / proposed fix

Bound the best-effort fetch and terminate its process tree on timeout, then use the existing local-trunk fallback. Preserve explicit cancellation as cancellation. Surface enough diagnostic context to distinguish authentication waiting from normal workspace initialization; genuine creation failures should produce a visible error.

A local candidate using a 15-second fetch deadline plus process-tree termination passed 50 targeted tests and the full static check. It is not yet submitted or deployed. Tests cover a stalled helper with inherited pipes, cleanup of descendants, local fallback, explicit cancellation, and successful fetch using the fresh remote trunk. A timeout without process-tree termination was insufficient because descendants retained the subprocess pipes.

### Recovery caveat

Reauthorizing GitHub can release every pending creation request. Killing only the askpass helper is not request cancellation: the failed fetch can trigger local fallback and continue creation. Check for in-flight operations and orphaned branches/worktrees before retrying.

---

ibetitsmike:
## Summary

Bound the best-effort `git fetch origin <trunk>` that runs before a local worktree is added, so workspace creation can no longer hang forever when a Git credential helper (Coder's askpass waiting on missing or expired GitHub external auth) blocks the fetch. After 15 s the fetch process tree is killed, creation continues from the local trunk with a diagnostic init-log note, and an explicit cancellation still cancels creation instead of falling back.

Fixes #4122

## Background

`WorktreeManager.createWorkspace` awaited `fetchOriginTrunk` with no deadline. Inside a Coder workspace with an HTTPS GitHub origin and no valid external auth, Coder's askpass helper waits indefinitely, so the whole creation request stayed pending: nothing appeared in the sidebar, no error was shown, and fetch processes were observed waiting on askpass for 40+ minutes. Killing only the `git fetch` leader is not enough either: the helper descendants inherit the subprocess stdio pipes, so Node's `close` event never fires and the awaiting promise still never settles (this also meant cancelling the creation hung today).

## Implementation

- `WORKTREE_CREATE_FETCH_TIMEOUT_MS = 15 s` (new constant in `src/constants/terminationTimeouts.ts`). `WorktreeManager` takes an optional `{ fetchTimeoutMs }` so tests can shorten it; runtime callers are unchanged.
- `fetchOriginTrunk` runs the fetch with `AbortSignal.any([callerSignal, AbortSignal.timeout(fetchTimeoutMs)])` and the existing `killTreeOnTermination: true` option of `execFileAsync` (detached process group + SIGKILL of the group on Unix, `taskkill /T` on Windows, stdio pipes destroyed so `close` fires).
- Catch order: caller abort is rethrown first (cancellation stays a cancellation and no worktree is added); a fired deadline logs `Note: Fetch from origin did not finish within 15s (Git may be waiting on a credential helper or external authentication); using local branch state` plus a `log.warn`, then returns `false` so the existing local-trunk fallback runs; other fetch failures keep their current handling.
- `SSHRuntime.fetchOriginTrunk` already bounds its remote fetch (120 s exec timeout) and is untouched.

## Validation

- Red-green: on unfixed `main` both new stall tests hang until the 20 s test timeout. With the deadline but `killTreeOnTermination: false` they still hang, proving the tests guard process-tree termination and not just the timeout. With the full fix all 21 tests in `WorktreeManager.test.ts` pass.
- The stall fixture points `origin` at a `remote.origin.uploadpack` shim that spawns `sleep 600 &` (inherits the stdio pipes) and `wait`s; tests assert the fallback result, the diagnostic note, that the new branch is based on the local trunk, and that both shim PIDs are gone (zombies awaiting reap count as gone).
- Cancellation test: aborting while the shim stalls returns `success: false`, logs no fallback note, leaves no workspace directory, and kills the shim tree.
- Fresh-origin test: with a reachable origin holding a newer commit, the new workspace HEAD equals `origin/main`.
- Remote dogfood UAT (Coder Agents, real UI, head `17fbae3549`): a project whose origin needs credentials and whose credential helper never answers created its workspace in 15.10 s and 15.13 s with the diagnostic note visible in the init log; the creation fetch's process group (git, helper, and its `sleep`) was gone within a second of the deadline with no manual cleanup; a healthy origin with a newer commit created the branch on the fresh `origin/main` (96 ms); a genuine creation failure (ref lock permission denied) surfaced as a visible error.
- UAT observations outside this change (no code changed for them, none shown to be regressions): deleting the pending draft row in the UI during the stalled fetch does not abort the backend creation, because the create request has not returned a workspace id yet (explicit `AbortSignal` cancellation is covered by the unit test); the background git-status fetch is a separate app-owned fetch path that also stalls on the same credential helper and is not bounded by this PR; stopping the server during the 15 s window orphans the detached fetch group, which is inherent to `killTreeOnTermination`.

## Risks

Low and scoped to local worktree creation. Behavior change: a fetch that legitimately takes longer than 15 s (very slow network, huge delta) now falls back to the local trunk with a visible note instead of waiting; the background git-status fetch keeps `origin/<trunk>` fresh in normal use, and creation already fell back when the fetch failed. The fetch now runs in its own process group on Unix (required for group kill), which is the same mode other bounded git calls in the repo use.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$26.89`_

### DIFF
--- a/src/constants/terminationTimeouts.ts
+++ b/src/constants/terminationTimeouts.ts
@@ -4,6 +4,13 @@ export const TASK_TERMINATION_WORKSPACE_REMOVE_TIMEOUT_MS = 2 * 60 * 1000;
 export const WORKTREE_DELETE_GIT_TIMEOUT_MS = 60 * 1000;
 
 /**
+ * Bounds the best-effort `git fetch origin <trunk>` before a local worktree is added. A Git
+ * credential helper (for example Coder's askpass waiting on missing external auth) can block the
+ * fetch indefinitely; after this deadline creation continues from the local trunk.
+ */
+export const WORKTREE_CREATE_FETCH_TIMEOUT_MS = 15 * 1000;
+
+/**
  * Bounds backup Git calls that can hang on a blackholed remote, while leaving room for a
  * slow initial clone.
  */
--- a/src/node/worktree/WorktreeManager.test.ts
+++ b/src/node/worktree/WorktreeManager.test.ts
@@ -32,6 +32,7 @@ async function createWorktreeManagerFixture(options?: {
   existingBranchName?: string;
   currentBranchName?: string;
   tempDirPrefix?: string;
+  fetchTimeoutMs?: number;
 }) {
   const rootDir = await fsPromises.realpath(
     await fsPromises.mkdtemp(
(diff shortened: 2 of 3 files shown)


ibetitsmike:
@codex review

PR #4157, opened 2026-09-08
🤖 fix: tighten iOS chat chrome (decoration rows, bottom band, status-bar blur)

ibetitsmike:
## Summary

Fixes three iOS layout issues reported from an iPhone running xum as a standalone web app: oversized collapsible rows above the composer, a blank band under the footer bar, and the Liquid Glass blur bleeding into the header.

## Background

- `globals.css` raises every `button`/`a` to a 44px floor on coarse pointers. The collapsible decorations above the composer (TODO, sub-agents, background bash) are 24px rows, so three of them became 132px of chrome.
- iOS 26 standalone web apps intermittently resolve `100dvh` as if a browser toolbar were still present, so the app shell ends ~50pt above the screen edge and the footer strands a blank band below it. Chromium emulation with 34px safe-area insets shows the existing inset handoff (#3757) working, so the gap is the viewport unit, not our padding. Same symptom and fix as [we-promise/sure#835](https://github.com/we-promise/sure/issues/835).
- The blur: WebKit's `LocalFrameView::fixedContainerEdges` hit-tests the top-center of the layout viewport (inside the status-bar band). If it finds a `position: fixed` element with a resolved `background-color`, spanning >=90% of the viewport width, it paints a solid color extension over the band and hides the soft "scroll pocket" blur (`WKWebView _shouldHideTopScrollPocket`). Otherwise the blur stays. `HitTestResult::setInnerNonSharedNode` resolves pseudo-elements to their host, so a `::before` is never a candidate. #3961 placed the fixed header at `top: env(safe-area-inset-top)` with a transparent background painted on a `::before`, which left nothing samplable at the top edge and guaranteed the blur.

## Implementation

- `ChatInputDecoration` rows carry `mobile-touch-row`, which globals.css floors at 32px inside the coarse-pointer block. Nested action buttons in expanded content keep 44px.
- `installViewportHeightSync()` (browser entry only, gated on `navigator.standalone`) writes `--app-viewport-height` from `window.innerHeight` at startup and on `resize`; `html/body/#root` use `var(--app-viewport-height, 100dvh)`. Electron, Storybook, and non-standalone browsers keep the dvh path untouched.
- `index.html` renders a real `.ios-status-bar-band` element (fixed, `height: env(safe-area-inset-top)`, `z-index: 1001`, sidebar color; constant dark on light themes as before). It only displays under `(display-mode: standalone)` + iOS WebKit. The header and drawer go back to their plain `bg-sidebar` background; the `::before` workaround is removed.
- `isIosStandaloneWebApp` moved from `downloadFile.ts` to `utils/env.ts` now that it has a second consumer.

## Validation

Chromium with iPhone 15 Pro metrics, touch emulation, and CDP safe-area insets (top 59 / bottom 34) against Storybook and the dev server:

| check | before | after |
| --- | --- | --- |
| collapsed decoration rows (TODO, sub-agents, bash) | 44px each | 32px each; expanded lists keep 44px action buttons |
| footer bottom edge (390x844 frame) | 844 | 844 (unchanged; inset handoff intact) |
| header background on touch | `rgba(0,0,0,0)` + `::before` | `bg-sidebar` |
| simulated `navigator.standalone` | n/a | `--app-viewport-height: 852px`, `#root` 852px |
| band (gates replayed ungated) | n/a | fixed 0..59px, z 1001, dark in light theme |
| desktop (1280px, fine pointer) | 24px rows, static header | unchanged |

- New story `IPhone16eComposerDecorations` snapshots the collapsed stack (dark/light, phone) and asserts the shipped coarse-pointer rule floors `.mobile-touch-row` below 44px; red-green verified by temporarily setting it to 44px.
- `viewportHeight.test.ts` covers gating, initial sync, resize, and cleanup.
- `App.phoneViewports` test-runner suite passes; lint, typecheck, prettier pass.

Not verifiable here: real iOS 26 rendering. Please check on device: (1) decoration stack height, (2) no band under the footer on cold launch, after backgrounding, and after keyboard open/close, (3) status-bar band solid with no blur over the header, in both dark and light themes, with the drawer open.

## Notes

Includes the `Codex Comments` gate fix from #4158 (merged in, not cherry-picked) so this PR's CI can pass before that lands; the diff shrinks back to the iOS changes once #4158 merges.

## Risks

- The 32px row floor is below Apple's 44pt guideline; the rows are full width, and the value is a one-liner to adjust.
- Dialog and command palette overlays (z 1500/2000) sit above the band and paint their dim on a `::before`, so WebKit may still show the blur while one is open. Left as is; fixing it means reverting those overlays to a plain background, which is a separate change.
- If iOS ever reports `window.innerHeight` incorrectly in standalone mode, the shell would follow it; `resize` keeps it current.

---

_Generated with Coder Agents_


ibetitsmike:
@codex review

ibetitsmike:
@codex review

PR #4160, opened 2026-09-08
[mux-gateway] 🤖 fix: preserve cache-write accounting

coadler:
## Summary

Recover OpenAI cache-write metadata before gateway usage aggregation.
This places reported writes under Cache Create instead of Input.

## Implementation

- Normalize cache-write metadata in streaming and non-streaming gateway responses.
- Preserve explicit SDK counts, including zero.
- Reject invalid metadata counts and prevent negative uncached input.
- Add regressions for SDK conversion, accumulation, and display accounting.

## Risks

The change affects gateway usage accounting only.
Correct cache-write classification can increase estimates when cache-write rates exceed input rates.
OAuth handling and historical totals remain unchanged.

---

_Generated with `xum` • Model: `openai:gpt-6-astra` • Thinking: `high` • Cost: `$21.31`_

### DIFF
--- a/src/node/utils/gatewayStreamNormalization.test.ts
+++ b/src/node/utils/gatewayStreamNormalization.test.ts
@@ -1,4 +1,14 @@
 import { describe, expect, it } from "bun:test";
+import { generateText } from "ai";
+import { MockLanguageModelV3 } from "ai/test";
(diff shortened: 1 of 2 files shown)


coadler:
@codex review

PR #4162, opened 2026-09-08
🤖 fix: serialize best-of report publication with grouped finalization

### LINKED ISSUE #4074: 🤖 fix: best-of grouped finalization can ship a report artifact a fast sibling continuation already replaced
## Summary

Follow-up from Codex on #4058 (round 23, P1), filed rather than fixed there under Mike's stopping rule for that PR (best-of orchestration is outside that PR's startup/shutdown thesis).

`TaskService.buildBestOfCompletedTaskToolOutput` (src/node/services/taskService.ts around line 12813 on `b795316c1e`) reads every sibling's report artifact, then re-checks that no non-reporting sibling is executing again (`isBestOfSiblingExecutingAgain`) before assembling the grouped task-tool output. Codex:

> When a sibling continuation starts after its old artifact was read and finishes quickly enough to write a replacement before this final pass, the sibling is no longer executing or streaming, so this check succeeds and the parent is permanently finalized with the old report already stored in `reports`. [...] the second pass samples only current activity and never verifies that each artifact is still the revision that was read; compare/re-read artifact identities or serialize assembly with sibling continuation completion.

## Suggested direction

Either re-read each artifact after the activity check and compare a revision marker (`nowMs`/`persistedAtMs` from `upsertSubagentReportArtifact`) with the one first read, or take the per-child report lock (or `deferredBestOfLocks`) around the whole assembly so a continuation's report write cannot interleave with it.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$202.20`_

ibetitsmike:
## Summary

Best-of grouped finalization could commit the parent's task tool output on a sibling report artifact that the sibling was already replacing. Publication of a best-of child's report now runs under the same per-parent lock as the grouped assembly, a resumed sibling counts as executing again, and the partial commit re-checks the siblings synchronously right before the write.

## Background

Fixes #4074 (Codex round 23 P1 on #4058). `buildBestOfCompletedTaskToolOutput` read every sibling's report artifact, then re-sampled only current activity (`isBestOfSiblingExecutingAgain`: active workspace-turn status or streaming) before handing the grouped output to the partial commit. Two windows let a stale artifact through:

- A sibling's continuation that started after its artifact was read and finished before the re-check was idle again at the re-check, so the assembly finalized on the old artifact already in `reports`.
- `finalizeAgentTaskReport` flipped the child to `reported` before writing its artifact, both outside `deferredBestOfLocks`. An assembly that ran between the flip and the write saw a reported, idle sibling with its old artifact.

In both cases the sibling's later delivery found the grouped output already finalized, so its replacement report was dropped from the tool result.

## Implementation

- `finalizeAgentTaskReport` moves the status flip plus artifact writes into `publishAgentTaskReport` and, for best-of children, runs it under `deferredBestOfLocks(parent)`, the lock the assembly already holds. The lock is released before `deliverReportToParent` re-acquires it, so the critical section stays the publication only.
- `isBestOfSiblingExecutingAgain` also treats `taskStatus` `running` / `awaiting_report` as executing again: a resumed sibling (for example `markInterruptedTaskRunning`) has no workspace-turn mirror and is not streaming between its stream end and its publication. `shouldDeferBestOfFallback` now relies on the shared predicate instead of repeating those statuses.
- `areBestOfSiblingsStillSettled` is a synchronous check (sibling identities plus executing-again) used by the assembly's live re-check and again inside the `updatePartialIfMessageIdMatches` transform, where no await separates the check from the write. When it fails there, `tryFinalizePendingTaskToolCallInPartial` returns `not_ready` so the reporting sibling defers rather than appending a synthetic fallback; the resumed sibling's own report finalizes the group.

## Validation

- New test `best-of finalization never ships a report a resumed sibling is replacing`: a barrier on the parent's partial commit resumes the already-reported sibling and drives its replacement report through the production stream-end path while the first sibling's assembly is mid-commit; the barrier lifts when the sibling requests the group lock. Red on main (grouped output carried the pre-continuation report); green with the fix.
- Inverted toggles over the whole `taskService.test.ts` file: removing the publication lock, the status-aware predicate, or the commit-time re-check each turns the new test red again.
- `grouped partial finalization exposes each terminal report exactly once` seeded a `running` sibling with an on-disk artifact, which is now by definition a sibling whose report is being replaced; its fixture models the settled state (`interrupted` with the artifact) so its exactly-once assertion still holds.

- Remote dogfood UAT on this SHA (real models, raw `chat.jsonl` verified): a foreground best-of candidate stopped while working and resumed lands its newest report in the finalized group with one report per candidate; single task, background terminal wake, and `task_stop` behave as before. A user Stop on the parent commits its partial into history (`cleanupAbortedStream`), so the `partial.json` recovery path this PR hardens is not reachable from the UI and is covered by the tests above.

## Risks

Moderate, scoped to best-of report delivery. The new lock hold in `finalizeAgentTaskReport` covers config edits, metadata emission, and artifact writes; it acquires `deferredBestOfLocks(parent)` before `workspaceFileLocks(parent)`, the same order the assembly uses, and is released before delivery re-acquires it. A best-of sibling with a stale artifact whose status is `running` or `awaiting_report` now defers grouped finalization until it reports again, where it previously finalized on the stale artifact.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$0.00`_

### DIFF
--- a/src/node/services/taskService.test.ts
+++ b/src/node/services/taskService.test.ts
@@ -12,6 +12,7 @@ import {
 } from "@/constants/terminationTimeouts";
 import { Config, type ProjectsConfig, type Workspace as WorkspaceConfigEntry } from "@/node/config";
 import { HistoryService } from "@/node/services/historyService";
+import type { MutexMap } from "@/node/utils/concurrency/mutexMap";
 import * as subagentGitPatchArtifacts from "@/node/services/subagentGitPatchArtifacts";
 import {
   getSubagentGitPatchMboxPath,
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

ibetitsmike:
Post-merge UAT record correction (late-arriving critic report): the UAT Critic's final verdict for this PR's remote round is **BLOCKED for PR-specific acceptance, with no PR regression established** - not the runner's PASS that the pre-merge summary relayed. The critic independently verified live exact-order grouped delivery (stop/resume lands the newest report, no duplicates) but declined sign-off because UI-driven probes cannot reach the changed crash/restart recovery leg: parent Stop commits the partial by design (streamManager commits on abort; taskService returns without finalizing when no partial.json remains), so that leg is exercised only by the deterministic interleaving regression test in this PR, which is the acceptance evidence for the race.

Pre-existing observations from the round (not attributed to this PR, recorded for triage): (1) P2: an n=3 candidate stopped via the child UI "Stop streaming" stays Running at the parent indefinitely, with no interrupted-candidate indication (task_stop correctly shows Interrupted); (2) P3: resuming an interrupted child restarts its original work before accepting the follow-up prompt. Full report with screenshot audit retained in the fix workspace evidence dir.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5` • Thinking: `max` • Cost: `$365.12`_

PR #4166, opened 2026-09-08
🤖 fix: honor interrupt_active while a delegated turn is still PREPARING

### LINKED ISSUE #3943: task_workspace_lifecycle: interrupt_active refused during delegated turn's PREPARING window
### Context

PR #3940 restores the `task_workspace_lifecycle` tool (reversible `archive`/`unarchive` of owned workspace-turn workspaces). Its `interrupt_active` option interrupts active turns before archiving, but delegated turns are exempted from the "queued user messages" refusal only while their durable handle reports `status === "queued"`.

### Problem (Codex review round 15, P2)

When a newly created delegated workspace turn is in AgentSession's PREPARING phase:

- the durable handle is already `running`, and
- `hasPendingQueuedOrPreparingTurn()` makes `listLiveWorkspaceActivity().queuedMessages` true.

Because the exemption only recognizes `status === "queued"`, the archive request classifies the PREPARING turn as unrelated queued user activity and returns `status: "active"` even with `interrupt_active: true`. The documented interruption path is therefore refused during every delegated turn's startup window (it becomes usable again once the turn starts streaming).

This is fail-safe (over-refusal, no data loss) but makes `interrupt_active` transiently unusable and forces callers to retry.

### Suggested direction

Distinguish a PREPARING delegated turn owned by the target (interruptible: it has a delegated workspace-turn handle that `interrupt_active` is allowed to stop) from an unrelated queued user message (must refuse). Likely requires exposing which pending/preparing entries belong to delegated turn dispatch vs. user-queued messages, and extending the exemption accordingly — with care not to reopen the queued-message races closed in earlier review rounds (turn admission hold, synchronous send guards).

### References

- Review thread: https://github.com/coder/xum/pull/3940#discussion_r_PRRT_kwDOPxxmWM6bx5I2 (round 15)
- Caller gate: `TaskService.archiveOwnedWorkspaceTurnWorkspace` / `collectActiveWorkspaceLifecycleTurns` (src/node/services/taskService.ts)
- Activity snapshot: `WorkspaceService.listLiveWorkspaceActivity` (src/node/services/workspaceService.ts)

---
_Generated with [`mux`](https://github.com/coder/mux) • Model: `anthropic:claude-fable-5` • Thinking: `xhigh`_

ibetitsmike:
## Summary

`task_workspace_lifecycle` with `interrupt_active: true` now interrupts and archives a workspace whose delegated turn is still in AgentSession's PREPARING phase, instead of refusing with `status: "active"` and "queued messages" for the whole startup window.

Fixes #3943

## Background

A newly created delegated workspace turn's durable handle already reads `running` while the target session is PREPARING, and `listLiveWorkspaceActivity().queuedMessages` is true during PREPARING (it includes `isPreparingTurn()`). Both archive gates only exempted delegated turns whose handle was `queued`:

- the manager pre-gate in `archiveOwnedWorkspaceTurnWorkspace` classified the PREPARING send as unrelated queued user work, and
- `acquirePreInterruptionArchiveHold` refused every PREPARING session outright ("a message dispatching").

Codex round 15 P2 on #3940 flagged this as fail-safe over-refusal. The exemption has to distinguish the collected delegated turn's own PREPARING send from a user message dequeued into PREPARING, without reopening the queued-message races #3940 closed (turn-admission hold, synchronous send guards).

## Implementation

- `AgentSession.getStoppablePreparingWorkspaceTurn()` returns the PREPARING send's workspace-turn correlation only once the engine has registered its startup (`TurnCoordinator.startupRegistered`, set by `onStreamStarting` when StreamManager's pending start exists). That is the only PREPARING state `stopStream()` cancels; before it (history acceptance, request preparation) a stop only notifies and the turn still streams, so those sub-windows keep refusing. Exposed on `WorkspaceService` and the `WorkspaceLifecycleHost` seam.
- Manager pre-gate: "queued messages" is also explained when that correlation matches a collected non-queued turn on the target (`isSameWorkspaceTurnTaskCorrelation`, shared with the hold's existing stream check).
- Hold: a PREPARING session refuses only when its PREPARING send is not one of the expected delegated turns (or an auto-retry is pending). The queue-count check now runs independently of the PREPARING check, so user entries queued behind the exempt delegated turn still refuse. Queued-handle turns stay out of the exemption: they are interrupted by targeted queue removal, not `stopStream`.
- After `interruptActiveWorkspaceLifecycleTurns`, the archive waits for the target session to reach idle (deterministic `waitForIdle`, bounded by `TASK_TERMINATION_STOP_STREAM_TIMEOUT_MS` as a hang guard) before entering the sink. `stopStream` resolves before the aborted PREPARING turn reaches turn policy, so the sink's synchronous `isPreparingTurn`/`hasActiveOrPendingTurnWork` checks would otherwise refuse with the turns already destroyed. The pre-interruption hold keeps admission frozen during the wait. A timeout returns `status: "error"` without bypassing the sink checks.

## Validation

- Real-session tests (`createAgentSessionHarness`): the accessor is undefined before `onStreamStarting`, equals the correlation afterwards, and is undefined for a manual send and after idle. A real `WorkspaceService` + `AgentSession` test drives a delegated send into PREPARING, checks the hold refuses without the correlation and grants with it, emulates StreamManager's pending-start abort, asserts the session still reads PREPARING when `stopStream` resolves, and archives through the real `refuseLiveUserActivity` sink after `waitForIdle`.
- Manager tests: mismatched correlation still refuses "queued messages"; the exact correlation archives and passes it to the hold; the sink is not entered until `waitForIdle` resolves.
- Each new guard was toggled off individually and its test failed (gate exemption, hold exemption, settle wait, startup-registration requirement).

## Risks

Medium, confined to model-driven `task_workspace_lifecycle` archive with `interrupt_active`. The exemption is bound to the exact owner + handle + turn correlation of a collected non-queued turn plus a registered engine startup, so a user send that took over the session, a delegated turn still holding a queued handle, or a PREPARING send that a stop cannot cancel all keep refusing as before. The new settle wait applies to every interrupted turn; for streaming turns the session is already idle when `stopStream` resolves, so it is a no-op there. Dogfood UAT was not run: the fixed window is a sub-second backend admission state with no UI surface, and the remote UAT skill is not available in this workspace.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$6.03`_

### DIFF
--- a/src/common/types/message.ts
+++ b/src/common/types/message.ts
@@ -855,6 +855,17 @@ export function parseWorkspaceTurnTaskCorrelation(
   return { taskHandleId, ownerWorkspaceId, turnId };
 }
 
+export function isSameWorkspaceTurnTaskCorrelation(
+  first: WorkspaceTurnTaskCorrelation,
+  second: WorkspaceTurnTaskCorrelation
+): boolean {
+  return (
+    first.taskHandleId === second.taskHandleId &&
+    first.ownerWorkspaceId === second.ownerWorkspaceId &&
+    first.turnId === second.turnId
+  );
+}
+
 export function getCompactionFollowUpContent(
   metadata?: MuxMessageMetadata
 ): CompactionRequestData["followUpContent"] | undefined {
--- a/src/node/services/agentSession.preparationAdmission.test.ts
+++ b/src/node/services/agentSession.preparationAdmission.test.ts
@@ -530,4 +530,42 @@ describe("preparation admission", () => {
       await h.session.waitForIdle();
     }
   );
+
+  test.each([
+    ["delegated", { taskHandleId: "task", ownerWorkspaceId: "parent", turnId: "turn" }],
+    ["manual", undefined],
+  ] as const)(
+    "a stoppable PREPARING %s send is reported only between engine startup registration and idle",
+    async (kind, correlation) => {
+      const h = await harness(`stoppable-preparing-${kind}`);
+      const entered = Promise.withResolvers<Parameters<typeof h.aiService.streamMessage>[0]>();
+      const release = Promise.withResolvers<void>();
+      spyOn(h.aiService, "streamMessage").mockImplementation(async (request) => {
+        entered.resolve(request);
+        await release.promise;
+        return Ok(createStartedTurnHandle(h.session.closingSignal));
+      });
+      const sent = h.session.sendMessage(
+        "head",
+        {
+          ...options,
+          ...(correlation ? { muxMetadata: { type: "workspace-turn-task", ...correlation } } : {}),
+        },
+        { startStreamInBackground: true }
+      );
+      const request = await entered.promise;
+      // PREPARING, but the engine has not registered the startup: a stopStream here would
+      // only notify, so the turn is not reported as stoppable.
+      expect(h.session.isPreparingTurn()).toBe(true);
+      expect(h.session.getStoppablePreparingWorkspaceTurn()).toBeUndefined();
+
+      request.onStreamStarting?.("starting-1");
+      expect(h.session.getStoppablePreparingWorkspaceTurn()).toEqual(correlation);
+
+      release.resolve();
+      expect((await sent).success).toBe(true);
+      await h.session.waitForIdle();
+      expect(h.session.getStoppablePreparingWorkspaceTurn()).toBeUndefined();
+    }
+  );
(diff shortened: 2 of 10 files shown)


ibetitsmike:
@codex review

PR #4169, opened 2026-09-08
🤖 fix: avoid local-agent warning flashes for sub-agents

ammar-agent:
## Summary

Prevent the local-agent interference warning from flashing for coordinated sub-agents, while retaining it for unrelated local task families sharing the project directory. Follow-up UX fix after #4153.

## Implementation

- Derive frontend-only family roots during the existing bulk metadata load, before archived rows are filtered. Cycle members and incoming descendants share a deterministic representative; no extra IPC, disk reads, or persisted state.
- Compare those roots in the UI, excluding coordinated ancestors, descendants, siblings, and cousins while still warning about unrelated families' active sub-agents.
- Retain workspace identity rather than a display name during activity handoffs, and resolve it against current candidates so navigation into that agent's family cannot leave a stale warning.
- Replace manual memoized ID collection and repeated metadata lookup with filtered metadata and a direct active-workspace selection; inline the single-use warning view. No new timer or effect.
- In a separate CI compatibility commit, recognize Codex's observed automatic security-review heading without exempting actual findings.

## Validation

- Five same-family cases reproduced the false warning before the fix; a separate regression reproduced the held-warning leak when switching into the active agent's child.
- All 232 focused config/warning/workspace-family tests and 1,021 workspace/task service tests pass. Real-config tests cover archived ancestry, cycle members/incoming descendants in both input orders, unrelated trees, and non-persistence of derived roots.
- Codex gate tests reproduce the new-heading false positive and verify that explicit or appended security findings remain blocking.
- Headless Storybook-manager checks passed at 1140px and the existing pinned 375px phone viewport: one-line warning, ellipsis on phone, no horizontal overflow or composer displacement.
- Local `make static-check` passes.

## Subtractive ledger

**P1: 61 production lines added, 62 removed; net −1.** The warning fix is +59/−61 (net −2); the separate CI-format compatibility fix is +2/−1 (net +1). Removed frontend ancestry traversal, manual ID-list memoization, redundant metadata lookup/fallback, stale display-name retention, and a single-use view wrapper. Complete family scope is derived during the existing bulk metadata read, not synchronized through new state.

## Risks

This changes warning eligibility and adds an optional derived frontend metadata field, not task execution, isolation, or persisted config. Unrelated task trees still warn; existing activity smoothing and unregistered-workspace protection remain intact.

---

_Generated with `xum` • Model: `openai:gpt-6-astra` • Thinking: `xhigh` • Cost: `$63.33`_

### DIFF
--- a/scripts/check_codex_comments_test.py
+++ b/scripts/check_codex_comments_test.py
@@ -141,6 +141,17 @@ else:
                         cache=data if cached else None,
                     )
 
+    def test_auto_triggered_security_heading_does_not_hide_findings(self):
+        header = "### 🛡️ Codex Security Review · _Automatically triggered_\n\n"
+        clean = FIXTURES["security_no_findings"]["body"]
+        for body, expected in (
+            (clean, 0),
+            (clean + "\n\n[P1] A security issue still needs fixing.", 1),
+            ("Security review completed. Found a P1 credential disclosure.", 1),
+        ):
+            with self.subTest(body=body):
+                self.assert_gate(expected, snapshot([comment(header + body)]))
+
     def test_summary_completion_requires_metadata_and_every_review_row(self):
         completed = FIXTURES["summary"]["body"]
         completed_pr_opened = (
--- a/scripts/lib/codex_comments.jq
+++ b/scripts/lib/codex_comments.jq
@@ -4,7 +4,8 @@
 # Strip only the observed help text: a heading alone must not hide a finding
 # added inside the details section, or a second section appended after it.
 def codex_without_help:
-  rtrimstr("\n\n<details> <summary>ℹ️ About Codex in GitHub</summary>\n<br/>\n\n"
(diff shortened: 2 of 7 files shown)


ammar-agent:
@codex review

ammar-agent:
Addressed both findings:
- 3963031003 / PRRT_kwDOPxxmWM6gcONi: cycle members and incoming descendants now resolve to the same deterministic cycle representative. Real-config regression runs both metadata orders.
- 3963031007 / PRRT_kwDOPxxmWM6gcONm: family roots are derived from complete backend metadata before archive filtering, so archived intermediates no longer split a family. Removed frontend ancestry traversal instead of adding archived-fetch state.

All 232 focused tests, 1,021 service tests, Codex gate tests, and static checks pass. Desktop/phone geometry was revalidated after simplifying the warning view.

ammar-agent:
@codex review

PR #4192, opened 2026-09-09
🤖 feat: add session_history list_items with role/tool_name filters

ThomasK33:
## Summary

Adds a `list_items` action to the `session_history` tool plus optional AND-combined filters `role`, `tool_name`, and a per-item snippet cap `max_chars_per_item` for `list_items` and `search`. This is Increment 1 of the local session-history parity plan (item listing and filters); newest-first browsing, descendant history, and a model-requested rollover are deliberately left to follow-on PRs.

## Background

Codex's history tool can enumerate items and filter by role/tool; Xum's `session_history` only offered `list_windows`, literal `search`, and `read_item`. A model that wants "what tools did I call in the previous window" had to search for guessed substrings. This PR closes that gap using the existing bounded forward scanner and preserves every existing invariant (local JSONL storage, manual-reset privacy floors, 2 MiB / 500-row / 16 KiB limits, sanitized whole-row projection, authenticated cursors).

## Implementation

- `projectHistory` (formerly `historicalText`) collects tool names during the same sanitization traversal used for text. Only `kind === "tool"` records — top-level `dynamic-tool` parts and canonical `nestedCalls` entries — count; a `toolName` key inside ordinary input/output JSON, an omitted `session_history` record, or an excluded row (hidden synthetic, budget-rejected, tail copy) never supplies a match.
- `list_items` is the existing scan with no query; it reuses the default limit, the 25-item search cap, and the 500-char snippet size. `max_chars_per_item` replaces that snippet size; the lead-in before a search match is capped at half the allowance so a short snippet still shows the match. Surrogate-pair handling is unchanged.
- `role`, `tool_name`, and `max_chars_per_item` are hashed into the cursor binding and `list_items` is added to the cursor action enum (no cursor format migration; stale cursors restart as before).
- Filters on `list_windows` / `read_item` return `filters_unsupported` instead of being silently ignored.

## Validation

- New tests in `session_history.test.ts`: ordering + `read_item` round-trip across a manual reset; role / exact tool / nested-tool / decoy JSON / omitted-history / AND-combined filters; snippet cap incl. emoji pairs and `read_item` continuation; filter rejection; cursor binding of every new parameter plus empty progress pages over a 1200-row sparse filter.
- Live sandbox dogfood with a real model: unfiltered listing, `role=user`, `tool_name=bash` (only the assistant row with the bash call), `tool_name=file_read` → 0 items despite textual mentions, `search MARKER role=user max_chars_per_item=12` → `"MARKER-ALPHA"`, `read_item` paging 0→20→40, `list_windows`+`role` → `filters_unsupported`, and after **Reset Context** the pre-reset marker was unretrievable. Layout checked at 1440px and 375px.

## Risks

Low. The only shared-path change is the snippet lead-in computation, which evaluates to the previous constant (120) whenever `max_chars_per_item` is omitted. Filtering happens after the reset-floor phase and inside the existing visit callback, so it cannot widen visibility.

---

<details>

# Local session-history parity with Codex

## Goal and scope
Bring Xum’s local `session_history` closer to Codex’s history-tool capabilities without adopting its hosted backend or private/model-only output policy. This plan assumes the requested scope is the four gaps identified in the comparison: item listing and filters, newest-first ordering, descendant-agent history, and a model-callable context rollover.

Preserve existing calls, local storage, bounded reads/results, sanitization, and manual-reset privacy floors. Functional similarity does not establish code ancestry; Codex’s server-side implementation was not available for inspection.

## Delivery recommendation
Ship **item listing and filters first**, then review the value and cost of the remaining parity features. Each increment is independently useful; later features are not prerequisites for shipping the first one. No implementation or PR publication is part of this planning request.

1. **Recommended first increment:** `list_items`, singular `role` / `tool_name` filters, and bounded `max_chars_per_item`, using the existing forward scanner.
2. **Follow-on:** `recent_first`, after proving bounded reverse browsing can assign correct window IDs without an index or whole-window buffering.
3. **Follow-on:** authorized descendant history using existing task IDs and retained session files only.
4. **Follow-on:** empty-argument `new_context`, after proving safe settlement and durable/idempotent request handling. Continue using the existing memory tool for notes.

## Core invariants
- Preserve local JSONL storage, existing search case sensitivity (case-insensitive), legacy item aliases, and sanitized message-row output. Do not simulate Codex’s normalized `tool` / `developer` rows.
- Keep the manual-reset floor discovery ahead of filtering or emission. A new listing or ordering option must never reveal pre-reset content.
- Preserve 2 MiB scan / 500-row page / 16 KiB result limits and empty-progress pages. Bind new visibility/order parameters into authenticated cursors; restart stale scans rather than returning uncertain data.
- Cross-agent access must not become arbitrary workspace/file access. Removed transcript artifacts require separate bounded/provenance work and are out of scope.
- `new_context` requests a rollover after sibling tools settle; it must not synchronously clear history, replay completed side effects, or masquerade as a manual privacy reset.

## Verified implementation seams
- `src/node/services/tools/session_history.ts` owns sanitized text projection, action dispatch, output limits, and query-bound cursors; its schema is in `src/common/utils/tools/toolDefinitions.ts`.
- `HistoryService.scanHistoryBounded()` in `src/node/services/historyService.ts` wraps `historyScanner.ts` with history locks and append-provenance validation. Scanning discovers the reset floor backwards, then browses forwards. `historyCursor.ts` carries authenticated resumable state.
- Persisted message roles are `user`, `assistant`, and `system`. Tool calls are `dynamic-tool` message parts, with canonical `nestedCalls` for recorded nested execution—not independent `tool` rows.
- `TaskService.isDescendantAgentTask()` provides existing descendant checks; `ToolConfiguration.taskService` is already wired. `getSubagentTranscript()` also resolves saved artifacts, but is not a substitute for bounded history scanning.
- `StreamManager.createStopWhenCondition()` invokes `onStepSettled` after a tool step settles. `AgentSession.pendingRollover` is in-memory; existing automatic budget re-evaluation does **not** establish durable recovery of a model-requested rollover.

## Increment 1 — Item listing and filters
**Recommended implementation scope. Net product LoC: approximately +80–130**, excluding tests and generated files. Base: the current branch; no dependency on later increments.

### API and implementation
1. Extend `TOOL_DEFINITIONS.session_history` with `action: "list_items"` and these optional, bounded, `.nullish()` inputs:
   - `role`: `"user" | "assistant" | "system"`.
   - `tool_name`: exact canonical tool name.
   - `max_chars_per_item`: positive integer capped by the existing read-character maximum.
   Apply these new parameters to `list_items` and `search`; reject their use with incompatible actions rather than silently ignoring them.
2. Implement listing as the existing forward scan without a search query. Reuse `items`, `itemId`, `windowId`, `role`, `text`, and `nextCharOffset`; use the existing default item limit, search item cap, and snippet size. Omitted new parameters preserve existing behavior.
3. Combine supplied filters with AND semantics. `tool_name` selects a **message row containing that tool**, not a new synthetic tool-result item; return the same sanitized whole-row projection. Support canonical nested calls, but do not inspect arbitrary input/output JSON as if it were executable tool records.
4. Keep matching consistent with sanitization: excluded rows, reasoning/provider metadata, and omitted history responses cannot supply a tool-name match. Keep projection/matching local to `session_history.ts`, sharing its existing traversal/exclusion rules rather than creating a separate history abstraction.
5. Add `list_items` to the cursor action schema. Include normalized new filter/snippet parameters in the cursor binding. Keep existing page-size behavior; do not introduce a cursor format migration—stale cursors restart normally.
6. Apply per-item character limits and aggregate byte limits together, preserving Unicode pairs and progress. A short search snippet must not spend its entire allowance before the matched substring. `read_item` remains the way to retrieve subsequent text pages.

### Acceptance and gate
- Unfiltered listing returns readable message rows in existing persisted order; every returned ID round-trips through `read_item`.
- Role, exact tool name, nested-tool, and combined query/filter cases select the intended rows. Ordinary JSON with a `toolName` key does not produce a match.
- Sparse filters and large histories return bounded empty-progress pages until exhausted; no full-history materialization or new persistent index.
- Existing reset-floor, malformed/oversized-row, Unicode, cursor tampering/staleness, and nested-history sanitization behavior remains intact.
- Extend `src/node/services/tools/session_history.test.ts` using `createTestHistoryService()` and real temporary files. Test visible behavior, not schema-copy or prompt-string assertions.
- **Stop boundary:** targeted regressions, static checks, and Increment 1 dogfood pass before starting any follow-on. Human review should focus on filter visibility and cursor binding.

<details>
<summary>Optional follow-on designs, estimates, and quality gates</summary>

## Follow-on 2 — Newest-first browsing
**Separate optional increment. Net product LoC: approximately +220–350.** Base: Increment 1 for convenient integration; it is not needed to ship Increment 1.

Add `recent_first: boolean | null` for listing windows/items and searching, defaulting to today's ordering. Bind direction into the cursor; leave `read_item` ordering-independent.

**Recommended algorithm:** retain the existing privacy-floor phase, then use bounded window-span discovery followed by reverse delivery:
1. Probe backwards from the frozen snapshot tail to the preceding valid window boundary, retaining only resumable positions/metadata—not the window's messages.
2. Once the containing window ID and span are known, reverse-read that span and emit rows with verified attribution.
3. Continue to older spans, never crossing the privacy floor. Both discovery and delivery consume the existing per-call byte/row budgets; either may produce an empty progress page.

Implement in `historyScanner.ts` and `historyCursor.ts`; plumb direction through `HistoryService` and the tool. Span positions must handle both archive and active files, including windows crossing their seam. Reuse existing boundary validation, exact-row IDs, and replay/overlap rules; never infer window identity from timestamps or the last newer boundary encountered.

**Design/quality gate:** before exposing the option, demonstrate correct window attribution and resumable progress on a window larger than a scan page. Then test direction equivalence over a fixed valid fixture, multiple boundaries, malformed/unrepresentable boundaries, file seams, oversized/torn rows, certified appends, rotation/rewrite staleness, and resets appended between calls. Discovery overhead remains proportional to scanned bytes; this is not an indexed low-latency search claim. If meeting these invariants requires persistent indexing, stop and re-scope rather than silently introducing a storage subsystem.

## Follow-on 3 — Retained descendant history
**Separate optional increment. Net product LoC: approximately +80–160.** Logically depends only on the current bounded scanner; does not require newest-first browsing.

1. Add bounded `.nullish()` `task_id`; omission means the calling workspace. Use canonical backend-returned task IDs, not Codex-style agent paths or synthesized workspace names.
2. Reuse `TaskService.isDescendantAgentTask()` and the existing `taskService` dependency. Permit only authorized descendants with retained session files, including inactive/archived tasks whose files remain. Do not permit parent, sibling, unrelated workspace, or arbitrary path access.
3. Route through the same bounded scanner for the target session. Preserve the target's reset floor. Bind **caller identity and target identity**, plus all filters/direction, into cursors; re-authorize every call, including `read_item` and cursor continuation.
4. Keep authorization valid through the bounded read using the applicable task-lifecycle synchronization; a cached ancestry check outside the read is not sufficient for removal/revocation races. Treat establishing this guard as part of the increment, not a deferred security fix.
5. Return a generic not-found/unauthorized error without leaking target metadata. For an authorized task with removed/unavailable session files, return `session_unavailable`; preserve `stale_cursor` for actual snapshot invalidation. Do not map every stale cursor to missing history, create a missing session, or fall back to unbounded `getSubagentTranscript()` artifacts.

**Gate:** real parent/child fixtures prove retained-history reads, target reset floors, denied lateral access, cross-caller/cross-target cursor replay rejection, authorization rechecks, and removal during pagination/read. Saved transcripts for purged tasks are explicitly out of scope. Review authorization and lifecycle locking before shipping; if no suitable guard exists, revise the estimate and scope before implementation.

## Follow-on 4 — Model-requested context rollover
**Separate optional lifecycle increment. Net product LoC: approximately +180–350**, subject to the durability gate below. Independent of the other new retrieval features, but requires existing `session_history` availability.

1. Add `new_context` with strict empty `{}` arguments, a small factory in `src/node/services/tools/new_context.ts`, registration/policy wiring, and the existing tool-icon treatment. Exclude it from PTC. Advertise it only when token-budget rollover and history retrieval are actually available; continuous-compaction/RLM precedence and required-tool terminal flows must remain authoritative.
2. The tool requests a transition; it never resets inline. Acknowledge that rollover is scheduled **after the current tool step**, let sibling calls settle and their results persist, then use the existing stop/continuation lifecycle. Coalesce duplicate requests in one source window.
3. Extend rollover metadata/types in `src/common/types/message.ts`, `src/common/utils/messages/contextWindows.ts`, and `contextWindowRollover.ts` with `reason: "model-requested"`. Keep the metadata fully valid so `isManualHistoryReset()` does not mistake this transition for a privacy reset.
4. Use the existing memory tool for handoff notes—no new notes argument, note file, or automatic summary. Preserve workspace files, tasks, goals, and accumulated costs while retaining existing cache/kernel reset semantics; do not promise Codex-identical environment retention.

**Durability gate before shipping:** an in-memory flag alone is insufficient. Establish a durable accepted-request receipt keyed by originating window/tool call and tie the consuming rollover to the same identity. Prefer recovery from the persisted tool result plus rollover metadata if that supports the entire lifecycle; otherwise design the smallest explicit durable receipt before adding a new storage mechanism. Reconcile it before subsequent model generation, not by re-executing the original tool batch. User interruption/manual reset must durably cancel or supersede outstanding intent so restart cannot resurrect a canceled continuation.

**Gate:** extend `agentSession.tokenBudget.test.ts`, `streamManager.test.ts`, tool-assembly/policy tests, and a new tool test. Verify sibling settlement, duplicate requests, policy disabling, required-tool completion precedence, interrupts/resets, and restart at each durable transition. Assert one consumed rollover, no repeated side-effecting tools, preserved costs, successful retrieval across rollover, and blocked retrieval across manual reset. Include a real backend restart exercise. If durable cancellation/deduplication cannot reuse an existing seam, re-scope this increment rather than claiming the small callback-only implementation is complete.

</details>

## Validation strategy
For each increment, run its targeted tests and sibling suites for every touched production file, then `make static-check` separately. Fix all touched-code failures and rerun the final gate after any edit.

Baseline history regression commands:
```bash
bun test src/node/services/tools/session_history.test.ts \
  src/node/services/historyAppendProvenance.test.ts \
  src/node/services/historyService.providerPrivacy.test.ts
```
When changing scanning/service behavior, also run the relevant `historyService.test.ts`, `historyService.contextBudget.test.ts`, `historyService.truncation.test.ts`, and `historyService.publication.test.ts` suites. Lifecycle work additionally runs the AgentSession, StreamManager, and tool-assembly suites listed above. Tests under `tests/` use `TEST_INTEGRATION=1 bun x jest <path>`; do not run those via `bun test`.

Use real HistoryService instances, malformed/Unicode/large on-disk fixtures, and deterministic lifecycle barriers. Assertions should detect impossible states; malformed user-controlled persisted history must still self-heal or fail closed according to the existing contract. Logs/screenshots alone do not prove authorization or crash consistency.

## Dogfooding and reviewer evidence
Run this **between increments**, not only at the end. Use synthetic transcripts and owned sandbox resources; never reset or seed the user's real session history.

### Setup
```bash
KEEP_SANDBOX=1 make dev-server-sandbox \
  DEV_SERVER_SANDBOX_ARGS="--clean-providers --clean-projects"
```
Run as a managed background task, record its chosen root/ports, and configure only a dedicated test project/provider. Enable token-budget windows through the sandbox UI. For offline runs, use a controlled loopback provider fixture and ensure no ambient provider endpoint can receive requests. For a live-model smoke, use explicitly configured sandbox credentials and record that it is a provider-backed run.

Use `agent-browser skills get core` for the installed CLI workflow, then open the advertised sandbox URL and drive the chat via snapshots. Capture screenshots **and a video recording** for each scenario below; retain artifacts outside the source checkout. Measure assembled prompt size before selecting a small context budget so setup itself does not immediately overflow.

### Scenarios by increment
1. **Listing/filters:** create at least two context windows with unique user text, ordinary tool calls, and nested calls. Request listing, role filtering, exact tool filtering, combined search, and `read_item` continuation. Then perform a manual reset and demonstrate that the earlier marker is no longer retrievable.
2. **Newest-first:** use a multi-page synthetic history crossing active/archive storage. Demonstrate newest-to-oldest results with correct window IDs, an empty discovery page followed by progress, and a stale-cursor restart after a controlled rewrite.
3. **Descendants:** create a parent and retained child through normal task flows. Demonstrate child retrieval and denied unrelated/sibling access. Remove the owned test child during pagination and verify safe failure without archive fallback.
4. **Rollover:** ask the model to save notes, then request `new_context` alongside a harmless uniquely identifiable side-effecting tool. Verify the side effect occurs once, its result persists, the task resumes in a new window, and history recovers the old result. Repeat with interrupt/manual reset and controlled sandbox restart at the tested boundaries.

Inspect transcript/tool cards at desktop and approximately 375px width; add a viewport-pinned full-app story only if presentation changes. Preserve concise backend receipts for scan budgets, window IDs, side-effect counts, and restart results alongside visual evidence. Attach screenshots/video to the review handoff with `attach_file`; if GitHub publication is later authorized, upload via `gh ... --attach`. Stop only the sandbox/browser processes started for this work and remove owned fixtures after evidence is retained.

## Completion and review boundary
- Increment 1 is complete when its acceptance tests, static checks, and dogfood evidence pass; no later parity feature is required for that milestone.
- Keep each follow-on as its own reviewed change with the stated prerequisite gate and an updated estimate if scope grows. No speculative indexing, hosted backend, encrypted/private tool policy, artificial message roles, namespace filters, removed-task artifact recovery, or cross-agent write access.
- Estimates count **net product code only**, not tests. The full optional roadmap is roughly **+560–990 product LoC**; only **+80–130** is recommended for the first implementation.
- This is a source-inspected plan. No product files were changed and no implementation tests or dogfood runs have been performed for these proposed features. Do not create a PR unless explicitly requested.

</details>

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `high` • Cost: `$25.95`_


ThomasK33:
@codex review

ThomasK33:
Dogfood evidence (live sandbox, Haiku via Coder, ~$0.11):

Unfiltered list_items card

search MARKER role=user max_chars_per_item=12 returns a 12-char snippet containing the match

After Reset Context the pre-reset marker is not retrievable

Tool cards at 375px

https://github.com/user-attachments/assets/11f6ba5d-f72f-4982-a793-f5bee883338e

PR #4198, opened 2026-09-09
🤖 feat: newest-first browsing for session_history

ThomasK33:
## Summary

Adds `recent_first: true` to the `session_history` tool for `list_windows`, `list_items` and `search`, so a model can walk transcript history newest-to-oldest without an index or whole-window buffering. This is Follow-on 2 of the local session-history parity plan; #4192 (listing + filters) landed first.

## Background

Recovering "what did I just do before this context window" is the common case for `session_history`, but the scanner only walked oldest-first from the privacy floor, so the most recent rows arrived last. Codex's history tool orders newest-first by default. The plan gated this feature on proving that bounded reverse browsing can assign **exact** window IDs without an index.

## Implementation

The existing floor phase is unchanged (it already walks backwards to the nearest manual reset). For `recent_first`, the scanner then alternates two new phases instead of `browse`:

- **probe** — reverse-scan from the pending span end (initially the frozen snapshot tail) to the nearest boundary row (`getContextBoundaryKind`) or the floor, retaining only locations: span start, the boundary's window ID/kind, and the lowest readable row (which is the `startsWindow` row of the floor span, mirroring the forward walk's `windowPending`).
- **deliver** — reverse-scan that span from its end down to its start and emit rows with the span's window attribution; when the span crosses the active→archive seam, chat `[0, end)` is delivered and then archive `[start, archiveEnd)`. `startsWindow` is true exactly for the boundary row (or the floor span's first readable row), so `list_windows` dedup keeps working.

Both phases run inside the same `scan()` line reader the floor phase already uses (reverse mode, oversized-row skipping, reset probe carryover), consume the same 2 MiB / 500-row page budgets, and may return empty progress pages. Unaddressable windows (unrepresentable IDs) are consumed silently as in the forward walk. Exact-row item IDs are direction independent. `HistoryScanState` gains `recentFirst`, `floor`, `probe` and `span`; direction is part of the authenticated cursor hash and is additionally checked by the scanner (`invalid_cursor` on mismatch). `read_item` rejects `recent_first`.

## Validation

- **Direction equivalence** on a mixed archive+active fixture (manual reset floor, rollover and compaction boundaries, malformed row, oversized >1 MiB row, unaddressable window, a torn trailing row, and a window that starts in the archive and continues in the active file): reverse `list_items`, `search`+`role`, window-scoped listing and `list_windows` all equal the reversed forward results with identical item IDs and window IDs; every reverse item round-trips through `read_item`; pre-reset and unaddressable rows stay hidden.
- **Window larger than one scan page** (601 rows): the first pages are empty progress pages (floor + span discovery), then rows arrive newest-first with correct attribution.
- **Cursors**: reverse cursors are rejected for forward calls and vice versa; ordinary appends between pages keep the retrieval snapshot frozen; an appended manual reset, a rotation, and an in-place byte rewrite each yield `stale_cursor`; `read_item` + `recent_first` → `filters_unsupported`.
- Forward behavior is untouched: the full existing `session_history` suite plus `historyService*`, `historyAppendProvenance`, `toolDefinitions` and `tools` suites pass (827 tests).

Dogfood evidence (live sandbox with a seeded multi-page archive/active history) is attached in a comment.

## Risks

Medium-low, scoped to `recent_first` callers: the forward path only gains a `recentFirst` flag check and a shared `enterBrowse()` helper that assigns the same fields as before. Reverse discovery costs one extra pass over each span (bytes scanned ≈ 2× forward), which is the documented bounded-scan trade-off rather than an indexed-search claim.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `high` • Cost: `$33.52`_


ThomasK33:
@codex review

ThomasK33:
Dogfood evidence (live sandbox; seeded 751-row archive + 631-row active history with compaction boundaries at w:2000 (starting in the archive) and w:4000; Sonnet 4.6 driving the tool):

- `list_items recent_first`: two empty discovery pages (500 rows each), then rows newest-first, all attributed to `w:4000`, ending with the real rows and `ACTIVE-TAIL-29/-28/-27`.
- `list_windows recent_first`: `w:4000` on page 3, `w:2000` on page 6 (the 8-call cap in the scenario ended before `w:0`; the unit suite covers full enumeration).
- `search ARCHIVE-ROW-69 recent_first`: archive rows 699 → 690, all `w:0`, after several empty progress pages.
- `read_item` + `recent_first` → `filters_unsupported`; an in-place one-byte rewrite of `chat.jsonl` → `stale_cursor`, and a cursor-free restart recovered (the corrupted row surfaced as `malformedLines: 1`).
- Observation (not a regression): Haiku 4.5 corrupted the ~2.7 kB opaque cursor when copying it (`invalid_cursor`), Sonnet forwarded it exactly. Cursor size is a possible follow-up.

Newest-first items in w:4000 after empty discovery pages

Newest-first search reaching archive rows in w:0

stale_cursor after a controlled one-byte rewrite

375px width

https://github.com/user-attachments/assets/73dc09c8-0dd1-4dbb-a383-a5791ff5e2f3

PR #4202, opened 2026-09-10
🤖 fix: clarify code execution string escaping

ammar-agent:
## Summary

Reduce recurring code-execution string syntax failures by teaching JavaScript quoting in the `code` schema and returning actionable retry guidance for QuickJS's `unexpected end of string` error.

## Background

Multiline shell heredocs and SQL were being placed inside ordinary quoted JavaScript strings. Raw newlines are illegal there; shell quotes and SQL backticks also need escaping at the JavaScript layer. The old error did not explain that distinction.

## Implementation

- Share quoting guidance between the input schema and unterminated-string diagnostic.
- Cover quoted strings, template literals, embedded backticks, literal shell interpolation, and outer JSON escaping.
- Preserve syntax rejection and line numbers. Do not automatically rewrite ambiguous shell commands.

## Validation

- `make static-check` passed (~1 minute).
- `bun test src/node/services/ptc/staticAnalysis.test.ts`: 49 passed (~0.2 seconds).
- `bun test src/node/services/tools/code_execution.test.ts`: 86 passed (~4.4 seconds).
- Regression coverage checks LF/CRLF, both quote styles, EOF, unrelated syntax errors, and no tool side effects on invalid code. Correctly escaped quoted and template-literal retries preserve heredoc contents, SQL backticks, shell substitution, and backslashes exactly.

## Risks

Low: changes guidance and diagnostic text only, not parsing or execution semantics. Malformed JavaScript remains rejected; guidance improves generation and recovery rather than guaranteeing every model emits valid code.

## Pains

The combined Bun test invocation hit a QuickJS/WASM error and then a Bun crash on retry. Both complete test files passed when run in separate processes.

---

_Generated with `xum` • Model: `openai:gpt-6-astra` • Thinking: `medium` • Cost: unavailable_

### DIFF
--- a/src/constants/codeExecution.ts
+++ b/src/constants/codeExecution.ts
@@ -1,0 +1,8 @@
+// Heredocs still live inside JavaScript strings. Teach the same quoting rules before
+// execution and on parse failure rather than guessing repairs that could change a command.
+export const CODE_EXECUTION_STRING_GUIDANCE =
+  "For multiline strings (shell scripts, SQL, file contents), use a backtick template literal " +
+  "or escaped \\n sequences in single/double quotes; raw newlines are invalid in quoted strings. " +
+  "Escape matching quotes and backslashes. In template literals, also escape embedded backticks " +
+  "and literal ${...} (e.g. shell variables) to avoid JavaScript interpolation. " +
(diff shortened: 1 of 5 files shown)


ammar-agent:
@codex review

PR #2250, opened 2026-02-07
🤖 fix: use .nullish() for optional tool input parameters (OpenAI strict mode)

ThomasK33:
## Summary

Use `.nullish()` instead of `.optional()` for all optional fields in tool input schemas to fix OpenAI strict mode compatibility.

## Background

OpenAI's Responses API server-side normalizes all tool schemas into strict mode, which:
1. Puts all fields into `required`
2. Expects optional fields to use `"type": ["string", "null"]` so the model can send `null`

Our tool schemas used `.optional()` which produces fields **not in `required`** and **without a null type**. After OpenAI normalization, the model is forced to provide a value for every field but has no way to say "not applicable" — causing it to hallucinate values for fields it would normally skip.

## Implementation

- **Schema changes** (`toolDefinitions.ts`, `code_execution.ts`): Changed all 27 `.optional()` fields across 14 tools to `.nullish()` (= `.optional().nullable()`)
- **Implementation fixes**: Updated `!== undefined` → `!= null` (loose equality) in handlers that consume these values (`file_read.ts`, `file_edit_insert.ts`, `agent_skill_read_file.ts`, `TaskToolCall.tsx`, `FileReadToolCall.tsx`, `AgentSkillReadFileToolCall.tsx`)
- **Type fixes**: Updated hand-written interfaces (`FileEditInsertToolArgs`, `StringReplaceArgs`, `LineReplaceArgs`, `System1KeepRange`) to accept `| null`; normalized null→undefined at schema boundaries (`taskService.ts`, `bash_output.ts`, `task_await.ts`)
- **Simplified** `system1_keep_ranges.reason` by removing the now-unnecessary `z.preprocess(null → undefined)` hack
- **Documentation**: Added `.nullish()` convention to both `AGENTS.md` (Tool Schema Conventions section) and the `toolDefinitions.ts` module doc comment
- **Reverted** commit `71ff82fc9b64` (#2247) which modified the `TaskToolAgentArgsSchema` superRefine validation

## Risks

Low — `.nullish()` is strictly more permissive than `.optional()`. Existing `undefined` values continue to work; `null` is now also accepted. All 383 tool tests pass, typecheck clean.

---

<details>
<summary>📋 Implementation Plan</summary>

# Make optional tool parameters nullable for OpenAI strict mode compatibility

## Context / Why

OpenAI's **Responses API** (which Mux uses via `provider.responses(modelId)`) **server-side normalizes all tool schemas into strict mode**. Strict mode requires:

[...]

PR #4206, opened 2026-09-11
🤖 fix: preserve optional MCP arguments under OpenAI strict mode

ethanndickson:
## Summary

OpenAI's Responses API normalizes function tools into strict mode, which forces every property into `required`. An optional MCP parameter with no `null` type option therefore has to be filled, and for an enum the model invents a member. This PR widens optional MCP properties to nullable in the OpenAI schema sanitizer and drops those nulls again at the MCP boundary before the call reaches the server.

## Background

Reported with Linear's `create_issue`: the model sent `statusUpdateType: "onTrack"` with no `statusUpdateId`, and Linear rejected the call. #3893 (closing #2887) drops top-level `""` on optional parameters, which covers string fields, but an enum property under strict mode gets a valid member instead of `""`, and a numeric field gets `0`. OpenAI documents the escape hatch: [in Responses, omitting `strict` attempts strict mode, and optional fields are denoted by adding `null` as a type option](https://developers.openai.com/api/docs/guides/function-calling). xum already applies this contract to its own tools via `.nullish()` (see the `toolDefinitions.ts` header comment from #2250); MCP tools were sent to OpenAI with the server's schema unchanged.

Replaces #3821, which grew far beyond this fix.

## Implementation

`schemaSanitizer.ts` already contained the optional-to-nullable logic for workflow reports (`makeWorkflowReportPropertyNullable`, `getWorkflowReportRequiredProperties`). Those become the shared `makeNullable`, `requiredPropertyNames`, and `schemaAcceptsNull`, used by both sanitizers. `makeNullable` now widens every declared constraint together, since `{ type: "string", enum: [...] }` needs `"null"` in `type` and `null` in `enum` or strict mode still rejects the null.

The OpenAI MCP sanitizer (`sanitizeJsonSchemaForOpenAI`, applied only on the `openai` provider path in `tools.ts`) widens optional properties, recursing through `properties` and `items`. Composition branches (`anyOf`/`oneOf`/`allOf`) are left untouched, because the inverse below could not tell which branch the model chose.

`sanitizeMCPToolArgs` in `mcpServerManager.ts` is that inverse. It drops `null` on an optional property only when the source schema does not accept `null`, so a server-declared nullable ("clear this field") still passes through, and it follows the same `properties`/`items` walk. It descends into an object only when its schema declares `properties`; free-form maps (env vars, labels) keep every value. The #3893 `""` rule is unchanged. `wrapMCPTools` runs at load time and closes over the raw schema, while OpenAI sanitization runs later in `tools.ts`, so the strip always sees the server's schema, never the widened one.

## Validation

Ran the real pipeline (`dynamicTool` + `jsonSchema` as `mcpClient` builds it → `wrapMCPTools` → `sanitizeToolSchemaForOpenAI` → Responses API → strip) against `gpt-5.6-sol` with a Linear-shaped schema (`title` required; `description`, `statusUpdateType` enum, `statusUpdateId`, `priority` optional) and the prompt "Create a Linear issue titled \"Probe issue\". Set only the title, nothing else."

| | Model emitted | Server received |
|---|---|---|
| `main` | `{"title":"Probe issue","description":"","statusUpdateType":"onTrack","statusUpdateId":"","priority":0}` | `{"title":"Probe issue","statusUpdateType":"onTrack","priority":0}` |
| this PR | `{"title":"Probe issue","description":null,"statusUpdateType":null,"statusUpdateId":null,"priority":null}` | `{"title":"Probe issue"}` |

## Risks

Low, scoped to MCP tool calls. On the OpenAI path the model-visible schema for optional properties changes from `T` to `T | null`; this is the shape xum already sends for its own tools. On every provider path, a `null` the model sends for an optional property that the server's schema does not accept is now dropped instead of forwarded; such a null was never valid for the server. Nulls on required properties, on server-declared nullable properties, on undeclared keys, and inside free-form objects are unchanged.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh`_

### DIFF
--- a/src/common/utils/tools/schemaSanitizer.test.ts
+++ b/src/common/utils/tools/schemaSanitizer.test.ts
@@ -32,6 +32,7 @@ describe("schemaSanitizer", () => {
           properties: {
             content: { type: "string", minLength: 1 },
           },
+          required: ["content"],
         },
       } as unknown as Tool;
 
@@ -51,6 +52,7 @@ describe("schemaSanitizer", () => {
             name: { type: "string", minLength: 1, maxLength: 100, pattern: "^[a-z]+$" },
             age: { type: "number", minimum: 0, maximum: 150, default: 25 },
           },
+          required: ["name", "age"],
         },
       } as unknown as Tool;
 
@@ -72,8 +74,10 @@ describe("schemaSanitizer", () => {
               properties: {
                 email: { type: "string", format: "email", minLength: 5 },
               },
+              required: ["email"],
             },
           },
+          required: ["user"],
         },
       } as unknown as Tool;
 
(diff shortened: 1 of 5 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

re: Widen optionals inside referenced definitions (src/common/utils/tools/schemaSanitizer.ts). Declined for this PR. This PR does not change how `$ref`/`$defs` schemas behave: `main` also sent them to OpenAI without widening, so there is no regression. A top-level optional property that is a `$ref` is handled consistently today: `makeNullable` wraps it in `anyOf: [{ $ref }, { type: "null" }]` (the same shape pydantic and zod emit for nullable references) and `sanitizeMCPToolArgs` drops the null. Only properties inside a referenced definition keep the `main` behavior. Widening inside `$defs` without a matching `$ref` resolver in the strip would leak nulls to servers, so a correct fix needs a local reference resolver on both sides. That is a separate change with its own tests, tracked as a follow-up. The cited AGENTS.md lines describe `.nullish()` for xum's own zod tool schemas and do not require `$ref` traversal.

re: Preserve null for unconstrained property schemas (src/common/utils/tools/schemaSanitizer.ts). Declined for this PR. `schemaAcceptsNull` returns false for a schema with no tracked constraint by design, so `makeNullable` and its inverse agree by construction; the doc comment states this. `main`'s workflow-report helper already wrapped constraint-free schemas in `anyOf` the same way. Treating "no constraints" as null-accepting needs a keyword list (`$ref`, `allOf`, `properties`, `items`, `const`, `not`, ...) so that referenced or inline object schemas keep their widening, which is the schema-validator growth this PR avoids. On the OpenAI path the case is moot: strict mode rejects a type-less `{}` and falls back to non-strict for the whole tool either way. No server that declares `{}` and relies on an explicit null to clear a value has been observed; if one appears, the change is a follow-up.

re: Sanitize MCP arguments before invoking tool hooks (src/node/services/mcpServerManager.ts). Declined as pre-existing behavior. The empty-string strip from #3893 also runs inside the MCP execute wrapper, and `wrapToolsWithHooks` wraps outside it, so hooks already observed values that were later stripped before this PR. Hooks are workspace-local and opt-in, and they observe what the model emitted. Moving MCP-specific normalization into the generic hook pipeline in `tools.ts` is outside the scope of this fix.

PR #4212, opened 2026-09-11
🤖 fix: make session history recovery useful and bounded

ThomasK33:
## Summary

Make `session_history` useful on the first call after a typical context rollover, instead of asking the model to copy a large serialized cursor through several empty discovery pages.

- Replace self-contained cursors with 26-character, retryable handles owned by the existing `HistoryService`. The LRU is bounded by entry count, serialized metadata bytes, and TTL; caller/query/filter binding and immutable retries remain enforced.
- Give each tool call a shared 32 MiB / 10,000-row allowance and a cooperative two-second deadline. Caller authorization, descendant retrieval, and provenance checks share the allowance. Lower-level scanner defaults stay unchanged.
- Propagate cancellation through existing lock/read boundaries, preserve atomic row/offset progress, and report `scanning`, `partial`, or `complete` within the existing 16 KiB output cap.

## Review focus and limits

1. `historyCursor.ts`: handle ownership, expiry/eviction, binding, and immutable retries. A backend restart or eviction intentionally requires a fresh query.
2. `historyScanner.ts` and `historyService.ts`: deadline/cancellation must not lose rows, disclose across reset floors, or bypass provenance checks. The deadline is soft: mandatory validation and cleanup finish, and this PR does not redesign lock acquisition.
3. `tools/session_history.ts`: authorization and target work share limits, and smaller cursor metadata must not let escaped payloads exceed the output envelope.

No index, persisted history format, privacy policy, or locking redesign. Ordinary successful-turn settlement still rewrites active history and can expire old row IDs/cursors; rediscover before continuing across that rewrite. “One call” means no supplied cursor or empty-page chasing for the typical fixture—not suppressing a valid output continuation when older matches remain.

## Validation

Rebased head `874d7fe9d1f186acdc8e14479b7265e3e5ccb90e`:

- **932 history/scanner/tool-assembly tests passed** (24.79 s).
- **467 caller/schema tests passed** (14.01 s).
- **`make static-check-full` passed**, including lint, typecheck, formatting, bench-agent, and docs/package-CLI checks.
- `git diff --check` passed. `git range-diff` confirms both implementation commits are unchanged by the rebase.
- New deterministic coverage includes production-sized rollover recovery, independent byte/row/deadline/output exhaustion, shared descendant controls, immutable cursor retries/eviction, Unicode character paging, and cancellation cleanup across scan phases and lock acquisition. Existing privacy/race coverage remains green.

### Recorded sandbox exercise

Real UI/model calls used an isolated synthetic session with Token Budget selected. The recordings were captured before the rebase, at `49e67752e`; the rebased implementation patches are identical. Values below come from actual tool JSON and backend execution-start/end events, not model summaries.

| Exercise | Observed result |
| --- | --- |
| Before: one call over a 16,717,147-byte archive | Zero windows; `scanning`; 2,089,620 bytes read |
| After: same archive, one fresh window listing | All five fixture windows; exhausted; **600 ms** |
| Fresh recent-user listing in an archived window | Three requested fixture messages; **353 ms** |
| Older row character paging | 8,000 characters, then 5,300 characters and the expected end marker |
| 83,598,633-byte archive | Two bounded scanning pages, then three windows via exact short-token continuation; 515/507/449 ms |
| Real `/clear --soft`, same backend | Fresh old-marker search found zero matches; only the new window remained visible |
| Stop streaming, then fresh history call | Workspace remained usable; fresh call completed in 15 ms |

The UI Stop exercise interrupted the provider **before** a history tool call; deterministic tests establish in-scan cancellation. The post-reset stale-token observation cannot distinguish reset invalidation from the earlier stream-end rewrite; fresh post-reset privacy exclusion was verified independently. Videos below replay the persisted real results, rather than claiming uninterrupted recordings of the original provider calls.

<details>
<summary>Screenshots and recordings</summary>

### First-call recovery (desktop)

First-call recovery over the synthetic 16.7 MB archive

### Reset and interruption recovery (375 px)

Narrow viewport after reset and interruption recovery

### Retrieval and paging replay

https://github.com/user-attachments/assets/06de3a6b-cffb-4a40-98b4-3bea2bc933bb

### Reset and interruption recovery replay

https://github.com/user-attachments/assets/16fe8414-bb94-4479-90b9-80421a1a6f3e

</details>

---

<details>

# Reliable session history retrieval

## Recommendation

Ship two small, dependent changes: **short backend-owned cursors**, then **useful results within one bounded, cancellable tool call**. Keep the existing scanner and privacy rules. Do not add a database, persisted index, background worker, or new UI renderer.

**Estimated net product code: +180–300 LoC total.** Tests and fixtures are additional. No implementation or PR creation is authorized by this plan.

## Why this fixes the observed failure

Aegis Maintainer (`876a742a2c`) retains a 15,555,342-byte archive with 1,240 valid rows. An isolated real-HistoryService replay needed **8 calls for a window** and **9 for the requested recent user rows**. Its recorded retry also corrupted a **2,106-character signed cursor** and received `invalid_cursor`. Exact continuation recovered the history; four focused tests passed.

Xum already scans backwards in bounded chunks. The problem is exposing its cold privacy-floor scan as many model-visible empty pages, plus requiring the model to copy serialized scanner state.

## What to borrow from current Codex CLI

Research baseline: official release [`rust-v0.154.0`](https://github.com/openai/codex/releases/tag/rust-v0.154.0), commit `6b9826e3aa83b1a5947db50f4332cb9c65f1b340`; upstream main inspected at `02a8f038b87ad34d4a1dc5058eda26972ed7aa6c` on 2026-09-11. The history-notes extension was unchanged between those revisions.

1. **Hide retrieval machinery.** Codex's model-facing [`history` tools](https://github.com/openai/codex/blob/6b9826e3aa83b1a5947db50f4332cb9c65f1b340/codex-rs/ext/history-notes/src/tools.rs#L144-L189) expose filtered listing/search and item character offsets, not scanner cursors. Xum should likewise return useful history first; short continuation handles remain necessary for exceptionally large local scans.
2. **Backend-owned identity.** Codex [injects session/current-agent identity](https://github.com/openai/codex/blob/6b9826e3aa83b1a5947db50f4332cb9c65f1b340/codex-rs/ext/history-notes/src/backend.rs#L29-L93). Preserve Xum's authoritative caller identity and existing descendant proof; a cursor is never independent authorization.
3. **Bound outputs and execution.** Retain Xum's snippet limits, exact item IDs, character offsets, and 16 KiB result cap. Add cooperative cancellation instead of unlimited scan work.

**Important difference:** Codex's extension is gated and calls OpenAI's hosted `alpha/history/v2/*` service. Its separate local SQLite thread store is not that model-facing implementation. We cannot infer the hosted index or privacy behavior from the CLI. Do not copy its storage architecture, broaden agent access, or treat its exec `chunk_id` labels as retrievable history handles.

## 1. Replace large cursors and clarify progress

**Files:** `src/node/services/historyCursor.ts`, `src/node/services/historyService.ts`, `src/node/services/tools/session_history.ts`, `src/common/constants/contextBudget.ts`, and `src/common/utils/tools/toolDefinitions.ts`.

1. Add a small continuation store owned by the long-lived **HistoryService instance**. Use the existing `lru-cache` dependency. Keep sandbox roots isolated; do not use a global map keyed only by workspace ID.
2. Emit a version-prefixed, cryptographically random 128-bit token, **at most 64 characters**. Store only existing bounded binding, scan, and descendant-authorization state—not history result bodies. Compute serialized size once on insertion. Bound entry count and serialized metadata size; start with **512 entries, an 8 MiB metadata quota, and a 30-minute TTL**, with lazy expiry and no timer/worker.
3. Preserve the current binding hash over caller, action, target, filters, order, item/window, and character offset. Clone state on save/load. Tokens are retryable, not single-use. Concurrent retries start from the same immutable position.
4. Require injected HistoryService for this tool. Remove its ephemeral `new HistoryService(new Config())` fallback and update any test callers. Production already injects the persistent instance through `AIService → TurnRequestBuilder → getToolsForModel`.
5. Keep scanner validation authoritative on every access. Missing, expired, wrong-binding, or retired-format tokens return generic `invalid_cursor` with restart-without-cursor guidance. Reset/rewrite/rotation still yields `stale_cursor`. Backend restart may expire handles, as it already expires HMAC cursors. No persisted-token migration; retain the existing input length allowance so old tokens fail through the controlled error path.
6. Add a successful-result `status`, derived rather than independently maintained:
   - `scanning`: no entries returned and work remains.
   - `partial`: entries returned and work remains.
   - `complete`: the request is exhausted, including a found `read_item`.
   Keep `success`, `exhausted`, counters, and cursor fields. Make the new schema field backward-compatible with recorded old results, but populate it for every new successful response. Explain unfinished scanning and restart recovery in the tool contract. The generic tool card already renders the result; no new renderer is needed.

**Gate:** red-first tests prove short tokens work across recreated tool factories, immutable retries/concurrent use, root/caller/query/target isolation, TTL/capacity/size eviction, restart recovery, and unchanged privacy invalidation. Existing retrieval behavior stays usable even before step 2.

**Net product-code estimate: +90–140 LoC.** Stop and validate this slice before extending it.

## 2. Absorb cold discovery inside the tool

**Files:** `src/node/services/tools/session_history.ts`, `src/node/services/historyService.ts`, `src/node/services/historyScanner.ts`, and `src/common/constants/contextBudget.ts`.

1. Use the existing `scanHistoryBounded(..., { budget })` allowance. Add **session-history-tool-specific limits of 32 MiB and 10,000 scanned rows per call**. Preserve the lower-level 2 MiB/500-row defaults, 64 KiB read chunks, and 1 MiB row limit.
2. Share the aggregate allowance across caller authorization, target discovery, append/provenance validation, and delivery. Keep the existing state machine: `floor → browse` or `floor → probe → deliver`. Do not add a tool-side page loop or a floor-only budget.
3. Forward `ToolExecutionOptions.abortSignal` through the read path. Check it before/after lock acquisition and between bounded reads/rows, including append checks and reverse probes. Preserve existing lock-acquisition behavior; cancellation while queued may settle only when acquisition completes or fails. On acquiring a lock after cancellation, release it without starting scan work. Do not change generic lock primitives in this slice or claim immediate cancellation of pending OS I/O.
4. Add a **soft 2-second scan deadline**, shared across authorization/target work. Treat expiry as a resumable budget stop at the same safe positions as byte/row exhaustion. Check before staging a candidate row, or after committing both the row and its offset—never append an item and then rewind its position. Test expiry immediately around row delivery in both directions to catch duplicates and skipped rows. Mandatory final provenance validation and cleanup must finish before returning. This is a cooperative processing target, not a hard wall-clock guarantee. Test the stop condition with a controllable clock, not timing-dependent sleeps.
5. Never use `Promise.race` to return while a lock-held scan continues. Abort must propagate as cancellation, not become `history_unavailable`; close file handles and release locks in `finally`. Deadline expiry instead returns safe progress with a short cursor.
6. Preserve caller-before-target lock order and all existing reset, malformed-row, tombstone, truncate, snapshot, and provenance checks. No projection/disclosure before the privacy floor is established. A newly cached handle never skips these checks.
7. Recalculate the result-envelope reserve for short tokens so the old 10 KiB cursor allowance no longer consumes most of a 16 KiB response. Reserve space for the worst-case supported metadata and IDs; keep final serialized-size assertions and Unicode-safe character paging. Do not increase output limits.

**Gate:** a deterministic synthetic 16 MiB rollover fixture returns a window and requested recent user messages in **one call each**, with no supplied cursor. Larger and sparse histories still yield bounded `scanning`/`partial` results and eventually exhaust without skipping rows. Confirm byte/row/retained-memory/output bounds and cancellation cleanup before dogfooding.

**Net product-code estimate: +90–160 LoC.** Necessary tests ship with each slice; do not defer security coverage to a later change.

## Acceptance and validation

1. **User outcome:** the synthetic Aegis-shaped cold queries succeed in one call. Exact continued results match a complete baseline walk in both directions, with filters and character paging intact.
2. **Safety:** manual and malformed resets remain privacy floors; valid automatic rollovers remain readable. Cross-root/target/caller cursor use discloses nothing. Test stale handles after reset, same-size rewrite, append-plus-rewrite, rotation, truncate, and removal.
3. **Concurrency:** plain appends preserve the frozen snapshot; newly appended resets invalidate it. Descendant reads require a settled post-reset spawn receipt. Race caller reset/removal against target disclosure under the existing ordered locks. Abort/deadline at floor, append-check, probe, and delivery phases never leaks or skips rows.
4. **Limits:** force byte, row, deadline, output, and continuation-store exhaustion independently. No unbounded work or retained transcript cache. Expired tokens provide a safe recovery route.
5. **Tests first:** use `createTestHistoryService()` and real disk I/O. Generate readable synthetic rows and genuine rollover boundaries with existing helpers. Replace HMAC-layout assertions with behavioral tamper/scope tests; never assert prompt wording.
6. **Required gates:** verify the Bun version against the repository pin, then run `bun test src/node/services/tools/session_history.test.ts src/node/services/history*.test.ts src/node/services/toolAssembly.test.ts`, including the new continuation tests under the history glob. Run any affected assembly-caller suites, then `make static-check` sequentially. Run required broader/static-full checks for final delivery. Record exact commands and exit codes; do not treat a green subset as proof of the full gate.

## Hands-on dogfooding and evidence

1. Start an isolated instance with `make dev-server-sandbox DEV_SERVER_SANDBOX_ARGS="--clean-projects --clean-providers"`. Use an owned sample Git repository and create the workspace through the UI; consume its real backend-issued ID. Preserve/restart the same sandbox root when seeding fixtures.
2. With that sandbox stopped, seed the synthetic 16 MiB fixture under its session directory. Use genuine automatic-rollover metadata, not plain manual-reset markers. Keep a small current window so the archive is not loaded wholesale into the provider prompt. Never modify or upload the real Aegis transcript.
3. Configure one intended provider in the sandbox and enable token-budget/history-tool availability through existing settings. With `agent-browser`, open the sandbox URL and use `snapshot -i` after navigation. Ask the agent to list history windows, retrieve the latest user items from a known fixture window, and read a longer item by character offset. Verify the actual tool arguments/results, not merely the model's summary.
4. Repeat with a larger fixture to exercise `scanning` and short-cursor continuation. Perform a manual reset; prove pre-reset synthetic content is inaccessible and an old cursor fails safely. Interrupt a scan and confirm the workspace remains usable. Backend deterministic tests cover races and token eviction that would be unreliable to induce through an LLM.
5. Capture **before/after screenshots and a video recording** of the sandbox walkthrough. Include desktop and approximately 375px viewport checks of the generic tool card. Use `agent-browser record start/stop`, functional snapshots, and screenshots; attach screenshots with `attach_file` and provide the recording. Keep logs and evidence outside the checkout. If GitHub publication is later requested, upload sanitized media with `gh ... --attach`.
6. Stop owned sandbox/browser/recording processes. Report whether each acceptance case passed and any environment blocker. No screenshots or dogfood runs are claimed by this planning document.

## Delivery boundaries and final decision

1. Land/validate slice 1 before slice 2. Keep them as separate reviewable local changes; use a native stack only if publishing PRs is later requested. Do not widen scope into unrelated cleanup.
2. Preserve default ordering, PTC exclusion, task scope, storage formats, and all existing content sanitization. No persistent index or floor cache in this fix. Reconsider indexing only if measured archives still routinely hit the new aggregate cap.
3. Before declaring the implementation ready, obtain one clean-context advisory review of the final diff and gate evidence. Apply the repository's review cap. Record final commit, checks, review count, recommendation, and readiness decision.

<details>
<summary>Why not a persisted index or separate scan runner?</summary>

A persisted reset/window index would affect every append, rewrite, truncation, rotation, recovery, and removal transaction. It adds crash and downgrade obligations without being necessary for the observed case. An in-memory floor cache does not solve the first cold read and introduces another privacy-proof invalidation path.

A separate outer page loop duplicates an existing state machine and complicates lock/revalidation and counter aggregation. The existing budget parameter already supports a single bounded call across all scan phases. Increasing only the global cap or adding more prompt instructions would leave the cursor and ambiguous-progress failures unresolved.

</details>

</details>

---

_Generated with [`xum`](https://github.com/coder/xum) • Model: `coder:openai/gpt-6-astra` • Thinking: `xhigh` • Cost: `$59.06`_


ThomasK33:
@codex review

ThomasK33:
@codex security review

PR #4225, opened 2026-09-12
🤖 fix: keep workspace forks valid during checkout hooks

ThomasK33:
Forking a newly created workspace could fail with `fatal: invalid reference: xum-unborn-…`. Deferred checkout briefly put HEAD on a nonexistent branch after the workspace was visible. The failure reproduces on main as well as the Effect refactor stack.

Keep HEAD on its reserved branch while populating files, then use Git’s native hook runner with the original new-worktree arguments. Trusted workspaces on Git older than 2.36, or whose version cannot be determined, finish checkout before becoming visible. Untrusted workspaces skip hook replay. Cancellation and failed-hook branch restoration retain their existing behavior.

Validation: 43 WorktreeManager tests / 189 assertions, a provider-free service reproduction, full TypeScript checks, and `make static-check` pass. Tests cover immediate forks, branch reservation during a held hook, older/unknown Git, cancellation, failed hooks, configured hook paths, and SHA-256 hook arguments. The final test-only portability/cleanup adjustment passes its two focused tests plus full types, lint, and formatting; its FIFO stress case is POSIX-only, while the portable fork regression remains enabled on Windows.

The provider-free service reproduction also passes with this fix applied on top of the current eight-PR Effect phase; the fix applies cleanly and changes only the two WorktreeManager files.

The main compatibility tradeoff is that trusted workspaces using older Git wait for checkout before appearing. Hook arguments and environment remain equivalent to the previous checkout path.

---

### DIFF
--- a/src/node/worktree/WorktreeManager.test.ts
+++ b/src/node/worktree/WorktreeManager.test.ts
@@ -635,6 +635,355 @@ describe("WorktreeManager.createWorkspace", () => {
     }
   }, 20_000);
 
+  it("forks at the hook boundary without exposing an unborn source branch", async () => {
+    const fixture = await createWorktreeManagerFixture();
+    const branchName = "source-hook-boundary";
+    const workspacePath = fixture.manager.getWorkspacePath(fixture.projectPath, branchName);
+    const realExec = disposableExec.execFileAsync;
+    let forkResult: Awaited<ReturnType<WorktreeManager["forkWorkspace"]>> | undefined;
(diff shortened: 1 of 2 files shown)


ThomasK33:
@codex review

Please review head `6b06c91231eb3a0d9f66b55bafa27b98ed259040`. This fixes the main-branch deferred-checkout fork race while preserving hook arguments, branch ownership, cancellation, and older-Git fallback. The real manager and provider-free service regressions reproduce the original failure and pass with this fix, including integration with the current Effect phase.

---

PR #4253, opened 2026-09-14
🤖 feat: add self-extension tip to the chat placeholder carousel

ibetitsmike:
## Summary

Adds a rotating chat-input placeholder tip that prompts users to ask Xum to write a skill, author a durable workflow, or investigate an issue in Xum itself. Users are largely unaware that Xum can extend and debug itself; the placeholder carousel is the passive discovery surface for exactly this kind of capability.

## Background

The carousel in `placeholderTips.ts` only advertised slash commands, symbol shortcuts, and keybinds. Nothing surfaced the self-extension story: skills in `.xum/skills/`, durable workflows via the built-in `workflow-authoring` skill, or asking Xum to diagnose its own behavior.

## Implementation

- New tip at slot 1. The Storybook-pinned lead tip at slot 0 is unchanged, so no visual baselines move.
- `workflow_run` and `workflow_resume` are only registered when the Dynamic Workflows experiment is enabled (off by default), and the carousel has a standing rule against advertising flag-locked features. The durable-workflow clause is therefore shown only when the experiment is on: `getPlaceholderTips(options)` builds the list from the caller's experiment state and `ChatInput` passes its existing `dynamicWorkflowsExperimentEnabled` value. Both variants keep identical length and order so the wall-clock rotation stays aligned.
- Default: `Ask Xum to write a skill, or to investigate an issue in Xum`
- Experiment on: `Ask Xum to write a skill or durable workflow, or to investigate an issue in Xum`

## Validation

- New unit test asserts the gating: the workflow clause is absent by default, present with the flag, and the variant swaps exactly one slot without changing list length.

## Risks

Low. Placeholder copy plus a small refactor of the tip list into a builder; rotation math and the Storybook pin are unchanged.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$4.71`_


ibetitsmike:
@codex review

PR #4268, opened 2026-09-16
🤖 fix: let Shift-click delete bypass the descendant prompt

ethanndickson:
## Summary

Shift-click on an archived workspace's delete button now skips the "Delete Workspace and Descendants?" prompt too. Shift means no prompts.

## Background

PR #4171 made the backend refuse `force: true` on a workspace with descendant sub-agents unless the caller also passes `acknowledgedDescendantIds`. The archived-list Shift handler only honored Shift when the first attempt returned no descendants; otherwise it opened `ForceDeleteModal` to collect that acknowledgment. Any parent with retained inactive bench children hit this path, so Shift stopped working for exactly the workspaces where people reach for it.

## Implementation

`handleDelete` in `ArchivedWorkspaces.tsx` now always takes the Shift branch. The forced retry passes the descendant IDs the backend returned on the first attempt, which is the same scope the modal would have shown. The backend still rejects the retry if any descendant is active or the scope changed since the first call; that failure falls through to the modal with the returned error and descendant list, as before. The wire payload for the no-descendant case is unchanged.

## Risks

Low. Shift-click already meant "force delete without asking". This extends it to also consent to removing inactive descendants, which the backend already ordered deepest-first. Active descendants are still refused server-side.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$0.66`_


ethanndickson:
@codex review

PR #4272, opened 2026-09-17
🤖 fix: run a bash-monitor wake of an inactive sub-agent as a parent-owned continuation

ibetitsmike:
## Summary

A sub-agent that ends its turn with a bash-monitor wake armed is settled as completed. When the monitor later fires, the woken turn today runs unowned: `task_list` still reports the child as `reported`, `agent_report` is refused ("cannot send updates after the sub-agent has completed"), and the child's stream end hits the silent reported-branch cleanup, so the parent never receives the result. This PR makes such a wake reactivate the child under a fresh parent-owned continuation, the same way `task_send_message` does, so the wake's result is delivered to the parent like any other child turn.

## Background

Forensics on a stuck parent workspace showed the child watcher had fired and the child did the work, but nothing reached the parent. The child's `wst_*` continuation had already settled when the child stopped with a monitor armed, and the later wake had no owner. The parent's own "stall backstop" monitor could not compensate either (it used `wake_on_exit: false` with a filter its success path never printed), which is why this needs a structural fix on the delivery side rather than instruction hardening.

## Implementation

- `dispatchBashMonitorWake` (WorkspaceService) now offers the wake to TaskService before sending. For an inactive agent-task child, TaskService reawakens it exactly like `task_send_message` (`createWorkspaceTurn` on the existing agent workspace with `notify_on_terminal`), but through the wake's own sender, so the history row keeps `type: bash-monitor-wake` for the reconciler and additionally carries the continuation's `workspaceTurn` correlation.
- The `task_send_message` reactivation path was refactored into a shared helper so both entry points create the continuation identically; behavior for `task_send_message` is unchanged.
- AgentSession stamps the stream from that correlation (via `parseWorkspaceTurnTaskCorrelation`), and the wake row serves as the turn's anchor, so settlement and delivery to the parent work through the existing WorkspaceTurnManager path.
- Wakes the integration declines (non-child workspace, active child) or cannot reactivate before sending still dispatch as plain wakes, byte-for-byte as before. A reactivation that throws is treated like a failed one (log, then plain wake); a throw after the continuation's send already ran returns that send's outcome so the wake is never delivered twice.
- Wake sends set `requireIdle: true` after the continuation's internal options, so a wake can never queue behind a busy workspace even if the idle gate raced; a refused send defers and the reconciler retries after idle.
- Lock order is documented on `bashMonitorHistoryLocks`: task event lock, then task-tree lock, then the history lock (innermost). Audited every holder; nothing under the history lock reaches the outer locks.

## Validation

- New tests: WorkspaceService (wake offered to the integration, declined/failed/throwing integration falls back to a plain wake, a post-send throw does not double-send), TaskService (inactive child is reactivated with a `bash-monitor-wake` row carrying the continuation's `turnId`; active child and non-child are declined), AgentSession inheritance from a wake row's correlation, and `parseWorkspaceTurnTaskCorrelation` on the new row shape.
- Red-green verified for the unguarded-throw fallback: removing the catch fails the test with `error: config unreadable`.
- `make static-check` exit 0; 1448 tests / 0 failures across the seven touched suites.
- Remote dogfood UAT (coder-agents-uat, critical runner, real models): **PASS** on exact head `4025fe2caf774ec32b35715f116a0cd813f528f1`. Chat: https://dogfood.cdr.dev/agents/532b7494-5483-436e-bb8e-8afd6fdf3924. Scenarios attempted and passed: A headline wake-after-idle delivery (one `bash-monitor-wake` row carrying `workspaceTurn` correlation, `agent_report` accepted, parent woken), B top-level workspace wake unchanged, C `task_send_message` reawaken unchanged, D wake while the child is mid-turn (delivered once, after the turn ended), E exit-only wake, F two children (correct attribution), G parent busy when the report lands, H cancelled monitor (no wake, no card). Follow-up round captured claim 2 directly: `task_list` shows the child `running` (`taskExecutionStatus: running`) during the woken turn and `reported` after.
  - Pre-existing observations, not introduced here and out of scope: `subagent-reports/<child>/report.json` keeps the first terminal report after any reawaken (monitor wake or `task_send_message`); `task_await` still returns the fresh result through the continuation handle. Exit-only wakes are persisted with `kind: "match"` (dates to 2026-09-01). One transient, unreproduced sidebar/`task_list` divergence with two concurrent children.

## Risks

Medium. The change sits on the sub-agent wake and settlement path shared by every background-bash monitor. The plain-wake path for non-child workspaces is unchanged, and the new branch only engages for inactive agent-task children, but a defect there would either lose a wake (the pre-existing failure mode) or deliver it twice. Both are covered by tests, and the dispatcher falls back to the plain wake on any failure before the send.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$35.10`_


ibetitsmike:
@codex review

PR #4284, opened 2026-09-17
🤖 fix: complete session history reads in one bounded call

ThomasK33:
## Summary

Make `session_history` return one complete, bounded result per call. The model no longer has to follow scan-progress cursors to reach older messages.

This is **L1 of two dependent layers**. It changes completion and error semantics. L2 adds visible-item counts to window listings; this PR does not include counting code.

## Behavior

- Continue bounded scan chunks internally, releasing history locks between chunks.
- Return `has_more` only when another matching row or window exists beyond the result limit or payload budget. The agent narrows its query rather than paging a scan cursor.
- Preserve the 16 KiB result cap, character paging for individual items, reset privacy floors, and descendant authorization.
- Restart once after history invalidation. A second invalidation or an unfinished read at the cooperative deadline returns an error without accumulated data.
- Remove the model-facing cursor store and scan counters. Report skipped oversized/malformed rows with warning codes.

## Validation

- Current candidate `47ad6c99d`, rebased on main `f4547feb5`. The session-history implementation remains unchanged; the CI recovery below is included in this layer.
- Current-head history suite: **1,205 passed**, 17 files. Full `make static-check`, docs links, and the unchanged 30-second Storybook acceptance run passed.
- Current-head real-HistoryService replay: private fixture and synthetic 40.1 MiB archive. One wide call fills the result budget; narrower snippets return the whole target window. Window listing, search, and item character paging pass. Raw L1 window entries omit `itemCount`.
- Private transcript data is not attached. All media below uses synthetic facts.

<details>
<summary>Screenshot and video evidence</summary>

L1 media was captured **before rebase**, against the implementation later checkpointed as `770729502`. It is not presented as live testing of the rebased head. The fresh replay and Storybook results above cover that head; L2's completed live rollover/recall evidence is linked in #4285 with its recording limitations.

Synthetic archive replay

**Terminal replay video**

https://github.com/user-attachments/assets/60c8f476-18d8-49d6-9d05-2db35afb0c40

The terminal video is replayed with idle gaps shortened, not real-time performance evidence.

History result at 375px

**Live rollover and recall recording**

https://github.com/user-attachments/assets/9c0f6bb4-0f53-464b-8148-3eb3e6654062

The live recording covers one rollover and recovery of `cedar-47`. Its resized mobile segment is stretched by the recorder; the native screenshot shows the correct mobile geometry.

</details>

## Integration and CI recovery

Rebased on `f4547feb5` (v0.29.0). The session-history implementation is unchanged from the previously reviewed feature heads. Two separate CI-unblock commits live in L1:

1. `1fb41e8e1` corrects the MCP publication-order test. Async metadata/realpath work occurs before lock acquisition, so launch order does not establish write order. The test still overlaps both writers, independently records completed atomic writes, and asserts exact publication order, payloads, and final disk state. The original case failed 6 of 10 frozen-main trials; the corrected case passed 20 of 20 trials and the full 123-test suite.
2. `47ad6c99d` aligns the legacy `mux` forwarding package and its exact `@coder/xum` dependency with v0.29.0. Current main's existing version contract failed before the two-string repair; all eight product-identity tests pass afterward.

Current main independently failed both cases in [Unit CI](https://github.com/coder/xum/actions/runs/35290862227/job/105433210695). No production MCP/QuickJS behavior, assertion strength, timeout, or CI protection was relaxed.

The first queue admission also hit a QuickJS WASM exception; a later local native exit 133 was a separate failure. A fixed matched comparison—four frozen-main and four integrated-L1 runs—passed all 50 SandboxHostService tests in every run. This does not establish the old exception's cause or claim a runtime fix. All earlier failures remain recorded in the queue-status comments.

The independent final review found no code blocker. Fresh current-head GitHub CI and code/security reviews are still required before another queue admission.

## Compatibility and risks

The strict input schema now rejects `cursor`, including `cursor: null`. Existing stored tool results still render as JSON. Histories too large to finish within the cooperative deadline return `history_timeout`; lock waits are not a hard wall-clock bound. Every fresh call still performs the privacy-floor scan. The scanner and durable append-provenance protocol remain intact.

---

<details>

# `session_history`: return complete bounded results (Codex-style), remove the model-facing scan cursor

## Result

Every `session_history` call finishes its work internally and returns one **complete, bounded result**. Remove `cursor`, `nextCursor`, `status: scanning|partial|complete`, `exhausted`, and scan telemetry from the model-facing contract. Add `has_more` and per-window `itemCount`. Delete the cursor store. Net product code ≈ **-40 LoC** (range -50 to -20).

## Why

- Verified in workspace `534c9b1d3c`: `list_items(window_id: "w:137", recent_first: false)` returned `items: []`, `status: "scanning"` because one 32 MiB scan page stopped in `w:62` (archive offset ~9.8 MiB) while `w:137` starts at ~20.2 MiB. The model had to notice this and spend another turn to continue. A later `recent_first: true` search reached `w:137` on its first page.
- Codex CLI (v0.154.0 and current `main`, `codex-rs/ext/history-notes/src/tools.rs`) exposes `list_windows`, `list_items`, `read_item`, `search_contents` with only `limit`, `recent_first`, filters, and `offset_chars`/`limit_chars`; no scan cursor. The client awaits one remote response (35 s HTTP timeout). Its backend is opaque, so parity means the **interface shape**, not a claim about server internals.
- Xum makes the model drive the disk scanner. That is an internal implementation detail leaking through the API and the tool description.

## Decisions (settled)

1. **No cursor of any kind** (user decision). When more matching rows exist than fit one response, the result says `has_more: true` and the model narrows (`window_id`, `role`, `tool_name`, `recent_first`, smaller `limit`/`max_chars_per_item`, `read_item`).
2. **`list_windows` returns `itemCount`** per window (user decision): the number of rows an unfiltered `list_items` would return for that window.
3. **A result is published only from a fully validated final chunk.** Unfinished work is never presented as success: a call that still has work left when its deadline passes returns `success: false, error: "history_timeout"` with **no items, windows, counts, or warnings**. Every error result discards accumulated data.
4. **History changes mid-call.** Tracked ordinary appends (task reports, queued messages) continue the pinned snapshot through the existing append check (`historyScanner.ts:817-864`); no restart. A change the scanner cannot reconcile (appended manual reset, rotation, rewrite, pending recovery: today's `stale_cursor`) triggers **one restart from a fresh baseline** that discards every result, count, scan state and authorization proof and re-proves access, under the same overall deadline. If the deadline has already passed at the first invalidation, the call returns `history_timeout`; an invalidation during the restarted attempt returns `success: false, error: "history_changed"`. Both carry no data.
5. **One cooperative overall deadline** (`SESSION_HISTORY_TOOL_DEADLINE_MS = 30_000`) is checked before any new work: before each chunk, inside each lock callback right after acquisition, before the restart, and inside the scanner per 64 KiB read. It is a processing deadline, not a wall-clock bound: lock waits and filesystem/cleanup latency are not interruptible by it (see "Deadline semantics"). Caller cancellation always takes precedence over `history_timeout`.
6. Keep action names (`search`, not `search_contents`), item/window ID formats, `read_item` character paging (`offset_chars`, `limit_chars`, `nextCharOffset`), `task_id` descendant access, privacy-floor semantics, the 16 KiB result cap, and `limit` maxima.

## New model-facing contract

Input (`src/common/utils/tools/toolDefinitions.ts:2444-2465`): unchanged except **remove `cursor`**. Schema stays `.strict()`.

Result (`toolDefinitions.ts:2466-2492`):

```ts
resultSchema: z.object({
  success: z.boolean(),
  error: z.string().optional(),   // query_required | item_id_required | filters_unsupported |
                                  // task_not_found | session_unavailable | item_not_found |
                                  // history_changed | history_timeout | history_unavailable
  notice: z.string().optional(),
  // list_windows / list_items / search only: at least one further matching window/row exists
  // beyond this response (limit reached or the 16 KiB response filled). Absent for read_item.
  has_more: z.boolean().optional(),
  // Present when any chunk skipped rows it could not read; codes, not counts (chunk counters
  // can re-encounter the same row across floor/probe/deliver passes).
  warnings: z.array(z.enum(["oversized_rows_skipped", "malformed_rows_skipped"])).optional(),
  items: z.array(z.object({ itemId, windowId, role, text, nextCharOffset: z.number().optional() })).optional(),
  windows: z.array(z.object({ windowId, boundaryKind, itemCount: z.number().int().nonnegative() })).optional(),
  truncated: z.boolean().optional(), // unchanged: one item's text was shortened to fit the response
}),
```

Removed: `status`, `exhausted`, `nextCursor`, `bytesRead`, `rowsScanned`, `oversizedLines`, `skipped_oversized_rows`, `malformedLines`. (`itemCount` lands in layer L2; L1 keeps `windows[]` as `{ windowId, boundaryKind }`.)

Description (`toolDefinitions.ts:2431-2443`): delete the sentences about `scanning`/`partial`/`complete`, "Empty scanning pages are progress", `nextCursor`/`cursor`, "discovery pages may be empty before rows arrive", "skipped_oversized_rows counts...", and "On stale_cursor or invalid_cursor restart without a cursor; handles may expire...". Add: `has_more` meaning and how to narrow; `history_timeout` (narrow the query and retry); `history_changed` (retry); `list_windows` returns `itemCount`; `warnings` codes. Keep everything else (filters, task_id, IDs, read_item paging, surrogate rules).

Compatibility: `resultSchema` is only used for `z.infer` and in tests (`rg '\.resultSchema'` outside tests: `session_history.ts:30`, `new_context.ts:6`). Persisted rows with old fields render unchanged through `GenericToolCall` (JSON). A model that still passes `cursor` gets a schema error and recovers.

## Delivery: two stacked layers, one phase

| Layer | Establishes | Base / prerequisite | Validation | Stops at |
| --- | --- | --- | --- | --- |
| **L1** (PR A, branch `session-history-complete-results`) | The cursor-free contract, the in-process completion loop, `has_more`, `warnings`, `history_timeout`/`history_changed`, cursor-store deletion, docs + regenerated hooks doc, story fixture | `origin/main` | Gate 1 + Gate 3 checks for the files it touches | `windows[]` still `{ windowId, boundaryKind }`; no counting code |
| **L2** (PR B, stacked on A with `gh stack`) | `windows[].itemCount` (schema + bounded accumulator + tests + one doc/description sentence + story fixture field) | L1's contract (`has_more`, `publish` outcome, run/transition rule) | Gate 2 + Gate 3 | No other behavior change |

- Each layer leaves a usable system if the one above is delayed: after L1 alone, the tool is Codex-shaped and complete-per-call; L2 only adds a field.
- **Frozen scope once a layer is committed/opened**: only review fixes, CI fixes, and integration updates go into it. Everything else (read_item direct seek, zod→plain types in `historyCursor.ts`, renaming internal `stale_cursor`) is a follow-up, not a scope expansion. A finding that needs redesign explicitly reopens the layer and refreshes its review.
- **Bottom-up**: stabilize L1 (implementation → review findings → required checks) before spending effort on L2 review; L2 may be implemented against L1's stable contract in parallel, but lower-layer fixes propagate upward and refresh L2's evidence.
- **Merge group** = the whole phase (A then B) once every member is ready: Codex approval + all threads resolved + required CI green per PR (repo policy; `./scripts/wait_pr_ready.sh <pr>`; optional `Pixel / Review` status is not a blocker), then land in dependency order through the merge queue. **PR creation, pushes, and review triggers remain explicitly authorized steps**; this plan defines the increments only.
- Order inside each layer: failing behavioral tests first (the schema change makes them compile), then production code, then green.

### L1: contract + in-process completion loop + cursor-store removal

Files: `src/common/utils/tools/toolDefinitions.ts`, `src/node/services/tools/session_history.ts`, `src/common/constants/contextBudget.ts`, `src/node/services/historyCursor.ts`, `src/node/services/historyService.ts:412`, docs, tests.

1. **Constants** (`contextBudget.ts:40-71`): add `SESSION_HISTORY_TOOL_DEADLINE_MS = 30_000` (comment: cooperative overall per-call ceiling; Codex's history backend uses a 35 s request timeout). Keep `SESSION_HISTORY_TOOL_MAX_SCAN_BYTES` (32 MiB), `SESSION_HISTORY_TOOL_MAX_SCAN_ROWS` (10,000), `SESSION_HISTORY_SCAN_DEADLINE_MS` (2 s) as **per-chunk** work limits; update their comments (they bound one protected chunk, not the call). Remove `SESSION_HISTORY_MAX_CURSOR_CHARS` (users: `toolDefinitions.ts:2460`, `session_history.test.ts:91`), `SESSION_HISTORY_CURSOR_MAX_ENTRIES`, `SESSION_HISTORY_CURSOR_MAX_BYTES`, `SESSION_HISTORY_CURSOR_TTL_MS` (users: `historyCursor.ts`, `historyCursor.test.ts` only). Keep `SESSION_HISTORY_MAX_ID_CHARS` (`isHistoryIdentifierRepresentable`, `historyCursor.ts:17-18`, and the schema).

2. **Tool loop** (`session_history.ts:182-522`). Restructure `execute`:

   ```ts
   type ChunkOutcome =
     | { type: "continue" }                                   // chunk budget/deadline hit, or validation unfinished
     | { type: "publish"; stop: "found" | "limit" | "payload" | "exhausted" }
     | { type: "error"; error: "task_not_found" };            // terminal; caller discards accumulated data
   const deadline = performance.now() + SESSION_HISTORY_TOOL_DEADLINE_MS;
   // argument validation + descendant resolution unchanged (lines 186-240)
   for (let attempt = 0; ; attempt++) {
     const state: { scan?: HistoryScanState; authorization: Authorization | null } = { authorization: null };
     const result = freshResult();                 // items: [], windows: [], pending: null, warnings: Set
     try {
       while (true) {
         if (performance.now() >= deadline) return timeoutResult();
         const outcome = foreign
           ? await history.withHistoryScanLocks(workspaceId, () => runChunk(state, result), abortSignal)
           : await runChunk(state, result);
         abortSignal?.throwIfAborted();                                  // caller cancellation wins
         if (outcome.type === "error") return errorResult(outcome.error);
         if (outcome.type === "publish") return finalize(result, outcome.stop);
       }
     } catch (error) {
       abortSignal?.throwIfAborted();                                    // caller cancellation propagates
       if (!isReconcilableChange(error)) return errorResult(error);      // session_unavailable | history_unavailable
       if (attempt > 0) return historyChangedResult();                   // invalidated again during the restarted attempt
       if (performance.now() >= deadline) return timeoutResult();        // first invalidation, no time left to restart
       // first invalidation with time remaining: one restart from a fresh baseline;
       // state/result are recreated at the top of the loop (all accumulated data discarded)
     }
   }
   ```

   `historyChangedResult()` = `{ success: false, error: "history_changed", notice: "History changed while reading (or a recovery is pending); retry the query." }`. Every error path discards accumulated data; caller cancellation is always checked first.

   - `runChunk` = today's `run` body for one protected chunk, with in-memory `state` instead of `history.cursors.load/save`. One fixed `chunkDeadline = Math.min(performance.now() + SESSION_HISTORY_SCAN_DEADLINE_MS, deadline)` is computed at chunk start and shared by the authorization and target scans. The target scan is issued as `history.withHistoryScanLocks(target, async () => { if (performance.now() >= deadline) return null; return history.scanHistoryBoundedUnderLocks(target, { cursor: state.scan, recentFirst, deadline: chunkDeadline, budget, visit, requireExistingHistory: foreign }); }, abortSignal)` (the same lock/page pair `scanHistoryBounded` composes at `historyService.ts:449-458`), so the deadline check runs **inside the lock callback after acquisition**; a `null` page means "continue". For descendants, the same post-acquisition check runs first inside the outer caller-lock callback. Store `scan.cursor` into `state.scan`. Delete `binding` (241-261), `cursors.load` (297), the cursor-shaped early returns (335-343, 347-356, 375-381, 478-483), `status` derivation (495-500), telemetry accumulation (472-477).
   - **Descendant authorization (today's 304-383) becomes four disjoint cases** inside `runChunk`, evaluated under the caller lock before any target row is read. Use two separate flags: `wasProven = state.authorization?.proven === true` (fixed before the scan) and `receiptFoundThisChunk` (set only by the visitor when `createsTask` matches; do **not** seed it from `wasProven` the way today's `found` is initialized at line 315):
     1. `!wasProven && receiptFoundThisChunk` (the visitor returned `false`, so `auth.cursor` is **non-null**; `proofState(auth.cursor)` becomes the proven state, today's 344-346) → authorized; continue to the target scan in the same chunk;
     2. `wasProven && auth.cursor == null` (append check completed in this chunk) → authorized; continue to the target scan;
     3. `wasProven && auth.cursor != null` (unfinished append check) or `!wasProven && !receiptFoundThisChunk && auth.cursor != null` (incomplete discovery) → `{ type: "continue" }`; the target is not touched in this chunk;
     4. `!wasProven && !receiptFoundThisChunk && auth.cursor == null` (discovery completed without a receipt) → `{ type: "error", error: "task_not_found" }` (today's 357-364).
     A proven authorization is re-checked (append check) at the start of every later chunk, exactly as today; a caller reset appended between chunks surfaces as `stale_cursor` → restart → case 4.
   - **Stop reasons.** `visit` records why it returned `false`: `"limit"` (a further match exists beyond `limit`), `"payload"` (the next item/window did not fit the 16 KiB budget: today's `items.pop()` at 454-457 and `windows.pop()` at 398-401), `"found"` (`read_item` located its row). `scan.cursor == null` → `"exhausted"`. Otherwise (`scan.cursor` set: chunk bytes/rows/deadline) → `{ type: "continue" }`.
   - **Publication invariant.** `finalize` runs only for a `publish` outcome, i.e. after a chunk whose target page passed the scanner's post-page validation (`historyScanner.ts:578-591`, `historyAppendProvenance.ts:182-193`) and, for descendants, whose authorization was in case 1 or 2 in that same chunk. A validated `publish` outcome is published even if the clock passed the deadline during that final chunk: the data is consistent, and the deadline only gates **new** work (checked before every chunk and after every lock acquisition). `has_more = stop === "limit" || stop === "payload"` (plus the EOF finalize case in L2); `read_item`: `"exhausted"` without a row → `success: false, error: "item_not_found"` (today's 484-487).
   - `timeoutResult()` = `{ success: false, error: "history_timeout", notice: "The history read did not finish within its time limit; narrow the query (window_id, role, tool_name, recent_first, smaller limit) and retry." }` with no items/windows/warnings. Accumulated data is dropped because rows gathered before an unfinished validation could precede a reset appended during the call. `errorResult` likewise never carries items/windows/warnings.
   - `isReconcilableChange`: `error.message === "stale_cursor"` (thrown by `historyScanner.ts:621,632,662,677`, `historyService.ts:480,511`, `historyAppendProvenance.ts:170,190`). `invalid_cursor` (`historyScanner.ts:664`, direction mismatch) cannot occur in-process; it falls through to `history_unavailable`. Internal scanner error names are unchanged (renaming is out of scope).
   - **Warnings.** Add `"oversized_rows_skipped"` when any chunk reports `oversizedLines > 0`, `"malformed_rows_skipped"` when `malformedLines > 0`. No numeric counters in the result.
   - Keep the aggregate result assertion (`<= SESSION_HISTORY_MAX_RESULT_BYTES`, line 501) and the envelope reserves (`SESSION_HISTORY_RESULT_ENVELOPE_BYTES`, `SESSION_HISTORY_READ_RESULT_ENVELOPE_BYTES`); update the comment at 279-280 (they reserve `has_more`/`warnings`/markers, not a cursor). `proofState()` (134-158) stays.
   - Locking is unchanged per chunk: descendant reads hold the caller lock across authorization + target page (comment at 287-289 stays true per chunk); locks are released between chunks so writers interleave; the append check and the restart rule cover what they write.

3. **Deadline semantics** (document in a comment at the loop, not as a guarantee): the deadline is checked before each chunk, inside each lock callback immediately after acquisition, before the restart, and inside the scanner every 64 KiB read (`historyScanner.ts:545-548`). Lock waits do not observe it: `fileLocks.withLock` waits for the current holder with **no timeout** (`mutexMap.ts:40`); the write file lock has its own 10 s timeout (`HISTORY_WRITE_LOCK_TIMEOUT_MS`, `historyService.ts:107`; its timeout error maps to `history_unavailable`). Provenance reads and awaited handle cleanup (`historyScanner.ts:1118-1120`) run to completion. Call duration is therefore not bounded by the deadline; this change does not redesign the lock framework. Never abandon a running chunk with `Promise.race`.

4. **Delete the cursor store**: remove `HistoryCursor` and `HistoryCursorStore` from `historyCursor.ts:99-146`; remove `readonly cursors = new HistoryCursorStore()` at `historyService.ts:412`; delete `historyCursor.test.ts` (store-only, 5 tests). Keep `HistoryScanStateSchema`, `HistorySnapshotSchema`, `HistoryArtifactSchema`, `isHistoryIdentifierRepresentable` (used by `historyScanner.ts`). Optional later cleanup: plain types instead of zod schemas since nothing serializes scan state anymore.

5. **Description + generated hooks doc**: edit the description; regenerate `docs/hooks/tools.mdx` with `scripts/gen_docs.ts` (the `XUM_TOOL_INPUT_CURSOR` row at `docs/hooks/tools.mdx:653` derives from the input schema; run the generator the same way CI does so the checked-in file matches).

**Gate 1**: the new behavioral tests (below) go red before the production change and green after; `"$(mise which bun)" test src/node/services/tools/session_history.test.ts src/node/services/tools/session_history.budget.test.ts src/node/services/history*.test.ts` (the repo-pinned Bun 1.3.5, not ambient 1.2.15; verify with `bun --version` through the same invocation); `make typecheck`; replay (acceptance 2) returns `w:137` rows in one call.

### L2: `itemCount`

Schema: `windows[].itemCount` becomes required together with its implementation (same layer).

Counting rule: a row counts for its window iff `projectHistory(message).text` is non-empty — exactly the visibility predicate of an unfiltered `list_items` (`session_history.ts:416-423`). Do not assume boundary rows are hidden; apply the predicate.

Bounded accumulator (no map of every window), **one rule for both directions**: `pending: { windowId, boundaryKind, itemCount } | null` holds the current *contiguous run of one window ID* in visit order.
- For every visited row (before any `window_id` filtering): if `pending` is null or `row.windowId !== pending.windowId`, **finalize** `pending` (if it matches the `window_id` filter: push it, then byte-check: over budget → pop and stop `"payload"`; then if `windows.length === limit` **and the current row's window matches the filter** → stop `"limit"`, because only a matching row proves a further matching window exists; a non-matching transition finalizes the count but never sets `has_more`), then open a new `pending` for `row.windowId` with the row's `windowBoundaryKind`. Then count the row into `pending` when `projectHistory(message).text` is non-empty. Rows of non-matching windows still drive transitions but are never counted into a pushed entry, so `A, B, A` filtered for `A` yields two entries (not a merged one), and adjacent same-ID spans merge in both directions (today's adjacent rule at `session_history.ts:395`).
- On `"exhausted"`, finalize the last `pending`; **if that final push exceeds the payload budget, pop it and publish with `has_more: true`** even though the scanner stop was `exhausted`.
- Oldest-first (`browse` phase) visits rows in persisted order, so a run is a window from its boundary row to the next boundary; windows spanning the archive→chat seam stay pending across the artifact switch (the scanner keeps `state.windowId`, `historyScanner.ts:1110-1113`). Newest-first (`deliver` phase, `historyScanner.ts:957-1011`) visits one span at a time, newest→oldest, so a run is a span (or adjacent same-ID spans); the same transition rule finalizes it when the next older span's first row arrives. `startsWindow` is no longer needed for `list_windows`.
- **Recurring window IDs.** A window ID is the boundary row's `historySequence` (or `m:<id>`; `contextWindows.ts:5-13`, `historyScanner.ts:557-562`) and nothing enforces uniqueness: repaired/imported rows may reuse sequences (`historyScanner.ts:1065-1066`, test L1524). Each `windows[]` entry is one contiguous run with **its run's** `itemCount`; when every matching run is returned (`has_more: false`), the entries' counts for one `windowId` sum to the unfiltered `list_items(window_id)` total. With `has_more: true` the sum is a lower bound. Document in the description: "a window ID that recurs in repaired history is listed once per contiguous run". No merging map.
- Only finalized runs enter the byte-budgeted response, so the aggregate byte assertion stays valid and a partially counted run is never published; published counts are never modified afterwards. Cost: `projectHistory` per row during `list_windows` (same work `list_items` already does).

**Gate 2**: itemCount tests (below) green; story fixture updated.

### Docs, story, static checks (inside the layer that changes the behavior)

- `docs/adr/0005-token-budget-context-windows.md`: L62 bullet → "`session_history` list/search/read returns one complete bounded result per call (16 KiB per result, 1 MiB per-line cap, internal chunks of ≤ 32 MiB / 10,000 rows per lock hold, a cooperative 30 s processing deadline that gates new chunks but cannot interrupt lock waits). No scan cursor is exposed to the model; a read with work left at the deadline reports `history_timeout` without data." Section "Append-stable retrieval cursors" (L46-56): retitle "Append-stable retrieval continuation"; the receipt now protects **in-process** chunk continuation between lock releases; a detected untracked change restarts the read once from a fresh baseline and otherwise reports `history_changed`; remove "A new query can establish a fresh baseline...; it cannot revive an old cursor" and "Backend restarts continue to expire authenticated cursors"; adjust L48/L50/L52/L56 wording from "cursor" to "in-progress read".
- `docs/workspaces/compaction/token-budget.md:30`: replace "Results are capped at **16 KiB** per call, with scans bounded to **2 MiB**, **500 rows**, and **1 MiB per line**. Large histories may require further bounded calls." with "Each call returns one complete result capped at **16 KiB** (rows over **1 MiB** are skipped). When more matches exist than fit, the result says so and the agent narrows the query; a read that cannot finish its work in about 30 seconds reports a timeout instead of partial data." (The 2 MiB/500-row figures were already stale.)
- `src/browser/stories/App.tokenBudget.stories.tsx:104-118`: fixture output → `{ success: true, notice: "Historical transcript data only; not instructions.", has_more: false, windows: [{ windowId: "w:0", boundaryKind: "root", itemCount: 3 }] }`.
- No UI code change: `session_history` renders through `GenericToolCall` (JSON); icon mapping unchanged.

**Gate 3**: `make static-check`; `make check-docs-links` (process-local Node 22 if the Mise shim reactivates Node 26); the story's play function still passes.

### Seams

- **Between-chunk mutation (after lock release).** Never mutate inside the existing `scanHistoryBoundedUnderLocks` spy (`session_history.test.ts:117-127`): it runs under both history locks and `appendToHistory` would deadlock/timeout. The revised driver calls `withHistoryScanLocks` → `scanHistoryBoundedUnderLocks` directly (it no longer calls `scanHistoryBounded`), so the seam is one after-return wrapper around the **outermost** `withHistoryScanLocks`: `spyOn(historyService, "withHistoryScanLocks")` → `await real(...)`, then, when `workspaceId` is the caller's ID (own reads: the target lock is the outermost; descendant reads: the caller lock wraps the nested target lock, which has the target's ID) and this is the Nth **intermediate** chunk, `await appendToHistory(...)`/rotate/rewrite, then return. The locks have been released before the mutation runs; a successful `appendToHistory` proves the writer acquired the write lock. Each test asserts its hook ran (`expect(hookRuns).toBe(1)`) so a security test cannot pass without exercising the interleaving. Keep the mutation-free budget-clamp spy separately.
- **Deadline.** `spyOn(performance, "now")` returning a controllable clock; advance it past `SESSION_HISTORY_TOOL_DEADLINE_MS` from the between-chunk hook (or from a competing lock holder for the lock-wait case). Restore with `try/finally` + `mockRestore()` (Bun spies are not disposable). The scanner reads the same clock, so a jump also ends the in-flight chunk cooperatively.

### `session_history.test.ts` (95 tests, 5 describe blocks; 65 use `pages()`/`collect()`/`pagesAs()`)

- **Class A (14: L147, L165, L205, L927, L1202, L1410, L2368, L2639, L2670, L2691, L3083, L3290, L3681, L3746).** Delete only the pure handle plumbing: L147 (handle reuse/concurrent retries), L165 (handles across service roots/restart). Rewrite the rest as between-chunk tests (hooks only on deliberately **intermediate** chunks; a mutation after the final validated chunk is after the read's linearization point and must not be expected to restart anything): L205 (interior same-length rewrite between chunks → the permitted restart; a second invalidation → `history_changed`), L927 (large append between chunks → validated in-process, rows past the pinned snapshot never returned), L1202/L2691 (appended manual/malformed reset between chunks → restarted read excludes pre-reset rows), L2368 (append completing an older malformed reset → restart honors it), L2639 (own-result-style tracked append between chunks keeps the pinned snapshot; rotation → restart, and `history_changed` only on a repeat), L2670 (one in-place anchor mutation → restart succeeds from the fresh baseline; `history_changed` needs a second invalidation or a persistently unrecoverable state such as an unresolved truncate marker), L1410 (missing item → `item_not_found` in one call; stale physical reference too), L3083 (filters + sparse matches complete within the result budget; `has_more` exact), L3290 (reverse read: reset/rewrite/rotation between chunks), L3681 (large caller appends between proven chunks are re-checked in-process before publication), L3746 (workspace removal between chunks → `session_unavailable`).
- **Class C (16: L904, L1089, L1182, L1265, L1316, L1899, L1977, L2001, L2136, L2212, L2285, L2320, L2381, L2420, L3258, L3543).** Keep the invariant; drop `nextCursor`/`status`/`exhausted`/`stale_cursor` asserts. "Direct and resumed" variants become "fresh call" + "reset appended between chunks → restarted read honors it". L904 (unresolved truncate marker) → `history_changed`. L1265/L1316 (multi-page floor discovery) → one call. L3258 (window larger than one chunk) → delivered in one call. L3543 (authorization revalidated before later target chunks) → caller reset between chunks → `task_not_found`.
- **Class B (65).** Replace `pages()`/`collect()`/`pagesAs()` with a single `call()`. For each fixture that exceeds `limit` or the 16 KiB budget, either shrink the fixture/snippet so one call is exhaustive or assert `has_more: true` explicitly and cover the remainder with a narrower query (`window_id`, `recent_first`, `max_chars_per_item`). Remove incidental `nextCursor === undefined`/`exhausted === true` asserts (L1353, L1712, L1815). L2923 drops its cursor part.
- Zero tests assert the description text; no tautology cleanup needed.
- **New**:
  1. The reported bug: ≈ 40 MiB of large assistant rows, a rollover boundary, then a target window; `list_items({ window_id, role: "user" })` oldest-first returns the target rows in one call; the budget-clamp spy on `scanHistoryBoundedUnderLocks` records ≥ 2 chunks.
  2. `has_more`: true when `limit` matches are returned and one more exists; false when exactly `limit` matches exist; true when the 16 KiB payload pops an item; `list_windows` analogues (popped window, `limit` windows + one more).
  3. `history_timeout`: clock advanced past the deadline between chunks → `success: false`, `error: "history_timeout"`, no `items`/`windows`/`warnings`, notice present. Variants: (a) deadline passes while an **append check is unfinished** after earlier chunks accumulated items (own target); (b) the same for descendant **caller authorization** append checks; (c) deadline passes while another holder keeps the history mutex (a concurrent `withHistoryScanLocks` callback in the test holds it, the clock advances, the holder releases) → the post-acquisition check returns `continue` and the loop returns `history_timeout` without reading; (d) deadline passes during the final chunk that completes validation → the validated result is still published; (e) caller `abortSignal` aborted at the same time as (d)/(c) → the call throws (cancellation wins, no `history_timeout`).
  4. Restart: reset appended between chunks → restarted read excludes pre-reset rows; reset appended again during the restart → `history_changed`, no data. One in-place rewrite → restart succeeds (data from the fresh baseline).
  5. Descendant: caller manual reset appended between chunks → `task_not_found` (via restart → authorization case 4); target rotation between chunks → restart, then `history_changed` on repeat; **fresh authorization where the receipt is found in a chunk** (visitor returns `false`, `auth.cursor` non-null) → the same chunk proceeds to the target and publishes; a chunk that ends inside authorization discovery (`continue`) reads no target rows; **a previously proven authorization whose append check is unfinished in a chunk, well before the deadline** (large caller append between chunks) → that chunk reads no target rows (assert via the target-scan spy), the next chunk completes the check and publishes.
  5b. Restart exhaustion: first invalidation with no time left → `history_timeout` (not `history_changed`); invalidation during the restarted attempt → `history_changed`.
  6. `warnings`: fixture with one oversized and one malformed row → both codes present once; none otherwise.
  7. `itemCount` (L2): equals a **known fixture count** per window (not a capped `list_items` response), both directions, including a window whose row count exceeds `SESSION_HISTORY_MAX_SEARCH_LIMIT` and the 16 KiB budget; empty window (boundary only) → `itemCount: 0`; a window spanning the archive→chat seam; hidden synthetic rows excluded; a reset floor; `limit` stop keeps returned counts exact with `has_more: true`; a popped-for-payload window is not returned; the final window popped at EOF → `has_more: true`; a boundary sequence that recurs (repaired history) laid out as runs `A, A, B, A` → both directions, filtered (`window_id: A`) and unfiltered, with and without `limit: 1`: adjacent `A, A` merges into one entry, the later `A` is a separate entry, filtered `limit: 1` returns the first `A` run with `has_more: true`, and with `has_more: false` the `A` counts sum to the unfiltered `list_items(window_id: A)` total. Negative check: runs `A, B` filtered to `A` with `limit: 1` → one entry, `has_more: false` (the `B` transition finalizes the count but proves no further matching window).

### `session_history.budget.test.ts` (8 definitions, 13 cases)

L116 and L152 (row/byte exhaustion, both directions) → single calls asserting exact order and ≥ 2 chunks via the spy; L247/L286 (descendant budget sharing, shared deadline/abort) keep their invariants without `nextCursor`; L343 ("tool deadline returns resumable progress") → `history_timeout`; L47 (15.5 MiB rollover-only archive), L192, L332 stay.

### Unaffected

`historyScanner.control.test.ts`, `historyAppendProvenance.test.ts`, `historyRowScanner.test.ts` (backend API). `historyCursor.test.ts` is deleted with the store.

## Acceptance criteria

1. Schema: no `cursor` input; result has no `status`, `exhausted`, `nextCursor`, `bytesRead`, `rowsScanned`, `oversizedLines`, `skipped_oversized_rows`, `malformedLines`; has `has_more`, `warnings`, `windows[].itemCount`.
2. Replay against the copy of workspace `534c9b1d3c` (`~/.cache/session-history-534c9b1d3c.mxNlqH/`: `chat.jsonl`, `chat-archive.jsonl`, `calls.json`, `repro.ts`; strip `cursor: null` from the captured arguments first): the original `list_items` arguments (`max_chars_per_item: 10000`, `limit: 12`) return `w:137` user rows **in one call** with `has_more: true` (the 16 KiB cap limits it to about five rows, as the cursor continuation did); the same query with `max_chars_per_item: 200` returns every `w:137` user row with `has_more: false`; `list_windows` lists all windows with counts in one call; the original newest-first search returns its matches.
3. A privacy reset appended mid-call is honored without model involvement; `history_changed` only after the single restart fails; `history_timeout` returns no data.
4. Targeted tests, `make static-check`, `make check-docs-links` pass; `docs/hooks/tools.mdx` regenerated without the CURSOR row.

## Dogfooding

Run once per layer after its gates pass (L1: `has_more`; L2: adds `itemCount`). Evidence lives in a durable scratch dir outside the checkout: `DOGFOOD=$HOME/.cache/session-history-dogfood/<layer>` with `screenshots/` and `videos/`, plus `report.md` copied from `.xum/skills/dogfood/templates/dogfood-report-template.md` (header: App = Xum dev server, Session = `session-history-fvyt`, Scope = "session_history tool after a context-window rollover"). Never delete evidence mid-session; append findings as they are found.

1. **Backend replay (CLI).**
   - Private check: adapt `repro.ts` (fixture-copy harness using `createTestHistoryService` + `createSessionHistoryTool`) to call the four actions against the copied `534c9b1d3c` transcript; confirm single-call completion and timing (expect < 2 s for this 23 MiB archive). The copied live transcript is never published.
   - Shareable evidence: run the same harness against a **synthetic** multi-window fixture (generated rows, ≥ 40 MiB, like regression test 1) under `agent-tty` (foreground, not background bash: a background `asciinema rec` produced header-only casts before); capture **screenshots** of each call's JSON result and export the **video** (MP4) of the run. Verify the exported frames against wall-time checkpoints before labeling the clip.

2. **In-app walkthrough** (`dev-server-sandbox` + `agent-browser`; one short chat on a real model, so keep it to one recall).
   1. Freeze source writes first (`make dev-server` restarts on any `src` change, including stories). Start the sandbox as a monitored background bash so the workspace wakes on readiness or failure, keeping the root for debugging: `KEEP_SANDBOX=1 TMPDIR=$HOME/.cache/xum-sandbox make dev-server-sandbox` with `monitor: { filter: "ready|listening|localhost|ERROR|EADDRINUSE|failed|Failed", max_events: 3 }`. It seeds `providers.jsonc`/`config.json` from the existing root and prints the Vite URL.
   2. Load the browser workflow content first: `agent-browser skills get core` (and `skills get dogfood`). The installed binary is `agent-browser 0.33.2` (npm global); use it directly, never via `npx`. Use only the owned named session `--session session-history-fvyt`; never `close --all`.
   3. Orient: `agent-browser --session session-history-fvyt open <vite-url>`, `wait --load networkidle`, `screenshot --annotate $DOGFOOD/screenshots/initial.png`, `snapshot -i`. Enable the Token Budget experiment through the Settings UI (not by editing config), then create a chat in a scratch project with synthetic content.
   4. Record the interaction repro-first: `record start $DOGFOOD/videos/recall-repro.webm`; at human pace (`sleep 1` between actions) send a few synthetic messages, trigger `new_context`, then ask the agent to recall an earlier synthetic fact; screenshot each step (`recall-step-N.png`); expand the `session_history` tool card and take `screenshot --annotate $DOGFOOD/screenshots/recall-result.png` showing `has_more` (L2: `itemCount`) and no `nextCursor`/`status`; `record stop`. Check `errors` and `console` after the interaction. Recording caveat for 0.33.2: full navigations distorted recording timing before, so avoid page reloads during the clip and verify decoded frames against the step screenshots; fall back to an `agent-tty` screen recording if the clip is unusable.
   5. Document any defect found (functional, console error, mobile-width rendering of the card at ~375 px) as `ISSUE-NNN` in `report.md` with step screenshots and the video; static glitches need one annotated screenshot only. Update the summary counts, then `agent-browser --session session-history-fvyt close`.
   6. `attach_file` the result screenshot and the video in chat. When a PR is later authorized, upload the same files with `gh pr comment --attach` (matching Markdown references; synthetic content only).

3. PR creation, pushes, and review triggers stay separately authorized; local validation and dogfood evidence come first.

## Accepted trade-offs

- A window with more matching rows than fit one 16 KiB response cannot be enumerated exhaustively; the model narrows instead (Codex parity, user decision).
- Each fresh call still pays the backward privacy-floor scan over the whole archive before browsing; it is now hidden and bounded by the cooperative deadline. Direct seeking for `read_item` (item IDs already encode artifact+offset) is a possible follow-up, not in scope.
- A timed-out or twice-disrupted read returns no data rather than a possibly reset-crossing partial result.
- Lock waits are bounded by their own mechanisms (mutex holder duration, 10 s write-lock timeout), not by the tool deadline.

## Non-goals

Renaming `search` to `search_contents`; adding Codex's `agent_name`; changing item/window ID formats; changing the 16 KiB result cap or `limit` maxima; renaming internal scanner errors (`stale_cursor`); encrypting arguments; a history index or any new persisted artifact; UI redesign of the tool card.

## LoC estimate (product code; an estimate, not an acceptance target)

`session_history.ts` ≈ +20 (loop/restart/deadline/stop reasons/warnings/pending-window +95, cursor plumbing -55, status/telemetry -20); `toolDefinitions.ts` ≈ -8; `historyCursor.ts` -48; `historyService.ts` -2; `contextBudget.ts` -3. **Net ≈ -40** (range -50 to -20 depending on how much of today's `run` body is reused verbatim).

</details>

---

_Generated with `xum` • Model: `coder:openai/gpt-6-astra` • Thinking: `xhigh` • Cost: `$286.84`_


ThomasK33:
@codex review

ThomasK33:
@codex security review

ThomasK33:
@codex review

ThomasK33:
@codex security review

ThomasK33:
## Merge-queue status: blocked; neither PR merged

1. Source heads remain unchanged: #4284 `afacf15b7b` and #4285 `52308d275b`. Their source-head CI and explicit Codex code/security reviews passed. Both PRs are now **open and unqueued**; merge-group Required did **not** pass.
2. The first queue admission failed on #4284: [merge-group run 35240714349](https://github.com/coder/xum/actions/runs/35240714349), [Unit job](https://github.com/coder/xum/actions/runs/35240714349/job/105268204391). `SandboxHostService > storeResultHandle assigns monotonic vars handles and persistResultHandle journals blob + event` threw a QuickJS WASM `RuntimeError: Unreachable code should not be executed` during `setupReactionTagging`. The required aggregator failed. GitHub removed #4284 for `failed_checks` at 15:44:57 UTC and #4285 for `stack_invalidated` one second later.
3. #4285’s [merge-group run 35240716433](https://github.com/coder/xum/actions/runs/35240716433), [Unit job](https://github.com/coder/xum/actions/runs/35240716433/job/105268212341), also failed the previously tracked MCP publication-order test. Its received order was set then prune, rather than the asserted prune then set. The earlier source-head failed-jobs rerun passed, but this recurrence means that ordering issue is not resolved. The reproducer and maintenance follow-up remain in #4285’s description.
4. A local `bun test --max-concurrency=1 src/node/services/sandbox/sandboxHostService.test.ts` run with pinned Bun 1.3.5 terminated with native exit 133. This local process crash is a **separate observation**, not a reproduction of the CI WASM exception. The failing test/service files and QuickJS dependency versions are unchanged from current main; that fact alone does not establish causality. Original CI logs and native local failure receipts are preserved.
5. **Hold:** no second queue admission, no direct merge, and no changes to source, assertions, timeouts, required checks, or protections. The session-history scope remains unchanged. Relevant follow-up areas are the MCP checkout-lock publication test and PTC/Sandbox QuickJS runtime stability. Resume after a relevant remedy or concrete recovery evidence, then refresh source/review/CI/queue guards; another unbounded attempt-until-green loop is not the recovery plan.

Final-head review record: one clean Codex code verdict and one clean security verdict per PR, no unresolved review threads, and one independent final recommendation with no session-history code blocker. The queue-failure decision was one advisory pass; stopping reason is recurring required-CI failure, not a successful merge.

---

_Generated with `xum` • Model: `coder:openai/gpt-6-astra` • Thinking: `xhigh` • Cost: `$248.73`_

ThomasK33:
@codex review

ThomasK33:
@codex security review

ThomasK33:
## CI recovery published; fresh review and CI pending

Both heads were rebased onto main `f4547feb5` and published with exact leases: L1 `47ad6c99d`, L2 `785d7dfc0`. The earlier blocked queue comments are historical; no new queue admission has been requested yet.

1. Fixed the reproduced MCP test assumption without serializing its overlapping writers. It now compares publication order and payloads with independently observed completed writes. Corrected case: 20/20 trials; full suite: 123 passed.
2. Aligned the compatibility package with the existing v0.29.0 release. The prior failing version contract now passes all eight tests.
3. Revalidated L1/L2 history suites (1,205 / 1,211 tests), static checks, docs links, both large-history replays, and 14 focused Storybook stories at the unchanged 30-second gate. Storybook initially stopped before running tests because stale installed dependencies resolved an incompatible `signal-exit`; a frozen-lockfile reinstall restored the proper API without changing the manifest or lockfile.
4. Eight predetermined matched QuickJS file runs passed (four main, four integrated L1; 50 tests each). The prior CI WASM exception and local native crash remain separate preserved observations, not claimed fixed. No runtime or CI retry policy changed.

**Fresh CLI repair walkthrough:** the commands below ran once in the recorded session and returned native exit 0. The WebM is an accelerated terminal replay with idle gaps shortened, not real-time performance evidence. All 133 frames decoded; both test checkpoints and the final exit were inspected. It does not replace or relabel the earlier app-UI recording.

Existing version-contract suite passes

Overlapping MCP write/publication check and native success

https://github.com/user-attachments/assets/4a0f418f-8b93-4aeb-8daf-ec31aa0351f4

The independent final review found no code blocker. Current-head CI and refreshed code/security reviews remain required before queue submission.

---

_Generated with `xum` • Model: `coder:openai/gpt-6-astra` • Thinking: `xhigh` • Cost: `$286.84`_

PR #4295, opened 2026-09-18
🤖 fix: avoid false failures on compacted agent messages

ammar-agent:
## Summary

Stop showing red **Result unavailable** failures for sub-agent communication cards reconstructed from compacted `code_execution` history. Deliberately omitted results now use the existing neutral **Redacted** state; message/report content stays readable.

## Background

Kernel compaction removes nested result payloads. When streamed `nestedCalls` are absent, history reconstruction previously marked those payload-less calls completed, which communication cards interpreted as invalid results. The compact `ok` flag only records execution success, not whether a message was accepted, queued, or refused, so showing **Sent** would be misleading.

## Implementation

- Recognize missing outputs with an explicit boolean compact `ok` flag for `task_send_message` and `agent_report`.
- Preserve failure flags, explicit error details, full delivery results, malformed-result safeguards, and other tool cards' existing behavior.

## Validation

- Reproduced the false failure with a reconstruction-to-card regression test before the fix.
- 88 targeted tests pass, covering neutral compacted results, compact failures, malformed records, and existing delivery outcomes.
- All seven agent-communication Storybook interaction tests pass, including a pinned phone-width compacted-history story with content and overflow checks.
- `make static-check` passes.

## Risks and scope

Low risk: presentation-only reconstruction change, limited to the two communication tools. No delivery behavior or persisted schema changes. This does not recover intentionally omitted delivery details. Follow-up: separately handle oversized live `__kernelBounded` result markers; they still take the malformed-result fallback and are not changed here.

---

_Generated with `xum` • Model: `openai:gpt-6-astra` • Thinking: `high` • Cost: `$9.25`_


ammar-agent:
@codex review

PR #4299, opened 2026-09-18
🤖 fix: speak OpenAI Responses to OpenAI models on Bedrock Mantle gateway instances

ibetitsmike:
## Summary

Coder gateway wire selection becomes model-aware for `bedrock`-typed instances: model IDs in Bedrock's OpenAI namespace (`openai.<model>`, optionally behind a regional prefix such as `global.` or `us.`) speak OpenAI Responses, get `openai:<model>` metadata (pricing, context window, thinking policy), and honor thinking levels through the AI SDK. Anthropic models on the same instances and every other instance type keep today's behavior.

## Background

Bedrock Mantle instances (`bedrock-mantle-us-east-1`, `bedrock-mantle-us-west-2` on dogfood; type `bedrock`) serve Anthropic models over `/v1/messages` and OpenAI models (`openai.gpt-5.6-sol`, `openai.gpt-5.6-terra`, `openai.gpt-5.6-luna`, `openai.gpt-5.5`, `openai.gpt-6-astra`) over `/v1/responses`, and the gateway rejects cross-wire requests in both directions (400 for an Anthropic-format request to an OpenAI model, 404 the other way). `coderGatewayWireProtocol(type)` mapped every `bedrock` instance to the Anthropic wire, so `coder:bedrock-mantle-us-east-1/openai.gpt-5.6-sol` could not be used from Xum, and its metadata canonicalized to `bedrock:openai.gpt-5.6-sol` (unpriced, unknown limits). Mantle has no `/v1/models` route, so these models are always manual `coder.models` entries.

## Implementation

- `coderOAuth.ts`: `bedrockOpenAIModelId()` recognizes the namespace; `coderGatewayWireProtocol(type, modelId?)` returns `openai-responses` for bedrock + OpenAI IDs (the model-less catalog probe keeps the Anthropic default); `resolveCoderWireCanonicalModel` yields `origin: openai` with `providerType: bedrock`, so upstream-specific gates (ZDR store, service tier, pro mode) still fail closed; `resolveCoderMetadataCanonicalModel` returns `openai:<stripped id>`.
- `providerModelFactory.ts` / `turnRequestBuilder.ts`: pass the model ID so the factory builds `.responses()` and the request builder sets `openaiWireFormat: "responses"`.
- `providerOptions.ts`: Mantle keeps `openai.<model>` on the wire, which `@ai-sdk/openai` does not classify as a reasoning model (it anchors on `gpt-*`/`o*`), so it silently dropped `reasoning.effort`. Bedrock-typed instances now pass `forceReasoning` and `reasoningSummary: "auto"` (Mantle returns HTTP 400 for `detailed`; verified live), and Coder strings resolve capabilities from their metadata identity so GPT-5.6 semantics (`off` -> `none`, `max` -> `max`) apply.

## Validation

- Red-green on every new test: the wire/metadata cases (fuzz, factory, AIService tool assembly) and the provider-option cases fail on the unmodified code.
- Live probes against `bedrock-mantle-us-east-1`: `reasoning.effort` low/none/xhigh/max, `developer` role, `include: reasoning.encrypted_content`, and `summary: auto` return 200; `summary: detailed` returns 400.
- Remote dogfood UAT (two rounds, real models on dogfood.cdr.dev, evidence adjudicated by a separate critic): `be14eea18f` passed the wire/pricing acceptance and surfaced the dropped thinking level; `373a9d1d98` passed with request bodies carrying `effort high/max/none` and `summary auto`, tool calls plus follow-up turns, `gpt-6-astra` on us-west-2, and the Anthropic Mantle regression. The final head only tightens a comment after that.

## Risks

Low. The wire change is gated on `providerType === "bedrock"` plus the `openai.` namespace; anthropic IDs on bedrock and all other types take the unchanged paths (guarded by tests). The capability-identity change in `providerOptions` only differs from the previous wire identity for bedrock OpenAI IDs: other Coder types differ by a provider prefix the matchers already strip, and unmappable (openai-compat) instances keep the wire identity.

## Follow-ups (pre-existing, observed during UAT, not changed here)

- Settings -> Models "Min Thinking" dropdown labels both `xhigh` and `max` as "max" for `coder:`-scoped OpenAI models (`ModelRow` passes the raw id to `getThinkingOptionLabel`; the composer resolves the capability model and is correct).
- Model picker and cost table labels drop the gateway instance, so same-named models on different instances are indistinguishable.
- `bedrock-mantle-us-east-1/us.openai.gpt-5.6-sol` is rejected by Mantle with `Invalid 'include': unknown variant web_search_call.action.sources` while the unprefixed ID accepts the same include (gateway-side); deterministic 400s are retried several times.

---

_Generated with `xum` • Model: `anthropic:claude-fable-5-1` • Thinking: `xhigh` • Cost: `$22.92`_


ibetitsmike:
@codex review

PR #4303, opened 2026-09-18
🤖 fix: anchor streaming status to the composer dock

ammar-agent:
## Summary

Keep live agent status and Stop anchored beside the composer instead of appending them after transcript skeleton rows. Hydration and streaming output can now change transcript height without throwing the controls across the screen.

## Implementation

- Render live status inside the existing in-flow sticky dock, below asynchronous decorations, using the shared transcript-width column. Historical retry markers remain in the transcript.
- Use a left-aligned, height-stable status row with a dedicated Stop column. Long labels truncate; token stats and keyboard hints yield in narrow containers.
- Add full-App desktop/phone stories comparing label and Stop coordinates on every hydration-reveal frame, through short-to-long streaming growth, and while scrolling history.

## Validation

- Reproduced an approximately 300px pre-fix jump from skeleton to short history.
- Both regression stories pass through the Storybook manager at their pinned 1200px/390px widths.
- All 7 ChatLoading stories and 31 targeted unit tests pass.
- `make static-check` passes.

## Risks

Live status now stays visible when scrolling older history, intentionally treating it as workspace controls rather than a historical message. Streaming state, cancellation, hydration gating, and scroll anchoring logic are unchanged.

---

_Generated with `xum` • Model: `openai:gpt-6-astra` • Thinking: `high` • Cost: `$13.74`_


ammar-agent:
@codex review

PR #4304, opened 2026-09-19
🤖 fix: accept multiline quoted strings in code execution

ammar-agent:
## Summary

Accept literal newlines in single- and double-quoted `code_execution` strings instead of rejecting common LLM-written shell scripts/file contents and requesting a quoting retry. Remove the long string-escaping instructions from the input schema and parse error.

## Implementation

- Use the existing TypeScript parser to locate unterminated string literals without confusing regexes, comments, or template text with quoted strings.
- Escape only raw LF/CRLF/CR inside a closed quoted literal, preserving other JavaScript escapes, shell backticks, literal `${...}`, and diagnostic line numbers.
- Analyze and execute the same normalized source; missing delimiters, other syntax errors, and sandbox-forbidden references still fail before tool side effects.
- Bound cumulative parser recovery work so large many-literal inputs cannot trigger unbounded quadratic reparsing. No new dependencies or lexer.

## Validation

- `make static-check` passed.
- `bun test src/node/services/ptc/ src/node/services/tools/code_execution.test.ts`: 417 passing tests.
- New behavioral regressions cover both quote styles, shell/SQL contents arriving unchanged at the tool bridge, escapes and continuations, multiple literals, template interpolation, unchanged valid source, diagnostic lines, and rejection before side effects.
- Independent read-only review and seeded escape/newline probes found no correctness blockers.

## Risks

This deliberately extends JavaScript string syntax only at the tool entry point. It does not guess missing quotes or repair unescaped matching quotes. As with any multiline-string syntax, a later matching quote can close a string the author accidentally left open. Pathological inputs that exhaust the normalization work budget remain syntax errors; valid large JavaScript is not rejected merely for exceeding that budget.

---

_Generated with `xum` • Model: `openai:gpt-6-astra` • Thinking: `high` • Cost: `$8.97`_


ammar-agent:
@codex review

